PR #4, opened 2022-09-30
Change component data wrappers to be data classes

jkozlowicz:


leandro-lucarella-frequenz:
I didn't go in detail through the PR yet, but my first impression is we really need to start specifying units. Is SoC in percentage? Is it in Wh, kWh? Etc.

I opened a discussion so we don't derail about that in this PR: https://github.com/frequenz-floss/frequenz-sdk-python/discussions/6

jkozlowicz:
Makes sense. Units are described in the protobuf messages, e.g. for `soc`:

```
class Data(google.protobuf.message.Message):
    """Battery data."""

    DESCRIPTOR: google.protobuf.descriptor.Descriptor

    DC_FIELD_NUMBER: builtins.int
    SOC_FIELD_NUMBER: builtins.int
    TEMPERATURE_FIELD_NUMBER: builtins.int
    HUMIDITY_FIELD_NUMBER: builtins.int
    @property
    def dc(self) -> frequenz.api.microgrid.common_pb2.DC:
        """DC electricity metrics."""
    @property
    def soc(self) -> frequenz.api.microgrid.common_pb2.MetricAggregation:
        """Battery's overall SoC.
        In percent (%).
        """
```

Shall I copy these over from the protobuf message docstrings?

leandro-lucarella-frequenz:
> Shall I copy these over from the protobuf message docstrings?

Yes, please! Until we decide how to represent units better I think is a good idea to have it at least in the documentation.

sahas-subramanian-frequenz:
It looks like the commits are not cryptographically signed.  Maybe because it was rebased from the UI?  

I wonder if we can get the DCO check to verify that the commits are signed, and not just that they have the "Signed-Off" line in the commit messages.

src/frequenz/sdk/microgrid/component_data.py
```diff
@@ -17,229 +19,218 @@
 from .component_states import EVChargerCableState
 
 
-class _BaseComponentData(ABC):
+@dataclass(frozen=True)
+class ComponentData(ABC):
```
leandro-lucarella-frequenz, line 23: We don't need now the `ABC` here, right?

src/frequenz/sdk/microgrid/component_data.py
```diff
@@ -17,229 +19,218 @@
...
+class MeterData(ComponentData):
+    """A wrapper class for holding meter data.
 
-class MeterData(_BaseComponentData):
-    """A wrapper class for holding meter data."""
+    component_id: the component ID
 
-    def __init__(self, raw: microgrid_pb.ComponentData) -> None:
-        """Create a `MeterData` instance.
+    timestamp: the timestamp of when the data was measured.
```
leandro-lucarella-frequenz: This documentation (`component_id` and `timestamp`) should go to `ComponentData`.

src/frequenz/sdk/microgrid/component_data.py
```diff
@@ -17,229 +19,218 @@
...
-        """
-        return self._raw.ts.ToDatetime(tzinfo=pytz.UTC)
 
+@dataclass(frozen=True)
+class MeterData(ComponentData):
+    """A wrapper class for holding meter data.
 
-class MeterData(_BaseComponentData):
-    """A wrapper class for holding meter data."""
+    component_id: the component ID
```
leandro-lucarella-frequenz: The diff makes it really hard to read, but the Docstring format is missing an `Attributes:` header:

```py
class MeterData(ComponentData):
    """A wrapper class for holding meter data.

    Attributes:
        component_id: the component ID
        ...
```

Not sure why this isn't caught by darglint or the other.

https://google.github.io/styleguide/pyguide.html#384-classes

leandro-lucarella-frequenz: The documentation is is pretty obvious, but maybe make it `component_id: the ID identifying this component in the microgrid`

src/frequenz/sdk/microgrid/component_data.py
```diff
@@ -72,7 +89,24 @@ def from_proto(raw: microgrid_pb.ComponentData) -> MeterData:
 
 @dataclass(frozen=True)
 class BatteryData(ComponentData):
-    """A wrapper class for holding battery data."""
+    """A wrapper class for holding battery data.
+
+    soc: battery's overall SoC in percent (%).
+
+    soc_lower_bound: the soc lower bound of the current bounds of the metric, as
```
sahas-subramanian-frequenz: I think the word `current bounds` in these descriptions can be confusing.

One option suitable for just SoC is "the SoC below which discharge command will be blocked by the system".  @tiyash-basu-frequenz does that make sense?

src/frequenz/sdk/microgrid/component_data.py
```diff
@@ -17,229 +19,218 @@
...
-    @property
-    def soc(self) -> float:
-        """Get current SoC level.
-
-        Returns:
-            SoC level.
-        """
-        return self._raw.battery.data.soc.avg
+    soc_lower_bound: the soc lower bound of the current bounds of the metric, as
+        imposed by the overall system.
```
leandro-lucarella-frequenz: Maybe add that this is also percentage (if it is percentage)? (same for other bounds)

src/frequenz/sdk/microgrid/component_data.py
```diff
@@ -17,229 +19,218 @@
...
-    def soc_upper_bound(self) -> float:
-        """Get SoC upper bound.
+    soc_upper_bound: the soc upper bound of the current bounds of the metric, as
+        imposed by the overall system.
 
-        Returns:
-            Maximum SoC level.
-        """
-        return self._raw.battery.data.soc.system_bounds.upper
+    capacity: the capacity of the battery
```
leandro-lucarella-frequenz: In which unit?

src/frequenz/sdk/microgrid/component_data.py
```diff
@@ -17,229 +19,218 @@
...
 
-    @property
-    def soc_lower_bound(self) -> float:
-        """Get SoC lower bound.
+    power_lower_bound: the power lower bound of the current bounds of the metric, as
+        imposed by the overall system.
```
leandro-lucarella-frequenz: In Watt? (same for other `power_...`)

leandro-lucarella-frequenz (commented):
I only made a comment on the first occurrence, but when/if fixed it should be fixed on all classes. (like add units to the documentation for all attributes, moving the `component_id` and `timestamp` documentation only to `ComponentData`, etc.)

leandro-lucarella-frequenz:
> I wonder if we can get the DCO check to verify that the commits are signed, and not just that they have the "Signed-Off" line in the commit messages.

The DCO bot is provided as is, but merging will be blocked if commits are not cryptographically signed, so having it on a check is not necessary (and it will be redundant in a way)

leandro-lucarella-frequenz:
@jakub-toptal BTW you can see the protobuf comments with less noise in the [`.proto` files](https://github.com/frequenz-floss/frequenz-api-microgrid/tree/v0.x.x/proto/frequenz/api/microgrid)

src/frequenz/sdk/microgrid/component_data.py
```diff
@@ -17,229 +19,218 @@
...
-    _raw: microgrid_pb.ComponentData
+    component_id: int
+    timestamp: datetime
 
-    @property
-    def component_id(self) -> int:
-        """Return the component id of the message.
+    @staticmethod
+    def from_proto(raw: microgrid_pb.ComponentData) -> ComponentData:
+        """Create ComponentData from a protobuf message.
```
leandro-lucarella-frequenz: This is very weird, one comment got lost from my last review.

I said that this should probably be:

```suggestion
from typing_extensions import Self
...

    @classmethod
    def from_proto(cls: Self, raw: microgrid_pb.ComponentData) -> Self:
        """Create ComponentData from a protobuf message."""
        ...
        
        return cls(...)
```

Otherwise if we need to subclass some component without adding new attributes for some reason, the `from_proto()` will return the wrong type.

Since [`Self`](https://docs.python.org/3.11/library/typing.html#typing.Self) was introduced in Python 3.11, we need to add [`typing_extensions`](https://pypi.org/project/typing-extensions/) as a dependency to be able to use this, but I think is fine, because is a package maintained by Python itself, so it should be stable and maintained and we are not requiring Python 3.11 as the minimum version for a while, so I think it would be a good addition to the tools we can use in the SDK in terms of typing.

src/frequenz/sdk/microgrid/component_data.py
jakub-toptal: I think we do, because it shouldn't be possible to create instances of `ComponentData`.

src/frequenz/sdk/microgrid/component_data.py
jakub-toptal: Actually, I don't know. The comment in the protobuf message only says:

> The capacity of the battery.

src/frequenz/sdk/microgrid/component_data.py
jakub-toptal: The corresponding protobuf message documentation says only:

```
class MetricAggregation(google.protobuf.message.Message):
    @property
    def system_bounds(self) -> global___Bounds:
        """The current bounds of the metric, as imposed by the overall system."""
```

````
class Bounds(google.protobuf.message.Message):
    lower: builtins.float
    """The lower bound."""
    upper: builtins.float
    """The upper bound."""
```

src/frequenz/sdk/microgrid/component_data.py
jakub-toptal: Technically, it makes sense to use `@classmethod` and `cls` as the first argument, but do you really see it coming in the near future that someone will be subclassing these? I thought we wanted to agree on these data classes so that everyone sticks to them and not subclasses them.

I can add it, but somehow adding a 3rd party package just to have a type annotation in 4 places seems a bit like an overkill.

src/frequenz/sdk/microgrid/component_data.py
jakub-toptal: Because the `Attributes` section isn't enforced by our checkers. Just look at other modules in the SDK—it is not there.

I didn't even know there's `Attributes` section, I just copied the docstrings from protobuf messages in the following format `attr: docstring`.

src/frequenz/sdk/microgrid/component_data.py
jakub-toptal: I merged these two docstrings:
https://github.com/frequenz-floss/frequenz-api-microgrid/blob/v0.x.x/proto/frequenz/api/microgrid/common.proto#L60

https://github.com/frequenz-floss/frequenz-api-microgrid/blob/v0.x.x/proto/frequenz/api/microgrid/common.proto#L15

src/frequenz/sdk/microgrid/component_data.py
jakub-toptal: Another thing is that the `Attributes` section will not work for us here, because we use `dataclasses`, which provide implicit constructor.

The `Attributes` section expect attributes to be declared explicitly in the `__init__` method like so:
```
class BatteryData:
    """A wrapper class for holding battery data.
   
    Attributes:
         soc: battery's overall SoC in percent (%).
    """

    def __init__(self):
        self.soc = 1
```

src/frequenz/sdk/microgrid/component_data.py
sahas-subramanian-frequenz: The protobuf structs are organized differently,  with a few structs that can be reused for everything,  so it has to provide generic documentation.  but in the sdk, we have specific fields,  like `soc_lower_bound`,  so we can explain exactly what they are.

jakub-toptal: Sure thing. You probably know better which attributes can be explained more precisely. Would you mind pointing these out and providing the correct documentation so that I can add it to the PR?

src/frequenz/sdk/microgrid/component_data.py
sahas-subramanian-frequenz: Capacity is always in `Wh` (Watt-hour)

src/frequenz/sdk/microgrid/component_data.py
sahas-subramanian-frequenz: Yup,  will do soon.

sahas-subramanian-frequenz:
Possible field descriptions:

## battery
soc_lower_bound: the SoC below which discharge commands will be blocked by the system.
soc_upper_bound: the SoC above which charge commands will be blocked by the system.
power_lower_bound: the maximum discharge power, in Watts, represented in the passive sign convention.  This will be a negative number, or zero if no discharging is possible.
power_upper_bound: the maximum charge power, in Watts, represented in the passive sign convention.  This will be a positive number, or zero if no charging is possbile.

## inverter
active_power_lower_bound: the maximum discharge power, in Watts, represented in the passive sign convention.  This will be a negative number, or zero if no discharging is possible.
active_power_upper_bound: the maximum charge power, in Watts, represented in the passive sign convention.  This will be a positive number, or zero if no charging is possible.

## everywhere
active_power: the 3-phase active power, in Watts, represented in the passive sign convention.

src/frequenz/sdk/microgrid/component_data.py
sahas-subramanian-frequenz: I made a separate comment.

src/frequenz/sdk/microgrid/component_data.py
leandro-lucarella-frequenz: Well, it is with this code:

```py
>>> from abc import ABC
>>> from dataclasses import dataclass
>>> @dataclass(frozen=True)
... class ComponentData(ABC):
...     pass
... 
>>> c = ComponentData()
>>> c
ComponentData()
```

Even just using a plain `ABC` works:
```py
>>> class A(ABC):
...     pass
... 
>>> a = A()
>>> a
<__main__.A object at 0x7f9a73ecd550>
```

It only fails if you define an `@abstract` method in the base class:

```py
>>> class B(ABC):
...     @abstractmethod
...     def b(self):
...             pass
... 
>>> b = B()
  File "<stdin>", line 1, in <module>
TypeError: Can't instantiate abstract class B with abstract methods b
```

I think the intention behind [`ABC`](https://docs.python.org/3/library/abc.html) is much more complicated than just avoiding instantiation (it allows overriding `issubclass()` and other advanced tricks).

Ah, but I just saw there is also an `@abstractclassmethod` so if you define that, then instantiation fails if any class in the hierarchy is missing it:

```py
>>> @dataclass(frozen=True)
... class ComponentData(ABC):
...     i: int
...     @abstractclassmethod
...     def from_proto(cls):
...             ...
... 
>>> c = ComponentData()
  File "<stdin>", line 1, in <module>
TypeError: Can't instantiate abstract class ComponentData with abstract methods from_proto
```

So adding the `@abstractclassmethod` should be enough :+1:

src/frequenz/sdk/microgrid/component_data.py
leandro-lucarella-frequenz: I would add it because:

* `typing_extension` is not 3rd party, is part of the Python foundation, it's just backporting of various typing improvements available in future Python versions, so it is maintained, very limited in scope, and temporary (once we have a sufficiently new Python version as the lowest Python version supported we can drop it), is not really a library adding new features.
* I don't think it will be used only here, Python typing system is still being stabilized, and new typing tools were added on each Python version. I think `Self` is super useful in many other places, and it also provides [`@dataclass_transform()`](https://peps.python.org/pep-0681/), which it might be the *right way* to fix the typing of the `@actor` decorator (I'm not 100% sure, I just looked at it quickly).
* About using the `@staticmethod`, it's true that it will probably not happen soon that we have a problem with it, but if it happens it might be a hard-to-catch issue, that might explode in weird ways, as types won't match but all will be a subclass of `ComponentData`, so I would rather avoid it even if we don't want to add `typing_extensions` (I think having the wrong type hint is better than returning a wrong type).

src/frequenz/sdk/microgrid/component_data.py
leandro-lucarella-frequenz: Oh, that's quite crappy, the style guide clearly allows documenting attributes that are not part of the the `__init__()`, it is in the example.

BTW, did you really tried this? Because I just did and it worked for me (based on the old version):

```diff
diff --git a/src/frequenz/sdk/microgrid/component_data.py b/src/frequenz/sdk/microgrid/component_data.py
index ffacc35..7d94e9f 100644
--- a/src/frequenz/sdk/microgrid/component_data.py
+++ b/src/frequenz/sdk/microgrid/component_data.py
@@ -18,7 +18,12 @@ from .component_states import EVChargerCableState
 
 
 class _BaseComponentData(ABC):
-    """A private base class for strongly typed component data classes."""
+    """A private base class for strongly typed component data classes.
+
+    Attributes:
+        component_id: The component ID
+
+    """
 
     _raw: microgrid_pb.ComponentData
 
```

```console
$ nox -s docstrings
nox > Running session docstrings
nox > Creating virtual environment (virtualenv) using python3.8 in .nox/docstrings
nox > python -m pip install pydocstyle darglint toml
nox > pydocstyle src tests examples benchmarks
nox > darglint src examples
nox > Session docstrings was successful.
```

src/frequenz/sdk/microgrid/component_data.py
leandro-lucarella-frequenz: To the SoC bounds I would still mention explicitly that it is also in percent to avoid any possibility of misinterpretation.

src/frequenz/sdk/microgrid/component_data.py
```diff
@@ -17,229 +19,209 @@
...
-    def soc_lower_bound(self) -> float:
-        """Get SoC lower bound.
+    soc: battery's overall SoC in percent (%).
 
-        Returns:
-            Minimum SoC level.
-        """
-        return self._raw.battery.data.soc.system_bounds.lower
+    soc_lower_bound: the SoC below which discharge commands will be blocked by the
+        system.
```
leandro-lucarella-frequenz: I would still add this is in percentage also explicitly to avoid any chance of confusion (and the same for all bounds).

leandro-lucarella-frequenz (commented):
Also minor, but I would add a `(W)` for Watt too (as in `in Watt (W)`), like with Volt and Ampere, just for consistency.

src/frequenz/sdk/microgrid/component_data.py
leandro-lucarella-frequenz: BTW, if we really want to avoid people instantiating `ComponentData` it might be a good idea to add a test for it, so it doesn't accidentally break :)

jakub-toptal: Good idea, I am adding a test. The good thing is that `mypy` will catch all instantiations of abstract classes:
```
************* Module tests.test_microgrid.test_component_data
tests/test_microgrid/test_component_data.py:20:8: E0110: Abstract class 'ComponentData' with abstract methods instantiated (abstract-class-instantiated)
```

src/frequenz/sdk/microgrid/component_data.py
jakub-toptal: You're right. It was `PyCharm` that was reporting a warning, but `darglint` works. Will add the `Attributes:` section.

image

leandro-lucarella-frequenz: Cool. We'll see when we render the docs what happens...

src/frequenz/sdk/microgrid/component_data.py
jakub-toptal: Done.

src/frequenz/sdk/microgrid/component_data.py
jakub-toptal: Added unit description in the docstring.

src/frequenz/sdk/microgrid/component_data.py
jakub-toptal: I tried, but it looks like it is supported starting from Python 3.11. `mypy` doesn't accept it either:
image

src/frequenz/sdk/microgrid/component_data.py
```diff
@@ -17,229 +19,204 @@
...
-        return self._raw.ts.ToDatetime(tzinfo=pytz.UTC)
-
-
-class MeterData(_BaseComponentData):
-    """A wrapper class for holding meter data."""
-
-    def __init__(self, raw: microgrid_pb.ComponentData) -> None:
-        """Create a `MeterData` instance.
+    @classmethod
+    @abstractmethod
```
leandro-lucarella-frequenz, line 54: Any reason not to use `abstractclassmethod`?

jakub-toptal: Because it is deprecated since py3.3:
https://docs.python.org/3/library/abc.html#abc.abstractclassmethod

leandro-lucarella-frequenz:
Marking this as blocked because until we have a decision on how the pipeline should work, we can't really decide on this one.

jakub-toptal:
> Marking this as blocked because until we have a decision on how the pipeline should work, we can't really decide on this one.

This doesn't really affect the data pipeline. It doesn't change anything, except for making the wrapper classes for `ComponentData` dataclasses instead of normal classes with properties. It makes it more convenient to work with these.

If for some reason we decide that the representation needs to change in the context of the data pipeline, we can just change dataclass-based classes instead of property-based classes.

leandro-lucarella-frequenz:
Well, is adding more processing each time you receive a message, because now you need to construct a dataclass for each grpc message. So I agree is nicer to work with in general but if we don't expect users to work with them directly, there is little value in adding the extra processing. 

And it does change something, how these objects are constructed, so it is even a breaking change (although I'm not sure if anybody else is using it).

It is IMHO an unnecessary change if these classes won't be omnipresent in the data pipeline or higher levels, this is why I think we should keep it on hold for now.

leandro-lucarella-frequenz:
The documentation improvements are quite nice though, that might also lean the balance towards merging it anyways :D

@ela-kotulska-frequenz @sahas-subramanian-frequenz any opinions on this? Do you know if these classes are constructed somewhere else outside of the SDK (so if the breaking change has any impact)?

jakub-toptal:
@ela-kotulska-frequenz @sahas-subramanian-frequenz please let us know what you think. If this is not a breaking change for our current SDK users, I'd go for it, because it makes it easier to work with, especially in tests. New classes don't need to be provided a protobuf message with raw data. Look at this example:

https://github.com/frequenz-floss/frequenz-sdk-python/blob/v0.x.x/tests/utils/data_generation.py#L28-L81

we have to create a quite complex protobuf message to get an instance of `BatteryData`. With the updated classes we could simply use something along these lines:
```
BatteryData(
    component_id=1,
    timestamp=ts,
    soc=1.0,
    soc_lower_bound=-5.0,
    soc_upper_bound=5.0,
    capacity=10.0,
    power_lower_bound=-20.0,
    power_upper_bound=20.0,
)
```

thomas-nicolai-frequenz:
> If this is not a breaking change for our current SDK users

We can always break things! Please don't take existing code into account. Its not important.

sahas-subramanian-frequenz:
> because now you need to construct a dataclass for each grpc message

Performance wise, they are both equal.  But it does make writing tests easier.  But we're losing the ability to get unexported fields from just `bat._raw.apparant_power`,  and not wait for next release.

If we drop `_raw`,  we'll lose that flexibility.  Maybe we keep a `_raw: Optional[ComponentData]`,  but also add a constructor so there are local fields?

src/frequenz/sdk/microgrid/component_data.py
```diff
@@ -6,240 +6,204 @@
...
-    _raw: microgrid_pb.ComponentData
+    component_id: int
+    timestamp: datetime
+    _raw: Optional[microgrid_pb.ComponentData] = field(default=None, init=False)
 
-    @property
-    def component_id(self) -> int:
-        """Return the component id of the message.
+    @staticmethod
+    def _set_raw(obj: ComponentData, raw: microgrid_pb.ComponentData) -> None:
```
sahas-subramanian-frequenz: Why does this have to be a static method?

It is cleaner to make it an instance method that takes `self`.

ela-kotulska-frequenz:
Could you convert it to `draft`?

src/frequenz/sdk/microgrid/component_data.py
```diff
@@ -6,240 +6,204 @@
...
-        """Create a `MeterData` instance.
+        Keep the dataclass frozen, but bypass it here to store the raw protobuf
+            message for in case it was needed later.
 
         Args:
+            obj: object to add the raw protobuf message to
             raw: raw component data as decoded from the wire.
         """
-        self._raw = raw
+        obj.__dict__["_raw"] = raw
```
ela-kotulska-frequenz: Why like this?
Why raw can't be passed in constructor?

src/frequenz/sdk/microgrid/component_data.py
```diff
@@ -52,7 +57,18 @@ def from_proto(cls, raw: microgrid_pb.ComponentData) -> ComponentData:
 
 @dataclass(frozen=True)
 class MeterData(ComponentData):
-    """A wrapper class for holding meter data."""
+    """A wrapper class for holding meter data.
+
+    Attributes:
+        active_power: the 3-phase active power, in Watts, represented in the passive
+            sign convention.
```
ela-kotulska-frequenz, line 69: This should also be:
            +ve power consumption, away from the grid.
            -ve power supply into the grid.

src/frequenz/sdk/microgrid/component_data.py
jakub-toptal: Because the idea was to exclude it from the constructor. If you want to instantiate from raw protobuf message, you would you `BatteryData.from_proto(raw)` instead.

src/frequenz/sdk/microgrid/component_data.py
jakub-toptal: Done

src/frequenz/sdk/microgrid/component_data.py
jakub-toptal: Done

src/frequenz/sdk/microgrid/component_data.py
jakub-toptal: Ok, I will add it (it is indeed documented like that in the protobuf message), but I was following the docstrings proposed above:
https://github.com/frequenz-floss/frequenz-sdk-python/pull/4#issuecomment-1267086625

src/frequenz/sdk/microgrid/component_data.py
sahas-subramanian-frequenz: `_raw` has to be optional, because we want to use these structs for the output from `resampling actor`, and it would make tests easier.

We also want to put `_raw` in the base class (`ComponentData`). Because these are all data classes with default `__init__` methods,   If we do that and take it as an optional constructor parameter,  it means all derived classes will have broken default constructors i.e params with default arguments appearing before params without default arguments.

The easy workaround for all of these is to have a special method for setting the `_raw` param.

leandro-lucarella-frequenz: This workaround should be very well documented, because is a hack that on the first sight look terribly wrong, is not trivial to realize why this is needed.

src/frequenz/sdk/microgrid/component_data.py
leandro-lucarella-frequenz: Can you create an issue in https://github.com/frequenz-floss/frequenz-api-microgrid (or even a PR)? If we see things that are wrong in the microgrid API, we should fix it!

src/frequenz/sdk/microgrid/component_data.py
```diff
@@ -6,240 +6,243 @@
...
+        component_id: the ID identifying this component in the microgrid
+        timestamp: the timestamp of when the data was measured.
+    """
 
-    @property
-    def component_id(self) -> int:
-        """Return the component id of the message.
+    component_id: int
+    timestamp: datetime
+    _raw: Optional[microgrid_pb.ComponentData] = field(default=None, init=False)
```
leandro-lucarella-frequenz: Please add a comment here explaining why we have the `_raw` data here, why is it optional and why is `init=False` being used.

src/frequenz/sdk/microgrid/component_data.py
```diff
@@ -6,240 +6,243 @@
...
-        return self._raw.ts.ToDatetime(tzinfo=pytz.UTC)
-
-
-class MeterData(_BaseComponentData):
-    """A wrapper class for holding meter data."""
-
-    def __init__(self, raw: microgrid_pb.ComponentData) -> None:
-        """Create a `MeterData` instance.
+        Keep the dataclass frozen, but bypass it here to store the raw protobuf
+            message for in case it was needed later.
```
leandro-lucarella-frequenz: Here also please be a bit more verbose, mention which alternatives we considered and why we decided for this approach.

src/frequenz/sdk/microgrid/component_data.py
leandro-lucarella-frequenz: Also thinking more about this, I think `_raw` should be a public member (`raw`). Even when it's a bit hidden, it is intended to be used by users, is not something internal.

src/frequenz/sdk/microgrid/component_data.py
```diff
@@ -6,240 +6,243 @@
...
-class _BaseComponentData(ABC):
-    """A private base class for strongly typed component data classes."""
+@dataclass(frozen=True)
+class ComponentData(ABC):
+    """A private base class for strongly typed component data classes.
 
-    _raw: microgrid_pb.ComponentData
+    Attributes:
+        component_id: the ID identifying this component in the microgrid
+        timestamp: the timestamp of when the data was measured.
```
leandro-lucarella-frequenz, line 28: Document `raw` too.

src/frequenz/sdk/microgrid/component_data.py
```diff
@@ -6,240 +6,243 @@
...
 
-        Returns:
-            Component id of the message.
-        """
-        return self._raw.id
-
-    @property
-    def timestamp(self) -> datetime:
-        """Return the timestamp of the message.
+    def _set_raw(self, raw: microgrid_pb.ComponentData) -> None:
```
leandro-lucarella-frequenz, line 41: This should stay protected, as we don't want users calling this themselves. For the users `raw` is still read-only.

leandro-lucarella-frequenz (commented):
Just a few minor last comments. Unblocking as the general feeling is this is worth it despite what happens with the individual metrics.

src/frequenz/sdk/microgrid/component_data.py
jakub-toptal: @leandro-lucarella-frequenz I think microgrid API is all good. The docstrings are complete there:
https://github.com/frequenz-floss/frequenz-api-microgrid/blob/v0.x.x/proto/frequenz/api/microgrid/common.proto#L214-L218

What Ela pointed out was only missing here. (except if we want to also add this part `represented in the passive sign convention.` to the docstring in microgrid-api, but I am not sure if it is needed.

src/frequenz/sdk/microgrid/component_data.py
jakub-toptal: I know that we don't want users to call this method, but what do you mean by `protected`? We cannot really make this method inaccessible to the users. The only thing we can do is to make it a private method, which it already is.

src/frequenz/sdk/microgrid/component_data.py
jakub-toptal: Added more verbose documentation.

src/frequenz/sdk/microgrid/component_data.py
jakub-toptal: Makes sense. Changed to a public member.

src/frequenz/sdk/microgrid/component_data.py
jakub-toptal: Done. Let me know if that's enough or sth else should be mentioned.

src/frequenz/sdk/microgrid/component_data.py
jakub-toptal: Done

ela-kotulska-frequenz (approved):

src/frequenz/sdk/microgrid/component_data.py
leandro-lucarella-frequenz: Great, thanks!

src/frequenz/sdk/microgrid/component_data.py
leandro-lucarella-frequenz: LGTM, thanks!

src/frequenz/sdk/microgrid/component_data.py
leandro-lucarella-frequenz: just using a  leading `_`, that's the convention in python. :)

I was just clarifying that even if `raw` should drop the `_`, this method shouldn't.

PR #14, opened 2022-10-04
Resampling actor

### LINKED ISSUE #5: Implement a data resampling actor
We need to implement a data resampling actor that takes care of receiving data from the microgrid and resamples it.

### Problem

The data can come from the microgrid at different intervals for different metrics. The data could potentially even come out of order, or there can be missing samples. Also formulas can be calculated using samples at different times for different metrics. Dealing with this is not easy so it shouldn't be done by applications or other use case actors directly.

### Proposed solution

Create a new resampling actor (`MicrogridDataResamplingActor`) that takes care of resampling metrics timeseries data from the `MicrogridApiClient`, so the resulting timeseries are separated by a fixed amount of time.

* The resampling should be done in a separate class (`MetricsGroupResampler`)
  * The actor should just be a thin wrapper over this resampling class.
  * The `MetricsGroupResampler` should use another class, `MetricResampler`, to handle each component's resampling.
  * All the requirements in this issue apply to the `MetricsGroupResampler` class (unless explicitly said otherwise).
  * The class will receive a `MicrogridApiClient` in the constructor to use as source for metrics data.

* The resampling rate should be configurable in the constructor (and there is only one for all the metrics being resampled).

* We start with a fixed size ring buffer for each component, storing the last samples that were received (`MetricResampler`).

* There is a timer configured to fire a new sample needs to be produced.

* There is a (configurable via the constructor) `maximum_data_age` that is the threshold to consider the data in the ring buffer relevant. Data older than `maximum_data_age` is considered irrelevant.

* When the timer fires, if there is no relevant data in the ring buffer, a special value `NaN` should be sent.

* When the timer fires, if there is relevant data in the ring buffer, the resampling is done by feeding the contents of the relevant data in the ring buffer to a resampling function (`MetricResampler`).
  * The resampling function should be configurable in the constructor
  * We'll use average by default for now
  * The resampling function signature should be: `resampling(samples: Sequence[(datetime, float)], resampling_period_s: float)`
    * This way it is possible for the resampling function to split samples into resampling windows if necessary.

* The data sent for a particular sampling window should have the **exact same timestamp** (i.e. the timestamp needs to be generate once per sampling window and reused in all outgoing messages for that sampling window).

* The timestamp should be the one at the end of the window (so the instant where the resampling timer fired). So if the current sampling window we are about to send if from 12:00:00 to 12:00:05 then the output timestamp should be 12:00:05.

* For now all channels to get resampled data will need to be created at the beginning of the application

#### Utils

```py
# probably in `frequenz.sdk.resampling`, but maybe Sample can go to a more general place

@dataclass
class ComponentMetricKey:
    metric_id: str  # Could be an enum or some ID
    component_id: int

def channel_name(prefix: str, metric: metric):
    return f'{prefix}_{metric.component_id}_{metric.metric_id}'

@dataclass
class Sample:
	timestamp: datetime
	value: float

ResamplingFunction = Callable[(samples: Sequence[Sample], resampling_period_s: float), float]
```

```py
# probably in `frequenz.sdk.core`

T = TypeVar('T')

# Look if python already offers some ring buffer implementation
class RingBuffer(Generic[T]):

    def __init__(self, size: int):
        ...

    def add(self, item: T) -> None:
        ...

    @property
    def items(self) -> Sequence[T]:
        ...

    # Future
    def resize(self, new_size: int) -> None:
        ...
```

#### `MetricResampler`

```py
# probably in `frequenz.sdk.resampling`
class MetricResampler:
    # Future:
    # * Calculate input characteristics: rate, volatily, min, max, etc.
    # * Dynamically decide on:
    #   * Buffer size
    #   * Resampling function to use
    _buffer: RingBuffer[Sample]
    _period_s: float
    _max_data_age_in_periods: float
    _function: ResamplingFunction

    def __init__(self, *,
            period_s: float,
            max_data_age_in_periods: float,
            function: ResamplingFunction,
            ):
        # * calculate buffer size based on the resampling_period_s and
        #   max_data_age_in_periods
        ...

    def add_sample(self, sample: Sample) -> None:
        ...

    def resample(self) -> float:
        ...
```

#### `MetricsGroupResampler`

```py
# probably in `frequenz.sdk.resampling`
class MetricsGroupResampler:
    _resamplers: Dict[ComponentMetricKey, MetricResampler]

    def __init__(self, *,
            period_s: float,
            max_data_age_in_periods: float,
            function: ResamplingFunction,
            ):
        # create the resamplers dict

    def add_metric(self, key: ComponentMetricKey) -> None:
        # create a metric resampler if it doesn't already exist and add it to self.metrics

    def remove_metric(self, key: ComponentMetricKey) -> None:
        # remove a metric from self.resamplers

    def add_sample(self, key: ComponentMetricKey, sample: Sample) -> None:
        # forward to MetricResampler.add_sample()

    def resample(self)-> dict[str, Sample]: # channel_id, sample
        timestamp = datetime.now()
        # loop over all metrics and call resample() and attach the timestamp to
        # create a new sample for each returned value
```

#### `MicrogridDataResamplingActor`

**NOTE:** How we create and pass input and output channels still needs to be decide and might change if we can do dynamic subscriptions.

```py
# probably in `frequenz.sdk.resampling` with an alias in `frequenz.sdk.actor`
class MicrogridDataResamplingActor:
    _resampler: MetricsGroupResampler
    _sampling_timer: Timer
    _input_receivers: Dict[ComponentMetricKey, Receiver[Sample]]
    _output_channels: Dict[ComponentMetricKey, Channel[Sample]]

    def __init__(self, input_receivers: Dict[ComponentMetricKey, Receiver[Sample]], *,
            start_time: Optional[datetime] = None,  # None means NOW, so only live data
            period_s: float = 0.2,
            max_data_age_in_periods: float = 3,
            function: ResamplingFunction = average,
            ):
        # create timer
        self._resampler = MetricsGroupResampler(period_s, function,
                metrics_buffer_size, max_data_age_in_periods)

    def subscribe(self, key: ComponentMetricKey) -> Receiver[Sample]:
        # if metric not in _output_channels:
        #    * (NOT NOW, in the future) create a new receiver from DataSourcingActor for the metric and add it to
        #      _input_receivers.
        #      We will need to keep an index of component_id because metrics for a single
        #      component come packed in one single object in the MicrogridApiClient (and the
        #      gRPC interface), the stream will receive all the metrics for a component_id.
        #    * create a new _output_channels for the metric
        # return a new receiver from _output_channels[metric]

    # Future
    def unsubscribe(self, subscription: ?) -> None:
        # TODO: Can this be done?

    def run(self):
        # for event in selec(timer, metric_receivers):
        # 	if event == timer:
        # 		call self.resampler.resample() and put the output into the output channels
        #   elif event == metric_receiver:
        # 		call self.resampler.add_sample()
```

#### Example usage

This assumes actors don't start when instantiated, but maybe it is possible to do this too even if they are already running.

```py
async def app_run() -> None:

    resampling_actor = ...(input_receivers, ...)

    battery_pool_actor = ...()
    for metric in battery_pool_actor.metrics:
        battery_pool_actor.attach(metric, resampling_actor.subscribe(metric))

    await actors_run(resampling_actor, battery_pool_actor)
```

### Questions

* Should we read directly from `MicrogridApiClient` or should we create a new `MicrogridDataReadingActor` or interface that just passes the data untouched?

  **Answer:** For now we use the `MicrogridApiClient` directly, but if possible (and makes sense) we should also extract this from `FormulaCalculator` and try to have the same subscription interface in the future we might split it out as an independent actor if message passing overhead is not an issue. See next steps, as it is related to the `DataSource` layer, see *Next steps*.

* Should a `NaN` be produced by the resampling class or by the resampling function? The former is more consistent and the latest is more flexible.

  **Answer:** `NaN` will be produced by the resampling class for consistency and simplicity, as we don't foresee any use cases where the resampling function will have to make that decision.

* Should we send one result per metric or one result with all the metric. The former makes it more complicated for users to compile the data for one sampling window (it will have to wait until the data for all metrics arrive and look at the timestamp to group them properly) but the later could send a lot of unneeded data if some user only needs the data for a couple of metrics. In both cases we'll need a timeout (smaller than the resampling rate probably) while receiving to detect if the resampling actor is broken or the messages channel is congested or something.

  If we want the output to have a common interface with the microgrid API we probably don't have other choice than to output individual component data in individual channels, but maybe this is a case for having the resampling actor a special case?

  **Answer:** We send one metric per channel to keep it simple (for the resampling actor) for now.

### Next steps

Not to do as part of this issue, but to have a more long term sight when implementing this.

* Infer the data rate of every individual input (#55):
  * Use the input timestamps to estimate the rate.
  * Use dynamic ring buffer sizes depending on the input data rate, so components with higher data rates can get bigger buffers (up to a configurable maximum) and lower data rates can get smaller buffers.
  * The input data rate could be used also to automatically pick a good resampling function (we can know, for example, if we are up or downsampling).

* ~Provide some helper function to create channels for metrics in batch~ (we'll see it if/when the need arrives)
  * Useful for providing standard channel names too
  * For example a function taking a set of tuples `(component_id, metric)` or a function taking a set of `(MicrogridApiClient, component_category, metric)` to create channels for some metric of all components of a particular category.

* Automatically chose a good resampling function based on metrics characteristics, like volatility (or input data rate, when available). (#57)

* ~Add dynamic metric subscription, the `MicrogridDataResamplingActor` will accept subscription request through a control channel and will return a channel for the subscribed metric to the use case actor requesting it.~ (done)

* ~Review `MicrogridApiClient` interface in the broader context of a common data ingestion interface that can be shared among, for example, `ReportingApiClient` and `LocalStorageClient`.~ (this is the data source actor: #30)

### Resources
  
image

image

jkozlowicz:
Initial implementation of the `ResamplingActor`.

- no reference to the `ComponentGraph` anymore as it is not needed for resampling
- `RingBuffer` will keep data points per component in sorted order and discard the oldest one to make room for the new one
- extrapolation logic to be provided from outside of `ResamplingActor` (conformance to `Resampler` interface is required, not sure if this is how it should look like, but didn't want to pass 4 separate methods to the `ResamplingActor` constructor)
- `DataPoint` class to be replaced by the `ComponentData` class from https://github.com/frequenz-floss/frequenz-sdk-python/pull/4

### DIFF
--- a/src/frequenz/sdk/data_ingestion/resampling/__init__.py
+++ b/src/frequenz/sdk/data_ingestion/resampling/__init__.py
@@ -1,0 +1,9 @@
+"""
+Tools for resampling streamed component data.
+
+Copyright
+Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+
+License
+MIT
+"""
--- a/src/frequenz/sdk/data_ingestion/resampling/resampler.py
+++ b/src/frequenz/sdk/data_ingestion/resampling/resampler.py
@@ -1,0 +1,170 @@
+"""
+Interface for a class that will implement the resampling logic.
+
+Copyright
+Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+
+License
+MIT
+"""
+import logging
+from abc import ABC, abstractmethod
+from datetime import datetime
+from typing import List
+
+from frequenz.sdk.data_ingestion.resampling.ring_buffer import DataPoint
+from frequenz.sdk.microgrid import Component, ComponentCategory
+
+logger = logging.Logger(__name__)
+
+
+class Resampler(ABC):
+    """Resampler interface."""
+
+    def resample_component_data(  # pylint: disable=too-many-arguments
+        self,
+        component: Component,
+        data_points: List[DataPoint],
+        resampling_window_start: datetime,
+        resampling_frequency: float,
+        max_component_data_delay: float = 5.0,
+    ) -> DataPoint:
+        """Delegate resampling of data points based on the component category.
+
+        Args:
+            component: component that the data points belong to
+            data_points: data points for a single battery from the ring buffer.
+            resampling_window_start: timestamp of the beginning of the resampling window.
+            resampling_frequency: what frequency should the original data points be
+                converted to
+            max_component_data_delay: amount of time in seconds describing how much
+                delayed data points in the ring buffer for `component_d` can be to still
+                be considered accurate enough for resampling
+
+        Returns:
+            Data points for a given battery resampled into a single sample.
+
+        Raises:
+            ValueError: if category of any component in the microgrid is not supported
+        """
+        if component.category == ComponentCategory.BATTERY:
+            return self.resample_battery_data(
+                data_points=data_points,
+                resampling_window_start=resampling_window_start,
+                resampling_frequency=resampling_frequency,
+                max_component_data_delay=max_component_data_delay,
+            )
+        if component.category == ComponentCategory.METER:
+            return self.resample_meter_data(
+                data_points=data_points,
+                resampling_window_start=resampling_window_start,
+                resampling_frequency=resampling_frequency,
+                max_component_data_delay=max_component_data_delay,
+            )
+        if component.category == ComponentCategory.INVERTER:
+            return self.resample_inverter_data(
+                data_points=data_points,
+                resampling_window_start=resampling_window_start,
+                resampling_frequency=resampling_frequency,
+                max_component_data_delay=max_component_data_delay,
+            )
+        if component.category == ComponentCategory.EV_CHARGER:
+            return self.resample_ev_charger_data(
+                data_points=data_points,
+                resampling_window_start=resampling_window_start,
+                resampling_frequency=resampling_frequency,
+                max_component_data_delay=max_component_data_delay,
+            )
+        raise ValueError(f"Unsupported component category: {component.category}")
+
+    @abstractmethod
+    def resample_battery_data(
+        self,
+        data_points: List[DataPoint],
+        resampling_window_start: datetime,
+        resampling_frequency: float,
+        max_component_data_delay: float = 5.0,
+    ) -> DataPoint:
+        """Resample battery data.
+
+        Args:
+            data_points: Data points for a single battery from the ring buffer.
+            resampling_window_start: timestamp of the beginning of the resampling window.
+            resampling_frequency: what frequency should the original data points be
+                converted to
+            max_component_data_delay: amount of time in seconds describing how much
+                delayed data points in the ring buffer for `component_d` can be to still
+                be considered accurate enough for resampling
+
+        Returns:
+            Data points for a given battery resampled into a single sample.
+        """
+
+    @abstractmethod
+    def resample_meter_data(
+        self,
+        data_points: List[DataPoint],
+        resampling_window_start: datetime,
+        resampling_frequency: float,
+        max_component_data_delay: float = 5.0,
+    ) -> DataPoint:
+        """Resample meter data.
+
+        Args:
+            data_points: Data points for a single meter from the ring buffer.
+            resampling_window_start: timestamp of the beginning of the resampling window.
+            resampling_frequency: what frequency should the original data points be
+                converted to
+            max_component_data_delay: amount of time in seconds describing how much
+                delayed data points in the ring buffer for `component_d` can be to still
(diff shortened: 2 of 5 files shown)


jakub-toptal:
I am closing this PR as it was based on a now outdated data pipeline diagram. Leaving it for reference, if we wanted to compare it with what will be implemented.

The new implementation for the resampling actor will be implemented in a separate PR as agreed. 
FYI @leandro-lucarella-frequenz @sahas-subramanian-frequenz

PR #25, opened 2022-10-12
Implement ChannelRegistry

### LINKED ISSUE #24: Creating output channels dynamically for the resampling actor
### What's needed?

The `Resampling` actor needs to send out dynamically created channel receivers (corresponding to dynamic component IDs) to use-case actors.  We don't have a way to do this currently.

### Proposed solution

The channel registry that was proposed a few months back would be a good solution to this problem.  It would work as follows:

We need a `ChannelRegistry` class that would create and own channels and provide channel senders and receivers to all actors that ask for them.  It would have `get_sender(chan_name)` and `get_receiver(chan_name)` methods that provide access to an existing channel,  or implicitly create one if a channel with the given name doesn't exist.

In single process contexts,  `ChannelRegistry` would hold a dictionary of channels,  and in multi-process contexts,  it would use an external service like redis or nats,  whichever we use for implementing multi-process channels.

In each process,  the `ChannelRegistry` instance can be accessed through the top level `MicrogridApi` module.

The `Resampling` actor would accept a channel receiver in its constructor,  through which external actors can send requests to it.  Requests would contain a component ID and a metric name.  The resampling actor and the external actor would construct a channel name based on the requested id and metric name,  and go to the `ChannelRegistry` to ask for a sender or receiver respectively,  and thus establish communication using `ChannelRegistry` as a broker.

If the resampling actor is already handling a newly requested metric,  it would discard the request.

### Use cases

This can be used in all cases where actors need to dynamically create channels.

_No response_

### Additional context

_No response_

sahas-subramanian-frequenz:
Closes #24

### DIFF
--- a/src/frequenz/sdk/actor/channel_registry.py
+++ b/src/frequenz/sdk/actor/channel_registry.py
@@ -1,0 +1,52 @@
+"""A class that would dynamically create, own and provide access to channels.
+
+Copyright
+Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+
+License
+MIT
+"""
+
+from typing import Any, Dict, Generic, TypeVar
+
+from frequenz.channels import Broadcast, Receiver, Sender
+
+T = TypeVar("T")
+
+
+class ChannelRegistry(Generic[T]):
+    """Dynamically creates, own and provide access to channels.
+
+    It can be used by actors to dynamically establish a communication channel
+    between each other.  Channels are identified by hashable key objects.
+    """
+
+    def __init__(self) -> None:
+        """Create a `ChannelRegistry` instance."""
+        self._channels: Dict[T, Broadcast[Any]] = {}
+
+    def get_sender(self, key: T) -> Sender[Any]:
+        """Get a sender to a dynamically created channel with the given key.
+
+        Args:
+            key: A key to identify the channel.
+
+        Returns:
+            A sender to a dynamically created channel with the given key.
+        """
+        if key not in self._channels:
+            self._channels[key] = Broadcast(f"dynamic-channel-{key}")
+        return self._channels[key].get_sender()
+
+    def get_receiver(self, key: T) -> Receiver[Any]:
+        """Get a receiver to a dynamically created channel with the given key.
+
+        Args:
+            key: A key to identify the channel.
+
+        Returns:
+            A receiver for a dynamically created channel with the given key.
+        """
+        if key not in self._channels:
+            self._channels[key] = Broadcast(f"dynamic-channel-{key}")
+        return self._channels[key].get_receiver()
--- a/src/frequenz/sdk/actor/__init__.py
+++ b/src/frequenz/sdk/actor/__init__.py
@@ -7,6 +7,7 @@ License
 MIT
 """
 
+from .channel_registry import ChannelRegistry
 from .decorator import actor
 
-__all__ = ["actor"]
+__all__ = ["actor", "ChannelRegistry"]
--- a/tests/actor/test_channel_registry.py
+++ b/tests/actor/test_channel_registry.py
@@ -1,0 +1,30 @@
+"""Tests for the ChannelRegistry.
+
+Copyright
+Copyright © 2021 Frequenz Energy-as-a-Service GmbH
+
+License
+MIT
+"""
+
+from frequenz.sdk.actor import ChannelRegistry
+
+
+async def test_channel_registry_with_string() -> None:
+    """Tests for ChannelRegistry, with string as key type."""
+    reg = ChannelRegistry[str]()
+
+    sender20 = reg.get_sender("20-hello")
+    receiver20 = reg.get_receiver("20-hello")
+
+    sender21 = reg.get_sender("21-hello")
+    receiver21 = reg.get_receiver("21-hello")
+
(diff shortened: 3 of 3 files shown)


matthias-wende-frequenz (dismissed):

src/frequenz/sdk/actor/channel_registry.py
```diff
@@ -0,0 +1,52 @@
...
+class ChannelRegistry(Generic[T]):
+    """Dynamically creates, own and provide access to channels.
+
+    It can be used by actors to dynamically establish a communication channel
+    between eachother.  Channels are identified by hashable key objects.
+    """
+
+    def __init__(self) -> None:
+        """Create a `ChannelRegistry` instance."""
+        self._channels: Dict[T, Broadcast[Any]] = {}
```
leandro-lucarella-frequenz, line 26: It might make more sense to parametrize over the type of channel (instead of hardcoding it to `Broadcast`).

Even when this probably only makes sense for `Broadcast` for now, it might change in the future. The ideal would be to have `Broadcast` as default, but defaults for generics will come in Python 3.12 ([PEP 696](https://peps.python.org/pep-0696/)).

This said, I'm fine with hardcoding `Broadcast` for this first version too. I did some experiments and it's not so easy to parametrize on channel type.

Also using `Any` as the channel's value circumvents the type system, it would be nice to find a way to have this more typed. I think this one is more important. Definitely more important than the dict key type.

tiyash-basu-frequenz (dismissed):

src/frequenz/sdk/actor/channel_registry.py
```diff
@@ -0,0 +1,52 @@
...
+        """Get a sender to a dynamically created channel with the given key.
+
+        Args:
+            key: A key to identify the channel.
+
+        Returns:
+            A sender to a dynamically created channel with the given key.
+        """
+        if key not in self._channels:
+            self._channels[key] = Broadcast(f"dynamic-channel-{key}")
```
leandro-lucarella-frequenz, line 38: I would add an argument to the constructor to set this prefix instead of having it hardcoded. We might want to have multiple registries

src/frequenz/sdk/actor/channel_registry.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class ChannelRegistry(Generic[T]):
+    """Dynamically creates, own and provide access to channels.
+
+    It can be used by actors to dynamically establish a communication channel
+    between eachother.  Channels are identified by hashable key objects.
+    """
+
+    def __init__(self) -> None:
```
leandro-lucarella-frequenz, line 24:
```suggestion
    def __init__(self, *, channel_names_prefix: str) -> None:
```

src/frequenz/sdk/actor/channel_registry.py
```diff
@@ -0,0 +1,52 @@
...
+from frequenz.channels import Broadcast, Receiver, Sender
+
+T = TypeVar("T")
+
+
+class ChannelRegistry(Generic[T]):
+    """Dynamically creates, own and provide access to channels.
+
+    It can be used by actors to dynamically establish a communication channel
+    between eachother.  Channels are identified by hashable key objects.
```
ela-kotulska-frequenz: > between each other.

src/frequenz/sdk/actor/channel_registry.py
```diff
@@ -0,0 +1,52 @@
...
+"""
+
+from typing import Any, Dict, Generic, TypeVar
+
+from frequenz.channels import Broadcast, Receiver, Sender
+
+T = TypeVar("T")
+
+
+class ChannelRegistry(Generic[T]):
```
ela-kotulska-frequenz, line 17: Is Generic[T] needed? We would like channels to be identified by `str`, yes?

matthias-wende-frequenz: Not necessarily. A channel can identified by a pair `ComponentId` and `MetricType`. But we don't know all use cases of the Registry, so let's leave it generic.

src/frequenz/sdk/actor/channel_registry.py
sahas-subramanian-frequenz: Fixed

ela-kotulska-frequenz (approved):

src/frequenz/sdk/actor/channel_registry.py
leandro-lucarella-frequenz: I had the same question. IMHO this adds a bit of unnecessary complexity (and will force us to create different registries for different key types). I think we can require users to just convert to a string themselves (`f"{component_id}_{metric_typ}"`).

leandro-lucarella-frequenz (commented):
I made a few comments, the part I'm most concerned is the use of `Any` as the value type for the channel.

Other than that, I really think this should go to the channels library instead of the SDK, this is generally useful for anyone using channels.

At some point it would be nice to have a way to have a unique global channel registry (doing that with typing can be also challenging so I think we are fine for now with having multiple registries).

leandro-lucarella-frequenz:
Oh, damn, I was too late :(

sahas-subramanian-frequenz:
> I was too late

no, those make sense.  I will make a new PR.

PR #29, opened 2022-10-13
ChannelRegistry improvements

shsms:
1. Remove generic key support in ChannelRegistry and use only strings
2. Add a `name` parameter for the ChannelRegistry class, and use it as a channel name prefix.

### DIFF
--- a/src/frequenz/sdk/actor/channel_registry.py
+++ b/src/frequenz/sdk/actor/channel_registry.py
@@ -19,8 +19,13 @@ class ChannelRegistry:
     between each other.  Channels are identified by string names.
     """
 
-    def __init__(self) -> None:
-        """Create a `ChannelRegistry` instance."""
+    def __init__(self, *, name: str) -> None:
+        """Create a `ChannelRegistry` instance.
+
+        Args:
+            name: A unique name for the registry.
+        """
+        self._name = name
         self._channels: Dict[str, Broadcast[Any]] = {}
 
     def get_sender(self, key: str) -> Sender[Any]:
@@ -33,7 +38,7 @@ class ChannelRegistry:
             A sender to a dynamically created channel with the given key.
         """
         if key not in self._channels:
-            self._channels[key] = Broadcast(f"dynamic-channel-{key}")
+            self._channels[key] = Broadcast(f"{self._name}-{key}")
         return self._channels[key].get_sender()
 
     def get_receiver(self, key: str) -> Receiver[Any]:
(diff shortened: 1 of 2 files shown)


src/frequenz/sdk/actor/channel_registry.py
```diff
@@ -7,25 +7,23 @@
...
-from typing import Any, Dict, Generic, TypeVar
+from typing import Any, Dict
 
 from frequenz.channels import Broadcast, Receiver, Sender
 
-T = TypeVar("T")
 
-
-class ChannelRegistry(Generic[T]):
+class ChannelRegistry:
```
ela-kotulska-frequenz, line 15: Why did you decide to remove Generic?

ela-kotulska-frequenz (dismissed):

src/frequenz/sdk/actor/channel_registry.py
sahas-subramanian-frequenz: Because Luca pointed out in the other issue that we probably have multiple key types that will need to share a single registry.

Also I realized that generic key types will work only as long as we are in the single process space.  And we will have to switch to strings if we make it multi-process.

So you were right earlier :-)

tests/actor/test_channel_registry.py
```diff
@@ -12,7 +12,7 @@
 
 async def test_channel_registry() -> None:
     """Tests for ChannelRegistry, with string as key type."""
-    reg = ChannelRegistry()
+    reg = ChannelRegistry(name="test-registry")
```
leandro-lucarella-frequenz, line 15: It looks strange that you changed the way you create the registry (you changed the channel naming scheme) and you didn't have to update any tests (no tests failed). I know it is probably not worth adding a test to check the channel name, specially because it's not even exposed, but pointing it out anyway just in case.

leandro-lucarella-frequenz (dismissed):

leandro-lucarella-frequenz:
Friendly reminder to remember to assign the project if there is no linked issue :)

tests/actor/test_channel_registry.py
sahas-subramanian-frequenz: Not sure I understand.  ChannelRegistry takes a `name` argument now,  for the same use as the `channel_name_prefix` that you had proposed in the earlier PR.  so without this change,  the tests would fail.

shsms:
Rebased to latest `v0.x.x`

leandro-lucarella-frequenz (approved):

tests/actor/test_channel_registry.py
leandro-lucarella-frequenz: Yeah, I just meant that the name is not included in the tests. Not important.

PR #50, opened 2022-10-21
Add timeout to all gRPC calls and set a default timeout

### LINKED ISSUE #38: Set time limit for every call to the Microgrid API
### What's needed?

If the Microgrid API does not respond in a given time limit then we should cancel the call, otherwise the program will stall forever.

### Proposed solution

Every call to the Microgrid API should have timeout (default 60 seconds).

### Use cases

_No response_

_No response_

### Additional context

_No response_

jkozlowicz:


benchmarks/data_ingestion/benchmark_microgrid_data.py
```diff
@@ -190,14 +190,15 @@ def __init__(self) -> None:
     def component_data_peekable(self) -> None:
         """Mock component_data_peekable method on MicrogridApiClient."""
 
-    async def components(self) -> Set[Component]:
+    async def components(self, timeout: float = 60.0) -> Set[Component]:
```
leandro-lucarella-frequenz: Maybe put the `60.0` in a *constant* here too?

tests/test_microgrid/test_timeout.py
```diff
@@ -0,0 +1,171 @@
...
+from .mock_api import MockGrpcServer, MockMicrogridServicer
+
+
+async def test_components_timeout(mocker: MockerFixture) -> None:
+    """Test if the components() method properly raises AioRpcError"""
+    servicer = MockMicrogridServicer()
+
+    # pylint: disable=unused-argument
+    def mock_list_components(request: ComponentFilter, context: Any) -> ComponentList:
+        time.sleep(0.3)
```
leandro-lucarella-frequenz: What are these small sleep/timeouts, `0.3`, `0.1` and `0.5`? Are we introducing new flaky tests? :worried: 

If we are not, then we should add comments about why the sleep/timeouts are there and why that value was picked.

jakub-toptal: These are arbitrary values. The argument to `sleep` (`0.3`) must be bigger than the argument passed to `timeout` (`0.1`) so that the timeout is triggered, and the last number must be greatest of all (`0.5`), because it is used to shutdown the `MockMicrogridServicer()` once the test is done. I just picked `0.3` and `0.1`, but could be something else. I can pull them out to some constants and add a comment there.

There's one test where I had to change the numbers, because the `<grpc-method>.write()` was used and I wasn't sure how to handle that case. Maybe someone knows, otherwise I will keep digging.

src/frequenz/sdk/microgrid/client.py
```diff
@@ -396,6 +422,7 @@ async def _expect_category(
     async def meter_data(
         self,
         component_id: int,
+        timeout: float = DEFAULT_TIMEOUT,
```
ela-kotulska-frequenz: I think it should be one global default timeout - just to make sure program won't wait forever.
Passing timeout as an argument would be problematic in case of channel communication.

jakub-toptal: Why would passing the timeout as an argument be problematic in the context of channels?

It is much easier and convenient to work with it as an argument in tests, because you can just pass it directly, without having to patch it.

tests/test_microgrid/test_timeout.py
leandro-lucarella-frequenz: Yes please, constant and comments clarifying this sounds good!

tests/test_microgrid/test_timeout.py
```diff
@@ -0,0 +1,185 @@
...
+# calls will raise an AioRpcError with status code equal to DEADLINE_EXCEEDED.
+GRPC_CALL_TIMEOUT: float = 0.1
+
+# How much late a response to a gRPC call should be. It is used to trigger a timeout
+# error and needs to be greater than `GRPC_CALL_TIMEOUT`.
+GRPC_SERVER_DELAY: float = 0.3
+
+# How long a mocked Microgrid server should be running for a single test function,
+# before it gets shut down so that the tests can finish.
+GRPC_SERVER_SHUTDOWN_DELAY: float = 1.0
```
leandro-lucarella-frequenz, line 39: Since we are awaiting for the function we are testing, I wonder if we really need to wait even more afterwards to shutdown the server? Can we just shut down immediately?

tests/test_microgrid/test_timeout.py
```diff
@@ -0,0 +1,185 @@
...
+    client = MicrogridGrpcClient(grpc_channel=grpc_channel, target=target)
+
+    with pytest.raises(grpc.aio.AioRpcError) as err_ctx:
+        # To be investigated: passing a timeout a little bit smaller than the sleep
+        # time in `mock_set_bounds` should be sufficient, but it doesn't work. Making
+        # the timeout much smaller than the sleep time does work. E.g.
+        # sleep => 0.3, timeout = 0.1 or even 0.001 doesn't work, but 0.00001 works
+        await client.set_bounds(component_id=1, lower=0.0, upper=10.0, timeout=0.00001)
+    assert err_ctx.value.code() == grpc.StatusCode.DEADLINE_EXCEEDED
+    await server.stop(GRPC_SERVER_SHUTDOWN_DELAY)
```
leandro-lucarella-frequenz: So what's the error here? It's not finishing after `timeout` with `DEADLINE_EXCEEDED`? Or is there any other error. In any case, I'm not familiar with gRPC, so I would have to start debugging from scratch to help here. Not sure if @ela-kotulska-frequenz or @sahas-subramanian-frequenz are more familiar and can help.

Tests are still failing though, not sure if because of this test or something else (haven't checked the logs).

src/frequenz/sdk/microgrid/client.py
ela-kotulska-frequenz: > Why would passing the timeout as an argument be problematic in the context of channels?

1. Because many actors would subscribe for the same channel.
What should happen if one actor subscribe for `meter_data with timeout = 2` and other actor would subscribe for `meter_data with timeout=60` ?
2. It is breaking change that is not needed now. Doing breaking changes for tests is problematic.

> It is much easier and convenient to work with it as an argument in tests, because you can just pass it directly, without having to patch it.

But you can import that DEFAUTL_TIMEOUT in the tests and override it, yes?

src/frequenz/sdk/microgrid/client.py
```diff
@@ -297,7 +315,8 @@ async def _component_data_task(
             )
             try:
                 call = self.api.GetComponentData(
-                    microgrid_pb.ComponentIdParam(id=component_id)
+                    microgrid_pb.ComponentIdParam(id=component_id),
+                    timeout=timeout,
```
sahas-subramanian-frequenz: I'm wondering if this timeout for the data stream will 
  - kill the stream after the timeout (this is undesirable, but we might not notice because we have retry code to restart the connection)
  - or kill the stream only if it hasn't transmitted any data for a while.

Have you tried this out already?

src/frequenz/sdk/microgrid/client.py
sahas-subramanian-frequenz: Having it as a single value also makes it easy to configure, from the UI for example.

tests/test_microgrid/test_timeout.py
ela-kotulska-frequenz: > await client.set_bounds(component_id=1, lower=0.0, upper=10.0, timeout=0.00001

What if the call finish in this short timeout? I think we should avoid this kind of tests...
There are many workaround. For example you could mock set_bounds method to sleep for 1 sec instead...

src/frequenz/sdk/microgrid/client.py
leandro-lucarella-frequenz: Playing devil's advocate, what would be the solution if for some reason you want only one particular call to have a smaller (or bigger) timeout? If we leave this option, the default could be also be still configured via the UI.

tests/test_microgrid/test_timeout.py
jakub-toptal: I think the application would terminate immediately without the shut down, but there could be some asyncio warnings that there are pending tasks. So this is more of a graceful shutdown instead of exiting right away. I actually adopted it from other tests:
https://github.com/frequenz-floss/frequenz-sdk-python/blob/v0.x.x/tests/test_microgrid/test_client.py#L130

src/frequenz/sdk/microgrid/client.py
jakub-toptal: > 1. Because many actors would subscribe for the same channel.
What should happen if one actor subscribe for meter_data with timeout = 2 and other actor would subscribe for meter_data with timeout=60 ?

Good point. Yeah, in this case, making it global would simplify it. Otherwise, it would have to be yet another constituent of a channel. 

I think there might be cases when actors that subscribe for data that might want to use different timeouts, but probably it will not happen soon.

> But you can import that `DEFAULT_TIMEOUT` in the tests and override it, yes?

I would have to patch it as a module variable, but should be doable.

Ok, so everyone agrees that it the timeout should be implemented as a global variable instead of passed as a param for now? I'll update the code once we make a decision that is the way to go.

benchmarks/data_ingestion/benchmark_microgrid_data.py
jakub-toptal: Done

tests/test_microgrid/test_timeout.py
jakub-toptal: Done

tests/test_microgrid/test_timeout.py
jakub-toptal: Do you mean `set_bounds` on the `client` instance? If I mock `set_bounds` then there will be no gRPC call.

I am keeping the client as it is, but only mocking the server response to take long enough to trigger a timeout error in the client.

jakub-toptal: @leandro-lucarella-frequenz 
> So what's the error here? It's not finishing after `timeout` with `DEADLINE_EXCEEDED`?

Yes, exactly. It says `did not raise`. I cannot find out why, because I am mocking the `SetBounds()` method on the server instance to sleep for `0.3`:
```
    def mock_set_bounds(
        request_iterator: Iterator[SetBoundsParam], context: Any
    ) -> Empty:
        time.sleep(GRPC_SERVER_DELAY)
        return Empty()

    mocker.patch.object(servicer, "SetBounds", mock_set_bounds)
    server = MockGrpcServer(servicer, port=57809)
    await server.start()
```

and I expect the timeout error to be raised, but it doesn't happen (when I reduced the timeout to a very small number like `0.000001` it does work, but not always):
`await client.set_bounds(component_id=1, lower=0.0, upper=10.0, timeout=0.00001)`
Obviously, setting this timeout is not a good solution, but I am just trying to figure out why it behaves like this... I added a comment and asked if any of you have an idea what can be the case here.

src/frequenz/sdk/microgrid/client.py
jakub-toptal: I haven't tried it yet. Will give it a go.

src/frequenz/sdk/microgrid/client.py
jakub-toptal: Done

tests/test_microgrid/test_timeout.py
jakub-toptal: As discussed with @sahas-subramanian-frequenz, I've removed the test for testing timeout of the `SetBounds` method (because it also involves establishing a connection to the remote gRPC server and it cannot be easily tested, unless someone has an idea how to test it).

src/frequenz/sdk/microgrid/client.py
jakub-toptal: FYI @leandro-lucarella-frequenz @ela-kotulska-frequenz, we discussed this with @sahas-subramanian-frequenz and decided to remove the timeout from the methods for getting component data, because the `GetComponentData` gets killed whenever a timeout error occurs and this block is re-executed:
https://github.com/frequenz-floss/frequenz-sdk-python/blob/v0.x.x/src/frequenz/sdk/microgrid/client.py#L299-L303

and this might cause some performance issues.

The method gets killed, because grpc timeout error falls into `grpc.aio.AioRpcError `category. If we don’t want to retry on timeout errors, we could inspect the status code of that `grpc.aio.AioRpcError` and take an appropriate action, if its status is equal to `grpc.StatusCode.DEADLINE_EXCEEDED`.

The consumer keeps consuming messages like so `async for msg in receiver` without any interruption or errors.

I checked that using the test function below (it has been removed from the code, because there's no timeout applied to `GetComponentData` anymore, but leaving it for reference). What it does is:
- it mocks the `GrpcServer` to keep sending some random battery data
- sending the 5th message on the server is delayed on purpose to cause a timeout error on the client side
- the client keeps consuming messages from the stream until it consumed some messages after the timeout already occurred

```python
@patch("frequenz.sdk.microgrid.client.DEFAULT_GRPC_CALL_TIMEOUT", GRPC_CALL_TIMEOUT)
async def test_get_component_data_stream(mocker: MockerFixture) -> None:
    """Test if the component data stream survives timeout errors"""
    servicer = MockMicrogridServicer()
    servicer.add_component(1, ComponentCategory.COMPONENT_CATEGORY_BATTERY)

    delayed_msg_index = 3

    # pylint: disable=unused-argument
    async def mock_component_data(
        request: ComponentIdParam, context: Any
    ) -> AsyncGenerator[ComponentData, None]:
        counter = 0
        while True:
            if counter == delayed_msg_index:
                await asyncio.sleep(GRPC_SERVER_DELAY)
            else:
                await asyncio.sleep(GRPC_CALL_TIMEOUT / 3)
            counter += 1
            timestamp = Timestamp()
            timestamp.GetCurrentTime()
            yield ComponentData(
                id=request.id,
                ts=timestamp,
                battery=Battery(
                    data=BatteryData(
                        soc=MetricAggregation(avg=random.uniform(1.0, 100.0)),
                    )
                ),
            )

    mocker.patch.object(servicer, "GetComponentData", mock_component_data)
    server = MockGrpcServer(servicer, port=57809)
    await server.start()

    target = "[::]:57809"
    grpc_channel = grpc.aio.insecure_channel(target)
    client = MicrogridGrpcClient(grpc_channel=grpc_channel, target=target)

    receiver = await client.battery_data(component_id=1)
    messages: List[int] = []
    while len(messages) < delayed_msg_index + 2:
        msg = await receiver.receive()
        messages.append(msg.component_id)

    await server.wait_for_termination(GRPC_SERVER_SHUTDOWN_DELAY)
```

leandro-lucarella-frequenz: So basically with streams there is no way to tell if a connection stopped getting data because there is really no data to transmit from the producing end or if the produced is actually stalled right? But if, for example, the connection is closed, we'll get and error and restart the streaming, correct?

If so, it sounds completely reasonable. I guess the only way around this is to enable some keep-alive mechanism, but probably not doable using gRPC as I would guess streams are one way only. We could have a `ping` call to see if the whole API server is up and responding on time, but beyond that it is probably too overkill.

benchmarks/data_ingestion/benchmark_microgrid_data.py
```diff
@@ -571,6 +571,6 @@ async def benchmark_multiple_formulas(n_formula: int = 100, n_msg: int = 100) ->
 
 
 if __name__ == "__main__":
-    # asyncio.run(benchmark_multiple_meters())
+    asyncio.run(benchmark_multiple_meters())
     # asyncio.run(benchmark_multiple_formulas())
-    asyncio.run(benchmark_multiple_batteries())
+    # asyncio.run(benchmark_multiple_batteries())
```
leandro-lucarella-frequenz: This can probably be left out of the PR, right?

leandro-lucarella-frequenz (commented):
Besides the small new comment, LGTM, but I leave the approval to Ela and Sahas that are more familiar with this code.

src/frequenz/sdk/microgrid/client.py
jakub-toptal: Yeah, good point. I think you are right that the streams are expected to produce data relatively frequently and if it doesn't happen for a longer time it perceived as a timeout on the client side, whether there was no data or if the device was broken.

Regarding the connection being closed:
- if a timeout error occurs, then the retry logic is executed and the stream is restored
- but the connection might also be closed by the producer? I remember there was an option to send a `None` to indicate the end of the stream? I think in this case the stream wouldn't be recreated

Yeah, the keep-alive mechanism sounds like something that could be useful in the future.

benchmarks/data_ingestion/benchmark_microgrid_data.py
jakub-toptal: Whoops, my bad—I was verifying that the benchmarks are working after making changes to the `MicrogridApiClient` (because they aren't included in the test suite).

src/frequenz/sdk/microgrid/client.py
sahas-subramanian-frequenz: The other concern was that if we keep reestablishing our streaming calls,  we'll end up missing datapoints that were sent between the timeout and the retry interval.  That would be highly undesirable,  even though the reconnection will be transparent to the user.

src/frequenz/sdk/microgrid/client.py
```diff
@@ -24,6 +24,9 @@
 from .connection import Connection
 from .retry import LinearBackoff, RetryStrategy
 
+# Default timeout applied to all gRPC calls
+DEFAULT_GRPC_CALL_TIMEOUT = 60.0
```
sahas-subramanian-frequenz, line 28: I thought we had decided to make this an instance variable of `MicrogridGrpcClient` in order to make it configurable.  Wasn't that the decision?

src/frequenz/sdk/microgrid/client.py
```diff
@@ -532,7 +565,12 @@ async def set_power(self, component_id: int, power_w: int) -> Empty:
             )
         return result
 
-    async def set_bounds(self, component_id: int, lower: float, upper: float) -> None:
+    async def set_bounds(
+        self,
+        component_id: int,
+        lower: float,
+        upper: float,
+    ) -> None:
```
sahas-subramanian-frequenz, line 573: there also seem to be a number of formatting changes that didn't get cleaned up.

ela-kotulska-frequenz (approved):

src/frequenz/sdk/microgrid/client.py
ela-kotulska-frequenz: hmm it depends. 
For example for the `set_power` call I use asyncio.wait with timeout anyway - because this timeout is too big.

If we plan to use `DEFAULT_GRPC_CALL_TIMEOUT` as default (preferable) then it would be better to make it constructor parameter and smaller.
But if we don't plan this, then this timeout is only to be extra safe. And there is no reason to make it configurable.

src/frequenz/sdk/microgrid/client.py
jakub-toptal: All checks passed so I assumed everything is fine.

PR #52, opened 2022-10-25
Add DataSourcingActor

### LINKED ISSUE #30: Implement a data sourcing actor
### What's needed?

We need to implement an actor that can expose data from different sources to the downstream layers.

Component data can come from different sources, e.g., the Microgrid API, the Reporting API, and parquet files in local storage.
The data format in these sources can be different.

There may be mutiple downstream actors, each of them reading from multiple sources.
If they all have mechanisms to convert data into one common format for further consumption, then:
* they would each need to understand how all the different data sources work, causing high coupling, and
* they would very likely run similar conversions, wasting CPU cycles

### Proposed solution

The solution would be to create an actor that
* can read multiple data sources, e.g., the Microgrid API, and parquet files in local storage,
* can expose data in a common format that abstracts the underlying data source away,
* can switch between input data streams seamlessly, based upon well-defined conditions (see below for details).

Note that in the first version (v1), this actor needs to consume data from the following sources only:
* Microgrid API
* parquet files in local storage

In order for it to identify the data channel and the data source(s), the requests to the actor should consists of the following parameters:
a. the component ID
b. the metric ID
c. the start timestamp, or the timestamp from which data should be read

If the start timestamp is in past, then the local parquet files should be read first. Once there is no more data in the parquet files, then data from the live data source (Microgrid API) should be streamed.

The top-level building blocks of the actor could be a set of classes as follows
(Note that these are just pseudocoded ideas to specify requirements. The actual implementation may vary):

```py
# This is the class that uniquely represents a metric channel.
@dataclass
class ChannelKey:
    # An enum or a unique ID representing the metric.
    metric_id: str

    # The component ID
    component_id: int

    # The time from which data should be consumed.
    start_ts: datetime
```
See `ComponentMetricKey` in [issue #5](https://github.com/frequenz-floss/frequenz-sdk-python/issues/5) for more details on `ChannelKey`.

```py
# This is the class that consumes and forwards data for one given channel.
# The actor can have a collection of the objects of this class.
class DataIngestor:
  _channel: Option[BroadcastChannel[float]]

  def __init_(self):
    # Initialize `_channel` as `None`
    self._channel = None
    ...

  def get_channel(self, data_key: ChannelKey) -> BroadcastChannel[float]:
    # This method is key here.
    #
    # Consume data from different sources, and stream them downstream via an
    # output channel. The user has to input the details of which
    # metric has to be consumed and from what time. The user gets a channel with
    # the desired data in return.
    #
    # If `data_key.start_ts` is `now()`, then read directly from the
    # Microgrid API.
    #
    # If `data_key.start_ts` is in the past, e.g., `now() - 30m`, then read the
    # parquet files until there is no more data there. Then read from the
    # microgrid API.
    #
    # If `data_key.start_ts` is in the future, e.g., `now() + 30m`, then wait
    # until the given timestamp is reached. Then read from the microgrid API.

    # Return `self._channel` if it already exists, otherwise create a new one.
    ...
```

The actor itself could be defined as follows:

```py
@dataclass
class DataRequestKey:
  # The details of the metric to read data for, and the timesamo from which the
  # data should be read.
  data_key: ChannelKey

  # Can have more elements as required.

class DataSourcingActor:
  # The idea here is to store previously returned channels, so that we reuse them
  # insted of creating new ones every time. The data structure can be a `dict`, or
  # can be some other custom data structure, depending upon the following
  # requirements:
  #
  # Two `ChannelKey` objects with `start_ts >= now()` are
  # effectively the same, since they consume live data from the Microgrid API.
  #
  # However, two `ChannelKey` objects that need historical data from the
  # same `start_ts` timestamp may not be the same. Assuming the two
  # `ChannelKey` objects are received at times 00:00:00 and 01:00:00,
  # then the 2nd one will still require to start from the the first historical
  # data sample, whereas the 1st one could be several data samples ahead or
  # reading live data by then.
  _data_ingestors: Dict[DataRequestKey, DataIngestor]

  def __init__(self):
    # Initialize the `_data_ingestors` object.
    self._data_ingestors = {}
    ...

  def get_channel(self, data_key: ChannelKey):
    data_req_key = DataRequestKey(data_key)

    # If a `DataIngestor`1 object has not yet been created, then create it.
    if not self._data_ingestors.contains(data_req_key):
      self._data_ingestors[data_req_key] = DataIngestor()

    # Return the channel from the `DataNgestor` instance.
    return self._data_ingestors[data_req_key].get_channel()
```

### Use cases

_No response_

_No response_

### Additional context

_No response_

shsms:
Closes #30

### DIFF
--- a/src/frequenz/sdk/actor/data_sourcing/data_sourcing.py
+++ b/src/frequenz/sdk/actor/data_sourcing/data_sourcing.py
@@ -1,0 +1,41 @@
+"""The DataSourcing Actor.
+
+Copyright
+Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+
+License
+MIT
+"""
+
+from frequenz.channels import Receiver
+
+from frequenz.sdk.actor import actor
+from frequenz.sdk.actor.channel_registry import ChannelRegistry
+from frequenz.sdk.data_pipeline import ComponentMetricRequest
+
+from .microgrid_api_source import MicrogridApiSource
+
+
+@actor
+class DataSourcingActor:
+    """An actor that provides data streams of metrics as time series."""
+
+    def __init__(
+        self,
+        request_receiver: Receiver[ComponentMetricRequest],
+        registry: ChannelRegistry,
+    ) -> None:
+        """Create a `DataSourcingActor` instance.
+
+        Args:
+            request_receiver: A channel receiver to accept metric requests from.
+            registry: A channel registry.  To be replaced by a singleton
+                instance.
+        """
+        self._request_receiver = request_receiver
+        self._microgrid_api_source = MicrogridApiSource(registry)
+
+    async def run(self) -> None:
+        """Run the actor."""
+        async for request in self._request_receiver:
+            await self._microgrid_api_source.add_metric(request)
--- a/src/frequenz/sdk/actor/data_sourcing/microgrid_api_source.py
+++ b/src/frequenz/sdk/actor/data_sourcing/microgrid_api_source.py
@@ -1,0 +1,366 @@
+"""The Microgrid API data source for the DataSourcingActor.
+
+Copyright
+Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+
+License
+MIT
+"""
+
+import asyncio
+import logging
+from typing import Any, Callable, Dict, List, Optional, Tuple
+
+from frequenz.channels import Receiver, Sender
+
+from frequenz.sdk.actor import ChannelRegistry
+from frequenz.sdk.data_pipeline import ComponentMetricId, ComponentMetricRequest, Sample
+from frequenz.sdk.microgrid import (
+    BatteryData,
+    ComponentCategory,
+    EVChargerData,
+    InverterData,
+    MeterData,
+    microgrid_api,
+)
+
+_MeterDataMethods: Dict[ComponentMetricId, Callable[[MeterData], float]] = {
+    ComponentMetricId.ACTIVE_POWER: lambda msg: msg.active_power,
+    ComponentMetricId.CURRENT_PHASE_1: lambda msg: msg.current_per_phase[0],
+    ComponentMetricId.CURRENT_PHASE_2: lambda msg: msg.current_per_phase[1],
+    ComponentMetricId.CURRENT_PHASE_3: lambda msg: msg.current_per_phase[2],
+    ComponentMetricId.VOLTAGE_PHASE_1: lambda msg: msg.voltage_per_phase[0],
+    ComponentMetricId.VOLTAGE_PHASE_2: lambda msg: msg.voltage_per_phase[1],
+    ComponentMetricId.VOLTAGE_PHASE_3: lambda msg: msg.voltage_per_phase[2],
+}
+
+_BatteryDataMethods: Dict[ComponentMetricId, Callable[[BatteryData], float]] = {
+    ComponentMetricId.SOC: lambda msg: msg.soc,
+    ComponentMetricId.SOC_LOWER_BOUND: lambda msg: msg.soc_lower_bound,
+    ComponentMetricId.SOC_UPPER_BOUND: lambda msg: msg.soc_upper_bound,
+    ComponentMetricId.CAPACITY: lambda msg: msg.capacity,
+    ComponentMetricId.POWER_LOWER_BOUND: lambda msg: msg.power_lower_bound,
+    ComponentMetricId.POWER_UPPER_BOUND: lambda msg: msg.power_upper_bound,
+}
+
+_InverterDataMethods: Dict[ComponentMetricId, Callable[[InverterData], float]] = {
+    ComponentMetricId.ACTIVE_POWER: lambda msg: msg.active_power,
+    ComponentMetricId.ACTIVE_POWER_LOWER_BOUND: lambda msg: msg.active_power_lower_bound,
+    ComponentMetricId.ACTIVE_POWER_UPPER_BOUND: lambda msg: msg.active_power_upper_bound,
+}
+
+_EVChargerDataMethods: Dict[ComponentMetricId, Callable[[EVChargerData], float]] = {
+    ComponentMetricId.ACTIVE_POWER: lambda msg: msg.active_power,
+    ComponentMetricId.CURRENT_PHASE_1: lambda msg: msg.current_per_phase[0],
+    ComponentMetricId.CURRENT_PHASE_2: lambda msg: msg.current_per_phase[1],
+    ComponentMetricId.CURRENT_PHASE_3: lambda msg: msg.current_per_phase[2],
+    ComponentMetricId.VOLTAGE_PHASE_1: lambda msg: msg.voltage_per_phase[0],
+    ComponentMetricId.VOLTAGE_PHASE_2: lambda msg: msg.voltage_per_phase[1],
+    ComponentMetricId.VOLTAGE_PHASE_3: lambda msg: msg.voltage_per_phase[2],
+}
+
+
+class MicrogridApiSource:
+    """Fetches requested metrics from the Microgrid API.
+
+    Used by the DataSourcingActor.
+    """
+
+    def __init__(
+        self,
+        registry: ChannelRegistry,
+    ) -> None:
+        """Create a `MicrogridApiSource` instance.
+
+        Args:
+            registry: A channel registry.  To be replaced by a singleton
+                instance.
+        """
+        self._comp_categories_cache: Dict[int, ComponentCategory] = {}
+        self.comp_data_receivers: Dict[int, Receiver[Any]] = {}
+        self.comp_data_tasks: Dict[int, asyncio.Task[None]] = {}
+        self._registry = registry
+        self._req_streaming_metrics: Dict[
+            int, Dict[ComponentMetricId, List[ComponentMetricRequest]]
+        ] = {}
+
+    async def _get_component_category(
+        self, comp_id: int
+    ) -> Optional[ComponentCategory]:
+        """Get the component category of the given component.
+
+        Args:
+            comp_id: Id of the requested component.
+
+        Returns:
+            The category of the given component, if it is a valid component, or
+            None otherwise.
+        """
+        if comp_id in self._comp_categories_cache:
+            return self._comp_categories_cache[comp_id]
+
+        api = microgrid_api.get().microgrid_api_client
+        for comp in await api.components():
+            self._comp_categories_cache[comp.component_id] = comp.category
+
+        if comp_id in self._comp_categories_cache:
+            return self._comp_categories_cache[comp_id]
+
+        return None
+
+    async def _check_battery_request(
+        self,
+        comp_id: int,
+        requests: Dict[ComponentMetricId, List[ComponentMetricRequest]],
+    ) -> None:
+        """Check if the requests are valid Battery metrics.
+
+        Raises:
+            ValueError: if the requested metric is not available for batteries.
+
+        Args:
+            comp_id: The id of the requested component.
+            requests: A list of metric requests received from external actors
+                for the given battery.
+        """
+        for metric in requests:
+            if metric not in _BatteryDataMethods:
+                raise ValueError(f"Unknown metric {metric} for Battery id {comp_id}")
+        if comp_id not in self.comp_data_receivers:
+            self.comp_data_receivers[
+                comp_id
+            ] = await microgrid_api.get().microgrid_api_client.battery_data(comp_id)
+
+    async def _check_ev_charger_request(
+        self,
+        comp_id: int,
+        requests: Dict[ComponentMetricId, List[ComponentMetricRequest]],
+    ) -> None:
+        """Check if the requests are valid EV Charger metrics.
+
+        Raises:
+            ValueError: if the requested metric is not available for ev charger.
+
+        Args:
+            comp_id: The id of the requested component.
+            requests: A list of metric requests received from external actors
+                for the given EV Charger.
+        """
+        for metric in requests:
+            if metric not in _EVChargerDataMethods:
+                raise ValueError(f"Unknown metric {metric} for EvCharger id {comp_id}")
+        if comp_id not in self.comp_data_receivers:
+            self.comp_data_receivers[
+                comp_id
+            ] = await microgrid_api.get().microgrid_api_client.ev_charger_data(comp_id)
+
+    async def _check_inverter_request(
+        self,
+        comp_id: int,
+        requests: Dict[ComponentMetricId, List[ComponentMetricRequest]],
+    ) -> None:
+        """Check if the requests are valid Inverter metrics.
+
+        Raises:
+            ValueError: if the requested metric is not available for inverters.
+
+        Args:
+            comp_id: The id of the requested component.
+            requests: A list of metric requests received from external actors
+                for the given inverter.
+        """
+        for metric in requests:
+            if metric not in _InverterDataMethods:
+                raise ValueError(f"Unknown metric {metric} for Inverter id {comp_id}")
+        if comp_id not in self.comp_data_receivers:
+            self.comp_data_receivers[
+                comp_id
+            ] = await microgrid_api.get().microgrid_api_client.inverter_data(comp_id)
+
+    async def _check_meter_request(
+        self,
+        comp_id: int,
+        requests: Dict[ComponentMetricId, List[ComponentMetricRequest]],
+    ) -> None:
+        """Check if the requests are valid Meter metrics.
+
+        Raises:
+            ValueError: if the requested metric is not available for meters.
+
+        Args:
+            comp_id: The id of the requested component.
+            requests: A list of metric requests received from external actors
+                for the given meter.
+        """
+        for metric in requests:
+            if metric not in _MeterDataMethods:
+                raise ValueError(f"Unknown metric {metric} for Meter id {comp_id}")
+        if comp_id not in self.comp_data_receivers:
+            self.comp_data_receivers[
+                comp_id
+            ] = await microgrid_api.get().microgrid_api_client.meter_data(comp_id)
+
+    async def _check_requested_component_and_metrics(
+        self,
+        comp_id: int,
+        category: ComponentCategory,
+        requests: Dict[ComponentMetricId, List[ComponentMetricRequest]],
+    ) -> None:
+        """Check if the requested component and metrics are valid.
+
+        Raises:
+            ValueError: if the category is unknown or if the requested metric
+                is unavailable to the given category.
+
+        Args:
+            comp_id: The id of the requested component.
+            category: The category of the requested component.
+            requests: A list of metric requests received from external actors
+                for the given component.
+        """
+        if comp_id in self.comp_data_receivers:
+            return
+
+        if category == ComponentCategory.BATTERY:
+            await self._check_battery_request(comp_id, requests)
+        elif category == ComponentCategory.EV_CHARGER:
+            await self._check_ev_charger_request(comp_id, requests)
+        elif category == ComponentCategory.INVERTER:
+            await self._check_inverter_request(comp_id, requests)
+        elif category == ComponentCategory.METER:
+            await self._check_meter_request(comp_id, requests)
+        else:
+            raise ValueError(f"Unknown component category {category}")
+
+    def _get_data_extraction_method(
+        self, category: ComponentCategory, metric: ComponentMetricId
+    ) -> Callable[[Any], float]:
+        """Get the data extraction method for the given metric.
+
+        Raises:
+            ValueError: if the category is unknown.
+
+        Args:
+            category: The category of the component.
+            metric: The metric for which we need an extraction method.
+
+        Returns:
+            A method that accepts a `ComponentData` object and returns a float
+            representing the given metric.
+        """
+        if category == ComponentCategory.BATTERY:
+            return _BatteryDataMethods[metric]
+        if category == ComponentCategory.INVERTER:
+            return _InverterDataMethods[metric]
+        if category == ComponentCategory.METER:
+            return _MeterDataMethods[metric]
+        if category == ComponentCategory.EV_CHARGER:
+            return _EVChargerDataMethods[metric]
+        raise ValueError(f"Unknown component category {category}")
+
+    def _get_metric_senders(
+        self,
+        category: ComponentCategory,
+        requests: Dict[ComponentMetricId, List[ComponentMetricRequest]],
+    ) -> List[Tuple[Callable[[Any], float], List[Sender[Sample]]]]:
+        """Get channel senders from the channel registry for each requested metric.
+
+        Args:
+            category: The category of the component.
+            requests: A list of metric requests received from external actors
+                for a certain component.
+
+        Returns:
+            A dictionary of output metric names to channel senders from the
+            channel registry.
+        """
+        return [
+            (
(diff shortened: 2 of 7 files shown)


src/frequenz/sdk/actor/data_sourcing/common.py
```diff
@@ -0,0 +1,66 @@
+"""Common types for the DataSourcingActor.
```
leandro-lucarella-frequenz: What about `messages.py` or `request.py`? I'm not crazy about modules called `common` because it's hard to figure out what could be inside (and with time they tend to have a tendency to be a mess full of unrelated stuff).

src/frequenz/sdk/actor/data_sourcing/common.py
```diff
@@ -0,0 +1,66 @@
...
+    def get_channel_name(self) -> str:
+        """Return a channel name constructed from Self.
+
+        This channel name can be used by the sending side and receiving sides to
+        identify the right channel from the ChannelRegistry.
+
+        Returns:
+            A string denoting a channel name.
+        """
+        return f"{self.component_id}::{self.metric.name}::{self.start_time}::{self.channel_uuid}"
```
leandro-lucarella-frequenz: I still don't fully understand why we want to construct the channel name like this if users are supposed to pass a UUID. If there is a UUID involved, then then channel name can simply be the UUID, it will be also known by both parties. Is all the other information in the channel name only used for debugging purposes?

At some point it was mentioned that `channel_uuid` was going to be named more like `channel_namespace`, which gives more the impression that 2 different users could use the same channel if they both use the same *namespace*, but if it is a UUID it looks like nobody will ever use the same channel name, even if subscribing to the same metric from the same component.

That said, I'm also OK with keeping it as it is until we gather more experience and a better view on how this will be used by the `*Pool` and `LogicalMeter` actors.

src/frequenz/sdk/actor/data_sourcing/common.py
```diff
@@ -0,0 +1,66 @@
...
+MIT
+"""
+
+from dataclasses import dataclass
+from datetime import datetime
+from enum import Enum
+from typing import Optional
+
+
+class ComponentMetric(Enum):
```
leandro-lucarella-frequenz: I would name this `ComponentMetricId`, as we said that the name of field of a component is a [*Metric ID*](https://github.com/frequenz-floss/frequenz-sdk-python/wiki/Glossary#metric-id). It could also be simply `MetricId`, but I'm fine with making it more explicit that is a metric ID of components.

src/frequenz/sdk/actor/data_sourcing/common.py
```diff
@@ -0,0 +1,66 @@
...
+
+    # a uuid to identify the output channel from the channel registry.
+    channel_uuid: str
+    # id of the requested component
+    component_id: int
+    # `metric_name` would be attribute names on BatteryData, etc.  Example
+    # value: `soc`, `current_phase_1`, etc.  For individual phase current and
+    # voltage values, we need to add additional attributes to `MeterData` and
+    # `EVChargerData`.
+    metric: ComponentMetric
```
leandro-lucarella-frequenz: I would also name this `metric_id`.

src/frequenz/sdk/actor/data_sourcing/microgrid_source.py
```diff
@@ -0,0 +1,373 @@
...
+        Raises:
+            TypeError: if the requested metric is not available for batteries.
+
+        Args:
+            comp_id: The id of the requested component.
+            requests: A list of metric requests received from external actors
+                for the given battery.
+        """
+        for metric in requests.keys():
+            if metric not in {
```
leandro-lucarella-frequenz: Why not `metric not in _BatteryDataMethods`? (and so on with all categories)?

src/frequenz/sdk/actor/data_sourcing/microgrid_source.py
```diff
@@ -0,0 +1,373 @@
+"""The Microgrid API data source for the DataSourcingActor.
```
leandro-lucarella-frequenz: I would name this file `micrigrid_api_source` too, to make it clear this is about API (live) data, as data coming from parquet files will also be *microgrid* data.

src/frequenz/sdk/actor/data_sourcing/microgrid_source.py
```diff
@@ -0,0 +1,373 @@
...
+                actor.
+        """
+        comp_id = request.component_id
+
+        if comp_id not in self._req_streaming_metrics:
+            self._req_streaming_metrics[comp_id] = {request.metric: [request]}
+        elif request.metric not in self._req_streaming_metrics[comp_id]:
+            self._req_streaming_metrics[comp_id][request.metric] = [request]
+        else:
+            self._req_streaming_metrics[comp_id][request.metric].append(request)
```
leandro-lucarella-frequenz: pro-tip: You can use [`dict.setdefault()`](_BatteryDataMethods) for this.

```suggestion
        self._req_streaming_metrics.setdefault(comp_id, {}).setdefault(request.metric, []).append(request)
```

Not sure if it makes it more clear or more opaque, though :)

src/frequenz/sdk/actor/data_sourcing/common.py
leandro-lucarella-frequenz: This is first required by the `MicrogridApiSource`, so I would move it directly to `microgrid_source.py`.

src/frequenz/sdk/actor/data_sourcing/common.py
```diff
@@ -0,0 +1,66 @@
...
+
+    POWER_LOWER_BOUND = "power_lower_bound"
+    POWER_UPPER_BOUND = "power_upper_bound"
+
+    ACTIVE_POWER_LOWER_BOUND = "active_power_lower_bound"
+    ACTIVE_POWER_UPPER_BOUND = "active_power_upper_bound"
+
+
+@dataclass
+class ComponentMetricRequest:
```
leandro-lucarella-frequenz: This is first required by the `DataSourcingActor`, so I would move it to `data_sourcing.py`.

This way we can also completely get rid of the `common` module. Both `ComponentMetricRequest` and `ComponentMetricId` can be exposed to the user via `__init__`, so they can do `from frequenz.sdk.actor.data_sourcing import ComponentMetricRequest` for example, and they don't care where it is initially defined.

src/frequenz/sdk/actor/data_sourcing/data_sourcing.py
```diff
@@ -0,0 +1,79 @@
...
+            comp_id: Id of the requested component.
+
+        Returns:
+            The category of the given component, if it is a valid component, or
+            None otherwise.
+        """
+        if comp_id in self._comp_categories:
+            return self._comp_categories[comp_id]
+
+        api = microgrid_api.get().microgrid_api_client
```
leandro-lucarella-frequenz: Why is the category resolution done at this level if it uses the microgrid API client? Shouldn't it be done by the `MicrogridApiSource` so the actor itself have no direct dependency to the microgrid API?

When getting data from parquet files at some point we'll need to check the category in some historical description of the microgrid too, right?

src/frequenz/sdk/actor/data_sourcing/microgrid_source.py
```diff
@@ -0,0 +1,373 @@
...
+    ComponentMetric.CURRENT_PHASE_2: lambda msg: msg.current_per_phase[1],
+    ComponentMetric.CURRENT_PHASE_3: lambda msg: msg.current_per_phase[2],
+    ComponentMetric.VOLTAGE_PHASE_1: lambda msg: msg.voltage_per_phase[0],
+    ComponentMetric.VOLTAGE_PHASE_2: lambda msg: msg.voltage_per_phase[1],
+    ComponentMetric.VOLTAGE_PHASE_3: lambda msg: msg.voltage_per_phase[2],
+}
+
+
+## FIXME: does `src/frequenz/sdk/data_pipeline/types.py` sounds like a good
+## place for the `Sample` class?
```
leandro-lucarella-frequenz: Good question.

I initially wrote a very long message with lots of different approaches with pros and cons, but I don't think it is really adding any value because I didn't come up with any solution I like better than what you are suggesting :laughing: 

So I would say let's go with `src/frequenz/sdk/data_pipeline/types.py` for now decide later if there is a better place.

src/frequenz/sdk/actor/data_sourcing/microgrid_source.py
```diff
@@ -0,0 +1,373 @@
...
+@dataclass(frozen=True)
+class Sample:
+    """Describes a single observed floating point value at a specific timestamp."""
```
leandro-lucarella-frequenz: As I said in the resampling PR, maybe we can just say `measurement` to be more consistent with the glossary. Also maybe mention why and when the value could be `None`:

```suggestion
    """A measurement taken at a particular point in time.
    
    The `value` could be `None` if a component is malfunctioning or data is lacking
    for another reason, but a sample still needs to be sent to have a coherent view
    on a group of component metrics for a particular timestamp.
    """
```

src/frequenz/sdk/actor/data_sourcing/microgrid_source.py
```diff
@@ -0,0 +1,373 @@
...
+    """Fetches requested metrics from the Microgrid API.
+
+    Used by the DataSourcingActor.
+    """
+
+    def __init__(
+        self,
+        registry: ChannelRegistry,
+    ) -> None:
+        """Create a `_MicrogridApiSource` instance.
```
leandro-lucarella-frequenz:
```suggestion
        """Create a `MicrogridApiSource` instance.
```

src/frequenz/sdk/actor/data_sourcing/microgrid_source.py
```diff
@@ -0,0 +1,374 @@
...
+        for metric in requests.keys():
+            if metric not in {
+                ComponentMetric.SOC,
+                ComponentMetric.SOC_LOWER_BOUND,
+                ComponentMetric.SOC_UPPER_BOUND,
+                ComponentMetric.CAPACITY,
+                ComponentMetric.POWER_LOWER_BOUND,
+                ComponentMetric.POWER_UPPER_BOUND,
+            }:
+                raise TypeError(f"Unknown metric {metric} for Battery id {comp_id}")
```
leandro-lucarella-frequenz: Nitpick: I would use a `ValueError` here, as `TypeError` are supposed to be mismatches on the type system.

src/frequenz/sdk/actor/data_sourcing/microgrid_source.py
```diff
@@ -0,0 +1,374 @@
...
+        if category == ComponentCategory.BATTERY:
+            await self._check_battery_request(comp_id, requests)
+        elif category == ComponentCategory.EV_CHARGER:
+            await self._check_ev_charger_request(comp_id, requests)
+        elif category == ComponentCategory.INVERTER:
+            await self._check_inverter_request(comp_id, requests)
+        elif category == ComponentCategory.METER:
+            await self._check_meter_request(comp_id, requests)
+        else:
+            raise TypeError(f"Unknown component category {category}")
```
leandro-lucarella-frequenz: This should be a `ValueError`.

src/frequenz/sdk/actor/data_sourcing/microgrid_source.py
```diff
@@ -0,0 +1,374 @@
...
+        """
+        if category == ComponentCategory.BATTERY:
+            return _BatteryDataMethods[metric]
+        if category == ComponentCategory.INVERTER:
+            return _InverterDataMethods[metric]
+        if category == ComponentCategory.METER:
+            return _MeterDataMethods[metric]
+        if category == ComponentCategory.EV_CHARGER:
+            return _EVChargerDataMethods[metric]
+        raise TypeError(f"Unknown component category {category}")
```
leandro-lucarella-frequenz: This should be a `ValueError`.

src/frequenz/sdk/actor/data_sourcing/microgrid_source.py
```diff
@@ -0,0 +1,374 @@
...
+        """
+        comp_id = request.component_id
+
+        if comp_id not in self._req_streaming_metrics:
+            self._req_streaming_metrics[comp_id] = {request.metric: [request]}
+        elif request.metric not in self._req_streaming_metrics[comp_id]:
+            self._req_streaming_metrics[comp_id][request.metric] = [request]
+        else:
+            self._req_streaming_metrics[comp_id][request.metric].append(request)
+
```
leandro-lucarella-frequenz: Looking at `_update_strems()` a possible optimization here would be to check if the requested metric is already handled and avoid calling `_update_strems()` to avoid cancelling and re-creating the task.

leandro-lucarella-frequenz (commented):
The microgrid source was a bit hard to follow, not sure if there is a simpler way to do it because I guess it has quite a bit of inherent complexity, but maybe more comments can be added. I feel I lacked a more high-level overview when reviewing about what are the responsibilities of each part, or the general algorithm, or even comments on the main attributes for microgrid source class.

src/frequenz/sdk/actor/data_sourcing/common.py
sahas-subramanian-frequenz: I was going to rename this to `namespace`, but forgot.  Will do in the next changeset.

src/frequenz/sdk/actor/data_sourcing/common.py
shsms: done

src/frequenz/sdk/actor/data_sourcing/common.py
shsms: Both `ComponentMetricId` and `ComponentMetricRequest` need to be accessible for other actors,  and will probably be accepted by the resampling actor too?  So I put it in the ~`frequenz/sdk/types.py`~ (EDIT: I meant `frequenz/sdk/data_pipeline/types.py`).  Does that sound ok?

src/frequenz/sdk/actor/data_sourcing/common.py
shsms: done

src/frequenz/sdk/actor/data_sourcing/common.py
shsms: done

src/frequenz/sdk/actor/data_sourcing/data_sourcing.py
shsms: I think the decision was that the parquet files will on the edge will get migrated when the component IDs change, etc.  So I think the microgrid api will be the source of this information for historical data as well.

shsms: But I guess you're right,  the microgrid api won't be available when running the actor on the cloud to just read historical data.  I will move it into `MicrogridApiSource` tomorrow.

src/frequenz/sdk/actor/data_sourcing/microgrid_source.py
shsms: done

src/frequenz/sdk/actor/data_sourcing/microgrid_source.py
shsms: done

src/frequenz/sdk/actor/data_sourcing/microgrid_source.py
shsms: done

src/frequenz/sdk/actor/data_sourcing/microgrid_source.py
shsms: done

src/frequenz/sdk/actor/data_sourcing/microgrid_source.py
shsms: oh yes,  it is because I wrote these checks first,  before I wrote the data extraction methods.  Fixed now, thanks for noticing. :-)

src/frequenz/sdk/actor/data_sourcing/microgrid_source.py
shsms: done

src/frequenz/sdk/actor/data_sourcing/microgrid_source.py
shsms: done

src/frequenz/sdk/actor/data_sourcing/microgrid_source.py
shsms: done

src/frequenz/sdk/actor/data_sourcing/microgrid_source.py
shsms: it is more readable,  and people can look up the method,  updated, thanks. :-)

src/frequenz/sdk/actor/data_sourcing/microgrid_source.py
shsms: will do tomorrow.

shsms (commented):
I will add some more comments tomorrow as well.

src/frequenz/sdk/actor/data_sourcing/microgrid_api_source.py
```diff
@@ -0,0 +1,330 @@
...
+
+        Raises:
+            ValueError: if the requested metric is not available for batteries.
+
+        Args:
+            comp_id: The id of the requested component.
+            requests: A list of metric requests received from external actors
+                for the given battery.
+        """
+        for metric in requests.keys():
```
leandro-lucarella-frequenz: Nitpick: this is a bit redundant, `x in dict` is looking if `x` is in the keys of the dict already :)

```suggestion
        for metric in requests:
```

src/frequenz/sdk/actor/data_sourcing/common.py
leandro-lucarella-frequenz: For me it doesn't sound like a general interface, I see it more owned by the data sourcing actor, even if other actors use it. It is completely coupled with how the data sourcing actor works and will have to be updated if the data sourcing actor *interface* changes, this is why I think they belong together.

Yes, probably the resampling actor will use it, but it can expose an alias, because eventually the resampling actor could have other parameters in the request (I was thinking that maybe, just **maybe**, we could want at some point to request a metric to be resampled with a particular resampling function after we talked about resampling a `xxx_max` metric for example). In that case it will have to define it's own request.

All that said, it is still a style issue I guess, so I won't oppose if you want to leave it like this. :)

src/frequenz/sdk/actor/data_sourcing/microgrid_source.py
leandro-lucarella-frequenz: OK, I forgot to mention but I think the PR is fine without this too, it was more of a thought and to check if I was understanding correctly, but if it is a low hanging fruit and want to add it now I won't oppose either :)

src/frequenz/sdk/actor/data_sourcing/common.py
leandro-lucarella-frequenz: Closing so we have the discussion only in one place: https://github.com/frequenz-floss/frequenz-sdk-python/pull/52#discussion_r1005486650

src/frequenz/sdk/actor/data_sourcing/common.py
leandro-lucarella-frequenz: Closing so we have all the discussion about where to put these types in one place: https://github.com/frequenz-floss/frequenz-sdk-python/pull/52#discussion_r1005486650

src/frequenz/sdk/actor/data_sourcing/microgrid_api_source.py
sahas-subramanian-frequenz: This is one of the stupid things about python that lead to confusing bugs for me.  If the language always insists on `.keys()`,  that would have been amazing.  But I will remove it, to be consistent.

src/frequenz/sdk/actor/data_sourcing/common.py
shsms: The `DataSourcingActor` will also be the first source of `Sample`s.  What do you think about just `frequenz/sdk/actor/data_sourcing/types.py` for all three?

(Also I had a typo in the path I mentioned earlier, that's now fixed)

src/frequenz/sdk/actor/data_sourcing/data_sourcing.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+@actor
+class DataSourcingActor:
+    """An actor that provides data streams of metrics as time series."""
+
+    def __init__(
+        self,
+        request_receiver: Receiver[ComponentMetricRequest],
+        registry: ChannelRegistry,
```
sahas-subramanian-frequenz, line 26: we probably need a global singleton registry as well,  before this PR can be merged.

leandro-lucarella-frequenz: Why? I mean why it is a requirement for this to be merged?

sahas-subramanian-frequenz: Because we're passing the registry as an argument to an actor,  and that's not ideal.

We can of course merge,  but then we'll have to come back and fix this afterwards.

src/frequenz/sdk/actor/data_sourcing/data_sourcing.py
sahas-subramanian-frequenz: done

src/frequenz/sdk/actor/data_sourcing/microgrid_source.py
sahas-subramanian-frequenz: this is done for now.  Reusing channels will get complicated when we start adding parquet as a source,  but we can deal that then.

src/frequenz/sdk/actor/data_sourcing/microgrid_api_source.py
sahas-subramanian-frequenz: done

src/frequenz/sdk/actor/data_sourcing/common.py
leandro-lucarella-frequenz: Haha, I thought about `Sample` but since I didn't have a good case for it I left it out in the hope that you won't notice :trollface: 

OK, now that you forced me to think about it harder, I found out why `Sample` is different for me: If we ever make the resampler agnostic of the data sourcing actor, `Sample` will be a more general concept, where other sourcing actors will also use `Sample`, but not `ComponentMetricId` and `ComponentMetricRequest`. This is why I think `Sample` is a more general data pipeline concept, not necessarily tied to the data sourcing actor, while the other too I don't see them being used outside of it (independently I mean, not to communicate with the data sourcing actor).

But since this is an hypothetical case for now, I would be fine to add it too to the data sourcing actor. At the end I think we'll probably need to re-arrange a thing of two as we advance more with the data pipeline and the structure is more clear.

About `types`, I'm not crazy about general modules like `common` and `types` but `types` is definitely better as it is more scoped than `common`. So personally I would still have `ComponentMetricId` and `ComponentMetricRequest` very close to the class that gives live to those classes (again, for me `Sample` is different), but `types` is fine too.

leandro-lucarella-frequenz (commented):
LGTM, I can approve after the fixups are squashed (otherwise I will have to approve again anyway :)

leandro-lucarella-frequenz (approved):

PR #54, opened 2022-10-31
Indent multi-line return value descriptions in docstrings

shsms:


leandro-lucarella-frequenz (commented):
I guess you did this with an automatic tool and this is why it is also *affecting* `Args:`, `Raises:` and some body function, right? It might be a good idea to mention this in the commit too, but it is just a nitpick so I'm fine with leaving it as it is too.

But I would delay merging this until the documentation generation is merged (it should be soon-ish), so we can actually check the results are now correct, or did you somehow tested it?

sahas-subramanian-frequenz:
> I guess you did this with an automatic tool and this is why it is also affecting Args:, Raises: and some body function, right? It might be a good idea to mention this in the commit too, but it is just a nitpick so I'm fine with leaving it as it is too.

I've updated the commit message.

> did you somehow tested it

no,  do you have the documentation generation code in your fork?  maybe that can be rebased on this to test?

leandro-lucarella-frequenz:
Sorry, not yet because I starting cleaning up and building the final config for the channels repo instead, so we can use the cross-reference here, so I will create a PR first to that repo.

leandro-lucarella-frequenz:
@sahas-subramanian-frequenz I created a (draft) PR for the channels repo: https://github.com/frequenz-floss/frequenz-channels-python/pull/25, it should be fairly easy to just apply the same patch to this repo and try it out if you are very eager to do so, otherwise we can just wait a bit longer to merge this :)

leandro-lucarella-frequenz (dismissed):
This was already tested in the channels repo, so let's merge it! :shipit:

shsms:
Rebased on to latest 0.x.x

sahas-subramanian-frequenz (approved):

PR #66, opened 2022-11-15
Add a logical meter implementation

### LINKED ISSUE #61: Add a `LogicalMeter` implementation
### What's needed?

A `LogicalMeter` implementation will have to support the follow (method) streams:

1. Grid meter power measurement
2. **Client load (sink/black hole - unknown consumer&producer)** DEPRECATED!
3. Battery meter power measurement
4. PV meter power measurement
5. EV meter power measurement
6. CHP meter power measurement
....

Their default formulas will be inferred from the component graph,  and the formulas can be overridden from the UI.

It should connect to the `ResamplingActor` to get resampled data for individual components.

### Tasks

- [x] #66 
    - [x] Formula engine that can apply simple formulas on `Sample` streams.
    - [x] Parsing of manual formulas specified in the UI
- [x] #102
- [x] #133 

cc: @leandro-lucarella-frequenz @tiyash-basu-frequenz @ela-kotulska-frequenz

shsms:
1. Adds a formula engine for running formulas on streaming data.  This also includes a parser for the manual formulas,  and a formula builder, for connecting streams from the resampling actor, to the formula engine.
2. A logical meter implementation that provides the public API for users to get high level metrics from.

Formulas can have Component IDs that are preceeded by a pound symbol(`#`), and these operators: `+, -, *, /, (, )`.  For example, the input string: `#20 + #5` is a formula for adding metrics from two components with ids 20 and 5.

Tracked in #61

### DIFF
--- a/src/frequenz/sdk/timeseries/logical_meter/__init__.py
+++ b/src/frequenz/sdk/timeseries/logical_meter/__init__.py
@@ -1,0 +1,8 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+
+"""A logical meter for calculating high level metrics for a microgrid."""
+
+from .logical_meter import LogicalMeter
+
+__all__ = ["LogicalMeter"]
--- a/src/frequenz/sdk/timeseries/logical_meter/__init__.py
+++ b/src/frequenz/sdk/timeseries/logical_meter/__init__.py
@@ -3,6 +3,6 @@
 
 """A logical meter for calculating high level metrics for a microgrid."""
 
-from .logical_meter import LogicalMeter
+from ._logical_meter import LogicalMeter
 
 __all__ = ["LogicalMeter"]
--- a/src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
+++ b/src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
@@ -1,0 +1,201 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+
+"""A formula engine that can apply formulas on streaming data."""
+
+from __future__ import annotations
+
+import asyncio
+from datetime import datetime
+from typing import Dict, List, Optional, Set
+
+from frequenz.channels import Receiver
+
+from .._sample import Sample
+from ._formula_steps import (
+    Adder,
+    Divider,
+    FormulaStep,
+    MetricFetcher,
+    Multiplier,
+    OpenParen,
+    Subtractor,
+)
+
+_operator_precedence = {
+    "(": 0,
+    "/": 1,
+    "*": 2,
+    "-": 3,
+    "+": 4,
+    ")": 5,
+}
+
+
+class FormulaEngine:
+    """A post-fix formula engine that operates on `Sample` receivers.
+
+    Operators and metrics need to be pushed into the engine in in-fix order, and they
+    get rearranged into post-fix in the engine.  This is done using the [Shunting yard
+    algorithm](https://en.wikipedia.org/wiki/Shunting_yard_algorithm).
+
+    Example:
+        To create an engine that adds the latest entries from two receivers, the
+        following calls need to be made:
+
+        ```python
+        engine = FormulaEngine()
+        engine.push_metric("metric_1", receiver_1)
+        engine.push_oper("+")
+        engine.push_metric("metric_2", receiver_2)
+        engine.finalize()
+        ```
+
+        and then every call to `engine.apply()` would fetch a value from each receiver,
+        add the values and return the result.
+    """
+
+    def __init__(
+        self,
+    ) -> None:
+        """Create a `FormulaEngine` instance."""
+        self._steps: List[FormulaStep] = []
+        self._build_stack: List[FormulaStep] = []
+        self._metric_fetchers: Dict[str, MetricFetcher] = {}
+        self._first_run = True
+
+    def push_oper(self, oper: str) -> None:
+        """Push an operator into the engine.
+
+        Args:
+            oper: One of these strings - "+", "-", "*", "/", "(", ")"
+        """
+        if self._build_stack and oper != "(":
+            op_prec = _operator_precedence[oper]
+            while self._build_stack:
+                prev_step = self._build_stack[-1]
+                if op_prec <= _operator_precedence[repr(prev_step)]:
+                    break
+                if oper == ")" and repr(prev_step) == "(":
+                    self._build_stack.pop()
+                    break
+                if repr(prev_step) == "(":
+                    break
+                self._steps.append(prev_step)
+                self._build_stack.pop()
+
+        if oper == "+":
+            self._build_stack.append(Adder())
+        elif oper == "-":
+            self._build_stack.append(Subtractor())
+        elif oper == "*":
+            self._build_stack.append(Multiplier())
+        elif oper == "/":
+            self._build_stack.append(Divider())
+        elif oper == "(":
+            self._build_stack.append(OpenParen())
+
+    def push_metric(self, name: str, data_stream: Receiver[Sample]) -> None:
+        """Push a metric receiver into the engine.
+
+        Args:
+            name: A name for the metric.
+            data_stream: A receiver to fetch this metric from.
+        """
+        fetcher = self._metric_fetchers.setdefault(
+            name, MetricFetcher(name, data_stream)
+        )
+        self._steps.append(fetcher)
+
+    def finalize(self) -> None:
+        """Finalize the formula engine.
+
+        This function must be called before calls to `apply` can be made.
+        """
+        while self._build_stack:
+            self._steps.append(self._build_stack.pop())
+
+    async def synchronize_metric_timestamps(
+        self, metrics: Set[asyncio.Task[Optional[Sample]]]
+    ) -> datetime:
+        """Synchronize the metric streams.
+
+        For synchronised streams like data from the `ComponentMetricsResamplingActor`,
+        this a call to this function is required only once, before the first set of
+        inputs are fetched.
+
+        Args:
+            metrics: The finished tasks from the first `fetch_next` calls to all the
+                `MetricFetcher`s.
+
+        Returns:
+            The timestamp of the latest metric value.
+
+        Raises:
+            RuntimeError: when some streams have no value, or when the synchronization
+                of timestamps fails.
+        """
+        metrics_by_ts: Dict[datetime, str] = {}
+        for metric in metrics:
+            result = metric.result()
+            name = metric.get_name()
+            if result is None:
+                raise RuntimeError(f"Stream closed for component: {name}")
+            metrics_by_ts[result.timestamp] = name
+        latest_ts = max(metrics_by_ts)
+
+        # fetch the metrics with non-latest timestamps again until we have the values
+        # for the same ts for all metrics.
+        for metric_ts, name in metrics_by_ts.items():
+            if metric_ts == latest_ts:
+                continue
+            fetcher = self._metric_fetchers[name]
+            while metric_ts < latest_ts:
+                next_val = await fetcher.fetch_next()
+                assert next_val is not None
+                metric_ts = next_val.timestamp
+            if metric_ts > latest_ts:
+                raise RuntimeError(
+                    "Unable to synchronize timestamps of resampled metrics"
+                )
+        self._first_run = False
+        return latest_ts
+
+    async def apply(self) -> Sample:
+        """Fetch the latest metrics, apply the formula once and return the result.
+
+        Returns:
+            The result of the formula.
+
+        Raises:
+            RuntimeError: if some samples didn't arrive, or if formula application
+                failed.
+        """
+        eval_stack: List[float] = []
+        ready_metrics, pending = await asyncio.wait(
+            [
+                asyncio.create_task(fetcher.fetch_next(), name=name)
+                for name, fetcher in self._metric_fetchers.items()
+            ],
+            return_when=asyncio.ALL_COMPLETED,
+        )
+
+        if pending or any(res.result() is None for res in iter(ready_metrics)):
+            raise RuntimeError("Some resampled metrics didn't arrive")
+
+        if self._first_run:
+            metric_ts = await self.synchronize_metric_timestamps(ready_metrics)
+        else:
+            res = next(iter(ready_metrics)).result()
+            assert res is not None
+            metric_ts = res.timestamp
+
+        for step in self._steps:
+            step.apply(eval_stack)
+
+        # if all steps were applied and the formula was correct, there should only be a
+        # single value in the evaluation stack, and that would be the formula result.
+        if len(eval_stack) != 1:
+            raise RuntimeError("Formula application failed.")
+
+        return Sample(metric_ts, eval_stack[0])
--- a/src/frequenz/sdk/timeseries/logical_meter/_formula_steps.py
+++ b/src/frequenz/sdk/timeseries/logical_meter/_formula_steps.py
@@ -1,0 +1,188 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+
+"""Steps for building formula engines with."""
+
+from abc import ABC, abstractmethod
+from typing import List, Optional
+
+from frequenz.channels import Receiver
+
+from .._sample import Sample
+
+
+class FormulaStep(ABC):
+    """Represents an individual step/stage in a formula.
+
+    Each step, when applied on to an evaluation stack, would pop its input parameters
+    from the stack and push its result back in.
+    """
+
+    @abstractmethod
+    def __repr__(self) -> str:
+        """Return a string representation of the step.
+
+        Returns:
+            A string representation of the step.
+        """
+
+    @abstractmethod
+    def apply(self, eval_stack: List[float]) -> None:
+        """Apply a formula operation on the eval_stack.
+
+        Args:
+            eval_stack: An evaluation stack, to apply the formula step on.
+        """
+
+
+class Adder(FormulaStep):
+    """A formula step for adding two values."""
+
+    def __repr__(self) -> str:
+        """Return a string representation of the step.
+
+        Returns:
+            A string representation of the step.
+        """
+        return "+"
+
+    def apply(self, eval_stack: List[float]) -> None:
+        """Extract two values from the stack, add them, push the result back in.
+
+        Args:
+            eval_stack: An evaluation stack, to apply the formula step on.
+        """
+        val2 = eval_stack.pop()
+        val1 = eval_stack.pop()
+        eval_stack.append(val1 + val2)
+
+
+class Subtractor(FormulaStep):
+    """A formula step for subtracting one value from another."""
+
+    def __repr__(self) -> str:
+        """Return a string representation of the step.
+
+        Returns:
+            A string representation of the step.
+        """
+        return "-"
+
+    def apply(self, eval_stack: List[float]) -> None:
+        """Extract two values from the stack, subtract them, push the result back in.
+
+        Args:
+            eval_stack: An evaluation stack, to apply the formula step on.
+        """
+        val2 = eval_stack.pop()
+        val1 = eval_stack.pop()
+        eval_stack.append(val1 - val2)
+
+
+class Multiplier(FormulaStep):
+    """A formula step for multiplying two values."""
+
+    def __repr__(self) -> str:
+        """Return a string representation of the step.
+
+        Returns:
+            A string representation of the step.
+        """
+        return "*"
+
+    def apply(self, eval_stack: List[float]) -> None:
+        """Extract two values from the stack, multiply them, push the result back in.
+
+        Args:
+            eval_stack: An evaluation stack, to apply the formula step on.
+        """
+        val2 = eval_stack.pop()
+        val1 = eval_stack.pop()
+        eval_stack.append(val1 * val2)
+
+
+class Divider(FormulaStep):
+    """A formula step for dividing one value by another."""
+
+    def __repr__(self) -> str:
+        """Return a string representation of the step.
+
+        Returns:
+            A string representation of the step.
+        """
+        return "/"
+
+    def apply(self, eval_stack: List[float]) -> None:
+        """Extract two values from the stack, divide them, push the result back in.
+
+        Args:
+            eval_stack: An evaluation stack, to apply the formula step on.
+        """
+        val2 = eval_stack.pop()
+        val1 = eval_stack.pop()
+        eval_stack.append(val1 / val2)
+
+
+class OpenParen(FormulaStep):
+    """A no-op formula step used while building a prefix formula engine.
+
+    Any OpenParen steps would get removed once a formula is built.
+    """
+
+    def __repr__(self) -> str:
+        """Return a string representation of the step.
+
+        Returns:
+            A string representation of the step.
+        """
+        return "("
+
+    def apply(self, _: List[float]) -> None:
+        """No-op."""
+
+
+class MetricFetcher(FormulaStep):
+    """A formula step for fetching a value from a metric Receiver."""
+
+    def __init__(self, name: str, stream: Receiver[Sample]) -> None:
+        """Create a `MetricFetcher` instance.
+
+        Args:
+            name: The name of the metric.
+            stream: A channel receiver from which to fetch samples.
+        """
+        self._name = name
+        self._stream = stream
+        self._next_value: Optional[Sample] = None
+
+    async def fetch_next(self) -> Optional[Sample]:
+        """Fetch the next value from the stream.
+
+        To be called before each call to `apply`.
+
+        Returns:
+            The fetched Sample.
+        """
+        self._next_value = await self._stream.receive()
+        return self._next_value
+
+    def __repr__(self) -> str:
+        """Return a string representation of the step.
+
+        Returns:
+            A string representation of the step.
+        """
+        return self._name
+
+    def apply(self, eval_stack: List[float]) -> None:
+        """Push the latest value from the stream into the evaluation stack.
+
+        Args:
+            eval_stack: An evaluation stack, to apply the formula step on.
+
+        Raises:
+            RuntimeError: No next value available to append.
+        """
+        if self._next_value is None or self._next_value.value is None:
+            raise RuntimeError("No next value available to append.")
+        eval_stack.append(self._next_value.value)
--- a/tests/timeseries/test_formula_engine.py
+++ b/tests/timeseries/test_formula_engine.py
@@ -1,0 +1,172 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+
+"""Tests for the FormulaEngine and the Tokenizer."""
+
+from datetime import datetime
+from typing import Dict, List, Tuple
+
+from frequenz.channels import Broadcast
+
+from frequenz.sdk.timeseries import Sample
+from frequenz.sdk.timeseries.logical_meter._formula_engine import FormulaEngine
+from frequenz.sdk.timeseries.logical_meter._tokenizer import Token, Tokenizer, TokenType
+
+
+class TestTokenizer:
+    """Tests for the Tokenizer."""
+
+    def test_1(self) -> None:
+        """Test the tokenization of the formula: "#10 + #20 - (#5 * #4)"."""
+        tokens = []
+        lexer = Tokenizer("#10 + #20 - (#5 * #4)")
+        for token in lexer:
+            tokens.append(token)
+        assert tokens == [
+            Token(TokenType.COMPONENT_METRIC, "10"),
+            Token(TokenType.OPER, "+"),
+            Token(TokenType.COMPONENT_METRIC, "20"),
+            Token(TokenType.OPER, "-"),
+            Token(TokenType.OPER, "("),
+            Token(TokenType.COMPONENT_METRIC, "5"),
+            Token(TokenType.OPER, "*"),
+            Token(TokenType.COMPONENT_METRIC, "4"),
+            Token(TokenType.OPER, ")"),
+        ]
+
+
+class TestFormulaEngine:
+    """Tests for the FormulaEngine."""
+
+    def setup(self) -> None:
+        """Initialize the channels required for a test.
+
+        Because we can't create a __init__ or multiple instances of the Test class, we
+        use the `setup` method as a constructor, and call it once before each test.
+        """
+        # pylint: disable=attribute-defined-outside-init
+        self.comp_2 = Broadcast[Sample]("")
+        self.comp_2_sender = self.comp_2.new_sender()
+
+        self.comp_4 = Broadcast[Sample]("")
+        self.comp_4_sender = self.comp_4.new_sender()
+
+        self.comp_5 = Broadcast[Sample]("")
+        self.comp_5_sender = self.comp_5.new_sender()
+        # pylint: enable=attribute-defined-outside-init
+
+    async def run_test(
+        self, formula: str, postfix: str, io_pairs: List[Tuple[List[float], float]]
+    ) -> None:
+        channels: Dict[str, Broadcast[Sample]] = {}
+        engine = FormulaEngine()
+        for token in Tokenizer(formula):
+            if token.type == TokenType.COMPONENT_METRIC:
+                if token.value not in channels:
+                    channels[token.value] = Broadcast(token.value)
+                engine.push_metric(
+                    f"#{token.value}", channels[token.value].new_receiver()
+                )
+            elif token.type == TokenType.OPER:
+                engine.push_oper(token.value)
+        engine.finalize()
+
+        assert repr(engine._steps) == postfix
+
+        now = datetime.now()
+        tests_passed = 0
+        for io_pair in io_pairs:
+            input, output = io_pair
+            [
+                await chan.new_sender().send(Sample(now, value))
+                for chan, value in zip(channels.values(), input)
+            ]
+            assert (await engine.apply()).value == output
+            tests_passed += 1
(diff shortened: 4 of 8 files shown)


leandro-lucarella-frequenz:
Assigning to v0.12.0 but with prio low to indicate we can release without it too.

shsms:
The source files need to be moved out of the `actor` package,  probably into `timeseries`?  Will do that after #68 is merged.

leandro-lucarella-frequenz:
> The source files need to be moved out of the `actor` package, probably into `timeseries`? Will do that after #68 is merged.

For stuff that works purely on *timeseries*, probably yes.

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
```diff
@@ -0,0 +1,454 @@
...
+        Returns:
+            A receiver to stream resampled data for the given component id.
+        """
+        request = ComponentMetricRequest(
+            self._namespace, component_id, self._metric_id, None
+        )
+        await self._resampler_subscription_sender.send(request)
+        return self._channel_registry.new_receiver(request.get_channel_name())
+
+    async def push_component_metric(self, component_id: int) -> None:
```
matthias-wende-frequenz: Is there any reason why this method isn't marked as private?

tests/timeseries/test_formula_engine.py
```diff
@@ -0,0 +1,152 @@
...
+    Token,
+    Tokenizer,
+    TokenType,
+)
+
+
+class TestTokenizer:
+    """Tests for the Tokenizer."""
+
+    def test_1(self) -> None:
```
matthias-wende-frequenz, line 19: I suggest to test all supported operators.

matthias-wende-frequenz (commented):
Formula engine reviewed! Next I'll add a review for the logical meter.

src/frequenz/sdk/timeseries/logical_meter/__init__.py
```diff
@@ -0,0 +1,12 @@
+"""A logical meter for calculating high level metrics for a microgrid.
+
+Copyright
+Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+
+License
+MIT
+"""
```
leandro-lucarella-frequenz:
```suggestion
# License: MIT
# Copyright © 2022 Frequenz Energy-as-a-Service GmbH

"""A logical meter for calculating high level metrics for a microgrid."""
```

Same for the rest files :)

src/frequenz/sdk/timeseries/logical_meter/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+
+from .logical_meter import LogicalMeter
```
leandro-lucarella-frequenz: If you are only exposing `LogicalMeter`, I would also make the whole module `logical_meter` hidden to keep the public interface clean:

```suggestion
from ._logical_meter import LogicalMeter
```

And also the whole package, so in `timeseries/__init__.py` I would do too an import of:

```py
from ._logical_meter import LogicalMeter
```

On top of what's already there.

Another option, if we want to give users access to the formula engine would be to keep the package `timeseries.logical_meter` visible and only hide `timeseries.logical_meter._logical_meter`.

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
```diff
@@ -0,0 +1,454 @@
...
+    "(": 0,
+    "/": 1,
+    "*": 2,
+    "-": 3,
+    "+": 4,
+    ")": 5,
+}
+
+
+class StringIter:
```
leandro-lucarella-frequenz: Maybe make it a `@dataclass(frozen=True)` to avoid having to expose all the properties manually yourself?

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
```diff
@@ -0,0 +1,454 @@
...
+
+    COMPONENT_METRIC = 0
+    OPER = 1
+
+
+@dataclass
+class Token:
+    """Represents a Token returned by the Tokenizer."""
+
+    type_: TokenType
```
leandro-lucarella-frequenz: Why not `type`? It is not a reserved word in Python.

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
```diff
@@ -0,0 +1,454 @@
...
+
+    def __init__(self, formula: str) -> None:
+        """Create a `Tokenizer` instance.
+
+        Args:
+            formula: The string formula to tokenize.
+        """
+        self._formula = StringIter(formula)
+
+    def read_unsigned_int(self) -> str:
```
leandro-lucarella-frequenz: This could also be private, right?

src/frequenz/sdk/timeseries/logical_meter/logical_meter.py
```diff
@@ -0,0 +1,111 @@
...
+from frequenz.channels import Broadcast, Receiver, Sender
+
+from ...actor import ChannelRegistry, ComponentMetricRequest
+from ...microgrid.component import ComponentMetricId
+from ...timeseries import Sample
+from ._formula_engine import FormulaBuilder, FormulaEngine
+
+
+class LogicalMeter:
+    """A logical meter for calculating high level metrics for a microgrid.
```
matthias-wende-frequenz: ... in a microgrid
would that sound better?

src/frequenz/sdk/timeseries/logical_meter/logical_meter.py
```diff
@@ -0,0 +1,111 @@
...
+
+    LogicalMeter instances need to be specified formulas for each of the metrics that
+    need to be calculated.
```
matthias-wende-frequenz: This sentence is confusing. A suspect it means that the user needs to specify formulas to a LM.

src/frequenz/sdk/timeseries/logical_meter/logical_meter.py
```diff
@@ -0,0 +1,111 @@
...
+
+        Args:
+            channel_registry: A channel registry instance shared with the resampling
+                actor.
+            resampler_subscription_sender: A sender for sending metric requests to the
+                resampling actor.
+        """
+        self._channel_registry = channel_registry
+        self._resampler_subscription_sender = resampler_subscription_sender
+        self._namespace = f"logical-meter-{uuid.uuid4()}"
```
matthias-wende-frequenz: I suggest to document where `uuid.uuid4()` comes from.

src/frequenz/sdk/timeseries/logical_meter/logical_meter.py
```diff
@@ -0,0 +1,111 @@
...
+
+        # Channels are created inside the class under the assumption that LogicalMeter
+        # won't be an actor.  If it is going to be an actor, then it can't own the
+        # channels anymore.
```
matthias-wende-frequenz: That comment looks outdated to me. If this PR gets merged then this is decided.

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
```diff
@@ -0,0 +1,454 @@
...
+        for step in self._steps:
+            step.apply(eval_stack)
+
+        if len(eval_stack) != 1:
+            raise ValueError("Formula application failed.")
+
+        return Sample(metric_ts, eval_stack[0])
+
+
+class FormulaBuilder:
```
matthias-wende-frequenz: Even though it's small, I think that this class deserves it's own file.

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
shsms: I was originally planning to build engines from component graph from the outside,  need to think about it separately.  The whole package is private though,  so it won't affect the public API of the logical meter.

tests/timeseries/test_formula_engine.py
shsms: I've added some more tests.  Are there any more cases that you can think of?

src/frequenz/sdk/timeseries/logical_meter/logical_meter.py
```diff
@@ -0,0 +1,111 @@
...
+        """Run the formula repeatedly and send the results to a channel.
+
+        Args:
+            formula: The formula to run.
+            sender: A sender for sending the formula results to.
+        """
+        while msg := await formula.apply():
+            await sender.send(msg)
+
+    async def start_formula(
```
matthias-wende-frequenz: Somehow I feel that the formula in the function name is not needed and `start` alone would be enough.

matthias-wende-frequenz (commented):
Just a couple of minor things. Otherwise it looks very good!

src/frequenz/sdk/timeseries/logical_meter/logical_meter.py
shsms: updated.

src/frequenz/sdk/timeseries/logical_meter/logical_meter.py
shsms: this was also a bit outdated because we've removed the `grid_power`, etc. methods temporarily.  Updated the comment.

src/frequenz/sdk/timeseries/logical_meter/logical_meter.py
shsms: done.

src/frequenz/sdk/timeseries/logical_meter/logical_meter.py
shsms: removed.

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
shsms: done

src/frequenz/sdk/timeseries/logical_meter/logical_meter.py
shsms: I'm not sure.  This would be the method that actors would use to start formulas,  and it would just start formulas.  So it would look like: `logical_meter.start_formula(...)`.  If we have `logical_meter.start(...)`,  that looks like we're trying to start the logical meter,  and not just one formula.

matthias-wende-frequenz: Makes sense. I haven't looked at it from that perspective ;).

leandro-lucarella-frequenz (commented):
Partial review ~ until the `FormulaEngine`

src/frequenz/sdk/timeseries/logical_meter/__init__.py
leandro-lucarella-frequenz: Ah, already updated.

src/frequenz/sdk/timeseries/logical_meter/__init__.py
leandro-lucarella-frequenz: Ah, already updated.

src/frequenz/sdk/timeseries/logical_meter/__init__.py
shsms: `from sdk.timeseries.logical_meter import LogicalMeter` is now possible.  `from sdk.timeseries import LogicalMeter` is running into cyclic imports for some reason.

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
shsms: I don't know,  I like the current approach better,  because we'd need a constructor anyway,  and one of the fields is exposed,  but not part of the constructor.  Otherwise we'd be making code harder to read in order to save a few lines.

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
shsms: true,  linter apparently complains only if we use `type` as a regular variable name,  no problem using it as a struct field.  updated.

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
shsms: yup, done.

src/frequenz/sdk/timeseries/logical_meter/__init__.py
leandro-lucarella-frequenz: I guess it comes from here:

Cyclic import (src.frequenz.sdk.actor -> src.frequenz.sdk.actor._data_sourcing -> src.frequenz.sdk.actor._data_sourcing.microgrid_api_source -> src.frequenz.sdk.timeseries -> src.frequenz.sdk.timeseries._logical_meter -> src.frequenz.sdk.timeseries._logical_meter._logical_meter -> src.frequenz.sdk.timeseries._logical_meter._formula_builder)

(there are more similar, I get these errors using `nox -R -s pylint`; strangely it is reported for `noxfile.py`, but I guess it's only because it is the last file in the command-line)

`_formula_builder` is importing `from ...actor import ChannelRegistry, ComponentMetricRequest`. I guess we always need to use the *leaf* nodes internally, it is a bit annoying as it will break more stuff if we move stuff internally but I guess it is fair trade-off to have a nice public interface.

So it should be `from ...actor._channel_registry import ChannelRegistry; from ..actor._microgrid_api_source import ComponentMetricRequest` and in `_microgrid_api_source` it should be `from ...timeseries._sample import Sample`.

leandro-lucarella-frequenz: Another solution would be to make package dependencies a DAG instead of a graph, so there are no cyclic dependencies between packages. In this case it looks like anything that it is using `actor`s shouldn't live in `timeseries`. So the `actor` package knows about `timeseries` but `timeseries` doesn't know about the `actor` package.

For now I would go with importing leaf modules only, to avoid having to rethink the package structure again, but in the long term I think not having cycling dependencies among internal packages would be the way to go (when restructuring the `micrigrid` package, did exactly this, it was based in which sub-modules knew about which other sub-modules, and this is why `graph` didn't ended up in the `component` sub-package, even when it is a *component* graph, but it implied cyclic dependencies between `client` and `component`).

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
leandro-lucarella-frequenz: For me the advantage of having `@dataclass(frozen=True)` is not the saving of a couple of lines but to make it very clear that it is an immutable class (when I read the decorator I immediately know that without reading all the code for it). But I agree that if it is not **really** a `dataclass`, but just an *immutable* class and you need to do some tricks to hide some attributes, then it might not be worth it.

src/frequenz/sdk/timeseries/logical_meter/__init__.py
leandro-lucarella-frequenz: With these changes it works:

<details>
<summary>Details</summary>


```console
 $ git status 
On branch logical-meter
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        modified:   src/frequenz/sdk/actor/_data_sourcing/microgrid_api_source.py
        modified:   src/frequenz/sdk/actor/_resampling.py
        modified:   src/frequenz/sdk/timeseries/__init__.py
        renamed:    src/frequenz/sdk/timeseries/logical_meter/__init__.py -> src/frequenz/sdk/timeseries/_logical_meter/__init__.py
        renamed:    src/frequenz/sdk/timeseries/logical_meter/_formula_builder.py -> src/frequenz/sdk/timeseries/_logical_meter/_formula_builder.py
        renamed:    src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py -> src/frequenz/sdk/timeseries/_logical_meter/_formula_engine.py
        renamed:    src/frequenz/sdk/timeseries/logical_meter/_formula_steps.py -> src/frequenz/sdk/timeseries/_logical_meter/_formula_steps.py
        renamed:    src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py -> src/frequenz/sdk/timeseries/_logical_meter/_logical_meter.py
        renamed:    src/frequenz/sdk/timeseries/logical_meter/_tokenizer.py -> src/frequenz/sdk/timeseries/_logical_meter/_tokenizer.py
```

```diff
diff --git a/src/frequenz/sdk/actor/_data_sourcing/microgrid_api_source.py b/src/frequenz/sdk/actor/_data_sourcing/microgrid_api_source.py
index f7fad2f..bad1895 100644
--- a/src/frequenz/sdk/actor/_data_sourcing/microgrid_api_source.py
+++ b/src/frequenz/sdk/actor/_data_sourcing/microgrid_api_source.py
@@ -20,7 +20,7 @@ from ...microgrid.component import (
     InverterData,
     MeterData,
 )
-from ...timeseries import Sample
+from ...timeseries._sample import Sample
 from .._channel_registry import ChannelRegistry
 
 
diff --git a/src/frequenz/sdk/actor/_resampling.py b/src/frequenz/sdk/actor/_resampling.py
index 604553c..2d45231 100644
--- a/src/frequenz/sdk/actor/_resampling.py
+++ b/src/frequenz/sdk/actor/_resampling.py
@@ -12,7 +12,8 @@ from typing import Dict, Sequence
 from frequenz.channels import Receiver, Sender
 from frequenz.channels.util import MergeNamed, Select, Timer
 
-from ..timeseries import GroupResampler, ResamplingFunction, Sample
+from ..timeseries._resampler import GroupResampler, ResamplingFunction
+from ..timeseries._sample import Sample
 from ._channel_registry import ChannelRegistry
 from ._data_sourcing import ComponentMetricRequest
 from ._decorator import actor
diff --git a/src/frequenz/sdk/timeseries/__init__.py b/src/frequenz/sdk/timeseries/__init__.py
index f52c7c5..a853e42 100644
--- a/src/frequenz/sdk/timeseries/__init__.py
+++ b/src/frequenz/sdk/timeseries/__init__.py
@@ -12,9 +12,11 @@ This module provides tools to operate on timeseries.
 
 from ._resampler import GroupResampler, Resampler, ResamplingFunction
 from ._sample import Sample
+from ._logical_meter import LogicalMeter
 
 __all__ = [
     "GroupResampler",
+    "LogicalMeter",
     "Resampler",
     "ResamplingFunction",
     "Sample",
diff --git a/src/frequenz/sdk/timeseries/logical_meter/__init__.py b/src/frequenz/sdk/timeseries/_logical_meter/__init__.py
similarity index 100%
rename from src/frequenz/sdk/timeseries/logical_meter/__init__.py
rename to src/frequenz/sdk/timeseries/_logical_meter/__init__.py
diff --git a/src/frequenz/sdk/timeseries/logical_meter/_formula_builder.py b/src/frequenz/sdk/timeseries/_logical_meter/_formula_builder.py
similarity index 96%
rename from src/frequenz/sdk/timeseries/logical_meter/_formula_builder.py
rename to src/frequenz/sdk/timeseries/_logical_meter/_formula_builder.py
index 7bd5c36..559b2ab 100644
--- a/src/frequenz/sdk/timeseries/logical_meter/_formula_builder.py
+++ b/src/frequenz/sdk/timeseries/_logical_meter/_formula_builder.py
@@ -6,7 +6,8 @@
 
 from frequenz.channels import Receiver, Sender
 
-from ...actor import ChannelRegistry, ComponentMetricRequest
+from ...actor._channel_registry import ChannelRegistry
+from ...actor._data_sourcing.microgrid_api_source import ComponentMetricRequest
 from ...microgrid.component import ComponentMetricId
 from .._sample import Sample
 from ._formula_engine import FormulaEngine
diff --git a/src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py b/src/frequenz/sdk/timeseries/_logical_meter/_formula_engine.py
similarity index 100%
rename from src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
rename to src/frequenz/sdk/timeseries/_logical_meter/_formula_engine.py
diff --git a/src/frequenz/sdk/timeseries/logical_meter/_formula_steps.py b/src/frequenz/sdk/timeseries/_logical_meter/_formula_steps.py
similarity index 100%
rename from src/frequenz/sdk/timeseries/logical_meter/_formula_steps.py
rename to src/frequenz/sdk/timeseries/_logical_meter/_formula_steps.py
diff --git a/src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py b/src/frequenz/sdk/timeseries/_logical_meter/_logical_meter.py
similarity index 96%
rename from src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py
rename to src/frequenz/sdk/timeseries/_logical_meter/_logical_meter.py
index ede2a43..8fd160f 100644
--- a/src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py
+++ b/src/frequenz/sdk/timeseries/_logical_meter/_logical_meter.py
@@ -9,7 +9,8 @@ from typing import Dict, List
 
 from frequenz.channels import Broadcast, Receiver, Sender
 
-from ...actor import ChannelRegistry, ComponentMetricRequest
+from ...actor._channel_registry import ChannelRegistry
+from ...actor._data_sourcing.microgrid_api_source import ComponentMetricRequest
 from ...microgrid.component import ComponentMetricId
 from .._sample import Sample
 from ._formula_builder import FormulaBuilder
diff --git a/src/frequenz/sdk/timeseries/logical_meter/_tokenizer.py b/src/frequenz/sdk/timeseries/_logical_meter/_tokenizer.py
similarity index 100%
rename from src/frequenz/sdk/timeseries/logical_meter/_tokenizer.py
rename to src/frequenz/sdk/timeseries/_logical_meter/_tokenizer.py
```

</details>

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
```diff
@@ -0,0 +1,200 @@
...
+            fetcher = self._metric_fetchers[name]
+        self._steps.append(fetcher)
+
+    def finalize(self) -> None:
+        """Finalize the formula engine.
+
+        This function must be called before calls to `apply` can be made.
+        """
+        while self._build_stack:
+            self._steps.append(self._build_stack.pop())
```
leandro-lucarella-frequenz, line 116: I'm not crazy about these patterns where you need to call X before calling Y, as it makes usage more complicated and error prone.

I wonder if it wouldn't make more sense that, for example, `finalize()` returns another object that only has the `apply()` function. Then this dependency is automatically enforced.

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
```diff
@@ -0,0 +1,200 @@
...
+        ready_metrics, pending = await asyncio.wait(
+            [
+                asyncio.create_task(fetcher.fetch_next(), name=name)
+                for name, fetcher in self._metric_fetchers.items()
+            ],
+            return_when=asyncio.ALL_COMPLETED,
+        )
+
+        if pending or any(res.result() is None for res in iter(ready_metrics)):
+            raise ValueError("Some resampled metrics didn't arrive")
```
leandro-lucarella-frequenz: `RuntimeError`?

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
```diff
@@ -0,0 +1,200 @@
...
+        for metric_ts, name in metrics_by_ts.items():
+            if metric_ts == latest_ts:
+                continue
+            fetcher = self._metric_fetchers[name]
+            while metric_ts < latest_ts:
+                next_val = await fetcher.fetch_next()
+                assert next_val is not None
+                metric_ts = next_val.timestamp
+            if metric_ts > latest_ts:
+                raise ValueError(
```
leandro-lucarella-frequenz: `RuntimeError`?

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
```diff
@@ -0,0 +1,200 @@
...
+        Raises:
+            ValueError: when some streams have no value, or when the synchronization of
+                timestamps fails.
+        """
+        metrics_by_ts: Dict[datetime, str] = {}
+        for metric in metrics:
+            result = metric.result()
+            name = metric.get_name()
+            if result is None:
+                raise ValueError(f"Stream closed for component: {name}")
```
leandro-lucarella-frequenz: `RuntimeError`?

src/frequenz/sdk/timeseries/logical_meter/__init__.py
shsms: Except downstream users will also have to import `from timeseries._sample import Sample` in their actors.  Or maybe not.

There is one more option which I like better:  move 'LogicalMeter' outside,  into a `data_pipeline` module.  `LogicalMeter` is not a timeseries primitive,  although it does produce timeseries,  just like some of the actors.  `BatteryPool`, etc can go inside `data_pipeline` as well.

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
```diff
@@ -0,0 +1,200 @@
...
+        else:
+            res = next(iter(ready_metrics)).result()
+            assert res is not None
+            metric_ts = res.timestamp
+
+        for step in self._steps:
+            step.apply(eval_stack)
+
+        if len(eval_stack) != 1:
+            raise ValueError("Formula application failed.")
```
leandro-lucarella-frequenz: `RuntimeError`?

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
```diff
@@ -0,0 +1,200 @@
...
+            metric_ts = await self.synchronize_metric_timestamps(ready_metrics)
+        else:
+            res = next(iter(ready_metrics)).result()
+            assert res is not None
+            metric_ts = res.timestamp
+
+        for step in self._steps:
+            step.apply(eval_stack)
+
+        if len(eval_stack) != 1:
```
leandro-lucarella-frequenz, line 198: Maybe add a comment about the result being left as the only value in the stack? It wasn't complete obvious to me on the first read.

src/frequenz/sdk/timeseries/logical_meter/_formula_steps.py
```diff
@@ -0,0 +1,197 @@
...
+        """Push the latest value from the stream into the evaluation stack.
+
+        Args:
+            eval_stack: An evaluation stack, to apply the formula step on.
+
+        Raises:
+            ValueError: No next value available to append.
+        """
+        if self._next_value is None or self._next_value.value is None:
+            raise ValueError("No next value available to append.")
```
leandro-lucarella-frequenz: `RuntimeError`?

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
```diff
@@ -0,0 +1,200 @@
...
+
+        Args:
+            name: A name for the metric.
+            data_stream: A receiver to fetch this metric from.
+        """
+        if name not in self._metric_fetchers:
+            fetcher = MetricFetcher(name, data_stream)
+            self._metric_fetchers[name] = fetcher
+        else:
+            fetcher = self._metric_fetchers[name]
```
leandro-lucarella-frequenz: `dict.setdefault()` is your friend :hugs: 

```suggestion
        fetcher = self._metric_fetchers.setdefault(name, MetricFetcher(name, data_stream))
```

src/frequenz/sdk/timeseries/logical_meter/_formula_steps.py
```diff
@@ -0,0 +1,197 @@
...
+        To be called before each call to `apply`.
+
+        Returns:
+            The fetched Sample.
+        """
+        self._next_value = await self._stream.receive()
+        return self._next_value
+
+    @property
+    def next_value(self) -> Optional[Sample]:
```
leandro-lucarella-frequenz: This doesn't seem to be used externally, so maybe remove the property?

leandro-lucarella-frequenz (commented):
Nice! Mostly very minor stuff, only one design-related comment, but being on internal classes it is also not very major.

I didn't look at the tests yet though, but since Matthias did I assume all major stuff is tested. If I find some time I'll try to have a look at them anyways.

src/frequenz/sdk/timeseries/logical_meter/__init__.py
leandro-lucarella-frequenz: > Except downstream users will also have to import from timeseries._sample import Sample in their actors. Or maybe not.

I don't think so.

> There is one more option which I like better: move 'LogicalMeter' outside, into a data_pipeline module. LogicalMeter is not a timeseries primitive, although it does produce timeseries, just like some of the actors. BatteryPool, etc can go inside data_pipeline as well.

For me it is in the realm of timeseries, as it takes timeseries as input (+ a formula explaining how to combine those timeseries) and produces a timeseries as output.

leandro-lucarella-frequenz:
```python
>>> from frequenz.sdk.timeseries import LogicalMeter
>>> from frequenz.sdk.actor import DataSourcingActor
>>> from frequenz.sdk.timeseries import Sample
>>> from frequenz.sdk.actor import actor
>>> 
```

It works because no we resolved the cyclic imports internally, as long as the user doesn't introduce more cycles themselves, the SDK shouldn't introduce any.

But again for the future, I would avoid having cyclic sub-package dependencies altogether. And maybe a way is to move stuff in `timeseries` that is using the `actor` package somewhere else, but I would leave that restructuring for the future, when we have more stuff so we can picture a better structure instead of doing futurology now.

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
shsms: made an issue: https://github.com/frequenz-floss/frequenz-sdk-python/issues/82

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
shsms: done

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
shsms: done

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
shsms: done

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
shsms: done

shsms: done

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
shsms: done

src/frequenz/sdk/timeseries/logical_meter/_formula_steps.py
shsms: done

src/frequenz/sdk/timeseries/logical_meter/_formula_steps.py
shsms: done

src/frequenz/sdk/timeseries/logical_meter/__init__.py
shsms: thanks for the patch! done

shsms: looks like it is still broken though.  luckily I created a separate commit.  Maybe move `Sample` to near `ComponentMetricRequest` or `ComponentMetricRequest` near `Sample`?

shsms: Is the patch that you provided the final version that works for you?  Or were there changes after you pasted it here?

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
shsms: probably can be done when we move `FormulaEngine` outside so it can be reused with `BatteryPool`

src/frequenz/sdk/timeseries/logical_meter/__init__.py
shsms: @leandro-lucarella-frequenz I've created a separate PR with your patch and removed it from this PR,  so that this can be taken in,  and we can look into the paths next week: https://github.com/frequenz-floss/frequenz-sdk-python/pull/83

@matthias-wende-frequenz or @leandro-lucarella-frequenz could you please approve this PR whoever sees this first?

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
leandro-lucarella-frequenz: Ah, the class was not even really immutable though, because `_pos` changes, so never mind.

src/frequenz/sdk/timeseries/logical_meter/__init__.py
leandro-lucarella-frequenz: Ah, sorry, I didn't run the whole tests, I only did the quick import check. It looks like you are right, because Python needs to initialize the package `actor` to import, for example `actor._data_sourcing...` so it still runs into cyclic dependencies in some cases.

I guess there is no way around to avoid cyclic dependencies between sub-packages then, which will require us to shuffle things around a bit and clearly separate packages based on what they depend on, or have a more nested public interface where parent packages don't expose symbols of sub-packages.

Anyway, definitely something requiring more discussion and not for this PR.

leandro-lucarella-frequenz: * #84

leandro-lucarella-frequenz (approved):

PR #75, opened 2022-11-23
Expose additional fields from `ComponentData` streams

### LINKED ISSUE #37: BMS temperature data support
### What's needed?

Battery temperature data is needed for lynch in order to decide whether to (dis)charge early within a 15min window.  Lynch will use the **max** temperature of all batteries in the microgrid. Hence only one value(highest temperature out of all batteries) is necessary for this usecase.

### Proposed solution

Either one value outputted which gives the max temperature or temperature of each battery. Both solutions work for this usecase.

### Use cases

Peak shaving (GMM)

_No response_

### Additional context

Even though for peak shaving only the max temperature is required, usecases in the future might need separate temperatures for each battery.

### LINKED ISSUE #48: Add frequency to meter data
### What's needed?

Add frequency to meter data

### Proposed solution

`frequency = msg.inner.__dict__["_raw"].meter.data.ac.frequency.value` should be part of meter data rather than needing to use `_raw`.

### Use cases

FCR

`frequency = msg.inner.__dict__["_raw"].meter.data.ac.frequency.value`

### Additional context

_No response_

shsms:
This PR exposes the following fields:
 - `frequency` from `MeterData` objects
 - `temperature_max` from `BatteryData` objects

This PR also temporarily switches to an older version of the channels library.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -23,6 +23,10 @@ time.
 
 <!-- Here goes the main new features and examples or instructions on how to use them -->
 
+* `MeterData` objects now expose the AC `frequency` measured by the meter.
+* `BatteryData` objects now expose the temperature of the hottest block in the
+  battery as `temperature_max`
+
 ## Bug Fixes
 
 <!-- Here goes notable bug fixes that are worth a special mention or explanation -->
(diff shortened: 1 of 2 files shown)


leandro-lucarella-frequenz (dismissed):
Maybe add an entry to the release notes.

sahas-subramanian-frequenz:
> Maybe add an entry to the release notes.

done.

leandro-lucarella-frequenz (approved):

PR #83, opened 2022-11-25
Make `logical_meter` module private

shsms:


leandro-lucarella-frequenz:
We are dropping this in favor of a more long term solution as discussed in #84.

Issue #84, opened 2022-11-25
Resolve potential cyclic import issues


leandro-lucarella-frequenz:
To have a clean public interface we are defining some top level sub-packages that expose classes in some sub-packages or sub-modules, but some of these top-level packages have or can have cyclic dependencies. For example, with the introduction of the `LogicalMeter`, which uses the `DataSourcingActor`, we have the `actor` package depending on the `timeseries` packages and viceversa (currently we are not exposing the `LogicalMeter` directly in `timeseries` because of this.

There are 2 main paths that I can see here:

1. Avoid cyclic dependencies between packages.
2. Don't expose sub-modules symbols in packages (and use them instead as namespaces for other modules).

I noticed some other issues with approach 1., like in `actor`, it doesn't look right to expose  classes supporting actor implementations (like `Request` / `Response` in `PowerDistirbutingActor` or `ComponentMetricRequest` in `DataSourcingActor`) and it makes sense to have a namespace for these (this is actually why the `PowerDistributingActor` ended up with its own sub-module).

This also can allow for shorter class names, for example: `actor.power_distributing.Actor`, which also plays nicely with Google coding standard (which we are supposed to be using) if we use `from frequenz.sdk.actor import power_distributing; pd = power_distributing.Actor(...); pd_sender.send(power_distributing.Request(....))`.

It is a bit of a drastic change, but it might make sense.

Adopting 1. might look as a less drastic change, but in reality it introduces artificial limitations in how we want to structure our API.

_Coming from @leandro-lucarella-frequenz in https://github.com/frequenz-floss/frequenz-sdk-python/pull/66#discussion_r1032685146_

leandro-lucarella-frequenz:
Related to:
* #83 (which follows option 1)

leandro-lucarella-frequenz:
I'm more and more convinced that option **2** is the way to go, as I think we'll keep bumping into these issues again and again if we go with option **1**. Also as the SDK grows, I think approach **1** will end up with packages with too many symbols in them, making it even more likely for cyclic dependencies to happen. This goes beyond the discussion about if classes used to communicate with actors should live in the actor module or not.

I know this would mean another change in the package structure, and I take the blame on it, I think I did the previous round too rushed to have the cleanup before the release without thinking it more thoroughly.

[...]

PR #86, opened 2022-11-28
Make `logical_meter` module private

leandro-lucarella-frequenz:
This is based on #83, adding an extra commit to move `.timeseries.Sample` to `.core.Sample`.
    
This is mainly to break a dependency cycle between actors and timeseries utilities. Any actor that work on timeseries needs to `import` `Sample`, which lived in `.timeseries`, but `.timeseries` also import all utilities working on timeseries (like `_resampler` and `_logical_meter`) that in turn use some actors.

Moving Sample to a (really) leaf module solves the issue.

That said, I consider this a temporary patch and not a long term solution, to find a long term solution we have this issue:

* #84


leandro-lucarella-frequenz:
We are dropping this in favor of a more long term solution as discussed in #84.

PR #92, opened 2022-11-29
PowerDistributor should not block user command.

### LINKED ISSUE #91: PowerDistributor blocks user commands if all components are considered as broken
### What happened?

PowerDistributor blocks user commands if all components are considered as broken.

Better detection if components is broken or not: https://github.com/frequenz-floss/frequenz-sdk-python/issues/33
As a workaround in this PR:
If all batteries are considered as broken, then user should anyway be able to use them.

### Affected version(s)

v0.12.0

### Affected part(s)

Battery power distribution (part:power-distribution)

### Extra information

_No response_

ela-kotulska-frequenz:
This is workaround for the bug.
Previously if all components were considered as broken, then all user commands were blocked for 30 seconds. With this workaround if all components are considered as broken then we will try to use them anyway.

### DIFF
--- a/src/frequenz/sdk/actor/power_distributing.py
+++ b/src/frequenz/sdk/actor/power_distributing.py
@@ -91,6 +91,34 @@ class _BrokenComponents:
         """
         self._timeout_sec = timeout_sec
 
+    def get_working_subset(self, components_ids: Set[int]) -> Set[int]:
+        """Get subset of batteries that are not marked as broken.
+
+        If all given batteries are broken, then mark them as working and return them.
+        This is temporary workaround to not block user command.
+
+        Args:
+            components_ids: set of component ids
+
+        Returns:
+            Subset of given components_ids with working components.
+        """
+        working = set(filter(lambda cid: not self.is_broken(cid), components_ids))
+
+        if len(working) == 0:
+            _logger.warning(
+                "All requested components: %s are marked as broken. "
+                "Marking them as working to not block command.",
+                str(components_ids),
+            )
+
+            for cid in components_ids:
+                self._broken.pop(cid, None)
+
+            working = components_ids
+
+        return working
+
     def is_broken(self, component_id: int) -> bool:
         """Check if component is marked as broken.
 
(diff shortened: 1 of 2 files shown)


tests/actor/test_power_distributing.py
```diff
@@ -738,3 +740,43 @@ async def test_power_distributor_stale_all_components_message(
...
+            20 * [datetime.fromisoformat("2001-01-01T00:00:31+00:00")]
+        )
+
+        datetime_mock.now.side_effect = expected_datetime
+
+        # After 30 seconds components should be considered as working
+        broken = _BrokenComponents(30)
+
+        for i in range(3):
+            broken.mark_as_broken(i)
```
sahas-subramanian-frequenz: interesting,  pylint always complains to me when I use single letter variable names.

src/frequenz/sdk/actor/power_distributing.py
```diff
@@ -91,6 +91,34 @@ def update_retry(self, timeout_sec: float) -> None:
...
+        """Get subset of batteries that are not marked as broken.
+
+        If all given batteries are broken, then mark them as working and return them.
+        This is temporary workaround to not block user command.
+
+        Args:
+            components_ids (Set[int]): set of component ids
+
+        Returns:
+            Set[int]: subset of given components_ids with working components.
```
sahas-subramanian-frequenz: we are not specifying types for args or returns anymore.

tests/actor/test_power_distributing.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/actor/power_distributing.py
ela-kotulska-frequenz: True, I have to set it in VsCode.

ela-kotulska-frequenz: Done

tests/actor/test_power_distributing.py
ela-kotulska-frequenz: Maybe it is related with python version number?
What python you are using?

sahas-subramanian-frequenz: it is fine,  I was just worried that pylint might not be running on the test files for some reason and we are not catching other issues as well.  I will test it separately.

sahas-subramanian-frequenz (approved):

tests/actor/test_power_distributing.py
leandro-lucarella-frequenz: it probably have an exception for `for` loops.

PR #95, opened 2022-11-30
Support handling `None` values in LogicalMeter/FormulaEngine

shsms:
The Resampling actor sends `None` values out when there is no data from a component.  Currently the logical meter raises an exception if it sees a `None` value from any of its streams,  which kills the LogicalMeter.  This is incorrect.

`None` values should instead be treated either as `None` or `0` based on context.

To this end, a new parameter `nones_are_zeros` is introduced in the LogicalMeter and FormulaEngine.

If this field is False,  then when there are None values in any of the input streams,  the calculated values will be None.

When it is True, the None values will be treated as 0.0 when the formula is evaluated.

### DIFF
--- a/src/frequenz/sdk/timeseries/logical_meter/_formula_steps.py
+++ b/src/frequenz/sdk/timeseries/logical_meter/_formula_steps.py
@@ -27,7 +27,7 @@ class FormulaStep(ABC):
         """
 
     @abstractmethod
-    def apply(self, eval_stack: List[float]) -> None:
+    def apply(self, eval_stack: List[Optional[float]]) -> None:
         """Apply a formula operation on the eval_stack.
 
         Args:
@@ -46,7 +46,7 @@ class Adder(FormulaStep):
         """
         return "+"
 
-    def apply(self, eval_stack: List[float]) -> None:
+    def apply(self, eval_stack: List[Optional[float]]) -> None:
         """Extract two values from the stack, add them, push the result back in.
 
         Args:
@@ -54,7 +54,11 @@ class Adder(FormulaStep):
         """
         val2 = eval_stack.pop()
         val1 = eval_stack.pop()
-        eval_stack.append(val1 + val2)
+        if val1 is None or val2 is None:
+            res = None
+        else:
+            res = val1 + val2
+        eval_stack.append(res)
 
 
 class Subtractor(FormulaStep):
@@ -68,7 +72,7 @@ class Subtractor(FormulaStep):
         """
         return "-"
 
-    def apply(self, eval_stack: List[float]) -> None:
+    def apply(self, eval_stack: List[Optional[float]]) -> None:
         """Extract two values from the stack, subtract them, push the result back in.
 
         Args:
@@ -76,7 +80,11 @@ class Subtractor(FormulaStep):
         """
         val2 = eval_stack.pop()
         val1 = eval_stack.pop()
-        eval_stack.append(val1 - val2)
+        if val1 is None or val2 is None:
+            res = None
+        else:
+            res = val1 - val2
+        eval_stack.append(res)
 
 
 class Multiplier(FormulaStep):
@@ -90,7 +98,7 @@ class Multiplier(FormulaStep):
         """
         return "*"
 
-    def apply(self, eval_stack: List[float]) -> None:
+    def apply(self, eval_stack: List[Optional[float]]) -> None:
         """Extract two values from the stack, multiply them, push the result back in.
 
         Args:
@@ -98,7 +106,11 @@ class Multiplier(FormulaStep):
         """
         val2 = eval_stack.pop()
         val1 = eval_stack.pop()
-        eval_stack.append(val1 * val2)
+        if val1 is None or val2 is None:
+            res = None
+        else:
+            res = val1 * val2
+        eval_stack.append(res)
 
 
 class Divider(FormulaStep):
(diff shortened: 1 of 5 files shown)


matthias-wende-frequenz (approved):

leandro-lucarella-frequenz:
Playing a bit devils advocate here, but:

> When it is True, the None values will be treated as 0.0 when the formula is evaluated.

Isn't this repeating the conceptual issue with `MicrogridData` about adding a default value? I guess this is added to be able to reuse the `FormulaEngine` by the pools, but the idea with the pools was to exclude faulty components from the formula altogether, not having a formula with a value replaced with 0.

Specially if formulas support general multiplication, it is starting to feel like we are building our own sympy here.  True, limited and stream-based, but still much more generic than we talked in the design talks.

I wonder if we are going in the right direction or if we'll end up with the same issues as with `MicrogridData` but using our own thing.

src/frequenz/sdk/timeseries/logical_meter/_formula_steps.py
```diff
@@ -46,15 +46,19 @@ def __repr__(self) -> str:
...
+    def apply(self, eval_stack: List[Optional[float]]) -> None:
         """Extract two values from the stack, add them, push the result back in.
 
         Args:
             eval_stack: An evaluation stack, to apply the formula step on.
         """
         val2 = eval_stack.pop()
         val1 = eval_stack.pop()
-        eval_stack.append(val1 + val2)
+        if val1 is None or val2 is None:
```
leandro-lucarella-frequenz, line 57: While you are at this, it should probably be a good idea to also treat *NaN*'s specially (probably the same as `None`). You can check using `math.isnan()`.

(same for all steps)

**OR**, maybe replace `None` with *NaN* and it should have the same effect of propagating `None` manually as you are doing here, but it will be done automatically. Then you only need to convert the final *NaN* in the stack to `None` and you are set. This option is easier to write (as you don't have to touch every step) but it might be harder read (to figure out what's going on).

src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py
```diff
@@ -60,7 +60,7 @@ async def _engine_from_formula_string(
             self._resampler_subscription_sender,
             metric_id,
         )
-        return await builder.from_string(formula)
+        return await builder.from_string(formula, False)
```
leandro-lucarella-frequenz: Reading this, I recommend making the `nones_are_zeros` arguments a keyword-only argument (`def f(..., *, nones_are_zeros: bool`), otherwise is very hard to know what's going on on the caller site without having to go see the function definition.

leandro-lucarella-frequenz (commented):
I see the `LogicalMeter` also accepts the option `nones_are_zeros`, which makes me even more confused why are we even allowing this?

sahas-subramanian-frequenz:
> but still much more generic than we talked in the design talks.

yes,  but it is the generic bring-your-own-formula version,  where we are not making any decisions anyway.

As soon as we have specific formula methods,  this flexibility can go away.  The `FormulaEngine` is internal anyway.

>but the idea with the pools was to exclude faulty components from the formula altogether, not having a formula with a value replaced with 0.

yes,  but we exclude the faulty components by replacing them with 0.  This is trivial if the formulas are generated by us from the component graph,  but if it a formula is provided by the user,  there's nothing else we can do anyway.  We just have to make sure the division-by-zero doesn't kill the logical meter, and instead send `None` if that's the case and print a warning.

That's why the resampler keeps sending `None` values at the same interval as the other values. And it would mean that devices get added back as soon as they become available again,  with no extra checks/code.

src/frequenz/sdk/timeseries/logical_meter/_formula_steps.py
sahas-subramanian-frequenz: Because this PR is already merged,  I'm making separate issues for your comments: https://github.com/frequenz-floss/frequenz-sdk-python/issues/99

src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py
sahas-subramanian-frequenz: https://github.com/frequenz-floss/frequenz-sdk-python/issues/100

sahas-subramanian-frequenz:
> I see the LogicalMeter also accepts the option nones_are_zeros, which makes me even more confused why are we even allowing this?

it is only temporary, so that it can be used to calculate battery metrics as well,  until *Pools are ready.

leandro-lucarella-frequenz:
> it is only temporary, so that it can be used to calculate battery metrics as well, until *Pools are ready.

Ah, OK, I see, then it would be good to add an issue about removing this when the time comes, as it is something that could be overlooked and stay there forever. I would also add in the `Args:` docs that this flags shouldn't be used and will be removed in the future?

> yes, but we exclude the faulty components by replacing them with 0. This is trivial if the formulas are generated by us from the component graph, but if it a formula is provided by the user, there's nothing else we can do anyway. We just have to make sure the division-by-zero doesn't kill the logical meter, and instead send `None` if that's the case and print a warning.
> 
> That's why the resampler keeps sending None values at the same interval as the other values. And it would mean that devices get added back as soon as they become available again, with no extra checks/code.

But I thought for pools we are not allowing formulas, just a list of components that take part in the pool, there is no need to use any formulas for pools.

As long his is all temporary to be able to fill the gap until we have the *Pool classes, otherwise maybe we should have a talk to regroup to make sure we are all heading in the same direction.

Issue #31, opened 2022-10-14
Add a class that manages a buffer to get all the needed resampled data


leandro-lucarella-frequenz:
### What's needed?

We need to be able to wait until all metrics for a particular timestamp of resampled data have arrived. We know they will eventually arrive, but we need to store the metrics somewhere until we have all to do the needed calculation.

### Proposed solution

Add a class that manages this buffer. This class should own the buffer that holds all the data points that arrived so far for a particular timestamp, and have a method to tell when all the expected data points arrived for this particular timestamp, so whatever calculation needs to be made with the data can be done.

### Use cases

Many users of the resampling actor, like the `LogicalMeter` or `BatteryPool` will need to do this.

### Additional context

Example draft implementation:

```py
@dataclass
class Sample:
    timestamp: datetime
    value: float

class ResampledTimeSeriesWindow:
    _samples: Dict[TimeSeriesId, Sample]

    def __init__(self):
        self._samples = {}

    def process_sample(self, timeseries_id: TimeSeriesId, sample: Sample) -> Optional[Dict[TimeSeriesId, Sample]]:
        result = None
        if self._samples and self._samples[0].timestamp != sample.timestamp:
            result = self._samples
            self._samples = {}
        self._samples[timeseries_id] = sample
        return result

window = ResampledTimeSeriesWindow()

while True:
    ts, sample = get_sample()
    results = window.process_sample(ts, sample)
    if results:
        # We have a complete window
        process_window(results)
```

leandro-lucarella-frequenz:
FYI @sahas-subramanian-frequenz in case you can come up with something while addressing #66

sahas-subramanian-frequenz:
I already have a solution for doing that in #66:  https://github.com/frequenz-floss/frequenz-sdk-python/pull/66/commits/0b613ae7c325acac917396f7c9612ee202e6be74#diff-a1aa5eb2cd80b9e0c2cc053be5369a7e7abd327594ebec482827295f54534997R67-R87

It is very modular,  but also specific to scenarios where multiple metrics are needed for a single formula,  so we should be able to reuse it in BatteryPool also.  I will write up some documentation for it soon.

leandro-lucarella-frequenz:
I don't think that's what this issue is about. This issue is about basically passing a set of sample sources and get one resample window of data at the time.

leandro-lucarella-frequenz:
I think a simpler interface would be:

```py
[...]

PR #96, opened 2022-12-01
Split the formula execution logic from the formula building logic 

### LINKED ISSUE #82: Refactor FormulaEngine to provide a safer interface
> I'm not crazy about these patterns where you need to call X before calling Y, as it makes usage more complicated and error prone.

I wonder if it wouldn't make more sense that, for example, `finalize()` returns another object that only has the `apply()` function. Then this dependency is automatically enforced.

_Originally posted by @leandro-lucarella-frequenz in https://github.com/frequenz-floss/frequenz-sdk-python/pull/66#discussion_r1032460254_

shsms:
Splits the `FormulaEngine` class into:
  - `FormulaBuilder`
  - `FormulaEngine`

The `FormulaBuilder` class is used to build formulas, and has a
`build` method that returns a `FormulaEngine` instance.

Closes https://github.com/frequenz-floss/frequenz-sdk-python/issues/82

### DIFF
--- a/src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
+++ b/src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
@@ -35,94 +35,23 @@ _operator_precedence = {
 class FormulaEngine:
     """A post-fix formula engine that operates on `Sample` receivers.
 
-    Operators and metrics need to be pushed into the engine in in-fix order, and they
-    get rearranged into post-fix in the engine.  This is done using the [Shunting yard
-    algorithm](https://en.wikipedia.org/wiki/Shunting_yard_algorithm).
-
-    Example:
-        To create an engine that adds the latest entries from two receivers, the
-        following calls need to be made:
-
-        ```python
-        engine = FormulaEngine()
-        engine.push_metric("metric_1", receiver_1)
-        engine.push_oper("+")
-        engine.push_metric("metric_2", receiver_2)
-        engine.finalize()
-        ```
-
-        and then every call to `engine.apply()` would fetch a value from each receiver,
-        add the values and return the result.
+    Use the `FormulaBuilder` to create `FormulaEngine` instances.
     """
 
     def __init__(
-        self,
+        self, steps: List[FormulaStep], metric_fetchers: Dict[str, MetricFetcher]
     ) -> None:
-        """Create a `FormulaEngine` instance."""
-        self._steps: List[FormulaStep] = []
-        self._build_stack: List[FormulaStep] = []
-        self._metric_fetchers: Dict[str, MetricFetcher] = {}
+        """Create a `FormulaEngine` instance.
+
+        Args:
+            steps: Steps for the engine to execute, in post-fix order.
+            metric_fetchers: Fetchers for each metric stream the formula depends on.
+        """
+        self._steps = steps
+        self._metric_fetchers = metric_fetchers
         self._first_run = True
 
-    def push_oper(self, oper: str) -> None:
-        """Push an operator into the engine.
-
-        Args:
-            oper: One of these strings - "+", "-", "*", "/", "(", ")"
-        """
-        if self._build_stack and oper != "(":
-            op_prec = _operator_precedence[oper]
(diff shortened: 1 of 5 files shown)


src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
```diff
@@ -35,94 +35,23 @@
...
-        engine = FormulaEngine()
-        engine.push_metric("metric_1", receiver_1)
-        engine.push_oper("+")
-        engine.push_metric("metric_2", receiver_2)
-        engine.finalize()
-        ```
-
-        and then every call to `engine.apply()` would fetch a value from each receiver,
-        add the values and return the result.
+    Use the `FormulaBuilder` to create `FormulaEngine` instances.
```
leandro-lucarella-frequenz, line 38: Maybe it is worth adding a cross-reference?

```suggestion
    Use the [`FormulaBuilder`][frequenz.sdk.timeseries.logical_meter.FormulaBuilder] to create `FormulaEngine` instances.
```

leandro-lucarella-frequenz (commented):
Nice! I like it! Makes the `FormulaEngine` less cluttered too.

But please add a `RELEASE_NOTES` entry.

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
shsms: But Formula* are hidden classes,  this will be a broken reference.

shsms:
> But please add a `RELEASE_NOTES` entry.

but it is internal refactoring, with no impact on user experience.  do we want to add that category of changes to release notes as well?

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
leandro-lucarella-frequenz: Ah, OK, but at some point they will be exposed, right? Or the idea is for them to be always internal?

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
```diff
@@ -191,7 +120,7 @@ async def apply(self) -> Sample:
             raise RuntimeError("Some resampled metrics didn't arrive")
 
         if self._first_run:
-            metric_ts = await self.synchronize_metric_timestamps(ready_metrics)
+            metric_ts = await self._synchronize_metric_timestamps(ready_metrics)
```
ela-kotulska-frequenz, line 123: If user would run `apply` method twice without await, then it might happen that `synchronize_metric_timestamps` will be called twice.

Could you instead move setting `_first_run` here (before `await`):
```        
if self._first_run:
     self._first_run = False
     metric_ts = await self._synchronize_metric_timestamps(ready_metrics)
```

sahas-subramanian-frequenz: We just have to make sure it gets run at least once.  Additional times is fine,  but the one time is essential.  So we can't assume it finished,  because it could raise an exception in rare cases as well.  We'll at some point start calling it once every `N` messages or minutes,  just to make sure the stream remains synchronized.

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
sahas-subramanian-frequenz: I don't know of a reason to make it available to everyone yet.  So I guess it remains private until we have a reason to expose it?

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
leandro-lucarella-frequenz: Just FYI, off-topic but related, while writing the 2nd stage example, I also bumped into this issue (having to make sure I'm collecting samples from different sources but for the same resampling window), so I think it is worth pursuing #31 independently from the `FormulaEngine` (actually `FormulaEngine` could use it too).

ela-kotulska-frequenz (approved):

PR #97, opened 2022-12-02
Fix the disable_actor_auto_restart fixture

### LINKED ISSUE #41: Pytest fixture to change actors `restart_limit` doesn't work as expected
### What happened?

Originally created by @ela-kotulska-frequenz 

We added fixture to change the `restart_limit` in actors, but we didn't enable it for all actors in unit tests.

It looks like if we use this fixture, we can't pass any arguments to the tests functions.

Example:

https://github.com/frequenz-floss/frequenz-sdk-python/blob/20e1dffadcb728443190e1dcfae45d68545ba962/tests/configs/test_config_manager.py

If `class TestActorConfigManager:` derives after (`IsolatedAsyncioTestCase`) then the tests fails with error:
```
self = <tests.mos_framework.configs.test_config_manager.TestActorConfigManager testMethod=test_update>
func = <bound method TestActorConfigManager.test_update of <tests.mos_framework.configs.test_config_manager.TestActorConfigManager testMethod=test_update>>
args = (), kwargs = {}

    def _callMaybeAsync(self, func, /, *args, **kwargs):
        assert self._asyncioTestLoop is not None
>       ret = func(*args, **kwargs)
E       TypeError: test_update() missing 1 required positional argument: 'config_file'
```

The `restart_limit` is overridden correctly.

### Affected version(s)

All

### Affected part(s)

### Extra information

Fixture to change `restart_limit`:

https://github.com/frequenz-floss/frequenz-sdk-python/blob/20e1dffadcb728443190e1dcfae45d68545ba962/tests/conftest.py#L18-L31

leandro-lucarella-frequenz:
The fixture didn't always work because in cases where the actor kept running past the pytest session end, and failed then, the original restart limit was reestablished, running into the infinite loop again.

Maybe related to #41.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -15,13 +15,12 @@ SAFETY_TIMEOUT = 10.0
 def disable_actor_auto_restart():  # type: ignore
     """Disable auto-restart of actors while running tests.
 
-    Since this is auto-use, the yield part (and restore of the variable) is not
-    strictly needed, but we leave it as an example.
+    At some point we had a version that would set the limit back to the
+    original value but it doesn't work because some actors will keep running
+    even after the end of the session and fail after the original value was
+    reestablished, getting into an infinite loop again.
 
     Note: Test class must derive after unittest.IsolatedAsyncioTestCase.
     Otherwise this fixture won't run.
     """
-    original_restart_limit = _decorator.BaseActor.restart_limit
     _decorator.BaseActor.restart_limit = 0
-    yield
-    _decorator.BaseActor.restart_limit = original_restart_limit


leandro-lucarella-frequenz:
Hopefully it fixes #41, ~even when I wrote that issue I don't fully understand what is it really about, as the error that is shown in the issue description doesn't seem to match very well the title...~

It talks about some test deriving from `IsolatedAsyncioTestCase` but then the link doesn't point to a test deriving from it, so I think something got messed up when I wrote it.

In any case, I bumped with an issue with this fixture, which took me a long time to figure out because I kept running into the infinite loop, so this at least fixes something.

ela-kotulska-frequenz (dismissed):

leandro-lucarella-frequenz:
BTW @ela-kotulska-frequenz it seems like you wrote #41 originally. Re-reading it and looking at Jakub's comment, it doesn't look like this is fixing it. At least Jakub said one needs to use `IsolatedAsyncioTestCase` for the auto-fixture to be used.

If you could check if the error described in #41 still happens before and after applying this fix it might be useful to know if we can close #41 or not.

PR #103, opened 2022-12-06
Generate formulas automatically from the component graph

### LINKED ISSUE #102: Automatic formula generation from the component graph

shsms:
Closes #102

### DIFF
--- a/src/frequenz/sdk/microgrid/component/_component.py
+++ b/src/frequenz/sdk/microgrid/component/_component.py
@@ -7,8 +7,50 @@ from __future__ import annotations
 
 from dataclasses import dataclass
 from enum import Enum
+from typing import Optional
 
+import frequenz.api.microgrid.inverter_pb2 as inverter_pb
 import frequenz.api.microgrid.microgrid_pb2 as microgrid_pb
+
+
+class ComponentType(Enum):
+    """A base class from which individual component types are derived."""
+
+
+class InverterType(ComponentType):
+    """Enum representing inverter types."""
+
+    NONE = inverter_pb.Type.TYPE_UNSPECIFIED
+    BATTERY = inverter_pb.Type.TYPE_BATTERY
+    SOLAR = inverter_pb.Type.TYPE_SOLAR
+    HYBRID = inverter_pb.Type.TYPE_HYBRID
+
+
+def _component_type_from_protobuf(
+    component_category: microgrid_pb.ComponentCategory.ValueType,
+    component_type: inverter_pb.Type.ValueType,
+) -> Optional[ComponentType]:
+    """Convert a protobuf InverterType message to Component enum.
+
+    For internal-only use by the `microgrid` package.
+
+    Args:
+        component_category: category the type belongs to.
+        component_type: protobuf enum to convert.
+
+    Returns:
+        Enum value corresponding to the protobuf message.
+    """
+    # ComponentType values in the protobuf definition are not unique across categories
+    # as of v0.11.0, so we need to check the component category first, before doing any
+    # component type checks.
+    if component_category == microgrid_pb.ComponentCategory.COMPONENT_CATEGORY_INVERTER:
+        if not any(t.value == component_type for t in InverterType):
+            return None
+
+        return InverterType(component_type)
+
+    return None
 
 
 class ComponentCategory(Enum):
@@ -62,6 +104,7 @@ class Component:
 
     component_id: int
     category: ComponentCategory
+    type: Optional[ComponentType] = None
 
     def is_valid(self) -> bool:
         """Check if this instance contains valid data.
--- a/src/frequenz/sdk/timeseries/logical_meter/_formula_generators/_battery_power_formula.py
+++ b/src/frequenz/sdk/timeseries/logical_meter/_formula_generators/_battery_power_formula.py
@@ -1,0 +1,52 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+
+"""Formula generator from component graph for Grid Power."""
+
+from .....sdk import microgrid
+from ....microgrid.component import ComponentCategory, ComponentMetricId, InverterType
+from .._formula_engine import FormulaEngine
+from ._formula_generator import FormulaGenerator
+
+
+class BatteryPowerFormula(FormulaGenerator):
+    """Creates a formula engine from the component graph for calculating grid power."""
+
+    async def generate(
+        self,
+    ) -> FormulaEngine:
+        """Make a formula for the cumulative AC battery power of a microgrid.
+
+        The calculation is performed by adding the Active Powers of all the inverters
+        that are attached to batteries.
+
+        If there's no data coming from an inverter, that inverter's power will be
+        treated as 0.
+
+        Returns:
+            A formula engine that will calculate cumulative battery power values.
+
+        Raises:
+            RuntimeError: if there are no batteries in the component graph, or if they
+                don't have an inverter as a predecessor.
+        """
+        builder = self._get_builder(ComponentMetricId.ACTIVE_POWER)
+        component_graph = microgrid.get().component_graph
+        battery_inverters = list(
+            comp
+            for comp in component_graph.components()
+            if comp.category == ComponentCategory.INVERTER
+            and comp.type == InverterType.BATTERY
+        )
+
+        if not battery_inverters:
+            raise RuntimeError(
+                "Unable to find any battery inverters in the component graph."
+            )
+
+        for idx, comp in enumerate(battery_inverters):
+            if idx > 0:
+                builder.push_oper("+")
+            await builder.push_component_metric(comp.component_id, nones_are_zeros=True)
+
+        return builder.build()
--- a/src/frequenz/sdk/timeseries/logical_meter/_formula_generators/_battery_power_formula.py
+++ b/src/frequenz/sdk/timeseries/logical_meter/_formula_generators/_battery_power_formula.py
@@ -6,7 +6,7 @@
 from .....sdk import microgrid
 from ....microgrid.component import ComponentCategory, ComponentMetricId, InverterType
 from .._formula_engine import FormulaEngine
-from ._formula_generator import FormulaGenerator
+from ._formula_generator import ComponentNotFound, FormulaGenerator
 
 
 class BatteryPowerFormula(FormulaGenerator):
@@ -27,8 +27,10 @@ class BatteryPowerFormula(FormulaGenerator):
             A formula engine that will calculate cumulative battery power values.
 
         Raises:
-            RuntimeError: if there are no batteries in the component graph, or if they
-                don't have an inverter as a predecessor.
+            ComponentNotFound: if there are no batteries in the component graph, or if
+                they don't have an inverter as a predecessor.
+            FormulaGenerationError: If a battery has a non-inverter predecessor
+                in the component graph.
         """
         builder = self._get_builder(ComponentMetricId.ACTIVE_POWER)
         component_graph = microgrid.get().component_graph
@@ -40,7 +42,7 @@ class BatteryPowerFormula(FormulaGenerator):
         )
 
         if not battery_inverters:
-            raise RuntimeError(
+            raise ComponentNotFound(
                 "Unable to find any battery inverters in the component graph."
             )
 
--- a/src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py
+++ b/src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py
@@ -17,7 +17,12 @@ from ...microgrid import ComponentGraph
 from ...microgrid.component import ComponentMetricId
 from .. import Sample
 from ._formula_engine import FormulaEngine
-from ._formula_generators import BatteryPowerFormula, FormulaGenerator, GridPowerFormula
+from ._formula_generators import (
+    BatteryPowerFormula,
+    FormulaGenerator,
+    GridPowerFormula,
+    PVPowerFormula,
+)
 from ._resampled_formula_builder import ResampledFormulaBuilder
 
 logger = logging.Logger(__name__)
@@ -170,3 +175,15 @@ class LogicalMeter:
 
         """
         return await self._get_formula_stream("battery_power", BatteryPowerFormula)
+
+    async def pv_power(self) -> Receiver[Sample]:
+        """Fetch the PV power production in the microgrid.
+
+        If a formula engine to calculate PV power production is not
+        already running, it will be started.  Else, we'll just get a new
+        receiver to the already existing data stream.
+
+        Returns:
+            A *new* receiver that will stream PV power production values.
+        """
+        return await self._get_formula_stream("pv_power", PVPowerFormula)
--- a/src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py
+++ b/src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py
@@ -3,17 +3,21 @@
 
 """A logical meter for calculating high level metrics for a microgrid."""
 
+from __future__ import annotations
+
 import asyncio
 import logging
 import uuid
-from typing import Dict, List
+from typing import Dict, List, Type
 
 from frequenz.channels import Broadcast, Receiver, Sender
 
 from ...actor import ChannelRegistry, ComponentMetricRequest
+from ...microgrid import ComponentGraph
 from ...microgrid.component import ComponentMetricId
 from .. import Sample
 from ._formula_engine import FormulaEngine
+from ._formula_generators import BatteryPowerFormula, FormulaGenerator, GridPowerFormula
 from ._resampled_formula_builder import ResampledFormulaBuilder
 
 logger = logging.Logger(__name__)
@@ -35,6 +39,7 @@ class LogicalMeter:
         self,
         channel_registry: ChannelRegistry,
         resampler_subscription_sender: Sender[ComponentMetricRequest],
+        component_graph: ComponentGraph,
     ) -> None:
         """Create a `LogicalMeter instance`.
 
@@ -43,6 +48,7 @@ class LogicalMeter:
                 actor.
             resampler_subscription_sender: A sender for sending metric requests to the
                 resampling actor.
+            component_graph: The component graph representing the microgrid.
         """
         self._channel_registry = channel_registry
         self._resampler_subscription_sender = resampler_subscription_sender
@@ -50,7 +56,7 @@ class LogicalMeter:
         # Use a randomly generated uuid to create a unique namespace name for the local
         # meter to use when communicating with the resampling actor.
         self._namespace = f"logical-meter-{uuid.uuid4()}"
-
+        self._component_graph = component_graph
         self._output_channels: Dict[str, Broadcast[Sample]] = {}
         self._tasks: List[asyncio.Task[None]] = []
 
@@ -118,3 +124,49 @@ class LogicalMeter:
             )
         )
         return out_chan.new_receiver()
+
+    async def _get_formula_stream(
+        self,
+        channel_key: str,
+        generator: Type[FormulaGenerator],
+    ) -> Receiver[Sample]:
+        if channel_key in self._output_channels:
+            return self._output_channels[channel_key].new_receiver()
+
+        formula_engine = await generator(
+            self._namespace, self._channel_registry, self._resampler_subscription_sender
+        ).generate()
+        out_chan = Broadcast[Sample](channel_key)
+        self._output_channels[channel_key] = out_chan
+        self._tasks.append(
+            asyncio.create_task(
+                self._run_formula(formula_engine, out_chan.new_sender())
+            )
+        )
+        return out_chan.new_receiver()
+
+    async def grid_power(self) -> Receiver[Sample]:
+        """Fetch the grid power for the microgrid.
+
+        If a formula engine to calculate grid power is not already running, it
+        will be started.  Else, we'll just get a new receiver to the already
+        existing data stream.
+
+        Returns:
+            A *new* receiver that will stream grid_power values.
+
+        """
+        return await self._get_formula_stream("grid_power", GridPowerFormula)
+
+    async def battery_power(self) -> Receiver[Sample]:
+        """Fetch the cumulative battery power in the microgrid.
+
+        If a formula engine to calculate cumulative battery power is not
+        already running, it will be started.  Else, we'll just get a new
+        receiver to the already existing data stream.
+
+        Returns:
+            A *new* receiver that will stream battery_power values.
+
+        """
+        return await self._get_formula_stream("battery_power", BatteryPowerFormula)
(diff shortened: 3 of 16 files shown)


shsms:
This PR currently implements just one `grid_power` formula.  I will add more in new commits,  but the other parts are already ready for a first look.

src/frequenz/sdk/microgrid/component/_component.py
```diff
@@ -62,6 +95,7 @@ class Component:
 
     component_id: int
     category: ComponentCategory
+    type: ComponentType = InverterType.NONE
```
ela-kotulska-frequenz: It is missleading that Category type has InverterType as default.
Should we have any default at all?
Do you plan to add type for all components in this PR and then remove this default?
If yes, then please comment it for easier PR :)

src/frequenz/sdk/timeseries/logical_meter/_resampled_formula_builder.py
```diff
@@ -59,7 +59,7 @@ async def _get_resampled_receiver(self, component_id: int) -> Receiver[Sample]:
         return self._channel_registry.new_receiver(request.get_channel_name())
 
     async def push_component_metric(
-        self, component_id: int, nones_are_zeros: bool
+        self, component_id: int, *, nones_are_zeros: bool
```
ela-kotulska-frequenz, line 62: What * mean? It is not described.

src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py
```diff
@@ -107,3 +114,27 @@ async def start_formula(
...
+        out_chan = Broadcast[Sample](channel_key)
+        self._output_channels[channel_key] = out_chan
+        self._tasks.append(
+            asyncio.create_task(
+                self._run_formula(formula_engine, out_chan.new_sender())
+            )
+        )
+        return out_chan.new_receiver()
+
+    # async def battery_power(self) -> Receiver[Sample]:
```
ela-kotulska-frequenz: Unnecessary comment?

src/frequenz/sdk/microgrid/component/_component.py
shsms: Yes,  I'd like to make it not default,  but `Component` objects are getting created literally everywhere in the codebase,  and so many in the tests,  so removing the default will be a huge change.  I already tried,  but gave up (temporarily) after around 500 lines.

Don't want to add types of other components yet either,  won't work for now,  the microgrid api needs an update for this to work.
 
I will try and make it an `Optional` field,  with `None` as default value.  That should work I think.

src/frequenz/sdk/timeseries/logical_meter/_resampled_formula_builder.py
shsms: It is part of the Python syntax.  just means that all the arguments after the `*` are keyword-only i.e., they can only be passed with the name, like `push_component_metric(10, nones_are_zeros = False)`.

https://peps.python.org/pep-3102/

src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py
shsms: It was supposed to be for the next commit.  I've moved it now.

ela-kotulska-frequenz:
For me looks good.
Waiting for more formulas to continue review :)

src/frequenz/sdk/microgrid/component/_component.py
```diff
@@ -8,9 +8,42 @@
 from dataclasses import dataclass
 from enum import Enum
 
+import frequenz.api.microgrid.inverter_pb2 as inverter_pb
 import frequenz.api.microgrid.microgrid_pb2 as microgrid_pb
 
 
+class ComponentType(Enum):
+    """A ABC from which individual component types are derived."""
```
leandro-lucarella-frequenz: This is an enum, not a ABC.

src/frequenz/sdk/microgrid/component/_component.py
```diff
@@ -8,9 +8,42 @@
...
+
+
+class InverterType(ComponentType):
```
leandro-lucarella-frequenz, line 20: Any reason to derive from an empty `Enum`?

src/frequenz/sdk/microgrid/component/_component.py
```diff
@@ -8,9 +8,42 @@
...
+
+    NONE = inverter_pb.Type.TYPE_UNSPECIFIED
+    BATTERY = inverter_pb.Type.TYPE_BATTERY
+    SOLAR = inverter_pb.Type.TYPE_SOLAR
+    HYBRID = inverter_pb.Type.TYPE_HYBRID
+
+
+def _component_type_from_protobuf(
+    component_type: inverter_pb.Type.ValueType,
+) -> ComponentType:
```
leandro-lucarella-frequenz: Am I crazy or the user of this function will have to forcibly cast the return to `InverterType` to be of any use?

src/frequenz/sdk/timeseries/logical_meter/_formula_generators.py
```diff
@@ -0,0 +1,93 @@
...
+from ._formula_engine import FormulaEngine
+from ._resampled_formula_builder import ResampledFormulaBuilder
+
+
+class FormulaGenerator:
+    """A class for generating formulas from the component graph."""
+
+    def __init__(
+        self,
+        namespace: str,
```
leandro-lucarella-frequenz: Maybe `request_namespace` or something that makes it more obvious what is referring to? Not sure, just a suggestion.

src/frequenz/sdk/timeseries/logical_meter/_formula_generators.py
```diff
@@ -0,0 +1,93 @@
...
+        # directly connected to the grid.
+        for idx, comp in enumerate(grid_successors):
+            if idx > 0:
+                builder.push_oper("+")
+
+            # Ensure the device has an `ACTIVE_POWER` metric.
+            if comp.category == ComponentCategory.INVERTER:
+                nones_are_zeros = True
+            elif comp.category == ComponentCategory.METER:
+                nones_are_zeros = False
```
leandro-lucarella-frequenz: Maybe explain why wants `None`s as zeros and the other doesn't?

src/frequenz/sdk/timeseries/logical_meter/_formula_generators.py
```diff
@@ -0,0 +1,93 @@
...
+            channel_registry: A channel registry instance shared with the resampling
+                actor.
+            resampler_subscription_sender: A sender for sending metric requests to the
+                resampling actor.
+        """
+        self._channel_registry = channel_registry
+        self._resampler_subscription_sender = resampler_subscription_sender
+        self._namespace = namespace
+
+    async def grid_power_formula(
```
leandro-lucarella-frequenz: Is the idea to keep adding formulas to this `FormulaGenerator`? If so I would recommend to go for a design that allows us to add more formulas without having all of them in the same file/class, as it could grow very big, as it happened with `MicrogridData`. Maybe having one class per formula generator instead?

src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py
```diff
@@ -32,6 +37,7 @@ def __init__(
         self,
         channel_registry: ChannelRegistry,
         resampler_subscription_sender: Sender[ComponentMetricRequest],
+        component_graph: ComponentGraph,
```
leandro-lucarella-frequenz, line 47: Shouldn't this be taken from the `microgrid.get()` singleton so it can pick up updates to the component graph? In any case it is fine for me to leave it like this for now, but I'm asking more to think about how we want things to work in the future, to make sure we all point at the same direction.

In any case, using `microgrid.get().component_graph` will not also help if the component graph is updated, because then the formula should also be regenerated. I'm starting to think that we need an actor or something that can notify (to whoever is listening) about component graph changes.

src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py
```diff
@@ -107,3 +117,32 @@ async def start_formula(
...
+        out_chan = Broadcast[Sample](channel_key)
+        self._output_channels[channel_key] = out_chan
+        self._tasks.append(
+            asyncio.create_task(
+                self._run_formula(formula_engine, out_chan.new_sender())
+            )
+        )
+        return out_chan.new_receiver()
+
+    async def grid_power(self) -> Receiver[Sample]:
```
leandro-lucarella-frequenz, line 153: I would try to make it more clear that this is actually creating a new receiver. At the method name level I think it could be difficult without making it look **really** ugly, but at least it should be pointed out very clearly in the docs.

src/frequenz/sdk/timeseries/logical_meter/_formula_generators.py
```diff
@@ -91,3 +91,61 @@ async def grid_power_formula(
...
+    async def battery_power_formula(
+        self, component_graph: ComponentGraph
+    ) -> FormulaEngine:
+        """Make a formula for the cumulative AC battery power of a microgrid.
+
+        The calculation is performed by adding the Active Powers of all the inverters
+        that are attached to batteries.
+
+        If there's no data coming from an inverter, that inverter's power will be
+        treated as 0.
```
leandro-lucarella-frequenz: This will be moved to the batter pool eventually, right?

src/frequenz/sdk/timeseries/logical_meter/_formula_generators.py
```diff
@@ -91,3 +91,61 @@ async def grid_power_formula(
...
+                don't have an inverter as a predecessor.
+        """
+        batteries = list(
+            comp
+            for comp in component_graph.components()
+            if comp.category == ComponentCategory.BATTERY
+        )
+
+        if not batteries:
+            raise RuntimeError("Unable to find any batteries in the component graph.")
```
leandro-lucarella-frequenz: I would add an exception like `ComponentNotFound` for these cases. I feel we are using too many generic exceptions so if at some point we need to start catching these exceptions and acting on them depending on the underlying issue, we'll have a **lot** of work converting all these `RuntimeError`s to something more meaningful all at once.

src/frequenz/sdk/microgrid/component/_component.py
shsms: This is done.

src/frequenz/sdk/microgrid/component/_component.py
shsms: Fixed.

src/frequenz/sdk/microgrid/component/_component.py
shsms: Just for typing.  Then we can have a single `type` field in the `Component` object, and assign `InverterType` or `MeterType` values to it and mypy will still be happy.  That way we'd also get an interface that follows the protobuf definition somewhat.

src/frequenz/sdk/microgrid/component/_component.py
shsms: A typehint from a super class doesn't change the type of the object being returned, in Python.  So the function still returns `InverterType.BATTERY` values that's directly comparable.  You're forgetting that we're using Python,  so yes, you are crazy \o/

src/frequenz/sdk/timeseries/logical_meter/_formula_generators.py
shsms: I think we need to qualify it only when we move out of the data pipeline.  It has a shared meaning in the data pipeline code,  so I'd say this is fine.

src/frequenz/sdk/timeseries/logical_meter/_formula_generators.py
shsms: That makes sense,  I'm working on it.

src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py
shsms: We can also subscribe to config changes from the component graph.  I guess we'll need a sprint to update all parts of the sdk once cloud sync starts to support the sdk,  and we start using the config manager.

src/frequenz/sdk/microgrid/component/_component.py
leandro-lucarella-frequenz: ~But what I don't understand is, if you are using a method that returns `ComponentType`, then you can't access any of the `InverterType` enum members, right? This is the part I don't understand~ Never mind :man_facepalming:

shsms: it doesn't return a `ComponentType`.  it returns an `InverterType` or `MeterType`, etc.  The type hint says `ComponentType`  and that keeps mypy happy,  because the return values do implement `ComponentType` (because it is empty).

src/frequenz/sdk/microgrid/component/_component.py
leandro-lucarella-frequenz: Ah, no, I remember this is Python, but the type checker is still a type checker and it won't let you do this:

```py
ct = _component_type_from_protobuf(...)
if ct == ct.BATTERY:
    ...
```

for example, but of course this is not what you want, the trick is you are still always comparing with the specialized type member:

```py
ct = _component_type_from_protobuf(...)
if ct == InverterType.BATTERY:
    ...
```

So you don't need to access any members in the return type, just compare it.

So yeah, all good, sorry about the brain fart :dash: 🤭

src/frequenz/sdk/timeseries/logical_meter/_formula_generators.py
shsms: done

src/frequenz/sdk/timeseries/logical_meter/_formula_generators.py
shsms: I've updated the comment.

src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py
shsms: Updated the docstring.

src/frequenz/sdk/timeseries/logical_meter/_formula_generators.py
shsms: No,  the battery *power* formula will remain in the logical meter.  Only the soc, sop, capacity, etc will go to the battery pool.

src/frequenz/sdk/timeseries/logical_meter/_formula_generators.py
shsms: also done.

tests/microgrid/mock_api.py
```diff
@@ -77,10 +78,22 @@ def __init__(
         self._latest_discharge: Optional[PowerLevelParam] = None
 
     def add_component(
-        self, component_id: int, component_category: ComponentCategory.V
+        self,
+        component_id: int,
+        component_category: ComponentCategory.V,
+        inverter_type: InverterType.V = InverterType.TYPE_UNSPECIFIED,
```
ela-kotulska-frequenz, line 84: `component_category` is expected as raw protobuf value.
`inverter_type` should be python Enum.
And the method creates struct of python Enums.

Maybe it would be more consistent if `inverter_type` would by expected as raw protobuf, too?

tests/microgrid/mock_api.py
```diff
@@ -77,10 +78,22 @@ def __init__(
...
+        if component_category == ComponentCategory.COMPONENT_CATEGORY_INVERTER:
+            self._components.append(
+                Component(
+                    id=component_id, category=component_category, inverter=inverter_type
+                )
+            )
+        else:
+            self._components.append(
+                Component(id=component_id, category=component_category)
+            )
```
ela-kotulska-frequenz, line 96: If `inverter_type` would be renamed to `component_type` and it would be raw protobuf (as mentioned above), then you will be able to remove all `if` here and use  `_component_type_from_protobuf` (that you implemented in the first commit).

src/frequenz/sdk/timeseries/logical_meter/_formula_generators.py
```diff
@@ -0,0 +1,93 @@
...
+            ),
+            None,
+        )
+
+        if grid_component is None:
+            raise RuntimeError(
+                "Unable to find a GRID component from the component graph."
+            )
+
+        grid_successors = component_graph.successors(grid_component.component_id)
```
ela-kotulska-frequenz: But if umg_meter will be by load side, then grid_successor will have only one element: JUNCTION. And then formula below won't work.

tests/timeseries/test_logical_meter.py
```diff
@@ -130,22 +127,178 @@ def next_msg(value: float) -> ComponentData:
...
+        servicer = await self._setup_service(mocker)
+        servicer.add_component(
+            1, microgrid_pb2.ComponentCategory.COMPONENT_CATEGORY_GRID
+        )
+        servicer.add_component(
+            4, microgrid_pb2.ComponentCategory.COMPONENT_CATEGORY_METER
+        )
+        servicer.add_component(
+            7, microgrid_pb2.ComponentCategory.COMPONENT_CATEGORY_METER
+        )
```
ela-kotulska-frequenz: Why there is no JUNCTION in this graph?

src/frequenz/sdk/timeseries/logical_meter/_formula_generators.py
ela-kotulska-frequenz: > the battery power formula will remain in the logical meter. 

What if user would like to know active power for subset of batteries?

tests/microgrid/mock_api.py
shsms: it is the protobuf type already.  This is the mock server,  can't use our sdk types here.

tests/microgrid/mock_api.py
shsms: No,  the `inverter` field in `Component` is a `oneof` defined [here](https://github.com/frequenz-floss/frequenz-api-microgrid/blob/v0.x.x/proto/frequenz/api/microgrid/microgrid.proto#L363-L369).  We can set it only if it is an inverter.  Else it will overwrite the value we set for a `battery` for example.  So it has to be done this way.

src/frequenz/sdk/timeseries/logical_meter/_formula_generators/_battery_power_formula.py
```diff
@@ -0,0 +1,69 @@
...
+            raise RuntimeError("Unable to find any batteries in the component graph.")
+
+        battery_inverters = list(
+            next(iter(component_graph.predecessors(bat.component_id)))
+            for bat in batteries
+        )
+
+        if len(batteries) != len(battery_inverters):
+            raise RuntimeError("Can't find inverters for all batteries.")
+
```
ela-kotulska-frequenz: Why not to look for ComponentCategory.INVERTER?
And then iterate once over returned list to find inverters with ComponentType.BATTERY?

src/frequenz/sdk/timeseries/logical_meter/_formula_generators.py
shsms: No,  we won't have JUNCTION components anymore,  components will just be directly connected.  This is be the component graph format for all locations that get added to the UI.  And v0.13.0 or higher of the sdk will not be deployed to locations that are not added to the UI.

tests/timeseries/test_logical_meter.py
shsms: Junction components are going away.

src/frequenz/sdk/timeseries/logical_meter/_formula_generators.py
shsms: Then they'll have to provide a custom formula.

shsms:
I will add some more tests as well, to this PR.

src/frequenz/sdk/timeseries/logical_meter/_formula_generators/_battery_power_formula.py
shsms: They are both the same right?  Do you think it is better to start from the inverters for some reason?

shsms:
This is ready for review again.  Sorry about the large number of changes.

src/frequenz/sdk/timeseries/logical_meter/_formula_generators/_battery_power_formula.py
ela-kotulska-frequenz: I am telling that you don't need batteries, here. :) 
```
inverters = list(
            comp
            for comp in component_graph.components()
            if comp.category == ComponentCategory.INVERTER and comp.type == InverterType.BATTERY
        )
for idx, comp in enumerate(battery_inverters):
   ....
```

src/frequenz/sdk/timeseries/logical_meter/_formula_generators/_battery_power_formula.py
```diff
@@ -38,22 +44,26 @@ async def generate(
...
+            )
 
         battery_inverters = list(
             next(iter(component_graph.predecessors(bat.component_id)))
             for bat in batteries
         )
 
         if len(batteries) != len(battery_inverters):
-            raise RuntimeError("Can't find inverters for all batteries.")
+            raise ComponentNotFound("Can't find inverters for all batteries.")
```
ela-kotulska-frequenz: Don't you think it should be handled in ComponentGraph?

src/frequenz/sdk/timeseries/logical_meter/_formula_generators/_battery_power_formula.py
shsms: oh that's true.  will update.

shsms: done

src/frequenz/sdk/timeseries/logical_meter/_formula_generators/_battery_power_formula.py
shsms: this is now obsolete,  after https://github.com/frequenz-floss/frequenz-sdk-python/pull/103#discussion_r1050667778

src/frequenz/sdk/timeseries/logical_meter/_formula_generators/_battery_power_formula.py
```diff
@@ -0,0 +1,58 @@
...
+            )
+
+        for idx, comp in enumerate(battery_inverters):
+            if idx > 0:
+                builder.push_oper("+")
+
+            if comp.category != ComponentCategory.INVERTER:
+                raise RuntimeError(
+                    f"Found a battery predecessor that's not an inverter: {comp}"
+                )
```
ela-kotulska-frequenz: It should be removed, now :)

shsms: oh wow 🤦🏽 done

ela-kotulska-frequenz (approved):

leandro-lucarella-frequenz:
Ah, this looks very nice with the new exceptions and one class per formula :heart:

PR #104, opened 2022-12-06
Add fine-grained `Result` types for the `PowerDistributingActor`

ela-kotulska-frequenz:
Improve power distributing results.

### DIFF
--- a/src/frequenz/sdk/actor/__init__.py
+++ b/src/frequenz/sdk/actor/__init__.py
@@ -3,7 +3,6 @@
 
 """A base class for creating simple composable actors."""
 
-from . import power_distributing
 from ._channel_registry import ChannelRegistry
 from ._config_managing import ConfigManagingActor
 from ._data_sourcing import ComponentMetricRequest, DataSourcingActor
@@ -17,5 +16,4 @@ __all__ = [
     "ConfigManagingActor",
     "DataSourcingActor",
     "actor",
-    "power_distributing",
 ]
--- a/src/frequenz/sdk/actor/power_distributing/__init__.py
+++ b/src/frequenz/sdk/actor/power_distributing/__init__.py
@@ -10,7 +10,17 @@ It also provides all the secondary features that should be used to communicate w
 PowerDistributingActor and send requests for charging or discharging power.
 """
 
-from .power_distributing import PowerDistributingActor, Result
+from .power_distributing import PowerDistributingActor
 from .request import Request
+from .result import Error, Ignored, OutOfBound, PartialFailure, Result, Success
 
-__all__ = ["PowerDistributingActor", "Request", "Result"]
+__all__ = [
+    "PowerDistributingActor",
+    "Request",
+    "Result",
+    "Error",
+    "Success",
+    "Ignored",
+    "OutOfBound",
+    "PartialFailure",
+]
--- a/src/frequenz/sdk/actor/power_distributing/__init__.py
+++ b/src/frequenz/sdk/actor/power_distributing/__init__.py
@@ -1,0 +1,15 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+
+"""This module provides feature to set power between many batteries.
+
+Distributing power is very important to keep the microgrid ready
+for the power requirements.
+This module provides PowerDistributingActor that knows how to distribute power.
+It also provides all the secondary features that should be used to communicate with
+PowerDistributingActor and send requests for charging or discharging power.
+"""
+
+from .power_distributing import PowerDistributingActor, Request, Result
+
+__all__ = ["PowerDistributingActor", "Request", "Result"]
--- a/src/frequenz/sdk/actor/power_distributing/__init__.py
+++ b/src/frequenz/sdk/actor/power_distributing/__init__.py
@@ -10,6 +10,7 @@ It also provides all the secondary features that should be used to communicate w
 PowerDistributingActor and send requests for charging or discharging power.
 """
 
-from .power_distributing import PowerDistributingActor, Request, Result
+from .power_distributing import PowerDistributingActor, Result
+from .request import Request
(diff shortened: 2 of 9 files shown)


src/frequenz/sdk/actor/power_distributing/result.py
```diff
@@ -0,0 +1,111 @@
...
+            succeed_power: How much power out of requested were successfully set.
+            outside_bound: How much power was ignored, because was outside power
+                bound. If Request.adjust_power = False, then this value is always 0.
+        """
+        super().__init__(request)
+        self.succeed_power: int = succeed_power
+        self.outside_bound: int = outside_bound
+
+
+class AnyFail(Result):
```
shsms: Why is it called `AnyFail`?

src/frequenz/sdk/actor/power_distributing/result.py
```diff
@@ -0,0 +1,111 @@
...
+    ) -> None:
+        """Create class instance.
+
+        Args:
+            request: Initial_request
+            succeed_power: How much power was successfully set.
+            failed_power: How much power failed.
+            failed_batteries: What batteries failed.
+            outside_bound: How much power was ignored, because was outside power
+                bound. If Request.adjust_power = False, then this value is always 0.
```
shsms: Maybe call the field `excess_power`,  with description something like: "part of the requested power that could not be fulfilled, because it was outside available power bounds."?

and maybe you want to say `==` instead of a single =.

src/frequenz/sdk/actor/power_distributing/result.py
```diff
@@ -0,0 +1,111 @@
...
+
+        Args:
+            request: Initial_request
+            msg: Error message telling why error happened.
+        """
+        super().__init__(request)
+        self.msg: str = msg
+
+
+class OutOfBound(Result):
```
shsms, line 100: OutOfBounds?

sahas-subramanian-frequenz:
I tried to improve the title of the PR also,  but feel free to change it if it is not accurate.

src/frequenz/sdk/actor/power_distributing/result.py
ela-kotulska-frequenz: Example:
5 batteries are charged.
One of them fail ( it won't response or it will response with error).
Then AnyFail error will be send. Because at least one battery failed.

shsms: Does `PartialFailure` work better?

src/frequenz/sdk/actor/power_distributing/result.py
ela-kotulska-frequenz: Yes. By default PowerDistributingActor adjust the given power to the bounds.
But Tom told PowerDistributingActor should not adjust power.
I am not agree with it but I add extra field Request.adjust_power: bool. Default to True. If user will set it to False, and requested power will be > bounds, then request won't be executed and OutOfBound will be returned.

shsms: No,  I meant,  maybe you want to add an `s` at the end,  instead of just `OutOfBound`. :-)

but makes sense otherwise.

ela-kotulska-frequenz: No one is using it.
And with async user is not sure when the request will be executd, so it makes no sense.
So maybe I could remove it in some later feature.

ela-kotulska-frequenz: There is only one bound. 
One for upper bound and one for lower bound :)
In the  message I give only one Bound. So it make sense to make it `OutOfBound`

src/frequenz/sdk/actor/power_distributing/result.py
ela-kotulska-frequenz: Sounds better, thanks! Done.

src/frequenz/sdk/actor/power_distributing/result.py
ela-kotulska-frequenz: Thanks! Done

shsms:
LGTM,  I'll approve as soon as you rebase.

src/frequenz/sdk/actor/power_distributing/__init__.py
```diff
@@ -0,0 +1,8 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+
+"""A base module for creating and using power distributing features."""
```
shsms: I guess we should put the more detailed documentation here, because this would now become the top level docs.

https://frequenz-floss.github.io/frequenz-sdk-python/v0.14/reference/frequenz/sdk/actor/power_distributing/#frequenz.sdk.actor.power_distributing

shsms (commented):
lgtm,  except for one documentation change.

src/frequenz/sdk/actor/power_distributing/__init__.py
ela-kotulska-frequenz: Done :)

src/frequenz/sdk/actor/__init__.py
```diff
@@ -3,12 +3,12 @@
 
 """A base class for creating simple composable actors."""
 
-from . import power_distributing
 from ._channel_registry import ChannelRegistry
 from ._config_managing import ConfigManagingActor
 from ._data_sourcing import ComponentMetricRequest, DataSourcingActor
 from ._decorator import actor
 from ._resampling import ComponentMetricsResamplingActor
+from .power_distributing import power_distributing
```
shsms: Is this accurate?  This is like not using the `__init__.py` file that you made.

ela-kotulska-frequenz: Done

sahas-subramanian-frequenz (approved):

PR #105, opened 2022-12-07
Prevent LogicalMeter from crashing on formula application failure

shsms:


sahas-subramanian-frequenz (approved):
Approving myself because this is an urgent fix for production code,  and the tests pass,  and it is a trivial change.

src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py
```diff
@@ -71,8 +74,12 @@ async def _run_formula(
...
             sender: A sender for sending the formula results to.
         """
-        while msg := await formula.apply():
-            await sender.send(msg)
+        while True:
+            try:
+                msg = await formula.apply()
+                await sender.send(msg)
+            except Exception as err:  # pylint: disable=broad-except
+                logger.warning("Formula application failed: %s", err)
```
leandro-lucarella-frequenz, line 82: If you stop receiving this will get stuck in an infinite active loop sending `None` all over the place, right? I guess this is not what we want?

ela-kotulska-frequenz: 1. What failed? `send` or `apply` method?
If only `apply`, then I would use: `try-except-else`.

2. Luca is right. Shouldn't we send `None` in that case?

shsms: This also needs to handle asyncio.CancelledError.  I will make an update PR.

> If you stop receiving this will get stuck in an infinite active loop sending None all over the place, right? I guess this is not what we want?

I don't understand. In "If you stop receiving ...", do you mean 
  1. if the user stops receiving from the logical meter?  then we'll just keep sending the values to the channel and it will print warnings saying the channel is overflowing.
  2. if the logical meter stops receiving data from the resamplers because the resampler crashed (and didn't come back), then the logical meter would just block.
  3. The channels between the resampling actor and the logical meter are owned by the channel registry,  so unless people do nasty things,  the channels won't get closed.

shsms: > What failed? send or apply method?
> If only apply, then I would use: try-except-else.

only `apply` failed.  what should we put in the `else` block?

shsms: I've created https://github.com/frequenz-floss/frequenz-sdk-python/pull/106.

If there are more things from this discussion,  I'll add them to that PR as well.

leandro-lucarella-frequenz: I mean if `formula.apply()` returned `None`, which I guessed it meant you can't apply that formula anymore and you need to stop the loop, based on the old code, but without looking at `formula.apply()`, but now I see that it can only return `Sample`, so the old code could effectively only stop if an exception was raised. So never mind from my side.

Still, if the formula has a persistent error we'll just litter the logs if we blindly keep going forward, but I guess there is not much else that we can do about it (spamming or dying).

PR #116, opened 2022-12-13
Simple ringbuffer with tests

mathias-baumann-frequenz:
First draft, more tests to come, also performance test.
next step is testing the slicing parts.

Signed-off-by: Marenz <mathias.baumann@frequenz.com>

### DIFF
--- a/src/frequenz/sdk/util/ringbuffer.py
+++ b/src/frequenz/sdk/util/ringbuffer.py
@@ -1,0 +1,332 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+
+"""Ringbuffer implementation with focus on time & memory efficiency."""
+
+from __future__ import annotations
+
+from copy import deepcopy
+from datetime import datetime, timedelta
+from typing import Any, Generic, Sequence, TypeVar, Union
+
+import numpy as np
+
+T = TypeVar("T")
+
+Container = Union[list, np.ndarray]
+
+
+class RingBuffer(Generic[T]):
+    """A ring buffer with a fixed size.
+
+    Should work with most backends, tested with list and np.ndarrays.
+    """
+
+    def __init__(self, container: Container) -> None:
+        """Initialize the ring buffer with the given container.
+
+        Args:
+            container: Container to store the data in.
+        """
+        self._container = container
+        self._write_index = 0
+        self._read_index = 0
+        self._len = 0
+
+    def __len__(self) -> int:
+        """Get current amount of elements.
+
+        Returns:
+            The amount of items that this container currently holds.
+        """
+        return self._len
+
+    @property
+    def maxlen(self) -> int:
+        """Get the max length.
+
+        Returns:
+            The max amount of items this container can hold.
+        """
+        return len(self._container)
+
+    def push(self, value: T) -> int:
+        """Push a new value into the ring buffer.
+
+        Args:
+            value: Value to push into the ring buffer.
+
+        Returns:
+            The index in the ringbuffer.
+        """
+        if self._len == len(self._container):
+            # Move read position one forward, dropping the oldest written value
+            self._read_index = self._wrap(self._read_index + 1)
+        else:
+            self._len += 1
+
+        self._container[self._write_index] = value
+        value_index = self._write_index
+        self._write_index = self._wrap(self._write_index + 1)
+
+        return value_index
+
+    def pop(self) -> T:
+        """Remove the oldest value from the ring buffer and return it.
+
+        Raises:
+            IndexError: when no elements exist to pop.
+
+        Returns:
+            Oldest value found in the ring buffer.
+        """
+        if self._len == 0:
+            raise IndexError()
+
+        val = self._container[self._read_index]
+        self._read_index = (self._read_index + 1) % len(self._container)
+        self._len -= 1
+
+        return val
+
+    @property
+    def is_full(self) -> bool:
+        """Check if the container is full.
+
+        Returns:
+            True when the container is full.
+        """
+        return len(self) == len(self._container)
+
+    def __setitem__(self, index: int | slice, value: T | Sequence[T]) -> None:
+        """Write the given value to the requested position.
+
+        Args:
+            index: Position to write the value to.
+            value: Value to write.
+        """
+        if isinstance(index, int):
+            self._container[self._wrap(index + self._read_index)] = value
+        else:
+            wstart = self._wrap(index.start)
+            wstop = self._wrap(index.stop)
+            if wstop < wstart:
+                self._container[wstart:] = value[:self.maxlen-wstart]
+                self._container[:wstop] = value[self.maxlen-wstart:]
+            else
+                self._container[wstart:wstop] = value
+
+
+
+    def __getitem__(self, index_or_slice: int | slice) -> T | Container:
+        """Request a value or slice.
+
+        Does not support wrap-around or copying of data.
+
+        Args:
+            index_or_slice: Index or slice specification of the requested data
+
+        Returns:
+            the value at the given index or value range at the given slice.
+        """
+        return self._container[index_or_slice]
+
+    def _wrap(self, index: int) -> int:
+        return index % len(self._container)
+
+
+class OrderedRingBuffer(Generic[T]):
+    """Time aware ringbuffer that keeps its entries sorted time."""
+
+    def __init__(
+        self,
+        buffer: Any,
+        sample_rate: int,
+        window_border: datetime = datetime(1, 1, 1),
+    ) -> None:
+        """Initialize the time aware ringbuffer.
+
+        Args:
+            buffer: instance of a buffer container to use internally
+            sample_rate: resolution of the incoming timestamps in
+                seconds
+            window_border: datetime depicting point in time to use as border
+                beginning, useful to make data start at the beginning of the day or
+                hour.
+        """
+        self._buffer = buffer
+        self._sample_rate = sample_rate
+        self._window_start = window_border
+
+        self._missing_windows = []
+        self._datetime_newest = datetime.min
+        self._datetime_oldest = datetime.max
+
+    @property
+    def maxlen(self) -> int:
+        """Get the max length.
+
+        Returns:
+            the max amount of items this container can hold.
+        """
+        return len(self._buffer)
+
+    def update(self, timestamp: datetime, value: T, missing: bool = False) -> None:
+        """Update the buffer with a new value for the given timestamp.
+
+        Args:
+            timestamp: Timestamp of the new value
+            value: value to add
+            missing: if true, the given timestamp will be recorded as missing.
+                The value will still be written.
+
+            Returns:
+                Nothing.
+        """
+        # Update timestamps
+        self._datetime_newest = max(self._datetime_newest, timestamp)
+        self._datetime_oldest = min(self._datetime_oldest, timestamp)
+
+        if self._datetime_oldest < self._datetime_newest - timedelta(
+            seconds=len(self._buffer) * self._sample_rate
+        ):
+            self._datetime_oldest = self._datetime_newest - timedelta(
+                len(self._buffer) * self._sample_rate
+            )
+
+        # Update data
+        insert_index = self.datetime_to_index(timestamp)
+
+        self._buffer[insert_index] = value
+
+        # Update list of missing windows
+        #
+        # We always append to the last pending window.
+        # A window is pending when end is None
+        if missing:
+            # Create new if no pending window
+            if (
+                len(self._missing_windows) == 0
+                or self._missing_windows[-1].end is not None
+            ):
+                self._missing_windows.append({"start": timestamp, "end": None})
+        elif len(self._missing_windows) > 0:
+            # Finalize a pending window
+            if self._missing_windows[-1].end is None:
+                self._missing_windows[-1].end = timestamp
+
+        # Delete out-to-date windows
+        if len(self._missing_windows) > 0 and self._missing_windows[0].end is not None:
+            if self._missing_windows[0].end <= self._datetime_oldest:
+                self._missing_windows = self._missing_windows[1:]
+
+    def datetime_to_index(self, timestamp: datetime) -> int:
+        """Convert the given timestamp to an index.
+
+        Throws an index error when the timestamp is not found within this
+        buffer.
+
+        Args:
+            timestamp: Timestamp to convert.
+
+        Raises:
+            IndexError: when requesting a timestamp outside the range this container holds
+
+        Returns:
+            index where the value for the given timestamp can be found.
+        """
+        if self._datetime_newest < timestamp or timestamp < self._datetime_oldest:
+            raise IndexError(
+                f"Requested timestamp {timestamp} is is "
+                f"outside the range [{self._datetime_oldest} - {self._datetime_newest}]"
+            )
+
+        return self._wrap(int(abs((self._window_start - timestamp).total_seconds())))
+
+    def window(self, start: datetime, end: datetime, force_copy: bool = False) -> Container:
+        """Request a view on the data between start timestamp and end timestamp.
+
+        Will return a copy in the following cases:
+        * The requested time period is crossing the start/end of the buffer
+        * The requested time period contains missing entries.
+        * The force_copy parameter was set to True (default False)
+
+        This means, if the caller needs to modify the data to account for
+        missing entries, they can safely do so.
+
+        Args:
+            start: start time of the window
+            end: end time of the window
+
+        Returns:
+            the requested window
+        """
+        assert start < end
+
+        start_index = self.datetime_to_index(start)
+        end_index = self.datetime_to_index(end)
+
+        # Requested window wraps around the ends
+        if start_index > end_index:
+            window = self._buffer[start_index:]
+
+            if end_index > 0:
+                if isinstance(self._buffer, list):
+                    window += self._buffer[0:end_index]
+                else:
+                    window = np.concatenate((window, self._buffer[0:end_index]))
+            return window
+
+        def in_window(window):
+            if window.start <= start < window.end:
+                return True
+            if window.start <= end < window.end:
+                return True
+
+            return False
+
+        # Return a copy if there are none-values in the data
+        if any(map(in_window, self._missing_windows)) or force_copy:
+            return deepcopy(self._buffer[start_index:end_index])
+
+        return self._buffer[start_index:end_index]
+
+    def _wrap(self, index: int) -> int:
+        """Normalize the given index to fit in the buffer by wrapping it around.
+
+        Args:
+            index: index to normalize.
+
+        Returns:
+            an index that will be within maxlen.
+        """
+        return index % self.maxlen
+
+    def __getitem__(self, index_or_slice: int | slice) -> T | Sequence[T]:
+        """Get item or slice at requested position.
+
+        No wrapping of the index will be done.
+
+        Args:
+            index_or_slice: Index or slice specification of the requested data.
+
+        Returns:
+            The requested value or slice.
+        """
+        return self._buffer[index_or_slice]
+
+    def __len__(self) -> int:
+        """Return the amount of items that this container currently holds.
+
+        Returns:
+            The length.
+        """
+        if self._datetime_newest == datetime.min:
+            return 0
+
+        start_index = self.datetime_to_index(self._datetime_oldest)
+        end_index = self.datetime_to_index(self._datetime_newest)
+
+        if end_index < start_index:
+            return len(self._buffer) - start_index + end_index
+        return start_index - end_index
--- a/tests/utils/ringbuffer.py
+++ b/tests/utils/ringbuffer.py
@@ -1,0 +1,140 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+
+"""Tests for the `Ringbuffer` class."""
+
+from __future__ import annotations
+
+import random
+from datetime import datetime
+from itertools import cycle, islice
+from typing import TypeVar
+
+import numpy as np
+import pytest
+
+from frequenz.sdk.util.ringbuffer import OrderedRingBuffer, RingBuffer
+
+T = TypeVar("T")
+
+
+@pytest.mark.parametrize(
+    "buffer",
+    [
+        RingBuffer[int]([0] * 50000),
+        RingBuffer[int](np.empty(shape=(50000,), dtype=np.float64)),
+    ],
+)
+def test_simple_push_pop(buffer: RingBuffer[int]) -> None:
+    """Test simple pushing/popping of RingBuffer."""
+    for i in range(buffer.maxlen):
+        buffer.push(i)
+
+    assert len(buffer) == buffer.maxlen
+
+    for i in range(buffer.maxlen):
+        assert i == buffer.pop()
+
+
+@pytest.mark.parametrize(
+    "buffer",
+    [
+        RingBuffer[int]([0] * 50000),
+        RingBuffer[int](np.empty(shape=(50000,), dtype=np.float64)),
+    ],
+)
+def test_push_pop_over_limit(buffer: RingBuffer[int]) -> None:
+    """Test pushing over the limit and the expected loss of data."""
+    over_limit_pushes = 1000
+
+    for i in range(buffer.maxlen + over_limit_pushes):
+        buffer.push(i)
+
+    assert len(buffer) == buffer.maxlen
+
+    for i in range(buffer.maxlen):
+        assert i + over_limit_pushes == buffer.pop()
+
+    assert len(buffer) == 0
+
+
+@pytest.mark.parametrize(
+    "buffer, element_type",
+    [
+        (RingBuffer[int]([0] * 5000), int),
+        (RingBuffer[float](np.empty(shape=(5000,), dtype=np.float64)), float),
+    ],
+)
+def test_slicing(buffer: RingBuffer[T], element_type: type) -> None:
+    """Test slicing method."""
+    for i in range(buffer.maxlen):
+        buffer.push(element_type(i))
+
+    # Wrap in extra list() otherwise pytest complains about numpy arrays
+    # pylint: disable=protected-access
+    assert list(buffer._container) == list(range(buffer.maxlen))
+
+
+@pytest.mark.parametrize(
+    "buffer",
+    [
+        OrderedRingBuffer[int]([0] * 24, 1),
+        OrderedRingBuffer[float](np.empty(shape=(24,), dtype=np.float64), 1),
+    ],
+)
+def test_timestamp_ringbuffer(buffer: OrderedRingBuffer[float | int]) -> None:
+    """Test ordered ring buffer."""
+    size = buffer.maxlen
+
+    #    import pdb; pdb.set_trace()
+    random.seed(0)
+
+    # Push in random order
+    for i in random.sample(range(size), size):
+        buffer.update(datetime.fromtimestamp(200 + i), i)
+
+    # Check all possible window sizes and start positions
+    for i in range(size):
+        for j in range(1, size):
+            start = datetime.fromtimestamp(200 + i)
+            end = datetime.fromtimestamp(200 + j + i)
+
+            tmp = list(islice(cycle(range(0, size)), i, i + j))
+            assert list(buffer.window(start, end)) == list(tmp)
+
+
+@pytest.mark.parametrize(
+    "buffer",
+    [
+        (OrderedRingBuffer[float]([0] * 24, 1)),
+        (OrderedRingBuffer[float](np.empty(shape=(24,), dtype=np.float64), 1)),
+    ],
+)
+def test_timestamp_ringbuffer_overwrite(buffer: OrderedRingBuffer[float | int]) -> None:
+    """Test overwrite behavior and correctness."""
+    size = buffer.maxlen
+
+    #    import pdb; pdb.set_trace()
+    random.seed(0)
+
+    # Push in random order
+    for i in random.sample(range(size), size):
+        buffer.update(datetime.fromtimestamp(200 + i), i)
+
+    # Push the same amount twice
+    for i in random.sample(range(size), size):
+        buffer.update(datetime.fromtimestamp(200 + i), i * 2)
+
+    # Check all possible window sizes and start positions
+    for i in range(size):
+        for j in range(1, size):
+            start = datetime.fromtimestamp(200 + i)
+            end = datetime.fromtimestamp(200 + j + i)
+
+            tmp = islice(cycle(range(0, size * 2, 2)), i, i + j)
+            # assert list(buffer.window(start, end)) == list(tmp)
+            actual: float
+            for actual, expectation in zip(buffer.window(start, end), tmp):
+                assert actual == expectation
+
+            assert j == len(buffer.window(start, end))
--- a/benchmarks/util/benchmark_ringbuffer.py
+++ b/benchmarks/util/benchmark_ringbuffer.py
@@ -1,0 +1,168 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+
+"""Performance test for the `Ringbuffer` class."""
+
+import random
+import timeit
+from datetime import datetime, timedelta
+from typing import TypeVar
+
+import numpy as np
+
+from frequenz.sdk.util.ringbuffer import OrderedRingBuffer
+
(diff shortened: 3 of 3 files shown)


src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,111 @@
...
+import numpy as np
+from itertools import chain
+
+
+T = TypeVar('T')
+
+class AbstractRingBufferContainer(Generic[T], ABC):
+    def __init__(self, size):
+        self.size = size
+
```
ela-kotulska-frequenz: I wonder if we should define any max_size (even the large one) like 108000 (half of day in seconds).
We won't allow user to store data for half of day. 
@mathias-baumann-frequenz  @sahas-subramanian-frequenz  any opinion?

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,111 @@
...
+    @abstractmethod
+    def write(self, value: T, index) -> None:
```
ela-kotulska-frequenz: 1. `index` needs type :) 
2. Each argument should be described. Please look at code in this repo.

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,111 @@
...
+
+class ListContainer(AbstractRingBufferContainer, ABC):
+    def __init__(self, size: int):
+        AbstractRingBufferContainer.__init__(self, size)
+        self.data = list([T] * size)
+
+
+    def write(self, value: T, index: int) -> None:
+        """"""
+        self.data[index] = value
```
ela-kotulska-frequenz: 1. You can also override `__setitem__` :) 
2. Maybe check if index < len(self.data)?

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,111 @@
...
+    def __init__(self, size):
+        self.size = size
+
+    @abstractmethod
+    def write(self, value: T, index) -> None:
+        """Write to value to the requested index"""
+        pass
+
+    @abstractmethod
+    def __getitem__(self, index):
```
ela-kotulska-frequenz: index needs type.

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,111 @@
...
+        """Returns the value at the requested index"""
+        pass
+
+    def __len__(self):
+        return self.size
+
+class RingBuffer(Generic[T]):
+
+    # class syntax
```
ela-kotulska-frequenz: what means: `# class syntax`

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,111 @@
...
+        self.read_index = 0
+        self.size = 0
+
+    def __len__(self):
+        return self.size
+
+    def push(self, value: T) -> int:
+        """Pushes a new value into the ring buffer, returns the index in the ringbuffer"""
+
+        if self.size == self.container.size:
```
ela-kotulska-frequenz: You could override `len` operator for `AbstractRingBufferContainer`

src/frequenz/sdk/util/ringbuffer.py
Marenz: The size here is basically the max size, as it is a non-growing container that we pass to the ring queue

src/frequenz/sdk/util/ringbuffer.py
mathias-baumann-frequenz: Just a left over from a code example :D

src/frequenz/sdk/util/ringbuffer.py
shsms: I don't think we need a max-size,  because these will be fixed-size buffers and users are required to specify a size every time they create an instance,  so we're eliminating the possibility of it using up all memory without the developer being aware of it or having no control over it.

mathias-baumann-frequenz:
I pushed an update that shows a basic working of the timestamp buffer. The numpy variant still has a problem that I am currently looking into.

More logic tests are to come as well as performance tests

mathias-baumann-frequenz:
```
Array: ........................................
List:  ........................................
Time to fill 29 days with data:
Array: 0.09143387260000964 seconds
List:  0.08543843417501193 seconds
Diff:  0.005995438424997709

       =========================================
Array: ........................................
List:  ........................................
Filling 29 days and running average & mean on every day:
Array: 0.09528444537500036 seconds
List:  0.12714749602500888 seconds
Diff:  -0.03186305065000852
```

benchmarks/benchmark_ringbuffer.py
```diff
@@ -0,0 +1,135 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+
+"""
+Performance test for the `Ringbuffer` class
```
leandro-lucarella-frequenz: Please mirror the directory structure in `src` for benchmarks so it is easier to figure out which benchmark is benchmarking what. In this case it should be `benchmarks/util/ringbuffer.py`

benchmarks/benchmark_ringbuffer.py
```diff
@@ -0,0 +1,135 @@
...
+
+def fill_buffer(days, buffer):
+    """Fill the given buffer up to the given amount of days, one sample per
+    minute"""
+    random.seed(0)
+    basetime = datetime(2022, 1, 1)
+
+    for day in range(days):
+        # Push in random order
+        for i in random.sample(range(24 * 60), 24 * 60):
```
leandro-lucarella-frequenz: Maybe it is worth having a `MINUTES_IN_A_DAY` too?

benchmarks/benchmark_ringbuffer.py
```diff
@@ -0,0 +1,135 @@
...
+
+    for _ in range(5):
+        for day in range(days):
+            minutes = buffer.window(
+                basetime + timedelta(days=day), basetime + timedelta(days=day + 1)
+            )
+
+            total += np.average(minutes)
+            (f"{basetime + timedelta(days=day)} average {np.average(minutes)}")
+            (f"{basetime + timedelta(days=day)} median {np.median(minutes)}")
```
leandro-lucarella-frequenz: What are these strings laying around there?

benchmarks/benchmark_ringbuffer.py
```diff
@@ -0,0 +1,135 @@
...
+print("\nList:  ", end="")
+duration_list = timeit.Timer(test_29_days_slicing_list).timeit(number=num_runs)
+print("")
+
+print(
+    f"Filling 29 days and running average & mean on every day:\n\t"
+    + f"Array: {duration_array/num_runs} seconds\n\t"
+    + f"List:  {duration_list/num_runs} seconds\n\t"
+    + f"Diff:  {duration_array/num_runs -  duration_list/num_runs}"
+)
```
leandro-lucarella-frequenz: You could consider putting this in a `main()` function and then the classic:

```py
if __name__ == "__main__":
    main()
```

At the end.

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,240 @@
...
+
+T = TypeVar("T")
+
+def RingBufferArray(size: int):
+    return RingBuffer[T](np.empty(shape=(size,), dtype=np.float64))
+
+
+
+class RingBuffer(Generic[T]):
+    def __init__(self, container: Any):
```
leandro-lucarella-frequenz: You probably want to have some restrictions on the container, like `MutableSequence`?

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,240 @@
...
+import numpy as np
+
+T = TypeVar("T")
+
+def RingBufferArray(size: int):
+    return RingBuffer[T](np.empty(shape=(size,), dtype=np.float64))
+
+
+
+class RingBuffer(Generic[T]):
```
leandro-lucarella-frequenz, line 19: You might want to inherit from `Sequence[T]` instead?

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,240 @@
...
+    def __init__(self, container: Any):
+        """Initialize the ring buffer with the given container.
+
+        Args:
+            container: Container to store the data in.
+        """
+        self.container = container
+        self.write_index = 0
+        self.read_index = 0
+        self.size = 0
```
leandro-lucarella-frequenz: Ideally all internal representation should be private, so it also doesn't appear in the docs.

```suggestion
        self._container = container
        self._write_index = 0
        self._read_index = 0
        self._size = 0
```

I would also rename `_size` to `_len` as IMHO size is a bit ambiguous, it is capacity? it is size in bytes?

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,240 @@
...
+
+    def __len__(self):
+        """Return the amount of items that this container currently holds."""
+        return self.size
+
+    def max_size(self):
```
leandro-lucarella-frequenz: Even when it doesn't comply with the naming convention, you could consider calling this `maxsize` to match `deque` (not sure if the rest of the interface is compatible with `deque`, if it isn't then maybe it doesn't make sense to have the same name as in `deque`).

I called it `capacity` in the resampler buffer (which will **not** be compatible with `deque`). I don't have a strong preference but maybe we should be at least consistent in the SDK.

In any case this should be a `@property`.

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,240 @@
...
+from abc import ABC, abstractmethod
+from datetime import datetime, timedelta
+from typing import Any, Generic, Sequence, TypeVar
+
+import numpy as np
+
+T = TypeVar("T")
+
+def RingBufferArray(size: int):
+    return RingBuffer[T](np.empty(shape=(size,), dtype=np.float64))
```
leandro-lucarella-frequenz: Did you try this? How can `T` be inferred here? Also you are fixing the type to `np.float64`, right? Maybe it should be `RingBuffer[np.float64]`?

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,240 @@
...
+        if self.size == 0:
+            raise
+
+        val = self.container[self.read_index]
+        self.read_index = (self.read_index + 1) % len(self.container)
+        self.size -= 1
+
+        return val
+
+    def wrap(self, index) -> int:
```
leandro-lucarella-frequenz: I would make this private (`_wrap`).

src/frequenz/sdk/util/ringbuffer.py
leandro-lucarella-frequenz:
```suggestion
    def max_size(self) -> int:
```

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,240 @@
...
+
+        Args:
+            container: Container to store the data in.
+        """
+        self.container = container
+        self.write_index = 0
+        self.read_index = 0
+        self.size = 0
+
+    def __len__(self):
```
leandro-lucarella-frequenz:
```suggestion
    def __len__(self) -> int:
```

src/frequenz/sdk/util/ringbuffer.py
leandro-lucarella-frequenz:
```suggestion
    def __init__(self, container: Any) -> None:
```

src/frequenz/sdk/util/ringbuffer.py
leandro-lucarella-frequenz:
```suggestion
    def wrap(self, index: int) -> int:
```

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,240 @@
...
+
+        return val
+
+    def wrap(self, index) -> int:
+        return index % len(self.container)
+
+    def full(self) -> bool:
+        return len(self) == len(self.container)
+
+    def __setitem__(self, index: int | slice, value: T | Sequence[T]) -> None:
```
leandro-lucarella-frequenz, line 101: You need to:

```py
```

to be able to use the `|` syntax in Python 3.8.

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,240 @@
...
+        self.container[index] = value
```
leandro-lucarella-frequenz: This is probably broken? Shouldn't you shift the indexes based on the read and write index?

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,240 @@
...
+
+    def __getitem__(self, index_or_slice: int | slice) -> T | Sequence[T]:
+        """Returns the value at the given index or value range at the given
+        slice. Does not support wrap-around or copying of data."""
+        return self.container[index_or_slice]
```
leandro-lucarella-frequenz: Same here. If you don't support doing the shifting, I wouldn't support indexing at all, it doesn't make much sense without doing the shifting.

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+class OrderedRingBuffer(Generic[T]):
+    """Time aware ringbuffer that keeps its entries sorted time."""
```
leandro-lucarella-frequenz, line 139: I don't think this is necessary, the only **potential** use case (AFAIK) is the resampler, and will use another buffer implementation.

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,240 @@
...
+            raise IndexError()
+
+        return self.wrap(int(abs((self.window_start - datetime).total_seconds())))
+
+    def window(self, start: datetime, end: datetime) -> Any:
+        """Request a view on the data between start timestamp and end timestamp.
+
+        Will return a copy in the following cases:
+            * The requested time period is crossing the start/end of the buffer
+            * The requested time period contains missing entries.
```
leandro-lucarella-frequenz: This should be plain markdown, AFAIK:
```suggestion
        * The requested time period is crossing the start/end of the buffer
        * The requested time period contains missing entries.
```

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,240 @@
...
+
+    def datetime_to_index(self, datetime: datetime) -> int:
+        """Convert the given timestamp to an index.
+
+        Throws an index error when the timestamp is not found within this
+        buffer.
+
+        Args:
+            datetime: Timestamp to convert.
+        Returns:
```
leandro-lucarella-frequenz, line 235: Not sure if it's necessary but we usually leave and empty line between different sections.

```suggestion

        Returns:
```

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,240 @@
...
+            * The requested time period is crossing the start/end of the buffer
+            * The requested time period contains missing entries.
+
+        This means, if the caller needs to modify the data to account for
+        missing entries, they can safely do so.
+
+        Args:
+            start: start time of the window
+            end: end time of the window
+        Returns:
```
leandro-lucarella-frequenz, line 261:
```suggestion

        Returns:
```

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,240 @@
...
+
+        self.missing_windows = []
+        self.datetime_newest = datetime.min
+        self.datetime_oldest = datetime.max
+
+    def max_size(self):
+        """Return the max amount of items this container can hold."""
+        return len(self.buffer)
+
+    def update(self, datetime: datetime, value: T, missing: bool = False):
```
leandro-lucarella-frequenz:
```suggestion
    def update(self, datetime: datetime, value: T, missing: bool = False) -> None:
```

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,240 @@
...
+        """
+        self.buffer = buffer
+        self.resolution_in_seconds = resolution_in_seconds
+        self.window_start = window_border
+
+        self.missing_windows = []
+        self.datetime_newest = datetime.min
+        self.datetime_oldest = datetime.max
+
+    def max_size(self):
```
leandro-lucarella-frequenz:
```suggestion
    @property
    def max_size(self) -> int:
```

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,240 @@
...
+            datetime: Timestamp to convert.
+        Returns:
+            index where the value for the given timestamp can be found.
+        """
+        if datetime < self.datetime_oldest:
+            raise IndexError()
+
+        return self.wrap(int(abs((self.window_start - datetime).total_seconds())))
+
+    def window(self, start: datetime, end: datetime) -> Any:
```
leandro-lucarella-frequenz: Maybe return `Sequence[T]` instead (try to avoid `Any` whenever possible as it kills type checking):
```suggestion
    def window(self, start: datetime, end: datetime) -> Sequence[T]:
```

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,240 @@
...
+
+        # Return a copy if there are none-values in the data
+        if any(map(in_window, self.missing_windows)):
+            import copy
+
+            return copy.deepcopy(self.buffer[start_index:end_index])
+
+        return self.buffer[start_index:end_index]
+
+    def wrap(self, index: int):
```
leandro-lucarella-frequenz:
```suggestion
    def wrap(self, index: int) -> int:
```

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,240 @@
...
+
+        Args:
+            index: index to normalize.
+
+        Returns:
+            an index that will be within max_size.
+        """
+        return index % self.max_size()
+
+    def __getitem__(self, index_or_slice: int | slice) -> T | Sequence[T]:
```
leandro-lucarella-frequenz, line 305: Same as before, better not provide this if the implementation is not *correct* IMHO.

tests/utils/ringbuffer.py
```diff
@@ -0,0 +1,162 @@
...
+    for i in range(len(container)):
+        assert i == buffer.pop()
+
+
+def test_simple_push_pop_list() -> None:
+    _test_simple_push_pop([0] * 50000)
+
+
+def test_simple_push_pop_array() -> None:
+    _test_simple_push_pop(np.empty(shape=(50000,), dtype=np.float64))
```
leandro-lucarella-frequenz: `@pytest.parametrize()` is your friend: https://docs.pytest.org/en/6.2.x/parametrize.html

(untested but something like this should work)

```suggestion
@pytest.parametrize("container", [
    [0] * 50000,
    np.empty(shape=(50000,), dtype=np.float64),
])
def test_simple_push_pop(container):
    buffer = RingBuffer(container)

    for i in range(len(container)):
        buffer.push(i)

    assert len(buffer) == len(container)

    for i in range(len(container)):
        assert i == buffer.pop()
```

tests/utils/ringbuffer.py
```diff
@@ -0,0 +1,162 @@
...
+        for j in range(1, size):
+            start = datetime.fromtimestamp(200 + i)
+            end = datetime.fromtimestamp(200 + j + i)
+            from itertools import cycle, islice
+
+            tmp = list(islice(cycle(range(0, size)), i, i + j))
+            assert list(buffer.window(start, end)) == list(tmp)
+
+
+# map(lambda x: datetime(1, 1, 1, second=x), range(size))
```
leandro-lucarella-frequenz: :thinking:

leandro-lucarella-frequenz (commented):
A **lot** of the comments are just me putting the `pylint` or `mypy` hat, so it is not as bad as it looks :laughing:

leandro-lucarella-frequenz:
Beware of damn GitHub compressing comments and hiding them, make sure to click when it says "N hidden conversations. Load more".

benchmarks/benchmark_ringbuffer.py
mathias-baumann-frequenz: I was just trying to prevent python from optimizing them away

src/frequenz/sdk/util/ringbuffer.py
mathias-baumann-frequenz: I'd love to, but I found no type that would fit so far. Will investigate again.

src/frequenz/sdk/util/ringbuffer.py
mathias-baumann-frequenz: No, this is still a WIP. In fact, I have currently no idea how to do this part

src/frequenz/sdk/util/ringbuffer.py
mathias-baumann-frequenz: Well, this and getitem provide a raw access to the buffer. I can certainly imagine cases where that is needed.
I can also just remove them, otherwise they aren't really needed...

src/frequenz/sdk/util/ringbuffer.py
mathias-baumann-frequenz: Hu? You mean the whole class is not necessary? It is clearly specified in the ticket that a sorted ringbuffer is requested...

src/frequenz/sdk/util/ringbuffer.py
mathias-baumann-frequenz: Well, it is correct insofar as that it provides direct access to the buffer..

benchmarks/benchmark_ringbuffer.py
mathias-baumann-frequenz: I didn't see `power_distribution` in the `src` dir so I assumed it would be more.. arbitrary..

src/frequenz/sdk/util/ringbuffer.py
mathias-baumann-frequenz: Do we actually want our buffer to act as a Sequence?

src/frequenz/sdk/util/ringbuffer.py
mathias-baumann-frequenz:
```
tests/utils/ringbuffer.py:19: error: Argument 1 to "RingBuffer" has incompatible type "ndarray[Any, dtype[f
loating[_64Bit]]]"; expected "MutableSequence[<nothing>]"  [arg-type]
```

benchmarks/benchmark_ringbuffer.py
leandro-lucarella-frequenz: I think at some point that one was top-level, we probably forgot to update the `benchmarks` structure when we moved it.

benchmarks/benchmark_ringbuffer.py
leandro-lucarella-frequenz: I don't get it? You want to test the performance of building a formatted string?

src/frequenz/sdk/util/ringbuffer.py
leandro-lucarella-frequenz: Can you show how you changed the code to get that message?

mathias-baumann-frequenz: I just changed `Any` to `MutableBuffer[T]`

src/frequenz/sdk/util/ringbuffer.py
leandro-lucarella-frequenz: If we are providing `__getitem__` and `__len__` I don't see why not, all the work for it to be a `Sequence` is already done. But well, you weren't really providing `__getitem__` with shifting, so I guess it depends on that.

src/frequenz/sdk/util/ringbuffer.py
leandro-lucarella-frequenz: If `np.empty(shape=(size,), dtype=np.float64)` returns elements of type `np.float64` then using `RingBuffer[np.float64]` should work. In any case feel free to ask if there is something related to typing that you can't achieve, it takes some time to get used to a dynamic language with type hints :D

src/frequenz/sdk/util/ringbuffer.py
leandro-lucarella-frequenz: If you want to provide raw access to the underlying container, than I would make a property to get `self._container` and let people operate on that, this way is much clearer that there is no funny index adjustment going on when you slice and such. If I do:

```py
ring_buffer[0] = 1
```

I definitely expect to write the *first* element in the **ring buffer**, not the underlying container.

src/frequenz/sdk/util/ringbuffer.py
leandro-lucarella-frequenz: The use case I documented there was for resampling, in case at some point we get out-of-order data from a data source, but it is a hypothetical very low priority case. Sorry if it is not clear in the issue, I think we mentioned it in live conversations but we should have made it more clear in the issue too if it's not.

In any case it looks like @matthias-wende-frequenz might have a use case for this in the prediction pipeline.

src/frequenz/sdk/util/ringbuffer.py
leandro-lucarella-frequenz: Same comment as above, I think it is more clear to provide access to the underlying container and let people slice that directly instead.

Another comment is you can improve type checking by adding `@overload`s, to specify which subset of types are valid in those unions. Here is an example: https://github.com/frequenz-floss/frequenz-sdk-python/pull/130/files#diff-43d601005020a1afe419acdfad091a0e813ddd9a6f2e1d4e0db73c030ff8d599R135-R143

src/frequenz/sdk/util/ringbuffer.py
mathias-baumann-frequenz: I went with `maxsize` for now

src/frequenz/sdk/util/ringbuffer.py
mathias-baumann-frequenz: The idea is more that the caller of the function can specify the type and I don't know how to pass that type parameter through

src/frequenz/sdk/util/ringbuffer.py
leandro-lucarella-frequenz: OK, yeah, it looks like it's not even a `Sequence` (although it should), but probably won't get all the methods needed by a `MutableSequence`: https://github.com/numpy/numpy/issues/2776.

So we have 2 options I can see:

1. Use an union: `MutableSquence[T] | np.ndarray`, which is a bit ugly to have this special casing, why only `np.ndarray` and not anything else that looks like a duck?
2. Use a [`Protocol`](https://docs.python.org/3/library/typing.html#typing.Protocol): just look at all the methods and attributes you need a container to have an write a `BufferContainer` protocol.

IMHO 2. is the correct option, but it is a bit of work that might not be justified at this point, so I would leave `Any` and create an issue to improve this in the future.

src/frequenz/sdk/util/ringbuffer.py
leandro-lucarella-frequenz: It's [`maxlen`](https://docs.python.org/3/library/collections.html#collections.deque.maxlen) in `deque`, sorry :grimacing:

src/frequenz/sdk/util/ringbuffer.py
leandro-lucarella-frequenz: I don't think there is a pretty way to do it with a type parameter because `np.ndarray` is so convoluted to create. I would leave it explicit:

```py
rb = RingBuffer[np.float64](np.empty(shape=(size,), dtype=np.float64))
```

Or have a concrete function to create a ring buffer from a `ndarray` like the one you have above:
```py
def RingBufferArray(size: int):
    return RingBuffer[np.float64](np.empty(shape=(size,), dtype=np.float64))
```

leandro-lucarella-frequenz: BTW, I would replace `size` with `len` or `length` everywhere, as those are really length of the array, not size.

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,321 @@
...
+
+
+class RingBuffer(Generic[T]):
+    """A ring buffer with a fixed size.
+
+    Should work with most backends, tested with list and np.ndarrays.
+    """
+
+    class NoElements(Exception):
+        """Exception class, thrown when no elements are available when calling pop()."""
```
leandro-lucarella-frequenz: The pythonic exception to use here is `IndexError`.

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,321 @@
...
+            raise RingBuffer.NoElements()
+
+        val = self._container[self._read_index]
+        self._read_index = (self._read_index + 1) % len(self._container)
+        self._len -= 1
+
+        return val
+
+    @property
+    def full(self) -> bool:
```
leandro-lucarella-frequenz: We try to stick to using something that looks like a question to name stuff that is a bool. What you had before was perfect: `is_full`.

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,321 @@
...
+        self._container = container
+        self._write_index = 0
+        self._read_index = 0
+        self._len = 0
+
+    def __len__(self) -> int:
+        """Get current amount of elements.
+
+        Returns:
+            the amount of items that this container currently holds.
```
leandro-lucarella-frequenz: Silly but try to always start these with caps for more consistency: "The amount of ...".

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,321 @@
...
+        Returns:
+            the max amount of items this container can hold.
+        """
+        return len(self._container)
+
+    def push(self, value: T) -> int:
+        """Push a new value into the ring buffer.
+
+        Args:
+            value: value to push into the ring buffer.
```
leandro-lucarella-frequenz: Same for `Args`: "value: Value to push..."

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,321 @@
...
+        value_index = self._write_index
+        self._write_index = self._wrap(self._write_index + 1)
+
+        return value_index
+
+    def pop(self) -> T:
+        """Remove the oldest value from the ring buffer and return it.
+
+        Raises:
+            NoElements: when no elements exist to pop.
```
leandro-lucarella-frequenz: Should be `IndexError`. I know it might sound weird but it is what people will expect, as this is what `list` and `deque` do.

leandro-lucarella-frequenz (commented):
There were a few pending unanswered comments from the previous review. I think all the remaining stuff is very minor. There are only 2 major topics:

1. If, when and how do we need the sorted buffer.
2. If we really need proper slicing (setting and getting). My impression is we don't need setting a slice. About getting, not sure, maybe, maybe it is enough with using [`islice`](https://docs.python.org/3/library/itertools.html#itertools.islice) and we save some complexity (which IMHO while implementing the resampling buffer was not at all minor, specially with the many cases if we include steps and negative values).

benchmarks/benchmark_ringbuffer.py
leandro-lucarella-frequenz: Not sure if it is intentional but the benchmark disappeared from the PR.

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,321 @@
...
+        return self._buffer[start_index:end_index]
+
+    def _wrap(self, index: int) -> int:
+        """Normalize the given index to fit in the buffer by wrapping it around.
+
+        Args:
+            index: index to normalize.
+
+        Returns:
+            an index that will be within max_size.
```
leandro-lucarella-frequenz: `maxlen`

benchmarks/benchmark_ringbuffer.py
mathias-baumann-frequenz: Yeah no, that was a bad renaming

mathias-baumann-frequenz:
Closing this to keep the branch around for the simple ring buffer, will create a new PR with only the sorted ring buffer

PR #117, opened 2022-12-13
Track status of the batteries

### LINKED ISSUE #33: Implement `BatteryPoolStatus` to track which batteries are working
### What's needed?

We need external instance that would track which batteries are working. 
It should be possible to get subset of working batteries anytime.
First it should be used by BatteryPool and PowerDistributor.

### Proposed solution

BatteryStatusActor should be Singleton that stores status of all battery-inverter pairs in the microgrid.

```py
class BatteryStatusActor:
def __init__(self):
        # Subscribe for the metrics below for internal use
        metrics = [".result.inverter.state.componentState",  ".result.battery.state.componentState", ".result.battery.state.relayState" ]
        metric_channels = # For each battery-inverter pair in Microgrid: Subscribe for `self.metrics`, each channel should be peekable. 

def run(self) -> Set[battery_id]:
    # Returns set of working batteries

    # Battery is consider as broken if:
    #    * inverter state in [UNSPECIFIED, UNAVAILABLE]
    #    * battery relay in [OPEM]
    #    * battery state in [ERROR]
    #    * component didn't send any data for some time ( how long? it should be configurable )
    #    * latest error level returned from `set_power` method is Critical.
    #    * battery-inverter didn't respond for the latest `set_power` call
    # If battery is considered as NOT working, then change it to working if:
    #    * state of the inverter or battery changed - valid state
    #    * RelayState or state of the battery changed
    #    * some time has passed (increase exponentially until reach 30 sec - this should be configurable)
    #    * component started to send data
  
    # If all batteries given in argument are considered as broken, then relax the conditions to below:
    # Battery is consider as broken if:
    #    * inverter state in [UNSPECIFIED, UNAVAILABLE]
    #    * battery relay in [CLOSED]
    #    * battery state in [ERROR]
    #    * component didn't send any data for some time
```
### Use cases

BatteryPool need subset of working batteries to return data (capacity, active_power) for working batteries only.
PowerDistributor needs subset of working batteries to charge/discharge only working batteries. 

BatteryStatus (poor version) is not part of PowerDistirbutor. But it has bugs and only PowerDistributor can use it.

### Additional context

_No response_

ela-kotulska-frequenz:
Add BatteryPoolStatus that tracks what batteries in pool are working or not.
Use BatteryPoolStatus in PowerDistributingActor temporary. 
In the future it will be used in BatteryPool.

### DIFF
--- a/src/frequenz/sdk/actor/power_distributing/_battery_pool_status.py
+++ b/src/frequenz/sdk/actor/power_distributing/_battery_pool_status.py
@@ -1,0 +1,124 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+"""Class that stores pool of batteries and manage them."""
+
+import asyncio
+import logging
+from typing import Set
+
+from ...microgrid._battery import BatteryStatus, StatusTracker
+from .result import PartialFailure, Result, Success
+
+_logger = logging.getLogger(__name__)
+
+
+class BatteryPoolStatus:
+    """Holds pool of batteries and returns data from them."""
+
+    def __init__(
+        self,
+        battery_ids: Set[int],
+        max_data_age_sec: float,
+        max_blocking_duration_sec: float,
+    ) -> None:
+        """Create partially initialized object instance.
+
+        Note:
+            Please call `async_init` method to fully initialize BatteryPoolStatus. Otherwise
+            it is not possible to use BatteryPoolStatus.
+
+        Args:
+            battery_ids: set of batteries ids that should be stored in pool.
+            max_data_age_sec: If component stopped sending data, then
+                this is the maximum time when its last message should be considered as
+                valid. After that time, component won't be used until it starts sending
+                data.
+            max_blocking_duration_sec: This value tell what should be the maximum
+                timeout used for blocking failing component.
+        """
+        self._batteries = {
+            id: StatusTracker(id, max_data_age_sec, max_blocking_duration_sec)
+            for id in battery_ids
+        }
+        self._init_method_called: bool = False
+
+    async def async_init(self) -> None:
+        """Init battery pool."""
+        await asyncio.gather(
+            *[bat.async_init() for bat in self._batteries.values()],
+            return_exceptions=True,
+        )
+        self._init_method_called = True
+
+    def get_working_batteries(self, battery_ids: Set[int]) -> Set[int]:
+        """Get subset of battery_ids with working batteries.
+
+        Args:
+            battery_ids: batteries ids
+
+        Raises:
+            RuntimeError: If `async_init` method was not called at the beginning to
+                initialized object.
+            KeyError: If any battery in the given batteries is not in the pool.
+
+        Returns:
+            Subset of given batteries with working batteries.
+        """
+        if not self._init_method_called:
+            raise RuntimeError(
+                "`async_init` method not called or not awaited. Run it before first use"
+            )
+
+        working: Set[int] = set()
+        uncertain: Set[int] = set()
+        for bat_id in battery_ids:
+            if bat_id not in battery_ids:
+                ids = str(self._batteries.keys())
+                raise KeyError(f"No battery {bat_id} in pool. All batteries: {ids}")
+            battery_status = self._batteries[bat_id].get_status()
+            if battery_status == BatteryStatus.WORKING:
+                working.add(bat_id)
+            elif battery_status == BatteryStatus.UNCERTAIN:
+                uncertain.add(bat_id)
+
+        if len(working) > 0:
+            return working
+
+        _logger.warning(
+            "There are no working batteries in %s. Falling back to using uncertain batteries %s.",
+            str(battery_ids),
+            str(uncertain),
+        )
+        return uncertain
+
+    def update_last_request_status(self, result: Result):
+        """Update batteries in pool based on the last result from the request.
+
+        Args:
+            result: Summary of what batteries failed and succeed in last request.
+
+        Raises:
+            RuntimeError: If `async_init` method was not called at the beginning to
+                initialize object.
+        """
+        if not self._init_method_called:
+            raise RuntimeError(
+                "`async_init` method not called or not awaited. Run it before first use"
+            )
+
+        if isinstance(result, Success):
+            for bat_id in result.used_batteries:
+                self._batteries[bat_id].unblock()
+
+        elif isinstance(result, PartialFailure):
+            for bat_id in result.failed_batteries:
+                duration = self._batteries[bat_id].block()
+                if duration > 0:
+                    _logger.warning(
+                        "Battery %d failed last response. Block it for %f sec",
+                        bat_id,
+                        duration,
+                    )
+
+            for bat_id in result.succeed_batteries:
+                self._batteries[bat_id].unblock()
--- a/src/frequenz/sdk/microgrid/_battery/_status.py
+++ b/src/frequenz/sdk/microgrid/_battery/_status.py
@@ -1,0 +1,372 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+"""Class to return battery status."""
+
+import asyncio
+import logging
+from dataclasses import dataclass
+from datetime import datetime, timedelta, timezone
+from enum import Enum
+from typing import Generic, Optional, Set, TypeVar, Union
+
+from frequenz.api.microgrid.battery_pb2 import ComponentState as BatteryComponentState
+from frequenz.api.microgrid.battery_pb2 import RelayState as BatteryRelayState
+from frequenz.api.microgrid.common_pb2 import ErrorLevel
+from frequenz.api.microgrid.inverter_pb2 import ComponentState as InverterComponentState
+from frequenz.channels import Peekable
+
+from .. import ComponentGraph
+from .. import get as get_microgrid
+from ..component import BatteryData, ComponentCategory, ComponentData, InverterData
+
+# Time needed to get first component data.
+# Constant for now. In future should be configurable in UI.
+_START_DELAY_SEC = 2.0
+
+_logger = logging.getLogger(__name__)
+
+T = TypeVar("T")
+
+
+class BatteryStatus(Enum):
+    """Tells if battery is can be used."""
+
+    # Component it is not working.
+    NOT_WORKING = 0
+    # Component should work, but it failed in last request. It is blocked for few
+    # seconds and it is not recommended to use it until it is necessary.
+    UNCERTAIN = 1
+    # Component is working
+    WORKING = 2
+
+
+@dataclass
+class _ComponentReceiver(Generic[T]):
+    component_id: int
+    receiver: Peekable[T]
+
+
+class StatusTracker:
+    """Class for tracking if battery is working."""
+
+    _battery_valid_relay: Set[BatteryRelayState.ValueType] = {
+        BatteryRelayState.RELAY_STATE_CLOSED
+    }
+    _battery_valid_state: Set[BatteryComponentState.ValueType] = {
+        BatteryComponentState.COMPONENT_STATE_IDLE,
+        BatteryComponentState.COMPONENT_STATE_CHARGING,
+        BatteryComponentState.COMPONENT_STATE_DISCHARGING,
+    }
+    _inverter_valid_state: Set[InverterComponentState.ValueType] = {
+        InverterComponentState.COMPONENT_STATE_STANDBY,
+        InverterComponentState.COMPONENT_STATE_IDLE,
+        InverterComponentState.COMPONENT_STATE_CHARGING,
+        InverterComponentState.COMPONENT_STATE_STANDBY,
+    }
+
+    def __init__(
+        self, battery_id: int, max_data_age_sec: float, max_blocking_duration_sec: float
+    ) -> None:
+        """Create class instance.
+
+        Args:
+            battery_id: Id of this battery
+            max_data_age_sec: If component stopped sending data, then
+                this is the maximum time when its last message should be considered as
+                valid. After that time, component won't be used until it starts sending
+                data.
+            max_blocking_duration_sec: This value tell what should be the maximum
+                timeout used for blocking failing component.
+        """
+        self._battery_id: int = battery_id
+        self._max_data_age: float = max_data_age_sec
+
+        self._init_method_called: bool = False
+        self._last_status: BatteryStatus = BatteryStatus.WORKING
+
+        self._min_blocking_duration_sec: float = 1.0
+        self._max_blocking_duration_sec: float = max_blocking_duration_sec
+        self._blocked_until: Optional[datetime] = None
+        self._last_blocking_duration_sec: float = self._min_blocking_duration_sec
+
+    @property
+    def battery_id(self) -> int:
+        """Get battery id.
+
+        Returns:
+            Battery id
+        """
+        return self._battery_id
+
+    async def async_init(self) -> None:
+        """Async part of constructor.
+
+        This should be called in order to subscribe for necessary data.
+
+        Raises:
+            RuntimeError: If given battery as no adjacent inverter.
+        """
+        component_graph = get_microgrid().component_graph
+        inverter_id = self._find_adjacent_inverter_id(component_graph)
+        if inverter_id is None:
+            raise RuntimeError(
+                f"Can't find inverter adjacent to battery: {self._battery_id}"
+            )
+
+        api_client = get_microgrid().api_client
+
+        bat_recv = await api_client.battery_data(self._battery_id)
+        # Defining battery_receiver and inverter receiver needs await.
+        # Because of that it is impossible to define it in constructor.
+        # Setting it to None first would require to check None condition in
+        # every call.
+        # pylint: disable=attribute-defined-outside-init
+        self._battery_receiver = _ComponentReceiver(
+            self._battery_id, bat_recv.into_peekable()
+        )
+        inv_recv = await api_client.inverter_data(inverter_id)
+        # pylint: disable=attribute-defined-outside-init
+        self._inverter_receiver = _ComponentReceiver(
+            inverter_id, inv_recv.into_peekable()
+        )
+
+        await asyncio.sleep(_START_DELAY_SEC)
+        self._init_method_called = True
+
+    def get_status(self) -> BatteryStatus:
+        """Return status of the battery.
+
+        The decision is made based on last message from battery and adjacent inverter
+        and result from last request.
+
+        Raises:
+            RuntimeError: If method `async_init` was not called or not awaited.
+
+        Returns:
+            Battery status
+        """
+        if not self._init_method_called:
+            raise RuntimeError(
+                "`async_init` method not called or not awaited. Run it before first use"
+            )
+
+        bat_msg = self._battery_receiver.receiver.peek()
+        inv_msg = self._inverter_receiver.receiver.peek()
+        if bat_msg is None or inv_msg is None:
+            if self._last_status == BatteryStatus.WORKING:
+                if not bat_msg:
+                    component_id = self._battery_id
+                else:
+                    component_id = self._inverter_receiver.component_id
+
+                _logger.warning(
+                    "None returned from component %d receiver.", component_id
+                )
+            self._last_status = BatteryStatus.NOT_WORKING
+            return self._last_status
+
+        older_message = bat_msg if bat_msg.timestamp < inv_msg.timestamp else inv_msg
+        is_msg_ok = (
+            self._is_message_reliable(older_message)
+            and self._is_battery_state_correct(bat_msg)
+            and self._is_inverter_state_correct(inv_msg)
+            and self._no_critical_error(bat_msg)
+            and self._no_critical_error(inv_msg)
+        )
+
+        if not is_msg_ok:
+            self._last_status = BatteryStatus.NOT_WORKING
+            return self._last_status
+
+        # Use battery as soon as its message is back correct.
+        if self._last_status == BatteryStatus.NOT_WORKING:
+            self.unblock()
+
+        if self.is_blocked():
+            self._last_status = BatteryStatus.UNCERTAIN
+        else:
+            self._last_status = BatteryStatus.WORKING
+
+        return self._last_status
+
+    def is_blocked(self) -> bool:
+        """Return if battery is blocked.
+
+        Battery can be blocked if last request for that battery failed.
+
+        Returns:
+            True if battery is blocked, False otherwise.
+        """
+        if self._blocked_until is None:
+            return False
+        return self._blocked_until > datetime.now(tz=timezone.utc)
+
+    def unblock(self) -> None:
+        """Unblock battery.
+
+        This will reset duration of the next blocking timeout.
+
+        Battery can be blocked using `self.block()` method.
+        """
+        self._blocked_until = None
+
+    def block(self) -> float:
+        """Block battery.
+
+        Battery can be unblocked using `self.unblock()` method.
+
+        Returns:
+            For how long (in seconds) the battery is blocked.
+        """
+        now = datetime.now(tz=timezone.utc)
+
+        if self._blocked_until is None:
+            self._last_blocking_duration_sec = self._min_blocking_duration_sec
+            self._blocked_until = now + timedelta(
+                seconds=self._last_blocking_duration_sec
+            )
+            return self._last_blocking_duration_sec
+
+        # If still blocked, then do nothing
+        if self._blocked_until > now:
+            return 0.0
+
+        # Increase blocking duration twice or until it reach the max.
+        self._last_blocking_duration_sec = min(
+            2 * self._last_blocking_duration_sec, self._max_blocking_duration_sec
+        )
+        self._blocked_until = now + timedelta(seconds=self._last_blocking_duration_sec)
+
+        return self._last_blocking_duration_sec
+
+    @property
+    def blocked_until(self) -> Optional[datetime]:
+        """Return timestamp when the battery will be unblocked.
+
+        Returns:
+            Timestamp when the battery will be unblocked. Return None if battery is
+                not blocked.
+        """
+        if self._blocked_until is None or self._blocked_until < datetime.now(
+            tz=timezone.utc
+        ):
+            return None
+        return self._blocked_until
+
+    def _no_critical_error(self, msg: Union[BatteryData, InverterData]) -> bool:
+        """Check if battery or inverter message has any critical error.
+
+        Args:
+            msg: message.
+
+        Returns:
+            True if message has no critical error, False otherwise.
+        """
+        critical = ErrorLevel.ERROR_LEVEL_CRITICAL
+        # pylint: disable=protected-access
+        critical_err = next((err for err in msg._errors if err.level == critical), None)
+        if critical_err is not None:
+            if self._last_status == BatteryStatus.WORKING:
+                _logger.warning(
+                    "Component %d has critical error: %s",
+                    msg.component_id,
+                    str(critical_err),
+                )
+            return False
+        return True
+
+    def _is_inverter_state_correct(self, msg: InverterData) -> bool:
+        """Check if inverter is in correct state from message.
+
+        Args:
+            msg: message
+
+        Returns:
+            True if inverter is in correct state. False otherwise.
+        """
+        # Component state is not exposed to the user.
+        # pylint: disable=protected-access
+        state = msg._component_state
+        if state not in StatusTracker._inverter_valid_state:
+            if self._last_status == BatteryStatus.WORKING:
+                _logger.warning(
+                    "Inverter %d has invalid state: %s",
+                    msg.component_id,
+                    str(state),
+                )
+            return False
+        return True
+
+    def _is_battery_state_correct(self, msg: BatteryData) -> bool:
+        """Check if battery is in correct state from message.
+
+        Args:
+            msg: message
+
+        Returns:
+            True if battery is in correct state. False otherwise.
+        """
+        # Component state is not exposed to the user.
+        # pylint: disable=protected-access
+        state = msg._component_state
+        if state not in StatusTracker._battery_valid_state:
+            if self._last_status == BatteryStatus.WORKING:
+                _logger.warning(
+                    "Battery %d has invalid state: %s",
+                    self._battery_id,
+                    str(state),
+                )
+            return False
+
+        # Component state is not exposed to the user.
+        # pylint: disable=protected-access
+        relay_state = msg._relay_state
+        if relay_state not in StatusTracker._battery_valid_relay:
+            if self._last_status == BatteryStatus.WORKING:
+                _logger.warning(
+                    "Battery %d has invalid relay state: %s",
+                    self._battery_id,
+                    str(relay_state),
+                )
+            return False
+        return True
+
+    def _is_message_reliable(self, message: ComponentData) -> bool:
+        """Check if message is too old to be considered as reliable.
+
+        Args:
+            message: message to check
+
+        Returns:
+            True if message is reliable, False otherwise.
+        """
+        now = datetime.now(tz=timezone.utc)
+        diff = (now - message.timestamp).total_seconds()
+        is_outdated = diff > self._max_data_age
+
+        if is_outdated and self._last_status == BatteryStatus.WORKING:
+            _logger.warning(
+                "Component %d stopped sending data. Last timestamp: %s.",
+                message.component_id,
+                str(message.timestamp),
+            )
+
+        return not is_outdated
+
+    def _find_adjacent_inverter_id(self, graph: ComponentGraph) -> Optional[int]:
+        """Find inverter adjacent to this battery.
+
+        Args:
+            graph: Component graph
+
+        Returns:
+            Id of the inverter. If battery hasn't adjacent inverter, then return None.
+        """
+        return next(
+            (
+                comp.component_id
+                for comp in graph.predecessors(self._battery_id)
+                if comp.category == ComponentCategory.INVERTER
+            ),
+            None,
+        )
--- a/src/frequenz/sdk/microgrid/component/_component_data.py
+++ b/src/frequenz/sdk/microgrid/component/_component_data.py
@@ -7,8 +7,10 @@ from __future__ import annotations
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
 from datetime import datetime
-from typing import Optional, Tuple
+from typing import List, Optional, Tuple
 
+import frequenz.api.microgrid.battery_pb2 as battery_pb
+import frequenz.api.microgrid.inverter_pb2 as inverter_pb
 import frequenz.api.microgrid.microgrid_pb2 as microgrid_pb
 import pytz
 
@@ -128,6 +130,9 @@ class BatteryData(ComponentData):
             charging is possible.
         temperature_max: the maximum temperature of all the blocks in a battery, in
             Celcius (°C).
+        _relay_state: state of the battery relay.
+        _component_state: state of the battery.
+        _errors: list of errors in protobuf struct.
     """
 
     soc: float
@@ -137,6 +142,9 @@ class BatteryData(ComponentData):
     power_lower_bound: float
     power_upper_bound: float
     temperature_max: float
+    _relay_state: battery_pb.RelayState.ValueType
+    _component_state: battery_pb.ComponentState.ValueType
+    _errors: List[battery_pb.Error]
 
     @classmethod
     def from_proto(cls, raw: microgrid_pb.ComponentData) -> BatteryData:
@@ -158,6 +166,9 @@ class BatteryData(ComponentData):
             power_lower_bound=raw.battery.data.dc.power.system_bounds.lower,
             power_upper_bound=raw.battery.data.dc.power.system_bounds.upper,
             temperature_max=raw.battery.data.temperature.max,
+            _relay_state=raw.battery.state.relay_state,
+            _component_state=raw.battery.state.component_state,
+            _errors=list(raw.battery.errors),
         )
         battery_data._set_raw(raw=raw)
         return battery_data
@@ -178,11 +189,15 @@ class InverterData(ComponentData):
         active_power_upper_bound: the maximum charge power, in Watts, represented in
             the passive sign convention. This will be a positive number, or zero if no
             charging is possible.
+        _component_state: state of the inverter.
+        _errors: list of errors from the component.
     """
 
     active_power: float
     active_power_lower_bound: float
     active_power_upper_bound: float
+    _component_state: inverter_pb.ComponentState.ValueType
+    _errors: List[inverter_pb.Error]
 
     @classmethod
     def from_proto(cls, raw: microgrid_pb.ComponentData) -> InverterData:
@@ -200,7 +215,10 @@ class InverterData(ComponentData):
             active_power=raw.inverter.data.ac.power_active.value,
             active_power_lower_bound=raw.inverter.data.ac.power_active.system_bounds.lower,
             active_power_upper_bound=raw.inverter.data.ac.power_active.system_bounds.upper,
+            _component_state=raw.inverter.state.component_state,
+            _errors=list(raw.inverter.errors),
         )
+
         inverter_data._set_raw(raw=raw)
         return inverter_data
 
--- a/src/frequenz/sdk/microgrid/component/_component_data.py
+++ b/src/frequenz/sdk/microgrid/component/_component_data.py
@@ -19,22 +19,21 @@ from ._component_states import EVChargerCableState
 
 @dataclass(frozen=True)
 class ComponentData(ABC):
-    """A private base class for strongly typed component data classes.
-
-    Attributes:
-        component_id: the ID identifying this component in the microgrid
-        timestamp: the timestamp of when the data was measured.
-        raw: raw component data as decoded from the wire
-    """
+    """A private base class for strongly typed component data classes."""
 
     component_id: int
+    """The ID identifying this component in the microgrid."""
+
     timestamp: datetime
+    """The timestamp of when the data was measured."""
+
     # The `raw` attribute is excluded from the constructor as it can only be provided
     # when instantiating `ComponentData` using the `from_proto` method, which reads
     # data from a protobuf message. The whole protobuf message is stored as the `raw`
     # attribute. When `ComponentData` is not instantiated from a protobuf message,
     # i.e. using the constructor, `raw` will be set to `None`.
     raw: Optional[microgrid_pb.ComponentData] = field(default=None, init=False)
+    """Raw component data as decoded from the wire."""
 
     def _set_raw(self, raw: microgrid_pb.ComponentData) -> None:
         """Store raw protobuf message.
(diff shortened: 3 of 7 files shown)


src/frequenz/sdk/microgrid/component/_component_errors.py
```diff
@@ -0,0 +1,151 @@
...
+        Returns:
+            Enum value corresponding to the protobuf message.
+        """
+        return ErrorLevel(error_lvl)
+
+
+class BatteryErrorCode(Enum):
+    """Battery error codes."""
+
+    ERROR_CODE_UNSPECIFIED = battery_pb.ErrorCode.ERROR_CODE_UNSPECIFIED
```
tiyash-basu-frequenz: This looks like a lot of duplication from protobuf here.
Will it be possible to just re-export the enum `battery_pb.ErrorCode` as `BatteryErrorCode`?

This comment extends also to other code in the files `_component_errors.py` and `_component_states.py`.

src/frequenz/sdk/actor/power_distributing/_battery_status.py
```diff
@@ -0,0 +1,364 @@
...
+class BatteryStatus:
+    """Track the status of battery and make decision which should be used."""
+
+    _battery_invalid_relay: Set[BatteryRelayState] = {
+        BatteryRelayState.RELAY_STATE_OPENED
+    }
+    _battery_invalid_state: Set[BatteryState] = {BatteryState.COMPONENT_STATE_ERROR}
+    _inverter_invalid_state: Set[InverterState] = {
+        InverterState.COMPONENT_STATE_UNSPECIFIED,
+        InverterState.COMPONENT_STATE_UNAVAILABLE,
```
tiyash-basu-frequenz: Here are the valid states:
`BatteryRelayState`s: `CLOSED`
`BatteryState`s: `IDLE`, `CHARGING`, `DISCHARGING`
`InverterState`s: `STANDBY`, `IDLE`, `CHARGING`, `DISCHARGING`

You could either specify the complements of these as invalid states, or directly specify these as valid states. I think it's easier if you go with the latter, because the list of valid states is more stable than the list of invalid states.

src/frequenz/sdk/actor/power_distributing/_battery_status.py
```diff
@@ -0,0 +1,364 @@
...
+        """
+        self._max_component_time_diff: float = max_component_time_diff
+        self._battery_ids = battery_pool
+        if len(battery_pool) == 0:
+            _logger.warning("Given set of batteries is empty.")
+
+        self._run_method_called = False
+        # Time needed to get first component data.
+        # Constant for now. In future should be configurable in UI.
+        self._start_delay: float = 2.0
```
tiyash-basu-frequenz: I'd suggest using a constant instead of `2.0`.  That way you can keep track of all the constants being used in this module in a single place.

src/frequenz/sdk/actor/power_distributing/_battery_status.py
```diff
@@ -0,0 +1,364 @@
...
+    BatteryState,
+    InverterState,
+)
+from ._battery_status_utils import BatInvReceivers, BlockedComponents
+from .result import PartialFailure, Result, Success
+
+_logger = logging.getLogger(__name__)
+
+
+class BatteryStatus:
```
tiyash-basu-frequenz: To me this looks like a collection of methods belonging to two different categories: ones that work on a single component, and ones that work on a collection of components.

So, I'd suggest splitting this into two classes for cleaner layers:
`BatteryPool`: has methods that work on a collection of battery IDs
`BatteryUtils`: has functions for  a single battery ID, like whether it's online or not, what is it's connected inverter, etc.

src/frequenz/sdk/actor/power_distributing/_battery_status.py
```diff
@@ -0,0 +1,364 @@
...
+        if len(self._receivers) == 0:
+            _logger.error(
+                "Can't find batteries to track status. Given batteries: %s",
+                str(self._battery_ids),
+            )
+
+        await asyncio.sleep(self._start_delay)
+        self._run_method_called = True
+
+    def _is_msg_correct(
```
tiyash-basu-frequenz: I'd rename this to `_is_battery_ready`.

You could also alter the method signature to have the battery ID instead of battery and inverter data samples, and just get the status of the inverter connected to the battery.

src/frequenz/sdk/actor/power_distributing/_battery_status.py
```diff
@@ -0,0 +1,364 @@
...
+                    _logger.info(
+                        "Battery %d failed last response. Block it for %f sec",
+                        bat_id,
+                        duration,
+                    )
+
+            for bat_id in result.succeed_batteries:
+                self._blocked_batteries.unblock(bat_id)
+
+    def _is_msg_too_old(
```
tiyash-basu-frequenz: Maybe `_is_data_outdated`?

src/frequenz/sdk/actor/power_distributing/_battery_status_utils.py
```diff
@@ -0,0 +1,131 @@
...
+        Args:
+            now: current time
+
+        Returns:
+            True if timer expired, false otherwise.
+        """
+        return self.expiration_time <= now
+
+
+class BlockedComponents:
```
tiyash-basu-frequenz: Maybe it makes sense to add this to component-specific classes, instead of being separate?

Then you could run functions like
```
if battery_0.is_blocked():
    bla()
```

You could also add a method `blocked_until()`, which returns the timestamp until when the component is blocked.

tiyash-basu-frequenz (commented):
As suggested, reviewed from commit #5 onwards.

src/frequenz/sdk/microgrid/component/_component_errors.py
ela-kotulska-frequenz: class name is BatteryErrorCode.
I would have to rename it to BatteryErrorCodePb or battery_error_code_pb. That is not very readable, too.

The modules could be separated into Battery, Inverter, Common (as in frequenz-api).
Now they are separated into component_erros, component_states etc.
But it is not discussed yet.

src/frequenz/sdk/actor/power_distributing/_battery_status_utils.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/actor/power_distributing/_battery_status.py
ela-kotulska-frequenz: done

src/frequenz/sdk/actor/power_distributing/_battery_status.py
ela-kotulska-frequenz: done

src/frequenz/sdk/actor/power_distributing/_battery_status.py
ela-kotulska-frequenz: done

src/frequenz/sdk/actor/power_distributing/_battery_status.py
ela-kotulska-frequenz: done

src/frequenz/sdk/actor/power_distributing/_battery_status.py
ela-kotulska-frequenz: done

src/frequenz/sdk/microgrid/component/_component_data.py
```diff
@@ -128,6 +134,11 @@ class BatteryData(ComponentData):
             charging is possible.
         temperature_max: the maximum temperature of all the blocks in a battery, in
             Celcius (°C).
+        relay_state: state of the battery relays. Each Relay is described in the
```
shsms: `Each *state* is ...`?

src/frequenz/sdk/microgrid/component/_component_errors.py
```diff
@@ -0,0 +1,151 @@
...
+import frequenz.api.microgrid.common_pb2 as common_pb
+import frequenz.api.microgrid.inverter_pb2 as inverter_pb
+
+
+class ErrorLevel(Enum):
+    """Component error level."""
+
+    ERROR_LEVEL_UNSPECIFIED = common_pb.ErrorLevel.ERROR_LEVEL_UNSPECIFIED
+    ERROR_LEVEL_WARN = common_pb.ErrorLevel.ERROR_LEVEL_WARN
+    ERROR_LEVEL_CRITICAL = common_pb.ErrorLevel.ERROR_LEVEL_CRITICAL
```
shsms: we don't need to have the `ERROR_LEVEL_` prefix in our enums.  That would be consistent with how it is done in other such enums as well.

src/frequenz/sdk/actor/power_distributing/battery_pool.py
```diff
@@ -0,0 +1,121 @@
...
+    def __init__(
+        self,
+        battery_ids: Set[int],
+        max_data_age_sec: float,
+        max_blocking_duration_sec: float,
+    ) -> None:
+        """Create object instance.
+
+        Note:
+            The object instance is fully created after running `run` method.
```
shsms: Another option is to have a `BatteryPool.new()` async class method that creates and returns a BatteryPool instance.  And in the class documentation say "use `pool = await BatteryPool.new(...)` to create".

``` python

class BatteryPool:
    """Use new() to create."""
    _CreatedThroughNew = object()

    def __init__(..., created_by: Any = None):
        """Not for direct use.  Use new() to create."""
        if created_by is not _CreatedThroughNew:
             raise RuntimeError(...)
        ... rest of init

    async def new(...) -> BatteryPool:
        await etc
        return BatteryPool(...., cls._CreatedThroughNew)

```

ela-kotulska-frequenz: Yes but then it will be impossible to create BatteryPool in any constructor.
So user would have to either:
1) In constructor define battery pool as: `battery_pool: Optional[BatteryPool] = None` -> Create battery pool in other(async) method -> before every use check None condition. (which would end up (in most cases) with ignoring pylint/mypy warnings)
2) Define BatteryPool outside constructor and ingore mypy/pylint warnings that object is defined outside `__init__`

So from usage perspective it is better to define it in this way :(
It is true , that user has to remember about `run` method. But at least it is easy to use.

src/frequenz/sdk/microgrid/component/_component_errors.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/microgrid/component/_component_data.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/actor/power_distributing/battery_pool.py
shsms: True,  okay. :-)

Then what do you think about the name `async_init` (or `init_batteries`, etc)?  `run` sounds like it would start a long running task,  but that's not the case.

shsms:
I guess the PR title needs to be changed as well?

src/frequenz/sdk/microgrid/component/_component_errors.py
shsms: there are a few in `_component_states.py` as well.

src/frequenz/sdk/actor/power_distributing/battery_pool.py
ela-kotulska-frequenz: I go with `init` because:
* `async_init` would look like: `async def async_init(self)` . 
* `init` is very close to `__init__` to so it makes sense it is part of the init step.
*   `init_batteries` - it is good, but can be misunderstood. User might wonder why he has to init batteries etc.
* BatteryPool is private class, now. We can rename `init` to anything else when we publish. And until we publish it,  it will be no harm to user :)

src/frequenz/sdk/microgrid/component/_component_errors.py
ela-kotulska-frequenz: Thanks, done! :)

src/frequenz/sdk/microgrid/component/_component_states.py
```diff
@@ -39,3 +41,83 @@ def from_pb(
...
+
+class BatteryRelayState(Enum):
+    """Battery relay states."""
+
+    RELAY_STATE_UNSPECIFIED = battery_pb.RelayState.RELAY_STATE_UNSPECIFIED
+    RELAY_STATE_OPENED = battery_pb.RelayState.RELAY_STATE_OPENED
+    RELAY_STATE_PRECHARGING = battery_pb.RelayState.RELAY_STATE_PRECHARGING
+    RELAY_STATE_CLOSED = battery_pb.RelayState.RELAY_STATE_CLOSED
+    RELAY_STATE_ERROR = battery_pb.RelayState.RELAY_STATE_ERROR
+    RELAY_STATE_LOCKED = battery_pb.RelayState.RELAY_STATE_LOCKED
```
shsms: here too, I guess the prefix can go away.

shsms (commented):
Just looked at the first commit again.  Will finish review tomorrow.

src/frequenz/sdk/microgrid/component/_component_errors.py
tiyash-basu-frequenz: Not sure why we'd need to rename it to BatteryErrorCodePb or battery_error_code_pb. The enums being created here have the same variants as the protobuf ones, so we could just re-use the protobuf enums directly. (maybe I am missing something)

Couldn't we do something like the following?
```
from frequenz.api.microgrid.battery_pb2 import ErrorCodes as BatteryErrorCodes
```

Or is the actual intention here to get rid of the prefixes in the variants (`COMPONENT_STATE_`, `ERROR_CODE_`, etc.)?

src/frequenz/sdk/_data_ingestion/battery_utils.py
```diff
@@ -0,0 +1,334 @@
...
+        )
+        # If message is correct, but it wasn't before, then unblock it.
+        # This will allow to use battery shortly after it start working.
+        if is_ready and not self._last_ready:
+            self.unblock()
+        self._last_ready = is_ready
+
+        return is_ready
+
+    def is_blocked(self) -> bool:
```
tiyash-basu-frequenz: From an UX perspective, it would seem that `!battery.is_ready()` would achieve the same purpose. But it would not, therefore introducing a bit of inconsistency.

Looks like the timestamp check is not a part of the `is_ready` method. If you add it there, then the `is_blocked` method can be removed.

src/frequenz/sdk/actor/power_distributing/battery_pool.py
```diff
@@ -0,0 +1,122 @@
...
+                ready.add(bat_id)
+            if self._batteries[bat_id].is_blocked():
+                blocked.add(bat_id)
+
+        working = ready - blocked
+
+        if len(working) > 0:
+            return working
+
+        _logger.warning("All batteries in %s are not working. Use blocked batteries.")
```
tiyash-basu-frequenz: The part `Use blocked batteries.` sounds like an instruction, and can be confusing. What is the intention here?

src/frequenz/sdk/actor/power_distributing/battery_pool.py
```diff
@@ -0,0 +1,122 @@
...
+
+        if isinstance(result, Success):
+            for bat_id in result.used_batteries:
+                self._batteries[bat_id].unblock()
+
+        elif isinstance(result, PartialFailure):
+            for bat_id in result.failed_batteries:
+                duration = self._batteries[bat_id].block()
+                if duration > 0:
+                    _logger.info(
```
tiyash-basu-frequenz: I'd suggest setting the log level to `warn` for this message.

tiyash-basu-frequenz (commented):
Thanks for addressing the last set of comments. I have a few more.

src/frequenz/sdk/microgrid/component/_component_errors.py
shsms: no,  we just don't want to expose protobuf generated code to users of the sdk.

src/frequenz/sdk/actor/power_distributing/battery_pool.py
shsms: Maybe something like "There are no working batteries in %s.  Falling back to using blocked batteries."?

src/frequenz/sdk/_data_ingestion/battery_utils.py
ela-kotulska-frequenz: But then I can't get batteries that are blocked and ready.

ela-kotulska-frequenz: I can't just move it, because then I won't be able to get batteries with correct message but are blocked.

I renamed `is_ready` to `get_state` that returns ENUM with information if battery is working, blocked or not working.

src/frequenz/sdk/actor/power_distributing/battery_pool.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/actor/power_distributing/battery_pool.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/microgrid/component/_component_errors.py
tiyash-basu-frequenz: > no, we just don't want to expose protobuf generated code to users of the sdk.

This decision is being enforced too strictly, and is causing us to duplicate enums.
I feel this is part of a broader discussion. I will suggest discussing this in a different channel, unblocking this PR.

tiyash-basu-frequenz (dismissed):

src/frequenz/sdk/actor/power_distributing/battery_pool.py
```diff
@@ -0,0 +1,125 @@
...
+from typing import Set
+
+from ..._data_ingestion.battery_utils import BatteryUtils, WorkingState
+from .result import PartialFailure, Result, Success
+
+_logger = logging.getLogger(__name__)
+
+
+class BatteryPool:
+    """Holds bool of batteries and returns data from them."""
```
matthias-wende-frequenz: pool ;)

src/frequenz/sdk/_data_ingestion/battery_utils.py
```diff
@@ -0,0 +1,361 @@
...
+
+    def get_state(self) -> WorkingState:
+        """Return if battery message indicates that battery can be used.
+
+        Raises:
+            RuntimeError: if method `init` was not called or not awaited.
+
+        Returns:
+            True if battery can be used, False otherwise.
+        """
```
shsms: The documentation seems to suggest that this returns a bool?

tests/actor/test_battery_pool.py
```diff
@@ -0,0 +1,929 @@
...
+
+import logging
+from dataclasses import dataclass
+from datetime import datetime, timedelta
+from typing import Callable, List, Optional, Set
+
+import frequenz.api.microgrid.microgrid_pb2 as microgrid_pb
+import pytz
+import time_machine
+from frequenz.api.microgrid.battery_pb2 import Battery as PbBatteryData
```
shsms: I think these should be aliased just as `PbBattery` for consistency,  because there's `from ...battery_pb2 import Data as PbBatteryData` somewhere else.

src/frequenz/sdk/actor/power_distributing/battery_pool.py
```diff
@@ -0,0 +1,125 @@
...
+                timeout used for blocking failing component.
+        """
+        self._batteries = {
+            id: BatteryUtils(id, max_data_age_sec, max_blocking_duration_sec)
+            for id in battery_ids
+        }
+        self._run_method_called: bool = False
+
+    async def init(self) -> None:
+        """Run battery pool."""
```
shsms: Just `init`, not `run`, I guess?

src/frequenz/sdk/actor/power_distributing/battery_pool.py
```diff
@@ -0,0 +1,125 @@
...
+        Args:
+            result: Summary of what batteries failed and succeed in last request.
+
+        Raises:
+            RuntimeError: If `run` method was not called at the beginning to initialize
+                object.
+        """
+        if not self._run_method_called:
+            raise RuntimeError(
+                "`run` method not called or not awaited. Run it before first use"
```
shsms: `_init_called`?  and `init method not called ...`.  In the docstring also.

src/frequenz/sdk/actor/power_distributing/battery_pool.py
```diff
@@ -0,0 +1,125 @@
...
+    def __init__(
+        self,
+        battery_ids: Set[int],
+        max_data_age_sec: float,
+        max_blocking_duration_sec: float,
+    ) -> None:
+        """Create partially initialized object instance.
+
+        Note:
+            Please call `run` method to fully initialize BatteryPool. Otherwise
```
shsms: `init`

src/frequenz/sdk/actor/power_distributing/battery_pool.py
```diff
@@ -0,0 +1,125 @@
...
+            RuntimeError: If `run` method was not called at the beginning to
+                initialized object.
+            KeyError: If any battery in the given batteries is not in the pool.
+
+        Returns:
+            Subset of given batteries with working batteries.
+        """
+        if not self._run_method_called:
+            raise RuntimeError(
+                "`run` method not called or not awaited. Run it before first use"
```
shsms: init

shsms (commented):
I think these classes should be called something else.  We have already reserved the name `BatteryPool` for the other thing that runs formulas on resampled data and sits in the `timeseries` module.

The class in this PR is really a helper class for the PowerDistributor, that operates on direct component data,  and is tightly coupled with the PowerDistributor.

These classes are really just about `Status` like the original name suggested,  and we might want to retain that word in the names.   Maybe something like "BatteryStatus" and "BatteryPoolStatus".

src/frequenz/sdk/actor/power_distributing/battery_pool.py
shsms: The only issue is that now when I say "the init method", it is ambiguous which one I mean.

We'll probably end up calling it the "async init" anyway.

> `async_init` would look like: `async def async_init(self)` .

Is that bad?

src/frequenz/sdk/_data_ingestion/battery_utils.py
```diff
@@ -0,0 +1,361 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+"""Class to store features for battery."""
```
leandro-lucarella-frequenz: The `_data_ingestion` package is supposed to die, what about putting this in `microgrid` instead? Also, what about making it private (`_battery_utils.py`) if it's not intended to be used by SDK users directly (or is it?). While we are at it, would it make sense to have a `battery` package (if we are going to add more modules related to battery handling)? Like `sdk.microgrid.battery.util`?

src/frequenz/sdk/_data_ingestion/battery_utils.py
```diff
@@ -0,0 +1,361 @@
...
+    WORKING = 2
+
+
+@dataclass
+class _ComponentReceiver(Generic[T]):
+    component_id: int
+    receiver: Peekable[T]
+
+
+class BatteryUtils:
```
leandro-lucarella-frequenz: I'm not crazy about this name. Isn't this mainly keeping track of the status of the battery? Why not `BatteryStatus`? Alternatives: `BatteryController`, `BatteryBlocker`, *util* is way too ambiguous and general IMHO.

src/frequenz/sdk/_data_ingestion/battery_utils.py
```diff
@@ -0,0 +1,361 @@
...
+        Battery can be blocked if last request for that battery failed.
+
+        Returns:
+            True if battery is blocked, False otherwise.
+        """
+        if self._blocked_until is None:
+            return False
+        return self._blocked_until > datetime.now(tz=pytz.UTC)
+
+    def unblock(self):
```
leandro-lucarella-frequenz: `-> None`, weird if it's not failing the linting...

src/frequenz/sdk/_data_ingestion/battery_utils.py
```diff
@@ -0,0 +1,361 @@
...
+
+        # Increase blocking duration twice or until it reach the max.
+        self._last_blocking_duration_sec = min(
+            2 * self._last_blocking_duration_sec, self._max_blocking_duration_sec
+        )
+        self._blocked_until = now + timedelta(seconds=self._last_blocking_duration_sec)
+
+        return self._last_blocking_duration_sec
+
+    def blocked_until(self) -> Optional[datetime]:
```
leandro-lucarella-frequenz: `@property`?

src/frequenz/sdk/_data_ingestion/battery_utils.py
```diff
@@ -0,0 +1,361 @@
...
+        """Return timestamp when the battery will be unblocked.
+
+        Returns:
+            Timestamp when the battery will be unblocked. Return None if battery is
+            not blocked.
```
leandro-lucarella-frequenz: Needs extra indentation to continue the line:
```py
        Returns:
            Timestamp when the battery will be unblocked. Return None if battery is
                not blocked.
```

src/frequenz/sdk/_data_ingestion/battery_utils.py
```diff
@@ -0,0 +1,361 @@
...
+            self.unblock()
+
+        if self._is_blocked():
+            self._last_state = WorkingState.UNCERTAIN
+        else:
+            self._last_state = WorkingState.WORKING
+
+        return self._last_state
+
+    def _is_blocked(self) -> bool:
```
leandro-lucarella-frequenz: Wouldn't it make sense to make this a public property? I know there is `blocked_util` too, but if you just want to know if it's blocked or not it feels weirder to do:
```py
if b.blocked_until is not None:
    print("is blocked")
```

src/frequenz/sdk/actor/power_distributing/battery_pool.py
```diff
@@ -0,0 +1,125 @@
...
+from asyncio import ALL_COMPLETED
+from typing import Set
+
+from ..._data_ingestion.battery_utils import BatteryUtils, WorkingState
+from .result import PartialFailure, Result, Success
+
+_logger = logging.getLogger(__name__)
+
+
+class BatteryPool:
```
leandro-lucarella-frequenz: (Small typo in the commit message: ***Int** the future it will be an actor that streams*)

I would put this class in `sdk.microgrid.battery` instead, as it is not an actor. IMHO actors should only be a thin wrapper over other reusable modules in the SDK.

Also, we need to start asking the question, is this class intended to be used by SDK users or it is internal, because if it internal it should be in a private module (starting with `_`).

src/frequenz/sdk/actor/power_distributing/battery_pool.py
```diff
@@ -0,0 +1,125 @@
...
+            max_blocking_duration_sec: This value tell what should be the maximum
+                timeout used for blocking failing component.
+        """
+        self._batteries = {
+            id: BatteryUtils(id, max_data_age_sec, max_blocking_duration_sec)
+            for id in battery_ids
+        }
+        self._run_method_called: bool = False
+
+    async def init(self) -> None:
```
leandro-lucarella-frequenz: I've seen this pattern twice now, and I wonder if users will **always** need to do:

```py
c = AsyncClass(...)
await c.init()
```

Then we should probably add an alternative async constructor instead:

```py

@classmethod
async def new(cls, ...) -> Self:  # Maybe this didn't work yet, even with typing_extensions, but we can use the concrete class for now
    o = cls(...)
    await o._init()
    return o
```

We could take it even further and forbid using the constructor:

```py
def __init__(self):
    # For documentation purposes maybe we can declare the instance attribute types
    raise NotImplementedError(f"Use {self.__class__.__name__}.new() instead")

@classmethod
async def new(cls, ...) -> Self:  # Maybe this didn't work yet, even with typing_extensions, but we can use the concrete class for now
    o = cls.__new__(cls)
    # initialize attributes
    await o._init()
    return o
```

Just an idea, but it would make it less error prone, not giving users the possibility to forget to call `init()`

src/frequenz/sdk/actor/power_distributing/battery_pool.py
leandro-lucarella-frequenz: Making instantiation and async init one step will also avoid having to do all these checks (or make them just an `assert` as users would have to be doing something really weird to get into a inconsistent state).

src/frequenz/sdk/_data_ingestion/battery_utils.py
shsms: For the other class also, it is also not really the data-pipeline BatteryPool that we have been talking about,  but something specific to the the PowerDistributor: https://github.com/frequenz-floss/frequenz-sdk-python/pull/117#pullrequestreview-1227363605

leandro-lucarella-frequenz: If it is stuff internal to the power distributor, then I think it should live in `sdk.power` (or `actor.power_distribution` if this really is only intended to be used by the actor and nobody else) as internal modules `_battery_status.py` and `_battery_pool.py`. For me it is still OK to use the name if it makes sense, but it should be definitely very clear that it is some internal implementation detail of the power distributor then.

shsms: No reason to reuse names and introduce ambiguities when we don't have to.

src/frequenz/sdk/actor/power_distributing/battery_pool.py
ela-kotulska-frequenz: @leandro-lucarella-frequenz please read answer below :) 
> Making instantiation and async init one step will also avoid having to do all these checks 

https://github.com/frequenz-floss/frequenz-sdk-python/pull/117#discussion_r1050787860

src/frequenz/sdk/_data_ingestion/battery_utils.py
ela-kotulska-frequenz: BatteryPool is now used in PowerDistributor. But in the future it won't.

PowerDistirbutor will not track what batteries are working in future. It will just distribute power over given batteries.
User will use BatteryPool.set_power(...). And BatteryPool will use PowerDistributor.

src/frequenz/sdk/actor/power_distributing/battery_pool.py
ela-kotulska-frequenz: Yes, but now it needs PowerDistirbutingResult. And PowerDistributing needs BatteryPool. And that is circular dependency  :( 
I the future it will be moved.
I will make it clear in commit message.

ela-kotulska-frequenz: > Also, we need to start asking the question, is this class intended to be used by SDK users or it is internal, because if it internal it should be in a private module (starting with _).

It is intended to be used by users in the future

src/frequenz/sdk/actor/power_distributing/battery_pool.py
ela-kotulska-frequenz: Yes we discussed it: and it is better to do it this way.
https://github.com/frequenz-floss/frequenz-sdk-python/pull/117#discussion_r1050787860

src/frequenz/sdk/actor/power_distributing/battery_pool.py
ela-kotulska-frequenz: async before function definition tells that this method is async.

ela-kotulska-frequenz: Ok I will rename it to async_init.

`async def async_init(self) ` - no, it is just lots of `async`

src/frequenz/sdk/actor/power_distributing/battery_pool.py
leandro-lucarella-frequenz: OK, as long as you are having what should be public interface and what should be internal and hidden, all good if we need to go through an intermediary step where things are not where they should in a final version :+1:

leandro-lucarella-frequenz: (OTOH as long as stuff is temporary I would still make the module hidden to make it clear for users that if they use that module it can disappear without notice at any time)

src/frequenz/sdk/actor/power_distributing/battery_pool.py
leandro-lucarella-frequenz: Ah, sorry I missed the conversation there. I will continue the discussion there.

leandro-lucarella-frequenz: Ah, sorry I missed the conversation there. I will continue the discussion there.

src/frequenz/sdk/actor/power_distributing/battery_pool.py
ela-kotulska-frequenz: Done

leandro-lucarella-frequenz: Damn GitHub, it lost my comment when I hit unresolve conversation :sob: 

I tried to say that I don't fully agree with not being able to use this class as other class attributes in constructors. There is another option, which is what happens in general with asyc. Async is viral, and if you want to use it, you need to make your stuff async and IMHO this should happen with constructors too, otherwise we are just making users need to write more error prone code.

So IMHO users that want to have a `BatteryPool` in their class should make their class *async-constructible* too, otherwise they will also need to split sync construction from async construction. We can add a utility class `AsyncConstructible` to allow users to also implement their own async constructible classes more easily, for example:

```py
class NotSyncConstructible(AssertError):
    pass


class AsyncConstructible(ABC):
    def __init__(self) -> None:
        raise NotSyncConstructible()

    @abstractmethod
    @classmethod
    async def async_new(cls) -> Self:  # or AsyncConstructible if it still doesn't work
        return cls.__new__(cls)


class BatteryPool(AsyncConstructible):
    def __init__(self) -> None:
        self.attr1: str
        """Documentation for attr1."""
        self.attr2: int
        """Documentation for attr2."""
        
        super().__init__()

    @classmethod
    def async_new(cls, ...) -> Self:  # or BatteryPool if it still doesn't work
        self = await super().async_new()
        self.attr1 = "lala"
        self.attr2 = 2
        return self
```

We just need to check if we can somehow document instance attributes.

All that said, I'm also fine to leave this PR as is and create a new issue for this, as it is not strictly related to the PR and it might need more discussion.

PS: I know this also add some error proneness, as users need to make sure to define their classes correctly, but at least you need to define a class (and get it right) only once, while if the error proneness is in the usage, you have to do it several times, having more chances to mess it up.

src/frequenz/sdk/microgrid/component/_component_states.py
ela-kotulska-frequenz: No enums anymore

src/frequenz/sdk/actor/power_distributing/battery_pool.py
leandro-lucarella-frequenz: Ah, sorry I missed the conversation there

src/frequenz/sdk/actor/power_distributing/battery_pool.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/_data_ingestion/battery_utils.py
ela-kotulska-frequenz: Done

tests/actor/test_battery_pool.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/actor/power_distributing/battery_pool.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/actor/power_distributing/battery_pool.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/actor/power_distributing/battery_pool.py
ela-kotulska-frequenz: done

src/frequenz/sdk/_data_ingestion/battery_utils.py
ela-kotulska-frequenz: Changed to `_status.py` and moved to `sdk.microgrid._battery._status`
New name for the class `StatusTracker`

src/frequenz/sdk/_data_ingestion/battery_utils.py
ela-kotulska-frequenz: BatteryPool is used in PowerDistributor temporary. It will be extended to use the formulas and exposed to user.

src/frequenz/sdk/_data_ingestion/battery_utils.py
ela-kotulska-frequenz: done

src/frequenz/sdk/_data_ingestion/battery_utils.py
ela-kotulska-frequenz: done

src/frequenz/sdk/_data_ingestion/battery_utils.py
ela-kotulska-frequenz: done

src/frequenz/sdk/_data_ingestion/battery_utils.py
ela-kotulska-frequenz: done

src/frequenz/sdk/actor/power_distributing/battery_pool.py
ela-kotulska-frequenz: OK I made it private

src/frequenz/sdk/actor/power_distributing/battery_pool.py
ela-kotulska-frequenz: > _init_called? and init method not called .... In the docstring also.

Done

src/frequenz/sdk/actor/power_distributing/battery_pool.py
ela-kotulska-frequenz:
```
class BatteryPool(AsyncConstructible):
    @classmethod
    def async_new(cls, ...) -> Self:  # or BatteryPool if it still doesn't work
        self = await super().__async_new()
        # Assign attributes, do more async stuff
        return self
```
About line     # Assign attributes, do more async stuff
If attribute was not assigned in method with self argument, then pylint doesn't recognize this attribute.  So there is lots of pylint errors that we have to quiet to make it work.

We should skip this topic in this PR and continue in different issue. 
https://github.com/frequenz-floss/frequenz-sdk-python/issues/135

ela-kotulska-frequenz:
Renamed BatteryPool to BatteryPoolStatus as discussed on the meeting.

src/frequenz/sdk/microgrid/component/_component_data.py
```diff
@@ -128,6 +130,9 @@ class BatteryData(ComponentData):
             charging is possible.
         temperature_max: the maximum temperature of all the blocks in a battery, in
             Celcius (°C).
+        _relay_state: state of the battery relay.
+        _component_state: state of the battery.
+        _errors: list of errors in protobuf struct.
```
leandro-lucarella-frequenz: It is OK for this PR, as there were already attributes documented like this, but in the future we need to move from documenting in `Attributes:` to documenting as docstrings where the attributes are defined, as the documentation tool we use renders the attributes much better if they are documented there.

src/frequenz/sdk/microgrid/component/_component_data.py
```diff
@@ -137,6 +142,9 @@ class BatteryData(ComponentData):
     power_lower_bound: float
     power_upper_bound: float
     temperature_max: float
+    _relay_state: battery_pb.RelayState.ValueType
+    _component_state: battery_pb.ComponentState.ValueType
+    _errors: List[battery_pb.Error]
```
leandro-lucarella-frequenz: So here, like:
```py
    _relay_state: battery_pb.RelayState.ValueType
    """State of the battery relay."""
    _component_state: battery_pb.ComponentState.ValueType
    """State of the battery."""
    _errors: List[battery_pb.Error]
    """List of errors in protobuf struct."""
```

With dataclasses it will say they are class attributes, but it is also marked as a dataclass so it is still easy for users to know they are really instance attributes.

src/frequenz/sdk/microgrid/_battery/_status.py
```diff
@@ -0,0 +1,373 @@
...
+        """Return if battery is blocked.
+
+        Battery can be blocked if last request for that battery failed.
+
+        Returns:
+            True if battery is blocked, False otherwise.
+        """
+        if self._blocked_until is None:
+            return False
+        return self._blocked_until > datetime.now(tz=pytz.UTC)
```
leandro-lucarella-frequenz: Nitpick: I would avoid using `pytz` only for `UTC`, as it is available in the stdlib as `datetime.timezone.utc`.

src/frequenz/sdk/actor/power_distributing/_battery_pool_status.py
```diff
@@ -0,0 +1,126 @@
...
+        }
+        self._init_method_called: bool = False
+
+    async def async_init(self) -> None:
+        """Init battery pool."""
+        tasks = [
+            asyncio.create_task(battery.async_init())
+            for battery in self._batteries.values()
+        ]
+        await asyncio.wait(tasks, return_when=ALL_COMPLETED)
```
leandro-lucarella-frequenz: Minor: I guess it could be jusut `asynio.gather()`, right?

tests/actor/test_power_distributing.py
```diff
@@ -232,14 +229,23 @@ async def test_power_distributor_one_user(
...
+            request_timeout_sec=SAFETY_TIMEOUT,
         )
 
         with mock.patch("asyncio.sleep", new_callable=AsyncMock):
             distributor = PowerDistributingActor(
                 mock_api, component_graph, {"user1": channel.service_handle}
             )
 
+            # Mock that all requested batteries are working.
+            distributor._battery_pool = MagicMock()
```
leandro-lucarella-frequenz: Something that we are not using very often but we probably should is the `spec` argument for mocks, it will validate that the mock is only used via attributes or calls that follow some object or class. For example: `distributor._battery_pool = MagicMock(spec=BatteryPoolStatus)` should create a mock object that already has a `async_init()` attribute and it is an `AsyncMock` because it knows the signature of the method, and will also validate that the arguments of the method follow the specs.

leandro-lucarella-frequenz (commented):
Some final minor comments. If they are addressed, great but otherwise from my side this can go as it is too. I will leave the approval pending because there were other reviewers that had comments.

src/frequenz/sdk/microgrid/component/_component_data.py
ela-kotulska-frequenz: Done (last commit).

src/frequenz/sdk/microgrid/component/_component_data.py
ela-kotulska-frequenz: Done (last commit).

src/frequenz/sdk/microgrid/_battery/_status.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/actor/power_distributing/_battery_pool_status.py

tests/actor/test_power_distributing.py
ela-kotulska-frequenz: I didn't know that. Thanks, done

sahas-subramanian-frequenz (commented):
LGTM,  I can approve once it is rebased.

sahas-subramanian-frequenz (approved):

PR #125, opened 2022-12-20
Fix warnings in unit tests

ela-kotulska-frequenz:
This PR cleans some warnings in pytest (Python 3.8).
More work is required to clean all warnings.
But this is enough for one PR.

### DIFF
--- a/tests/microgrid/test_client.py
+++ b/tests/microgrid/test_client.py
@@ -130,7 +130,7 @@ class TestMicrogridGrpcClient:
             }
 
         finally:
-            await server.stop(grace=1.0)
+            assert await server.graceful_shutdown()
 
     async def test_connections(self) -> None:
         servicer = mock_api.MockMicrogridServicer()
@@ -284,7 +284,7 @@ class TestMicrogridGrpcClient:
             }
 
         finally:
-            await server.stop(grace=1.0)
+            assert await server.graceful_shutdown()
 
     async def test_bad_connections(self) -> None:
         """Validate that the client does not apply connection filters itself."""
@@ -355,7 +355,7 @@ class TestMicrogridGrpcClient:
             )
 
         finally:
-            await server.stop(grace=1.0)
+            assert await server.graceful_shutdown()
 
     async def test_meter_data(self) -> None:
         servicer = mock_api.MockMicrogridServicer()
@@ -383,7 +383,7 @@ class TestMicrogridGrpcClient:
             await asyncio.sleep(0.2)
 
         finally:
-            await server.stop(0.1)
+            assert await server.graceful_shutdown()
 
         latest = peekable.peek()
         assert isinstance(latest, MeterData)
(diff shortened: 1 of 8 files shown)


tests/actor/test_channel_registry.py
```diff
@@ -3,9 +3,12 @@
 
 """Tests for the ChannelRegistry."""
 
+import pytest
+
 from frequenz.sdk.actor import ChannelRegistry
 
 
+@pytest.mark.asyncio
```
sahas-subramanian-frequenz: I think we already get these because we have them configured:  https://github.com/frequenz-floss/frequenz-sdk-python/blob/6ab8228f1227ac6a7742d26045135e1b38419cf8/pyproject.toml#L100

https://pytest-asyncio.readthedocs.io/en/latest/concepts.html#auto-mode

sahas-subramanian-frequenz (commented):
Thanks Ela,  for spending time on this.  I'm not sure we need the first commit,  but the other two make sense to me.

tests/actor/test_channel_registry.py
ela-kotulska-frequenz: True, thanks!
Commit dropped.

sahas-subramanian-frequenz (dismissed):
Wow,  hard to believe the test output can be so clean.

tests/microgrid/test_client.py
```diff
@@ -130,7 +130,7 @@ async def test_components(self) -> None:
             }
 
         finally:
-            await server.stop(grace=1.0)
+            assert await server.graceful_shutdown()
```
leandro-lucarella-frequenz, line 133: This is a tricky one IMHO, just because a mock server doesn't want to shut down, it doesn't really mean that the test failed. The other option is also not great, so for me it can go either way, but I wanted to put it out there just for general consideration.

leandro-lucarella-frequenz (commented):
Also very grateful you addressed these :heart: 

Just a couple text things, I don't get the PR title, what it "ut main"? Now that PR titles (and body) are used (by default) as the merge commit message it is important that it is clear.

Also super nitpick, commit titles should end without a full stop (".") :)

tests/microgrid/test_client.py
ela-kotulska-frequenz: Yes I know :( 
But it is the only way. Otherwise it will crash in random test :/

shsms: And I'd think having clean test outputs is a lot more valuable than being flexible towards mock server behaviour.

sahas-subramanian-frequenz (approved):

tests/microgrid/test_client.py
leandro-lucarella-frequenz: Yeah, the ideal would be to have an "error" result to know the test didn't fail but there was something about the test itself that went wrong. But I agree is better to fail the test than to succeed it.

PR #130, opened 2022-12-21
Implement a custom resampling buffer

leandro-lucarella-frequenz:
Add a custom resampling buffer

The resampling buffer uses a list as a backing container to implement a push-only ring buffer that can be also easily sliced.

Sadly the custom resampling buffer sucks. I think we need to implement it in C/Rust if we want to make it fast. Here are the results of the benchmark:

```
         OLD                           NEW                    NEW WITH BISECT          CUSTOM BUFFER WITH BISECT

resamples=10 samples=10      resamples=10 samples=10      resamples=10 samples=10      resamples=10 samples=10     
0.0008896420185919851        0.00062773801619187          0.0008012260077521205        0.0019748720005736686       
resamples=10 samples=100     resamples=10 samples=100     resamples=10 samples=100     resamples=10 samples=100    
0.007817161997081712         0.005761806009104475         0.006307974021183327         0.014143209002213553        
resamples=10 samples=1000    resamples=10 samples=1000    resamples=10 samples=1000    resamples=10 samples=1000   
0.07768873398890719          0.05851042701397091          0.0604277040110901           0.12536667400127044         
resamples=100 samples=10     resamples=100 samples=10     resamples=100 samples=10     resamples=100 samples=10    
0.008742492995224893         0.0062527229893021286        0.00808265499654226          0.021197415997448843        
resamples=100 samples=100    resamples=100 samples=100    resamples=100 samples=100    resamples=100 samples=100   
0.07808284999919124          0.057997508003609255         0.0624066719901748           0.13266414900135715         
resamples=100 samples=1000   resamples=100 samples=1000   resamples=100 samples=1000   resamples=100 samples=1000  
0.782658567011822            0.5870920980232768           0.6098103950207587           1.4532136470006662          
resamples=1000 samples=10    resamples=1000 samples=10    resamples=1000 samples=10    resamples=1000 samples=10   
0.08764891701866873          0.062448524025967345         0.07815460601705126          0.22267909600486746         
resamples=1000 samples=100   resamples=1000 samples=100   resamples=1000 samples=100   resamples=1000 samples=100  
0.78426024899818             0.5858371119829826           0.6357307220168877           1.5354863690008642          
resamples=1000 samples=1000  resamples=1000 samples=1000  resamples=1000 samples=1000  resamples=1000 samples=1000 
7.513815971993608            5.984694316983223            6.42200836900156             12.64786391400412           
```

So this PR is just opened for documentation purposes, but it won't be merged.

Issue #55, opened 2022-11-01
Make the resampler infer the data rate of the input


leandro-lucarella-frequenz:
### What's needed?

We need to infer the input data rate for the resampler for several reasons:

- Use dynamic ring buffer sizes depending on the input data rate, so components with higher data rates can get bigger buffers (up to a configurable maximum) and lower data rates can get smaller buffers.
- The input data rate could be used also to automatically pick a good resampling function (we can know, for example, if we are up or downsampling).

### Proposed solution

Use the input timestamps to estimate the input rate.

PR #131, opened 2022-12-21
Improve resampler buffer and config

leandro-lucarella-frequenz:
This add `typing_extensions` to the dependencies, even if it is not needed by this PR in particular.

We also move the resampler configuration to its own class, as it will start growing in options and otherwise there is too much duplication of a huge list of arguments that need to be repeated in multiple classes.

Finally we improve the performance of the resampling buffer (adding a benchmark to check it) and add a comment about the failed experiment to build our own custom resampling buffer.

### DIFF
--- a/benchmarks/timeseries/resampling.py
+++ b/benchmarks/timeseries/resampling.py
@@ -25,6 +25,7 @@ def _benchmark_resampling_helper(resamples: int, samples: int) -> None:
             resampling_period_s=1.0,
             max_data_age_in_periods=3.0,
             resampling_function=nop,
+            initial_buffer_len=samples * 3,
         )
     )
     now = datetime.now(timezone.utc)
--- a/benchmarks/timeseries/resampling.py
+++ b/benchmarks/timeseries/resampling.py
@@ -1,0 +1,51 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+
+"""Benchmark resampling."""
+
+from datetime import datetime, timedelta, timezone
+from timeit import timeit
+from typing import Sequence
+
+from frequenz.sdk.timeseries import Sample
+from frequenz.sdk.timeseries._resampling import ResamplerConfig, _ResamplingHelper
+
+
+def nop(  # pylint: disable=unused-argument
+    samples: Sequence[Sample], resampling_period_s: float
+) -> float:
+    """Return 0.0."""
+    return 0.0
+
+
+def _benchmark_resampling_helper(resamples: int, samples: int) -> None:
+    """Benchmark the resampling helper."""
+    helper = _ResamplingHelper(
+        ResamplerConfig(
+            resampling_period_s=1.0,
+            max_data_age_in_periods=3.0,
+            resampling_function=nop,
+        )
+    )
+    now = datetime.now(timezone.utc)
+
+    def _do_work() -> None:
+        nonlocal now
+        for _n_resample in range(resamples):
+            for _n_sample in range(samples):
+                now = now + timedelta(seconds=1 / samples)
+                helper.add_sample(Sample(now, 0.0))
+            helper.resample(now)
+
+    print(timeit(_do_work, number=5))
+
+
+def _benchmark() -> None:
+    for resamples in [10, 100, 1000]:
+        for samples in [10, 100, 1000]:
+            print(f"{resamples=} {samples=}")
+            _benchmark_resampling_helper(resamples, samples)
+
+
+if __name__ == "__main__":
+    _benchmark()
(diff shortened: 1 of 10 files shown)


leandro-lucarella-frequenz:
For a little bit of more context, I added `typing_extensions` because it was long overdue, this allow us to use more modern typing features that are not present in older Python versions. I want to use some features several times but thought "mmm, maybe it's not worth it just for this". But I think it is, because is an official extension that allow us to keep our code more forward compatible, is some sort of complement to `from __future__ import annotations`.

Some changes also might look a bit out of context now for 2 reasons:

* Failure of #130
* I also started this to address #55, so some changes will be useful to advance in that direction (config split).

benchmarks/timeseries/resampling.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def nop(  # pylint: disable=unused-argument
+    samples: Sequence[Sample], resampling_period_s: float
+) -> float:
+    """Return 0.0."""
+    return 0.0
+
+
+def _benchmark_resampling_helper(resamples: int, samples: int) -> None:
```
ela-kotulska-frequenz, line 21: Can we start writing in the benchmark files information when we last tested it and what was the result?
We could have short history with date and result.

benchmarks/timeseries/resampling.py
```diff
@@ -0,0 +1,51 @@
...
+        )
+    )
+    now = datetime.now(timezone.utc)
+
+    def _do_work() -> None:
+        nonlocal now
+        for _n_resample in range(resamples):
+            for _n_sample in range(samples):
+                now = now + timedelta(seconds=1 / samples)
+                helper.add_sample(Sample(now, 0.0))
```
ela-kotulska-frequenz, line 38: I think you should add random value here.
Otherwise it could be optimized.
PS. I am not sure if it can be optimized without Compiler... It is easier to add random value, then check it :)

src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -352,18 +347,23 @@ def resample(self, timestamp: datetime) -> Sample:
...
+        conf = self._config
+        minimum_relevant_timestamp = timestamp - timedelta(
+            seconds=conf.resampling_period_s * conf.max_data_age_in_periods
         )
-        self._remove_outdated_samples(threshold=threshold)
-
+        # We need to pass a dummy Sample to bisect because it only support
+        # specifying a key extraction function in Python 3.10, so we need to
+        # compare samples at the moment.
+        cut_index = bisect(self._buffer, Sample(minimum_relevant_timestamp, None))
```
ela-kotulska-frequenz, line 357: Sorry, I am outdated...
Why we need bisect in the ring buffer?

benchmarks/timeseries/resampling.py
leandro-lucarella-frequenz: That's a good idea, but I'm not sure I would write it in the benchmark file itself, maybe a Wiki? Also, if we do this, we need to take it with a **BIG** grain of salt, because results will be completely different when ran in different hardware, so at the end I'm not sure this will provide any useful information, unless we have some sort of standard hardware to run the benchmarks on.

benchmarks/timeseries/resampling.py
leandro-lucarella-frequenz: What could be optimized exactly? The `now` changes, and the value is not even used (the resampling function is a NOP). The whole Sample object for sure won't be optimized away :)

src/frequenz/sdk/timeseries/_resampling.py
leandro-lucarella-frequenz: Because now it is fixed size, so we add samples as they come and never removed them, only when they "drop out" of the circular buffer. If we stop receiving for example, the buffer will be filled, but with old data, so we need to make sure to pass only relevant data to the resampling function. I'm bisecting to find the oldest relevant timestamp in the buffer.

benchmarks/timeseries/resampling.py
ela-kotulska-frequenz: hmmm true... maybe `microgrid.sandbox.api.frequenz.io` for now?
I would write results in file. Otherwise everyone would forget it, and it will be harder to track history of benchmark files.

benchmarks/timeseries/resampling.py
ela-kotulska-frequenz: OWW sorry , Of course you are right :)

ela-kotulska-frequenz (dismissed):

benchmarks/timeseries/resampling.py
leandro-lucarella-frequenz: I created an internal issue to keep track of this to avoid blocking this PR with a topic that is much broader.

ela-kotulska-frequenz (approved):

PR #136, opened 2022-12-27
Make resampler infer the input sampling period and buffer length

### LINKED ISSUE #55: Make the resampler infer the data rate of the input

leandro-lucarella-frequenz:
The input sampling period is calculated by counting the total received samples, and dividing the total resampling time until the internal buffer is full by the total received samples.

To store all the source properties (sampling period, sampling start, total received samples) a new class is used (SourceProperties) and this information is passed to the resampling function, so it can have more information about the source to perform a proper resampling.

This calculation is performed once and then remains constant, but at some point we could add a timer to re-calculate it.

This also improves slightly the documentation and validation of the ResamplingConfig class.

The add_timeseries() methods now also takes a name as a way to identify different sources in log messages (the actor uses the channel name as timeseries name).

We also update the resampling buffer length based on the input sampling period, resampling period, and max data age in periods, so it can hold all the past data without wasting any space.

Fixes #55.

### DIFF
--- a/src/frequenz/sdk/timeseries/_resampling.py
+++ b/src/frequenz/sdk/timeseries/_resampling.py
@@ -32,6 +32,21 @@ sure all the requested past sampling periods can be stored.
 """
 
 
+DEFAULT_BUFFER_LEN_MAX = 1024
+"""Default maximum allowed buffer length.
+
+If a buffer length would get bigger than this, it will be truncated to this
+length.
+"""
+
+
+DEFAULT_BUFFER_LEN_WARN = 128
+"""Default minimum buffer length that will produce a warning.
+
+If a buffer length would get bigger than this, a warning will be logged.
+"""
+
+
 Source = AsyncIterator[Sample]
 """A source for a timeseries.
 
@@ -131,6 +146,51 @@ class ResamplerConfig:
     like sampling rate, to make sure all the requested past sampling periods
     can be stored.
     """
+
+    warn_buffer_len: int = DEFAULT_BUFFER_LEN_WARN
+    """The minimum length of the resampling buffer that will emit a warning.
+
+    If a buffer grows bigger than this value, it will emit a warning in the
+    logs, so buffers don't grow too big inadvertly.
+    """
+
+    max_buffer_len: int = DEFAULT_BUFFER_LEN_MAX
+    """The maximum length of the resampling buffer.
+
+    Buffers won't be allowed to grow beyond this point even if it would be
+    needed to keep all the requested past sampling periods. An error will be
+    emitted in the logs if the buffer length needs to be truncated to this
+    value.
+    """
+
+    def __post_init__(self) -> None:
+        """Check config values are valid.
+
+        Raises:
+            ValueError: if the initial buffer length is too small (less than 2)
+                or too big (more than `max_buffer_len`).
+        """
+        if self.initial_buffer_len < 2:
+            raise ValueError(
+                f"initial_buffer_len must be at least 2, got {self.initial_buffer_len}"
+            )
+        if self.initial_buffer_len > self.max_buffer_len:
+            raise ValueError(
+                f"initial_buffer_len be smaller than {self.max_buffer_len}, "
+                "got {self.initial_buffer_len}"
+            )
+        if self.initial_buffer_len > self.warn_buffer_len:
+            _logger.warning(
+                "initial_buffer_len (%s) is bigger than %s",
+                self.initial_buffer_len,
+                self.warn_buffer_len,
+            )
+        if self.initial_buffer_len > self.warn_buffer_len:
+            _logger.warning(
+                "initial_buffer_len (%s) is bigger than warn_buffer_len (%s)",
+                self.initial_buffer_len,
+                self.warn_buffer_len,
+            )
 
 
 class SourceStoppedError(RuntimeError):
--- a/src/frequenz/sdk/timeseries/_resampling.py
+++ b/src/frequenz/sdk/timeseries/_resampling.py
@@ -71,7 +71,7 @@ Args:
 """
 
 
-ResamplingFunction = Callable[[Sequence[Sample], float], float]
+ResamplingFunction = Callable[[Sequence[Sample], "ResamplerConfig"], float]
 """Resampling function type.
 
 A resampling function produces a new sample based on a list of pre-existing
@@ -86,9 +86,9 @@ new sample that is going to be produced will always be bigger than the biggest
 timestamp in the input data).
 
 Args:
-    input_samples (Sequence[Sample]): the sequence of pre-existing samples.
-    resampling_period_s (float): the period in seconds (i.e. how ofter a new sample is
-        produced.
+    input_samples (Sequence[Sample]): The sequence of pre-existing samples.
+    resampler_config (ResamplerConfig): The configuration of the resampling
+        calling this function.
 
 Returns:
     new_sample (float): The value of new sample produced after the resampling.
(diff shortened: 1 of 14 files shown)


src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -334,6 +480,94 @@ def add_sample(self, sample: Sample) -> None:
...
+        props = self._source_properties
+
+        # We only update it if we didn't before and we have enough data
+        if (
+            props.sampling_period_s is not None
+            or props.sampling_start is None
+            or props.received_samples
+            < config.resampling_period_s * config.max_data_age_in_periods
+            or len(self._buffer) < self._buffer.maxlen
+        ):
```
daniel-zullo-frequenz, line 519: should this also check `props.received_samples` is equal to zero? or is it redundant?

daniel-zullo-frequenz: it is redundant as `max_data_age_in_periods` is ensured to be greater than 1 and `resampling_period_s` is checked for positive

src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -334,6 +480,94 @@ def add_sample(self, sample: Sample) -> None:
...
+            return False
+
+        samples_time_delta = now - props.sampling_start
```
daniel-zullo-frequenz, line 522: I wonder if `now > props.sampling_start` should be checked/asserted to avoid a zero or negative delta.

src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -388,6 +480,42 @@ def add_sample(self, sample: Sample) -> None:
             sample: The sample to be added to the buffer.
         """
         self._buffer.append(sample)
+        if self._source_properties.sampling_start is None:
+            self._source_properties.sampling_start = sample.timestamp
+        self._source_properties.received_samples += 1
```
ela-kotulska-frequenz, line 491: What about:
```
if props.sampling_period_s is None:
   self._source_properties.received_samples += 1
```

src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -388,6 +480,42 @@ def add_sample(self, sample: Sample) -> None:
...
+            or len(self._buffer) < self._buffer.maxlen
+        ):
+            return False
+
+        samples_time_delta = now - props.sampling_start
+        props.sampling_period_s = (
+            samples_time_delta.total_seconds()
+        ) / props.received_samples
+
+        return True
```
ela-kotulska-frequenz: Could you print info log with `props`? To see in logs what is computed `sampling_period_s`

src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -517,6 +517,58 @@ def _update_source_sample_period(self, now: datetime) -> bool:
...
+            )
+
+        new_buffer_len = max(1, math.ceil(new_buffer_len))
+        if new_buffer_len > config.max_buffer_len:
+            _logger.warning(
+                "The new buffer length (%s) for timeseries %s is too big, using %s instead",
+                new_buffer_len,
+                self._name,
+                config.max_buffer_len,
+            )
```
ela-kotulska-frequenz: Shouldn't it be an error?
And the exception should be raised.

src/frequenz/sdk/timeseries/_resampling.py
leandro-lucarella-frequenz: I was thinking that maybe we want to re-calculate the input sampling rate in the future again, maybe regularly at some defined interval, and maybe we want to take into account the whole history (probably not) or just the last T time. For now I think it makes sense to just keep adding, because the variable counts all the samples received so far, it might be confusing that at some point it stops counting, but in the future we might want to stop or probably even better reset it to adjust the input sampling rate from now and then to make sure we are not sticking to a wrong period forever (if we were unlucky enough to happen to measure it when there was some networking issue or something like that).

src/frequenz/sdk/timeseries/_resampling.py
leandro-lucarella-frequenz: Yeah, I could add a check in case there is a race between the first sample is received and the timer to do the resampling fires :+1:

src/frequenz/sdk/timeseries/_resampling.py
leandro-lucarella-frequenz: Yeah, that's a good idea.

src/frequenz/sdk/timeseries/_resampling.py
leandro-lucarella-frequenz: I don't think an exception should be raised because we don't get anything from that, there isn't much to do by the caller if this happens (except maybe rebuild the whole thing increasing the `max_buffer_len` or adjusting the `resampling_period_s`, but those are more deployment things IMHO), but I could turn it into an error. It is not really fatal, but it is true that it can degrade the quality of the resampling if it happens so maybe it is better to shout louder.

Does this look acceptable?

leandro-lucarella-frequenz: Oh, I see you found a bug, my intention was to truncate the buffer to `max_buffer_len`, which I now see it's not being done! I will fix that too, and should probably add a test too.

leandro-lucarella-frequenz:
Something else I forgot to mention in the PR is to check if the `DEFAULT_XXX_BUFFER_LEN` make any sense, I assumed the typical input sampling period is about 0.2s and the typical resampling period is 1s, so with a window of 3, a length of 16 should be enough (default initial length), so 8 times more than that should be suspicious (warning length) and 64 times that looks very excessive, like something is really not right (error length).

@ela-kotulska-frequenz @sahas-subramanian-frequenz @idlir-shkurti-frequenz @tiyash-basu-frequenz maybe you can double check these numbers.

src/frequenz/sdk/timeseries/_resampling.py
leandro-lucarella-frequenz: Fixed.

src/frequenz/sdk/timeseries/_resampling.py
leandro-lucarella-frequenz: Fixed.

src/frequenz/sdk/timeseries/_resampling.py
leandro-lucarella-frequenz: Fixed (without raising an exception for now)

leandro-lucarella-frequenz:
I added a new commit to also fix a log of wrongly created loggers, that were getting me crazy as there was no way to see/capture the logs in the tests.

ela-kotulska-frequenz (dismissed):

tiyash-basu-frequenz:
>I assumed the typical input sampling period is about 0.2s and the typical resampling period is 1s, so with a window of 3, a length of 16 should be enough (default initial length), so 8 times more than that should be suspicious (warning length) and 64 times that looks very excessive, like something is really not right (error length).

Assuming 16 is the length of the buffer holding the 0.2s res samples, if we want a resampled window of 3x1 second, then it sounds alright. Also, 8 times and 64 times as warning and error thresholds sounds decent to me.

leandro-lucarella-frequenz:
Rebased and added a hack to the tests to cope with some spurious errors in logs from unrelated tests. Also enabled auto-merge.

leandro-lucarella-frequenz:
Unrelated, but FYI, I noticed now the tests are way slower than before. Not all of them, so my guess is, since I think I say some wait time to gather battery status in recent PRs, that maybe tests are now subject to this extra waiting time.

If this is the case we really need to start using `time-machine` and `async-solipsism` in all tests that depends on time where time can be faked via these 2 libraries. You can look at the resampling tests for an example.

sahas-subramanian-frequenz (approved):

PR #139, opened 2022-12-28
Sorted Ringbuffer

### LINKED ISSUE #58: Implement an efficient ring buffer
### What's needed?

We need a [ring buffer](https://en.wikipedia.org/wiki/Circular_buffer) that is efficient, memory and time wise.

### Time requirements

Ring buffers should be able to:

* Push data in O(1)
* Pop data in O(1)
* Random access in O(1)

Sometimes we will need to keep these buffers sorted (if data can come out-of-order), a sorted ring buffer should be able to:

* Push or Pop data preserving order in O(log N)
* The other operation should still be O(1)
* Ideally it should be possible to select which operation is O(1), so users can pick the best implementation to be used when it is forseen to have more pushes than pops or vice-versa.

### Memory requirements

* Ideally the data in the buffer is not scanned by the GC (this might be hard if the data that is stored is not plain data, like `str`, `int`, `float`, `tuple`).

* It should be memory-safe even if objects (non-plain data) is stored in them.

* We'll need to be careful that the data stored in the buffer is plain data (for example store a `tuple` instead of a class, as classes are tracked by the GC).

* Ideally it should be possible to copy the buffers using a simple memory copy.

### Proposed solution

Do some benchmarks testing these underlying data structures to store ring buffers:

* `list`: This is an *array* in Python. The problem is, it is tracked by the GC, and for sorted buffers we need to copy data around to insert sorted, so it might not be suitable for that

* `deque`: This is a doubly linked list in Python, [implemented](https://github.com/python/cpython/blob/75a6fadf369315b27e12f670e6295cf2c2cf7d7e/Modules/_collectionsmodule.c#L78-L82) as a list of ([64](https://github.com/python/cpython/blob/75a6fadf369315b27e12f670e6295cf2c2cf7d7e/Modules/_collectionsmodule.c#L27)) blocks with free lists, so for small buffers it should be pretty similar to a `list` and for bigger buffers it should be worse to do random access but better to do random insertions.

* `numpy.array`: This is an array that escapes the GC, but we need to depend on `numpy` to use it. Also we need to check if it couldn't be error prone if we store non-plain data in them, since they are not scanned by the GC, if we store objects only there, they might be potentially freed by the GC, leading to memory issues.

### Use cases

A lot of parts in the system will use ring buffers, some of them might be pretty big, like actors storing some window of historical data.

_No response_

### Additional context

Since Python uses a generational GC, and we expect buffers to be long lived, the buffers should be promoted to older generations so they should be rarely scanned after a while. Also absolutely all other objects in the program will still be handled normally by the GC, so unless ring buffers are a considerable portion of the total memory consumed by programs, the impact of optimizing the ring buffer might not be relevant anyway.

So maybe escaping the GC is not that important. This is why we need **realistic** benchmarks before worrying too much about this.

Here is a nice introduction about how Python's GC works: https://rushter.com/blog/python-garbage-collector/

mathias-baumann-frequenz:
fixes #58

Signed-off-by: Marenz <mathias.baumann@frequenz.com>

### DIFF
--- a/src/frequenz/sdk/timeseries/_ringbuffer.py
+++ b/src/frequenz/sdk/timeseries/_ringbuffer.py
@@ -1,0 +1,430 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Ringbuffer implementation with focus on time & memory efficiency."""
+
+from __future__ import annotations
+
+from copy import deepcopy
+from dataclasses import dataclass
+from datetime import datetime, timedelta
+from typing import Any, Generic, List, Sequence, TypeVar
+
+import numpy as np
+
+T = TypeVar("T")
+
+Container = TypeVar("Container", list, np.ndarray)
+
+
+@dataclass
+class Gap:
+    """A gap defines the range for which we haven't received values yet."""
+
+    start: datetime
+    """Start of the range, inclusive."""
+    end: datetime
+    """End of the range, exclusive."""
+
+    def contains(self, timestamp: datetime):
+        """Check if a given timestamp is inside this  gap.
+
+        Args:
+            timestamp: Timestamp to check.
+
+        Returns:
+            True if the timestamp is in the gap.
+        """
+        if self.start <= timestamp < self.end:
+            return True
+
+        return False
+
+
+class OrderedRingBuffer(Generic[T, Container]):
+    """Time aware ringbuffer that keeps its entries sorted by time."""
+
+    def __init__(
+        self,
+        buffer: Container,
+        sampling_period: timedelta,
+        time_index_alignment: datetime = datetime(1, 1, 1),
+    ) -> None:
+        """Initialize the time aware ringbuffer.
+
+        Args:
+            buffer: Instance of a buffer container to use internally.
+            sampling_period: Timedelta of the desired resampling period.
+            time_index_alignment: Arbitary point in time used to align
+                timestamped data with the index position in the buffer.
+                Used to make the data stored in the buffer align with the
+                beginning and end of the buffer borders.
+
+                For example, if the `time_index_alignment` is set to
+                "0001-01-01 12:00:00", and the `sampling_period` is set to
+                1 hour and the length of the buffer is 24, then the data
+                stored in the buffer could correspond to the time range from
+                "2022-01-01 12:00:00" to "2022-01-02 12:00:00" (date chosen
+                arbitrarily here).
+        """
+        self._buffer = buffer
+        self._sampling_period = sampling_period
+        self._time_index_alignment = time_index_alignment
+
+        self._gaps: list[Gap] = []
+        self._datetime_newest = datetime.min
+        self._datetime_oldest = datetime.max
+        self._time_range = (len(self._buffer) - 1) * sampling_period
+
+    @property
+    def sampling_period(self) -> timedelta:
+        """Return the sampling period of the ring buffer.
+
+        Returns:
+            Sampling period of the ring buffer.
+        """
+        return self._sampling_period
+
+    @property
+    def gaps(self) -> List[Gap]:
+        """Get the list of ranges for which no values were provided.
+
+        See definition of dataclass @Gaps for more info.
+
+        Returns:
+            List of gaps.
+        """
+        return self._gaps
+
+    @property
+    def maxlen(self) -> int:
+        """Get the max length.
+
+        Returns:
+            The max amount of items this container can hold.
+        """
+        return len(self._buffer)
+
+    def update(self, timestamp: datetime, value: T, missing: bool = False) -> None:
+        """Update the buffer with a new value for the given timestamp.
+
+        Args:
+            timestamp: Timestamp of the new value.
+            value: value to add.
+            missing: if true, the given timestamp will be recorded as missing.
+                The value will still be written.
+
+                Note: When relying on your own values to mark missing data (e.g.
+                `math.nan`), be advised that when `update()` is called with
+                timestamps newer than the current time + `sampling_period`
+                (as is the case for loading historical data after an app
+                restart), a gap of missing data exists.
+                This gap does not contain NaN values but simply the old invalid values.
+                The list of gaps returned by `gaps()` will reflect this.
+
+        Raises:
+            IndexError: When the timestamp to be added is too old.
+        """
+        # adjust timestamp to be exactly on the sample period time point
+        timestamp = self._normalize_timestamp(timestamp)
+
+        # Don't add outdated entries
+        if timestamp < self._datetime_oldest and self._datetime_oldest != datetime.max:
+            raise IndexError(
+                f"Timestamp too old (cut-off is at {self._datetime_oldest})."
+            )
+
+        # Update timestamps
+        prev_newest = self._datetime_newest
+        self._datetime_newest = max(self._datetime_newest, timestamp)
+        self._datetime_oldest = self._datetime_newest - self._time_range
+
+        # Update data
+        self._buffer[self.datetime_to_index(timestamp)] = value
+
+        self._update_gaps(timestamp, prev_newest, missing)
+
+    def datetime_to_index(
+        self, timestamp: datetime, allow_outside_range: bool = False
+    ) -> int:
+        """Convert the given timestamp to an index.
+
+        Args:
+            timestamp: Timestamp to convert.
+            allow_outside_range: If True, don't throw an exception when the
+                timestamp is outside our bounds
+
+        Raises:
+            IndexError: When requesting a timestamp outside the range this container holds.
+
+        Returns:
+            Index where the value for the given timestamp can be found.
+        """
+        timestamp = self._normalize_timestamp(timestamp)
+
+        if not allow_outside_range and (
+            self._datetime_newest + self._sampling_period < timestamp
+            or timestamp < self._datetime_oldest
+        ):
+            raise IndexError(
+                f"Requested timestamp {timestamp} is "
+                f"outside the range [{self._datetime_oldest} - {self._datetime_newest}]"
+            )
+
+        return self._wrap(
+            round(
+                (timestamp - self._time_index_alignment).total_seconds()
+                / self._sampling_period.total_seconds()
+            )
+        )
+
+    def window(
+        self, start: datetime, end: datetime, force_copy: bool = False
+    ) -> Container:
+        """Request a view on the data between start timestamp and end timestamp.
+
+        If the data is not used immediately it could be overwritten.
+        Always request a copy if you keep the data around for longer.
+
+        Will return a copy in the following cases:
+        * The requested time period is crossing the start/end of the buffer.
+        * The requested time period contains missing entries.
+        * The force_copy parameter was set to True (default False).
+
+        The first case can be avoided by using the appropriate
+        time_index_alignment value in the constructor so that the data lines up
+        with the start/end of the buffer.
+
+        This means, if the caller needs to modify the data to account for
+        missing entries, they can safely do so.
+
+        Args:
+            start: start time of the window.
+            end: end time of the window.
+            force_copy: optional, default False. If True, will always create a
+                copy of the data.
+
+        Raises:
+            IndexError: when requesting a window with invalid timestamps.
+
+        Returns:
+            The requested window
+        """
+        if start > end:
+            raise IndexError(
+                f"end parameter {end} has to predate start parameter {start}"
+            )
+
+        start_index = self.datetime_to_index(start)
+        end_index = self.datetime_to_index(end)
+
+        # Requested window wraps around the ends
+        if start_index >= end_index:
+            window: Any = self._buffer[start_index:]
+
+            if end_index > 0:
+                if isinstance(self._buffer, list):
+                    window += self._buffer[0:end_index]
+                else:
+                    window = np.concatenate((window, self._buffer[0:end_index]))
+            return window
+
+        def in_gap(gap) -> bool:
+            return gap.contains(start) or gap.contains(end)
+
+        # Return a copy if there are none-values in the data
+        if force_copy or any(map(in_gap, self._gaps)):
+            return deepcopy(self._buffer[start_index:end_index])
+
+        return self._buffer[start_index:end_index]
+
+    def is_missing(self, timestamp: datetime) -> bool:
+        """Check if the given timestamp falls within a gap.
+
+        Args:
+            timestamp: The timestamp to check for missing data.
+
+        Returns:
+            True if the given timestamp falls within a gap, False otherwise.
+        """
+        return any(map(lambda gap: gap.contains(timestamp), self._gaps))
+
+    def _update_gaps(
+        self, timestamp: datetime, newest: datetime, new_missing: bool
+    ) -> None:
+        """Update gap list with new timestamp.
+
+        Args:
+            timestamp: Timestamp of the new value.
+            newest: Timestamp of the newest value before the current update.
+            new_missing: if true, the given timestamp will be recorded as missing.
+
+        """
+        currently_missing = self.is_missing(timestamp)
+
+        if not new_missing:
+            # Replace all gaps with one if we went far into then future
+            if self._datetime_newest - newest >= self._time_range:
+                self._gaps = [
+                    Gap(start=self._datetime_oldest, end=self._datetime_newest)
+                ]
+                return
+
+            if not currently_missing and timestamp > newest + self._sampling_period:
+                self._gaps.append(
+                    Gap(start=newest + self._sampling_period, end=timestamp)
+                )
+
+        # New missing entry that is not already in a gap?
+        if new_missing:
+            if not currently_missing:
+                self._gaps.append(
+                    Gap(start=timestamp, end=timestamp + self._sampling_period)
+                )
+        elif len(self._gaps) > 0:
+            if currently_missing:
+                self._remove_gap(timestamp)
+
+        self._cleanup_gaps()
+
+    def _cleanup_gaps(self) -> None:
+        """Clean up the list of gaps.
+
+        * Merge existing gaps
+        * remove overlaps
+        * delete outdated gaps
+        """
+        self._gaps = sorted(self._gaps, key=lambda x: x.start.timestamp())
+
+        i = 0
+        while i < len(self._gaps) - 1:
+            w_1 = self._gaps[i]
+            w_2 = self._gaps[i + 1]
+            # Delete out-of-date gaps
+            if w_1.end <= self._datetime_oldest:
+                del self._gaps[i]
+            # If w2 is a subset of w1 we can delete it
+            elif w_1.start <= w_2.start and w_1.end >= w_2.end:
+                del self._gaps[i + 1]
+            # If the gaps are direct neighbors, merge them
+            elif w_1.end >= w_2.start:
+                w_1.end = w_2.end
+                del self._gaps[i + 1]
+            else:
+                i += 1
+
+    def _remove_gap(self, timestamp: datetime) -> None:
+        """Update the list of gaps to not contain the given timestamp.
+
+        Args:
+            timestamp: Timestamp that is no longer missing.
+        """
+        gap_index, gap = next(
+            filter(
+                lambda gap: gap[1].contains(timestamp),
+                enumerate(self._gaps),
+            ),
+            (0, None),
+        )
+
+        if gap is None:
+            return
+
+        if gap.start == timestamp:
+            # Is the whole gap consisting only of the timestamp?
+            if gap.end == timestamp + self._sampling_period:
+                del self._gaps[gap_index]
+            # Otherwise, make the gap smaller
+            else:
+                gap.start = timestamp + self._sampling_period
+        # Is the timestamp at the end? Shrinken by one then
+        elif gap.end - self._sampling_period == timestamp:
+            gap.end = timestamp
+        # Otherwise it must be in the middle and we need to create a new
+        # gap
+        else:
+            new_gap = deepcopy(gap)
+            gap.end = timestamp
+            new_gap.start = timestamp + self._sampling_period
+            self._gaps.append(new_gap)
+
+    def _normalize_timestamp(self, timestamp: datetime) -> datetime:
+        """Normalize the given timestamp to fall exactly on the resampling period.
+
+        Args:
+            timestamp: The timestamp to normalize.
+
+        Returns:
+            The normalized timestamp.
+        """
+        num_samples, remainder = divmod(
+            (timestamp - self._time_index_alignment), self._sampling_period
+        )
+
+        # Round towards the closer number and towards the even one in case of
+        # equal distance
+        if remainder != timedelta(0) and (
+            self._sampling_period / 2 == remainder
+            and num_samples % 2 != 0
+            or self._sampling_period / 2 < remainder
+        ):
+            num_samples += 1
+
+        normalized_timestamp = (
+            self._time_index_alignment + num_samples * self._sampling_period
+        )
+
+        return normalized_timestamp
+
+    def _wrap(self, index: int) -> int:
+        """Normalize the given index to fit in the buffer by wrapping it around.
+
+        Args:
+            index: index to normalize.
+
+        Returns:
+            An index that will be within maxlen.
+        """
+        return index % self.maxlen
+
+    def __setitem__(self, index_or_slice: int | slice, value: T) -> None:
+        """Set item or slice at requested position.
+
+        No wrapping of the index will be done.
+        Create a feature request if you require this function.
+
+        Args:
+            index_or_slice: Index or slice specification of the requested data.
+            value: Value to set at the given position.
+        """
+        self._buffer.__setitem__(index_or_slice, value)
+
+    def __getitem__(self, index_or_slice: int | slice) -> T | Sequence[T]:
+        """Get item or slice at requested position.
+
+        No wrapping of the index will be done.
+        Create a feature request if you require this function.
+
+        Args:
+            index_or_slice: Index or slice specification of the requested data.
+
+        Returns:
+            The requested value or slice.
+        """
+        return self._buffer.__getitem__(index_or_slice)
+
+    def __len__(self) -> int:
+        """Return the amount of items that this container currently holds.
+
+        Returns:
+            The length.
+        """
+        if self._datetime_newest == datetime.min:
+            return 0
+
+        start_index = self.datetime_to_index(self._datetime_oldest)
+        end_index = self.datetime_to_index(self._datetime_newest)
+
+        if end_index < start_index:
+            return len(self._buffer) - start_index + end_index
+        return start_index - end_index
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,11 +6,14 @@
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with --> 
+<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
 
 ## New Features
 
 <!-- Here goes the main new features and examples or instructions on how to use them -->
+
+* A new class `OrderedRingBuffer` is now available, providing a sorted ring buffer of datetime-value pairs with tracking of any values that have not yet been written.
+
 
 ## Bug Fixes
 
--- a/benchmarks/timeseries/benchmark_ringbuffer.py
+++ b/benchmarks/timeseries/benchmark_ringbuffer.py
@@ -1,0 +1,170 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Performance test for the `Ringbuffer` class."""
+
+import random
+import timeit
+from datetime import datetime, timedelta
+from typing import Any, TypeVar
+
+import numpy as np
+
+from frequenz.sdk.timeseries._ringbuffer import OrderedRingBuffer
+
+MINUTES_IN_A_DAY = 24 * 60
+MINUTES_IN_29_DAYS = 29 * MINUTES_IN_A_DAY
+
+
+T = TypeVar("T")
+
+
+def fill_buffer(
+    days: int, buffer: OrderedRingBuffer[T, Any], element_type: type
+) -> None:
+    """Fill the given buffer up to the given amount of days, one sample per minute."""
+    random.seed(0)
+    basetime = datetime(2022, 1, 1)
+
+    for day in range(days):
+        # Push in random order
+        for i in random.sample(range(MINUTES_IN_A_DAY), MINUTES_IN_A_DAY):
+            buffer.update(
+                basetime + timedelta(days=day, minutes=i, seconds=i % 3),
+                element_type(i),
+            )
+
+
+def test_days(days: int, buffer: OrderedRingBuffer[int, Any]) -> None:
+    """Fills a buffer completely up and then gets the data for each of the 29 days."""
+    print(".", end="", flush=True)
+
+    fill_buffer(days, buffer, int)
+
+    basetime = datetime(2022, 1, 1)
+
+    for day in range(days):
+        # pylint: disable=unused-variable
+        minutes = buffer.window(
+            basetime + timedelta(days=day), basetime + timedelta(days=day + 1)
+        )
+
+
+def test_slices(days: int, buffer: OrderedRingBuffer[T, Any]) -> None:
+    """Benchmark slicing.
+
+    Takes a buffer, fills it up and then excessively gets
+    the data for each day to calculate the average/median.
+    """
+    print(".", end="", flush=True)
+    fill_buffer(days, buffer, float)
+
+    # Chose uneven starting point so that for the first/last window data has to
+    # be copied
+    basetime = datetime(2022, 1, 1, 0, 5, 13, 88)
+
+    total_avg = 0.0
+    total_median = 0.0
+
+    for _ in range(5):
+        for day in range(days):
+            minutes = buffer.window(
+                basetime + timedelta(days=day), basetime + timedelta(days=day + 1)
+            )
+
+            total_avg += float(np.average(minutes))
+            total_median += float(np.median(minutes))
+
+
+def test_29_days_list() -> None:
+    """Run the 29 day test on the list backend."""
+    test_days(29, OrderedRingBuffer([0] * MINUTES_IN_29_DAYS, timedelta(minutes=1)))
+
+
+def test_29_days_array() -> None:
+    """Run the 29 day test on the array backend."""
+    test_days(
+        29,
+        OrderedRingBuffer(
+            np.empty(
+                shape=MINUTES_IN_29_DAYS,
+            ),
+            timedelta(minutes=1),
+        ),
+    )
+
+
+def test_29_days_slicing_list() -> None:
+    """Run slicing tests on list backend."""
+    test_slices(29, OrderedRingBuffer([0] * MINUTES_IN_29_DAYS, timedelta(minutes=1)))
+
+
+def test_29_days_slicing_array() -> None:
+    """Run slicing tests on array backend."""
+    test_slices(
+        29,
+        OrderedRingBuffer(
+            np.empty(
+                shape=MINUTES_IN_29_DAYS,
+            ),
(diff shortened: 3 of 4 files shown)


src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,221 @@
...
+
+Container = Union[list, np.ndarray]
+
+
+class OrderedRingBuffer(Generic[T]):
+    """Time aware ringbuffer that keeps its entries sorted time."""
+
+    def __init__(
+        self,
+        buffer: Any,
```
shsms: why not just say `Container` here, instead of `Any`?

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,221 @@
...
+        sample_rate: int,
+        window_border: datetime = datetime(1, 1, 1),
+    ) -> None:
+        """Initialize the time aware ringbuffer.
+
+        Args:
+            buffer: instance of a buffer container to use internally
+            sample_rate: resolution of the incoming timestamps in
+                seconds
```
matthias-wende-frequenz: in Hz ;)

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,221 @@
...
+
+            Returns:
+                Nothing.
+        """
+        # Update timestamps
+        self._datetime_newest = max(self._datetime_newest, timestamp)
+        self._datetime_oldest = min(self._datetime_oldest, timestamp)
+
+        if self._datetime_oldest < self._datetime_newest - timedelta(
+            seconds=len(self._buffer) * self._sample_rate
```
matthias-wende-frequenz: Sample rate is given by number of samples per second (as mentioned above in Hz). Thus this should be `len(self._buffer) / self._sample_rate`.

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,221 @@
...
+        ):
+            self._datetime_oldest = self._datetime_newest - timedelta(
+                len(self._buffer) * self._sample_rate
```
matthias-wende-frequenz: see comment above :)

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,221 @@
...
+                len(self._missing_windows) == 0
+                or self._missing_windows[-1].end is not None
+            ):
+                self._missing_windows.append({"start": timestamp, "end": None})
+        elif len(self._missing_windows) > 0:
+            # Finalize a pending window
+            if self._missing_windows[-1].end is None:
+                self._missing_windows[-1].end = timestamp
+
+        # Delete out-to-date windows
```
matthias-wende-frequenz: I suppose you mean out-of-date ;)

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,221 @@
...
+        Returns:
+            index where the value for the given timestamp can be found.
+        """
+        if self._datetime_newest < timestamp or timestamp < self._datetime_oldest:
+            raise IndexError(
+                f"Requested timestamp {timestamp} is is "
+                f"outside the range [{self._datetime_oldest} - {self._datetime_newest}]"
+            )
+
+        return self._wrap(int(abs((self._window_start - timestamp).total_seconds())))
```
matthias-wende-frequenz: ~Shouldn't this be divided by the sample rate?~
IIUC this formula only works for sample rate equal to 1Hz. With higher sampling rates we're having more then 1 element per second.

tests/utils/ringbuffer.py
```diff
@@ -0,0 +1,83 @@
...
+    [
+        OrderedRingBuffer[int]([0] * 24, 1),
+        OrderedRingBuffer[float](np.empty(shape=(24,), dtype=np.float64), 1),
+    ],
+)
+def test_timestamp_ringbuffer(buffer: OrderedRingBuffer[float | int]) -> None:
+    """Test ordered ring buffer."""
+    size = buffer.maxlen
+
+    #    import pdb; pdb.set_trace()
```
matthias-wende-frequenz: I guess this can be removed.

matthias-wende-frequenz (changes requested):
First part of the review is done. I suggest to add test for different sample rates.

tests/utils/ringbuffer.py
```diff
@@ -0,0 +1,83 @@
...
+        buffer.update(datetime.fromtimestamp(200 + i), i * 2)
+
+    # Check all possible window sizes and start positions
+    for i in range(size):
+        for j in range(1, size):
+            start = datetime.fromtimestamp(200 + i)
+            end = datetime.fromtimestamp(200 + j + i)
+
+            tmp = islice(cycle(range(0, size * 2, 2)), i, i + j)
+            # assert list(buffer.window(start, end)) == list(tmp)
```
matthias-wende-frequenz: I suppose that can be removed.

src/frequenz/sdk/util/ringbuffer.py
mathias-baumann-frequenz: Indeed, my test was all set to 1 and that's why I didn't catch it. Nice catch!

src/frequenz/sdk/util/ringbuffer.py
mathias-baumann-frequenz: I find this very unintuitive tbh, resolution sounds way more understandable

matthias-wende-frequenz: Why? It's just the reciprocal.

mathias-baumann-frequenz: unlike my brain );

ela-kotulska-frequenz:
Could you please:
* Assign yourself to this PR
* Assign Issue (Development tab)
* Assign milestone ( v0.17)
* Assign more labes - as this is not only tests

Please do that for all PRs

benchmarks/util/benchmark_ringbuffer.py
```diff
@@ -0,0 +1,168 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
```
ela-kotulska-frequenz: 2023 :)

mathias-baumann-frequenz:
I don't know what Development Tab you mean, but I am already assigned to the according issue https://github.com/frequenz-floss/frequenz-sdk-python/issues/58

Added the according issue number though

ela-kotulska-frequenz:
> don't know what Development Tab you mean, but I am already assigned to the according issue https://github.com/frequenz-floss/frequenz-sdk-python/issues/58

I meant that issues should be assign in tab named `Development` by right side. I see you already did it. Thanks!

mathias-baumann-frequenz:
> I meant that issues should be assign in tab named Development by right side. I see you already did it. Thanks!

Ah, now I see it. Never notice that before.

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+class OrderedRingBuffer(Generic[T]):
+    """Time aware ringbuffer that keeps its entries sorted by time."""
+
+    def __init__(
+        self,
+        buffer: Any,
+        sample_rate: float,
+        window_border: datetime = datetime(1, 1, 1),
```
ela-kotulska-frequenz: Why this default argument? Could you describe more what is window_border?

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,259 @@
...
+
+Container = Union[list, np.ndarray]
+
+
+class OrderedRingBuffer(Generic[T]):
+    """Time aware ringbuffer that keeps its entries sorted by time."""
+
+    def __init__(
+        self,
+        buffer: Any,
```
ela-kotulska-frequenz: Why `Any`.
If buffer is type `int` and I run self.maxlen() method, then it will raise an exception.

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,259 @@
...
+
+        return True
+
+    def datetime_to_index(
+        self, timestamp: datetime, allow_outside_range: bool = False
+    ) -> int:
+        """Convert the given timestamp to an index.
+
+        Throws an index error when the timestamp is not found within this
+        buffer.
```
ela-kotulska-frequenz: This is in `Raises` section already.

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,259 @@
...
+        self._datetime_oldest = datetime.max
+
+    @property
+    def missing_windows(self) -> list:
+        """Get the list of missing windows.
+
+        Returns:
+            list of missing windows.
+        """
+        return self._missing_windows
```
ela-kotulska-frequenz: Could you describe what is `missing_window`

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,259 @@
...
+        # Update timestamps
+        prev_newest = self._datetime_newest
+        self._datetime_newest = max(self._datetime_newest, timestamp)
+        self._datetime_oldest = self._datetime_newest - timedelta(
+            seconds=(len(self._buffer) - 1) / self._sample_rate
+        )
+
+        assert self.datetime_to_index(self._datetime_newest) != self.datetime_to_index(
+            self._datetime_oldest
+        )
```
ela-kotulska-frequenz: Why is it needed?

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,259 @@
...
+    @property
+    def maxlen(self) -> int:
+        """Get the max length.
+
+        Returns:
+            the max amount of items this container can hold.
+        """
+        return len(self._buffer)
+
+    def update(self, timestamp: datetime, value: T, missing: bool = False) -> bool:
```
ela-kotulska-frequenz: What does it mean that `value` is missing?  I am asking because I couldn't  find any test with `missing = True` .
Should it be None? Or it should has value?

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,259 @@
...
+        buffer: Any,
+        sample_rate: float,
+        window_border: datetime = datetime(1, 1, 1),
+    ) -> None:
+        """Initialize the time aware ringbuffer.
+
+        Args:
+            buffer: instance of a buffer container to use internally
+            sample_rate: resolution of the incoming timestamps in
+                hz (samples per second)
```
shsms: maybe I missed something,  but didn't we settle on sampling interval (or period)?

src/frequenz/sdk/util/ringbuffer.py
mathias-baumann-frequenz: I def. should add a test for `missing = True`, totally missed that.

Value can be anything you want, it will still be written but it's marked as "missing" in our `missing_windows()`.

The idea was to allow max. flexibility. Some users want to write Nan values, some want to use the avg of the previous one and for some the datatype might not even allow an easy "invalid" value (e.g. int values).

With this method it's independent of the value written.

src/frequenz/sdk/util/ringbuffer.py
mathias-baumann-frequenz: Well, it can't be `int` as it should be a container (as the docu says), more specifically list or nd array.

I tried using a more explicit type hint (e.g. `:list | np.ndarray`), but it proves rather difficult to do if you want to support multiple storage containers.

At the end i didn't find a solution that would work so far...

src/frequenz/sdk/util/ringbuffer.py
mathias-baumann-frequenz: missing windows are the gaps where we have no data points

src/frequenz/sdk/util/ringbuffer.py
mathias-baumann-frequenz: It's not "needed", it's just to make sure our timestamps are what we expect them to be.

src/frequenz/sdk/util/ringbuffer.py
mathias-baumann-frequenz: @matthias-wende-frequenz argued hard for this, argument was that it is used in the rest of the sdk... I personally would rather have the interval myself...

src/frequenz/sdk/util/ringbuffer.py
mathias-baumann-frequenz: I did try that and it lead to many problems, but trying it again now does not. :shrug:

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,259 @@
...
+        """Update the buffer with a new value for the given timestamp.
+
+        Args:
+            timestamp: Timestamp of the new value
+            value: value to add
+            missing: if true, the given timestamp will be recorded as missing.
+                The value will still be written.
+
+        Returns:
+            Nothing.
```
matthias-wende-frequenz: Returns a bool now ;)

src/frequenz/sdk/util/ringbuffer.py
ela-kotulska-frequenz: Did you try `typing.Union`?

src/frequenz/sdk/util/ringbuffer.py
matthias-wende-frequenz: I argued for it until we had a discussion and everyone else favored `resampling_period_s` as used in the resampler.

src/frequenz/sdk/util/ringbuffer.py
matthias-wende-frequenz: I didn't stumble upon that because I knew the context already before reading it, but I'd suggest to add it to the documentation somewhere.

src/frequenz/sdk/util/ringbuffer.py
matthias-wende-frequenz: but how would you specify what should be used?

src/frequenz/sdk/util/ringbuffer.py
matthias-wende-frequenz: AFAIU it's the fixed timestamp that's stored, to calculate the indexes of the incoming samples from. The problem with that approach is, that it's never updated and hence, given the ring buffer has fixed size, we are not able to keep it updated with latest incoming samples.
A simple "fix" would be to update it when wrapping around.

matthias-wende-frequenz: I suggest to remove the default argument. This argument is essential for this ringbuffer implementation and setting an arbitrary default doesn't make any sense.

mathias-baumann-frequenz: I don't think it's essential, rather a minor optimization. With this we basically control where in the buffer data will be written and it is (meant to be) used to make the data in the buffer start exactly on the border of a typical window size (e.g. one day for a 29 day buffer with minute resolution). 

Everything would still work if if this is an arbitrary value, but in some cases a tiny bit more copying might be happening (due to days overlapping in the end area)

src/frequenz/sdk/util/ringbuffer.py
mathias-baumann-frequenz: It is working now, not sure why it didn't earlier. Will updated when the other changes are done

src/frequenz/sdk/util/ringbuffer.py
mathias-baumann-frequenz: What do you mean? I am not sure what your "what" refers to and what that "what" would be used for?

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,276 @@
...
+    @property
+    def maxlen(self) -> int:
+        """Get the max length.
+
+        Returns:
+            the max amount of items this container can hold.
+        """
+        return len(self._buffer)
+
+    def update(self, timestamp: datetime, value: T, missing: bool = False) -> bool:
```
matthias-wende-frequenz: We could possibly use the Sample type defined [here](https://github.com/frequenz-floss/frequenz-sdk-python/blob/v0.x.x/src/frequenz/sdk/timeseries/_base_types.py)

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,399 @@
...
+
+        # Update data
+        self._buffer[self.datetime_to_index(timestamp)] = value
+
+        self._update_missing_windows(timestamp, prev_newest, missing)
+
+        return True
+
+    def _update_missing_windows_add_gap(
+        self, timestamp: datetime, newest: datetime
```
matthias-wende-frequenz: `newest` is not accessed

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,412 @@
...
+        self._time_index_alignment = time_index_alignment
+
+        self._missing_windows = []
+        self._datetime_newest = datetime.min
+        self._datetime_oldest = datetime.max
+        self._time_range = timedelta(
+            seconds=(len(self._buffer) - 1) * self._sampling_period_s
+        )
+
+    def normalize_timestamp(self, timestamp: datetime) -> datetime:
```
shsms: is this for only for internal use?  Maybe needs a `_` prefix?

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,412 @@
...
+        def in_window(window) -> bool:
+            return self._in_window(window, start) or self._in_window(window, end)
+
+        # Return a copy if there are none-values in the data
+        if force_copy or any(map(in_window, self._missing_windows)):
+            return deepcopy(self._buffer[start_index:end_index])
+
+        return self._buffer[start_index:end_index]
+
+    def is_missing(self, timestamp: datetime) -> bool:
```
shsms: is this part of the public api?

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,412 @@
...
+        self._datetime_oldest = self._datetime_newest - self._time_range
+
+        # Update data
+        self._buffer[self.datetime_to_index(timestamp)] = value
+
+        self._update_missing_windows(timestamp, prev_newest, missing)
+
+        return True
+
+    def datetime_to_index(
```
shsms: is this part of the public api?

shsms (commented):
I just tried generating docs for this and noticed there were a bunch of extra methods which might not have to be part of the public api.  Their names can be prefixed with a `_` to hide them.

I think this should also be moved to the `timeseries` package.

`normalize_timestamp` is getting called with every update,  so I would like to see if we can get by without it,  but I'm ok with keeping it for now and making an issue to into it later.

I would propose that we remove the generic `T` and specializing this for just `float` values as well,  but with this too,  there is no urgency,  and we can wait and see how we end up using it, and come back to it later.

src/frequenz/sdk/util/ringbuffer.py
matthias-wende-frequenz: I'm using it :)

src/frequenz/sdk/util/ringbuffer.py
matthias-wende-frequenz: IMO that can be private.

src/frequenz/sdk/util/ringbuffer.py
matthias-wende-frequenz: can be private

matthias-wende-frequenz (dismissed):
Apart from Sahas's that looks fine too me. I give an approve and we can address further things in follow up PRs.

src/frequenz/sdk/util/ringbuffer.py
mathias-baumann-frequenz: Hmmm, I thought someone might want to use it.. but sure, I can make it private..

src/frequenz/sdk/util/ringbuffer.py
mathias-baumann-frequenz: I thought it is rather useful, but sure, I can make it private

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,412 @@
...
+
+
+class OrderedRingBuffer(Generic[T, Container]):
+    """Time aware ringbuffer that keeps its entries sorted by time."""
+
+    def __init__(
+        self,
+        buffer: Container,
+        sampling_period_s: float,
+        time_index_alignment: datetime = datetime(1, 1, 1),
```
ela-kotulska-frequenz: When user has to specify that argument? 
`datetime(1, 1, 1)` seems to be always correct.

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,412 @@
...
+            seconds=round(
+                (timestamp - self._time_index_alignment).total_seconds()
+                / self._sampling_period_s
+            )
+            * self._sampling_period_s
+        )
+        return normalized_timestamp
+
+    @property
+    def missing_windows(self) -> list:
```
ela-kotulska-frequenz: Return type is missing
```
def missing_windows(self) -> List[Dict[str, datetime]]
```

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,412 @@
...
+        Raises:
+            IndexError: when requesting a timestamp outside the range this container holds
+
+        Returns:
+            index where the value for the given timestamp can be found.
+        """
+        timestamp = self.normalize_timestamp(timestamp)
+
+        if not allow_outside_range and (
+            self._datetime_newest + timedelta(seconds=1 * self._sampling_period_s)
```
ela-kotulska-frequenz: `timedelta(seconds=1 * self._sampling_period_s)` -> `timedelta(seconds=self._sampling_period_s)`

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,412 @@
...
+            < timestamp
+            or timestamp < self._datetime_oldest
+        ):
+            raise IndexError(
+                f"Requested timestamp {timestamp} is is "
```
ela-kotulska-frequenz: `Requested timestamp {timestamp} is is ` - > `Requested timestamp {timestamp} is `

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,412 @@
...
+            True if the entry was added. False if it was discarded for being too old.
+        """
+        # adjust timestamp to be exactly on the sample period time point
+        timestamp = self.normalize_timestamp(timestamp)
+
+        # Don't add outdated entries
+        if timestamp < self._datetime_oldest and self._datetime_oldest != datetime.max:
+            return False
+
+        # print(f"Normalizing {timestamp} to {self.normalize_timestamp(timestamp)}")
```
ela-kotulska-frequenz: remove ;)

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,412 @@
...
+            if not currently_missing and timestamp > newest + one_step:
+                self._missing_windows.append(
+                    {"start": newest + one_step, "end": timestamp}
+                )
+
+        # New missing entry that is not already in a window?
+        if missing:
+            if not currently_missing:
+                self._missing_windows.append(
+                    {"start": timestamp, "end": timestamp + one_step}
```
ela-kotulska-frequenz: Consider creating data class from it:

```python
@dataclass
class MissingWindow:
    start: timestamp
    end: timestamp
```

src/frequenz/sdk/util/ringbuffer.py
```diff
@@ -0,0 +1,412 @@
...
+        Args:
+            start: start time of the window.
+            end: end time of the window.
+            force_copy: optional, default False. If True, will always create a
+                copy of the data.
+
+        Returns:
+            the requested window
+        """
+        assert start < end
```
ela-kotulska-frequenz: Please don't use assert in source code

benchmarks/util/benchmark_ringbuffer.py
```diff
@@ -0,0 +1,170 @@
...
```
ela-kotulska-frequenz: 2023

ela-kotulska-frequenz:
Please update RELEASE_NOTES.md
https://github.com/frequenz-floss/frequenz-sdk-python/blob/v0.x.x/RELEASE_NOTES.md

src/frequenz/sdk/util/ringbuffer.py
mathias-baumann-frequenz: We had this discussion before, should we maybe have a call about this?

mathias-baumann-frequenz: Pasting the documentation here, though

```
            time_index_alignment: Arbitary point in time used to align
                timestamped data with the index position in the buffer.
                Used to make the data stored in the buffer align with the
                beginning and end of the buffer borders.

                For example, if the `time_index_alignment` is set to
                "0001-01-01 12:00:00", and the `sampling_period` is set to
                1 hour and the length of the buffer is 24, then the data
                stored in the buffer could correspond to the time range from
                "2022-01-01 12:00:00" to "2022-01-02 12:00:00" (date chosen
                arbitrarily here).
```

Does this explanation answer the question?

src/frequenz/sdk/util/ringbuffer.py
mathias-baumann-frequenz: This should be public, it can be used to check if a specific sample datetime is already written or not.

mathias-baumann-frequenz:
I believe this is in a good mergable state now. It may deserve another review though as quite a few things have changed again:

* The sample_period was changed from seconds(int) to timedelta, that way the user can use any time-unit they want to specify it
* We now support sample rates down to 1 microsecond, though I only tested with 200ms smallest so far.
* missing_windows was renamed to Gaps and it got its own explicit dataclass `Gap` (thangs for the suggestion @ela-kotulska-frequenz)
* Bug fixes, e.g. previously it would fail if the time_index_alignment was in the future (were as it shouldn't matter when)

I think I integrated all the feedback I got, except some of the still open comments here that I want to save for the next PR (the Sample datatype suggestion for example)

src/frequenz/sdk/timeseries/_ringbuffer.py
```diff
@@ -0,0 +1,431 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Ringbuffer implementation with focus on time & memory efficiency."""
+
+from __future__ import annotations
+
+from copy import deepcopy
+from datetime import datetime, timedelta
+from typing import Any, Dict, Generic, List, Sequence, TypeVar
```
matthias-wende-frequenz: Dict seems to be unused

src/frequenz/sdk/timeseries/_ringbuffer.py
```diff
@@ -0,0 +1,431 @@
...
+
+        Args:
+            timestamp: The timestamp to check for missing data.
+
+        Returns:
+            True if the given timestamp falls within a missing window, False otherwise.
+        """
+        return any(map(lambda window: self._in_window(window, timestamp), self._gaps))
+
+    def _update_missing_windows(
```
matthias-wende-frequenz: rename to `_update_gaps` ?

src/frequenz/sdk/timeseries/_ringbuffer.py
```diff
@@ -0,0 +1,431 @@
...
+            return self._in_window(window, start) or self._in_window(window, end)
+
+        # Return a copy if there are none-values in the data
+        if force_copy or any(map(in_window, self._gaps)):
+            return deepcopy(self._buffer[start_index:end_index])
+
+        return self._buffer[start_index:end_index]
+
+    def is_missing(self, timestamp: datetime) -> bool:
+        """Check if the given timestamp falls within a missing window.
```
matthias-wende-frequenz: ... falls within a gap

src/frequenz/sdk/timeseries/_ringbuffer.py
```diff
@@ -0,0 +1,431 @@
...
+
+        Returns:
+            True if the given timestamp falls within a missing window, False otherwise.
+        """
+        return any(map(lambda window: self._in_window(window, timestamp), self._gaps))
+
+    def _update_missing_windows(
+        self, timestamp: datetime, newest: datetime, new_missing: bool
+    ) -> None:
+        """Update missing windows list with new timestamp.
```
matthias-wende-frequenz: Update gaps...

src/frequenz/sdk/timeseries/_ringbuffer.py
```diff
@@ -0,0 +1,431 @@
...
+                self._gaps.append(
+                    Gap(start=timestamp, end=timestamp + self._sampling_period)
+                )
+        elif len(self._gaps) > 0:
+            if currently_missing:
+                self._update_missing_windows_remove_gap(timestamp)
+
+        self._cleanup_missing_windows()
+
+    def _cleanup_missing_windows(self) -> None:
```
matthias-wende-frequenz: `_cleanup_gaps`

src/frequenz/sdk/timeseries/_ringbuffer.py
```diff
@@ -0,0 +1,431 @@
...
+            elif w_1.start <= w_2.start and w_1.end >= w_2.end:
+                del self._gaps[i + 1]
+            # If the windows are direct neighbors, merge them
+            elif w_1.end >= w_2.start:
+                w_1.end = w_2.end
+                del self._gaps[i + 1]
+            else:
+                i += 1
+
+    def _update_missing_windows_remove_gap(self, timestamp: datetime) -> None:
```
matthias-wende-frequenz: gaps :)

src/frequenz/sdk/timeseries/_ringbuffer.py
```diff
@@ -0,0 +1,431 @@
...
+    @property
+    def maxlen(self) -> int:
+        """Get the max length.
+
+        Returns:
+            The max amount of items this container can hold.
+        """
+        return len(self._buffer)
+
+    def update(self, timestamp: datetime, value: T, missing: bool = False) -> bool:
```
matthias-wende-frequenz: This function raises in `TypeError` inherited from the `datetime_to_index` method.

matthias-wende-frequenz (commented):
some mainly cosmetic changes

src/frequenz/sdk/timeseries/_ringbuffer.py
mathias-baumann-frequenz: Any way to reproduce that?

src/frequenz/sdk/timeseries/_ringbuffer.py
```diff
@@ -0,0 +1,430 @@
...
+"""Ringbuffer implementation with focus on time & memory efficiency."""
+
+from __future__ import annotations
+
+from copy import deepcopy
+from datetime import datetime, timedelta
+from typing import Any, Generic, List, Sequence, TypeVar
+
+import numpy as np
+from attr import dataclass
```
matthias-wende-frequenz: I'd suggest to use `dataclass` from dataclasses, i.e.
`from dataclasses import dataclass`

src/frequenz/sdk/timeseries/_ringbuffer.py
matthias-wende-frequenz: It raises when the other method does. I meant it should be documented :)

mathias-baumann-frequenz: Actually no, it's correct like this. It can't raise that `TypeError` because the conditions checked in `datetime_to_index` are modified in update so that it would never throw

src/frequenz/sdk/timeseries/_ringbuffer.py
mathias-baumann-frequenz: Ah. The auto-import did that for me, I didn't question it :sweat_smile:

matthias-wende-frequenz (approved):

src/frequenz/sdk/timeseries/_ringbuffer.py
```diff
@@ -0,0 +1,430 @@
...
+@dataclass
+class Gap:
+    """A gap defines the range for which we haven't received values yet."""
+
+    start: datetime
+    """Start of the range, inclusive."""
+    end: datetime
+    """End of the range, exclusive."""
+
+    def contains(self, timestamp: datetime):
```
leandro-lucarella-frequenz, line 29: FYI you could also implement the `__contains__()` method to allow writing: `if date in gap: do_whatever()`.

src/frequenz/sdk/timeseries/_ringbuffer.py
```diff
@@ -0,0 +1,430 @@
...
+        prev_newest = self._datetime_newest
+        self._datetime_newest = max(self._datetime_newest, timestamp)
+        self._datetime_oldest = self._datetime_newest - self._time_range
+
+        # Update data
+        self._buffer[self.datetime_to_index(timestamp)] = value
+
+        self._update_gaps(timestamp, prev_newest, missing)
+
+    def datetime_to_index(
```
leandro-lucarella-frequenz, line 147: In python this is also commonly known as simply `index()` (for example `list` implements it).

src/frequenz/sdk/timeseries/_ringbuffer.py
```diff
@@ -0,0 +1,430 @@
...
+                arbitrarily here).
+        """
+        self._buffer = buffer
+        self._sampling_period = sampling_period
+        self._time_index_alignment = time_index_alignment
+
+        self._gaps: list[Gap] = []
+        self._datetime_newest = datetime.min
+        self._datetime_oldest = datetime.max
+        self._time_range = (len(self._buffer) - 1) * sampling_period
```
leandro-lucarella-frequenz, line 77: Almost all of these are missing type hints. I'm surprised it wasn't caught by the linters.

src/frequenz/sdk/timeseries/_ringbuffer.py
```diff
@@ -0,0 +1,430 @@
...
+                else:
+                    window = np.concatenate((window, self._buffer[0:end_index]))
+            return window
+
+        def in_gap(gap) -> bool:
+            return gap.contains(start) or gap.contains(end)
+
+        # Return a copy if there are none-values in the data
+        if force_copy or any(map(in_gap, self._gaps)):
+            return deepcopy(self._buffer[start_index:end_index])
```
leandro-lucarella-frequenz, line 237: I'm not sure `deepcopy()` is necessary nor desired. You can do a shallow copy by using `copy()` instead. Why was a deep copy preferred?

src/frequenz/sdk/timeseries/_ringbuffer.py
mathias-baumann-frequenz: Linters seem to care mostly about things outside of functions (parameters and returns).
Is it our policy to type-hint every single assignment then?

src/frequenz/sdk/timeseries/_ringbuffer.py
mathias-baumann-frequenz: Because we don't want a shallow copy because the user is expected to modify the data, e.g. to fill the gaps with float.nan or the avg of the previous values or whatever they desire

src/frequenz/sdk/timeseries/_ringbuffer.py
mathias-baumann-frequenz: That seems like a good idea

src/frequenz/sdk/timeseries/_ringbuffer.py
mathias-baumann-frequenz: Do you think we should rename it? :)

src/frequenz/sdk/timeseries/_ringbuffer.py
shsms: I'm pretty sure `mypy` can deduce the types for the fields from the type of the assigned values.  The only place where it can't deduce is for `self._gaps: list[Gap] = []`, because it is an empty list,  and in that case, a type has been provided.

PR #141, opened 2022-12-29
Fix formula precedence bug in the `FormulaEngine`

shsms:
Formulas with repeated operators like `#1 - #2 - #3` were getting calculated incorrectly as `#1 - (#2 - #3)`.

The correct post-fix notation would be `[#1, #2, -, #3, -]`,  but the formula engine was generating this instead: `[#1, #2, #3, -, -]`.

This is fixed by flushing the operators from the build stack not only if a previous operator has a higher precedence, but also if a previous operator has the same precedence as the current operator.


daniel-zullo-frequenz:
Code-wise LGTM. There is just a cosmetic in the commit message, a comma is missing in  `The correct post-fix notation would be `[frequenz-floss#1, frequenz-floss#2, -, frequenz-floss#3 -]...` that you might want to fix or not.

daniel-zullo-frequenz (dismissed):

matthias-wende-frequenz (dismissed):

matthias-wende-frequenz:
The commit message looks weird because github replaces hashes with links to issues/prs.

sahas-subramanian-frequenz:
> Code-wise LGTM. There is just a cosmetic in the commit message

Thanks Dani, fixed.

daniel-zullo-frequenz (approved):

PR #145, opened 2023-01-02
Define way to create object with async constructor

### LINKED ISSUE #135: Create object with async constructor
### What's needed?

Some classes needs to call await in the constructing step.
Problem:
* Python doesn't allow for it because __init__ can't be async.
* Python doesn't provide any  async_init method. 

### Proposed solution

@sahas-subramanian-frequenz 

```py

class BatteryPool:
    """Use new() to create."""
    _CreatedThroughNew = object()

    def __init__(..., created_by: Any = None):
        """Not for direct use.  Use new() to create."""
        if created_by is not _CreatedThroughNew:
             raise RuntimeError(...)
        ... rest of init

    async def new(...) -> BatteryPool:
        await etc
        return BatteryPool(...., cls._CreatedThroughNew)
```

@leandro-lucarella-frequenz 

```py
class NotSyncConstructible(AssertError):
    pass


class AsyncConstructible(ABC):
    def __init__(self) -> None:
        raise NotSyncConstructible()

    @abstractmethod
    @classmethod
    async def async_new(cls) -> Self:  # or AsyncConstructible if it still doesn't work
        return cls.__new__(cls)


class BatteryPool(AsyncConstructible):
    def __init__(self) -> None:        
        super().__init__()

        self.attr1: str
        """Documentation for attr1."""
        self.attr2: int
        """Documentation for attr2."""

    @classmethod
    def async_new(cls, ...) -> Self:  # or BatteryPool if it still doesn't work
        self = await super().async_new()
        self.attr1 = "lala"
        self.attr2 = 2
        return self
```

### Use cases

_No response_

_No response_

### Additional context

It comes from this discussion:

* https://github.com/frequenz-floss/frequenz-sdk-python/pull/117#discussion_r1050752095

ela-kotulska-frequenz:


src/frequenz/sdk/actor/power_distributing/_battery_pool_status.py
```diff
@@ -4,28 +4,35 @@
...
         battery_ids: Set[int],
         max_data_age_sec: float,
         max_blocking_duration_sec: float,
-    ) -> None:
-        """Create partially initialized object instance.
-
-        Note:
-            Please call `async_init` method to fully initialize BatteryPoolStatus. Otherwise
-            it is not possible to use BatteryPoolStatus.
+    ) -> "BatteryPoolStatus":
```
shsms: can you use `from __future__ import annotations` instead of the quotes?  I guess there's no need to write python 3.8 friendly code anymore.

src/frequenz/sdk/actor/power_distributing/_battery_pool_status.py
```diff
@@ -35,20 +42,21 @@ def __init__(
                 data.
             max_blocking_duration_sec: This value tell what should be the maximum
                 timeout used for blocking failing component.
+
+        Returns:
+            New instance of this class.
         """
+        self: "BatteryPoolStatus" = BatteryPoolStatus.__new__(cls)
```
shsms: here too.

src/frequenz/sdk/actor/power_distributing/power_distributing.py
```diff
@@ -164,15 +164,9 @@ def __init__(
...
-        )
-
-        self._battery_pool = BatteryPoolStatus(
-            battery_ids={battery.component_id for battery in batteries},
-            max_blocking_duration_sec=30.0,
-            max_data_age_sec=10.0,
-        )
+        # This will be temporary initialized in the `run` method.
+        # PowerDistributingActor won't use BatteryPoolStatus soon.
+        self._battery_pool: BatteryPoolStatus
```
shsms: I see just a single use of this variable outside the `run` method.  I think it might be cleaner to make it a local variable inside the `run` function,  and pass it as argument from there.

src/frequenz/sdk/util/asyncio.py
```diff
@@ -19,3 +20,22 @@ async def cancel_and_await(task: asyncio.Task) -> None:
...
+    def __init__(self) -> None:
+        """Raise error when object is created in sync way.
+
+        Raises:
+            NotSyncConstructible: If this method is called.
+        """
+        raise NotSyncConstructible(
+            "This object shouldn't be created with default constructor. ",
+            "Check class documentation for more information.",
+        )
```
shsms: maybe also add an abstract method, to standardize all uses?

``` python
@abstractmethod
async def async_new() -> Self:
  """Create an `AsyncConstructible` object."""
```

ela-kotulska-frequenz: I could do that, but the problem is that then number of constructor paramenters in async_new  method might different it he class that derives from it.

So in the `AsyncConstructible::async_new(cls: Type[T]) -> T `
But the classes that derives after it have different number of constructor arguments :confused:
And we get the pylint error:
```
Number of parameters was 1 in 'AsyncConstructible.async_new' and is now 4 in overridden 'StatusTracker.async_new' method
```
And the only way (I found) to avoid it is to create AsyncConstructibleConfig abstract class, and each class that derives from AsyncConstructible has to write its own config class that is passed to the constructor.

src/frequenz/sdk/actor/power_distributing/_battery_pool_status.py

src/frequenz/sdk/actor/power_distributing/_battery_pool_status.py

src/frequenz/sdk/actor/power_distributing/power_distributing.py
ela-kotulska-frequenz: Done ( in new commit)

src/frequenz/sdk/util/asyncio.py
ela-kotulska-frequenz: As discussed it stays like this for now

shsms (approved):

sahas-subramanian-frequenz (approved):

PR #151, opened 2023-01-04
Send 0 values from power formulas for non-existant component types

shsms:
Earlier, the automatically generated formulas for `pv_power` and `battery_power` were raising exception in cases where pv or battery inverters were not found in the component graph, preventing us from having a single high level formula that can be reused in all locations.

This is solved by subscribing to a non-existing component id, so that we can still send out 0 values at the same rate as the resampler.

### DIFF
--- a/src/frequenz/sdk/timeseries/logical_meter/_formula_generators/_pv_power_formula.py
+++ b/src/frequenz/sdk/timeseries/logical_meter/_formula_generators/_pv_power_formula.py
@@ -3,10 +3,14 @@
 
 """Formula generator for PV Power, from the component graph."""
 
+import logging
+
 from .....sdk import microgrid
 from ....microgrid.component import ComponentCategory, ComponentMetricId, InverterType
 from .._formula_engine import FormulaEngine
-from ._formula_generator import ComponentNotFound, FormulaGenerator
+from ._formula_generator import NON_EXISTING_COMPONENT_ID, FormulaGenerator
+
+logger = logging.getLogger(__name__)
 
 
 class PVPowerFormula(FormulaGenerator):
@@ -32,9 +36,18 @@ class PVPowerFormula(FormulaGenerator):
         )
 
         if not pv_inverters:
-            raise ComponentNotFound(
-                "Unable to find any PV inverters in the component graph."
+            logging.warning(
+                "Unable to find any PV inverters in the component graph. "
+                "Subscribing to the resampling actor with a non-existing "
+                "component id, so that `0` values are sent from the formula."
             )
+            # If there are no PV inverters, we have to send 0 values as the same
+            # frequency as the other streams.  So we subscribe with a non-existing
+            # component id, just to get a `None` message at the resampling interval.
+            await builder.push_component_metric(
+                NON_EXISTING_COMPONENT_ID, nones_are_zeros=True
+            )
+            return builder.build()
 
         for idx, comp in enumerate(pv_inverters):
             if idx > 0:
--- a/tests/timeseries/test_logical_meter.py
+++ b/tests/timeseries/test_logical_meter.py
@@ -253,7 +253,7 @@ class TestLogicalMeter:
         self,
         mocker: MockerFixture,
     ) -> None:
-        """Test the battery power and pv power formulas."""
+        """Test the composition of formulas."""
         mockgrid = await MockMicrogrid.new(mocker, grid_side_meter=False)
         mockgrid.add_batteries(3)
         mockgrid.add_solar_inverters(2)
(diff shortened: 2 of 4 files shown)


tests/timeseries/test_logical_meter.py
```diff
@@ -300,3 +300,65 @@ async def test_formula_composition(  # pylint: disable=too-many-locals
...
+            inv_pow = await inv_calc_recv.receive()
+
+            assert inv_pow == bat_pow
+            assert pv_pow.timestamp == inv_pow.timestamp and pv_pow.value == 0.0
+            count += 1
+
+        await mockgrid.cleanup()
+        await engine._stop()  # pylint: disable=protected-access
+
+        assert count == 10
```
ela-kotulska-frequenz, line 333: Can `count` be <  10?

src/frequenz/sdk/timeseries/logical_meter/_formula_generators/_pv_power_formula.py
```diff
@@ -32,9 +37,12 @@ async def generate(self) -> FormulaEngine:
...
         if not pv_inverters:
-            raise ComponentNotFound(
-                "Unable to find any PV inverters in the component graph."
-            )
+            logging.warning("Unable to find any PV inverters in the component graph.")
+            # If there are no PV inverters, we have to send 0 values as the same
+            # frequency as the other streams.  So we add a fake component, just to get a
+            # `None` message at the resampling interval.
+            await builder.push_component_metric(sys.maxsize, nones_are_zeros=True)
+            return builder.build()
```
ela-kotulska-frequenz: 1. Could you make `sys.maxsize` constant? `FAKE_COMPONENT_ID=sys.maxsize`
2. Could you extend warning  info log with information that you try to subscribe to non-existing component (with component id). I am asking because it will result in misleading error logs from MicrogridApi.
3. Please change description: `So we add a fake component...` ->  `So we subscribe for not existing component...`.  It took me while to check where you add component. :)

tests/timeseries/test_logical_meter.py
```diff
@@ -300,3 +300,65 @@ async def test_formula_composition(  # pylint: disable=too-many-locals
...
+
+    async def test_formula_composition_missing_bat(self, mocker: MockerFixture) -> None:
+        """Test the composition of formulas with missing battery power data."""
+        mockgrid = await MockMicrogrid.new(mocker, grid_side_meter=False)
+        mockgrid.add_solar_inverters(2)
+        request_sender, channel_registry = await mockgrid.start()
+        logical_meter = LogicalMeter(
```
ela-kotulska-frequenz, line 340: Did you check if that works if there is more not existing devices? Lets say no pv arrays and no ev chargers.

tests/timeseries/test_logical_meter.py
shsms: probably not,  but I like to do it where tests are run inside loops.  it is a defense against this case: `for _ in range(list):`,  but is also a way to keep it safe against future evolution of the test.

src/frequenz/sdk/timeseries/logical_meter/_formula_generators/_pv_power_formula.py
shsms: done all 3.

tests/timeseries/test_logical_meter.py
shsms: all relevant cases are covered,  but not that combination.  there's no need to test all possible combinations I think.

ela-kotulska-frequenz (approved):

leandro-lucarella-frequenz:
It is only me that feels this is extremely hacky? Is it supposed to be only a temporary solution?

shsms:
> It is only me that feels this is extremely hacky? 

It is just you,  I feel it is only moderately hacky. ;-)

> Is it supposed to be only a temporary solution?

I was thinking this can be a long term solution.  This is a reliable way to get the a continuous stream of None values.  

A possible optimization is for the resampler to check if `component id == NON_EXISTENT_COMPONENT_ID` and then not forward the request to the data source actor, but just keep sending None values out.

leandro-lucarella-frequenz:
> I was thinking this can be a long term solution. This is a reliable way to get the a continuous stream of None values.

Can't we have just that? An easy to create task that will just inject `None` values to a channel? I guess we'll need to also add a way to build a formula from something that doesn't have a component ID (like just a receiver).

I don't know, it feels like there should be a better way.

And I'd rather not expand this special casing and `NON_EXISTENT_COMPONENT_ID` to other parts of the SDK for now, at least until we are sure there is no better way. At least for now the hack is local and contained in the logical meter.

shsms:
> Can't we have just that? An easy to create task that will just inject None values to a channel?

It has to come from the resampling actor, because we want the None values at the same interval with the same timestamps as the actual resampled component streams.  Else it would break the formula engine.

And there's only one way to communicate with the formula engine.  We might be able to add a flag to `ComponentMetricRequest` instead,  in order to make it less hacky and more explicit,  that the resampling actor can check.

leandro-lucarella-frequenz:
Mmmm, okay, maybe if we somehow keep some internal state in the formula engine that one value is missing and we can fill the gap internally without having to subscribe to stuff and actually send `None` values via a channel. I understand it is extremely convenient to solve it this way, but for me the fact that we need to resort to this hack means we made a wrong decision about the `FormulaEngine` that is leading to this hack.

Probably not a high priority issue, but I think is something we should review at some point.

shsms:
> I understand it is extremely convenient to solve it this way, but for me the fact that we need to resort to this hack means we made a wrong decision about the FormulaEngine that is leading to this hack.

I don't know, in my opinion, handling this in the formula engine would be the hack, and relying on the resampler is the right approach.

The resampler is the component that makes the promise that it will send None values when there is no component data.  The formula engine relies on this promise.  It doesn't make sense to have Formula Engine make up data when the resampler already has the code to do it.

leandro-lucarella-frequenz:
OK, maybe I'm misunderstanding, so I'm looking at the code in more detail now.

So when someone requests a battery power formula, if the battery doesn't have an inverted, this information can't be provided, so you do all this so the `FormulaEngine` can return a zero for that formula, forever.

If this is correct, why isn't this just raising an exception? Why can you subscribe to something that doesn't really have any data to send?

leandro-lucarella-frequenz:
And in any case, if we really need this and the resampler must be able to produce some arbitrary value forever, then this speaks in favor of my approach of decoupling the resampler with the microgrid data, so we could just tell the resampler: Hey, plase resample this source that is just producing zeroes (and it is not connected to any component and doesn't require any fake component ID) :)

shsms:
> why isn't this just raising an exception? Why can you subscribe to something that doesn't really have any data to send?

That was somewhat of a compromise.  Exceptions would have made it less composable.  For example,  when someone writes the below code,  exceptions would make it very inconvenient:

``` python
    net_power_recv = (
        (
            await logical_meter.grid_power()
            - await logical_meter.battery_power()
            - await logical_meter.ev_charger_power()
            - await logical_meter.pv_power()
        )
        .build("net_power")
        .new_receiver()
    )

``` 

The `FormulaReceiver`s could instead have a flag that formula builders could use,  but that would mean calls to `ev_charger_power().receive()` would just block,  which is contrary to the no-gaps-after-resampling promise.

leandro-lucarella-frequenz:
> That was somewhat of a compromise.

OK, then we agree on this. :)

But you are still taking this from the perspective of the current design. My point is maybe there is a better design that avoids this issue completely. Or maybe not. I'm not suggesting to think about this now, I'm just identifying issues.

And I still think having to use a fake invalid component ID to subscribe to the resampler so it can send `None` values through a channel, that will be received and converted to zeroes by the formula engine[*] just because a battery doesn't have an inverter so we can't get the power is an issue. I'm not that convinced that it even justifies the inconvenience of having a more complicated composability in case of errors to be honest.

[*] Which we also wanted to avoid, because we said that unknown power shouldn't be zero power for the logical meter, it should be `None`, only the *Pools can decide that that missing components means zeros, right? Shouldn't `battery_power()` come from a `BatteryPool` in your example?

shsms:
> And I still think having to use a fake invalid component ID ... is an issue.

true, it is not minimal and relies on behaviour and not on an explicit feature of the resampler.

> Which we also wanted to avoid, because we said that unknown power shouldn't be zero power for the logical meter, it should be None, only the *Pools can decide that that missing components means zeros, right? 

Yes,  the formula generators already take care of this,  based on the type of component they are reading from.  Only missing measurements from physical meters will show up as `None`s.  Missing measurements from an inverter or ev charger will get converted to 0.

> Shouldn't battery_power() come from a BatteryPool in your example?

Yes,  that's coming soon.

PR #153, opened 2023-01-04
Wait for a short time after starting the test actors

shsms:
This is to fix some async scheduling issues in python 3.8 leading to
test failures.

### DIFF
--- a/tests/timeseries/mock_microgrid.py
+++ b/tests/timeseries/mock_microgrid.py
@@ -80,7 +80,9 @@ class MockMicrogrid:
         """
         await self._server.start()
         await asyncio.sleep(0.1)
-        return await self._init_client_and_actors()
+        ret = await self._init_client_and_actors()
+        await asyncio.sleep(0.1)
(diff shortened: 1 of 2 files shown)


ela-kotulska-frequenz (approved):

tests/timeseries/mock_microgrid.py
```diff
@@ -80,7 +80,9 @@ async def start(self) -> Tuple[Sender[ComponentMetricRequest], ChannelRegistry]:
         """
         await self._server.start()
         await asyncio.sleep(0.1)
-        return await self._init_client_and_actors()
+        ret = await self._init_client_and_actors()
+        await asyncio.sleep(0.1)
+        return ret
```
ela-kotulska-frequenz, line 85: Please try to remove first sleep() (aster server.start()) and check if that works.

shsms: will clean up as part of #155,  because it is quite slow to test.

PR #154, opened 2023-01-04
Adds Python 3.11 to CI as well as bump pyarrow version constrain

christianparpart:
The pyarrow-update is important to avoid conflicting pyarrow's indirect dependencies on numpy to a version that won't be available (or working) in newer Python installations.

Wheras the Python upgrade is made especially because more recent Linux distributions (Fedora 37) won't ship with an older version of Python.


sahas-subramanian-frequenz:
closing in favour of #157

PR #157, opened 2023-01-05
Support python 3.11

shsms:


.github/workflows/ci.yaml
```diff
@@ -49,7 +50,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: "3.10"
+          python-version: "3.11"
```
ela-kotulska-frequenz, line 53: does that mean that we release with python3.11?

.github/workflows/ci.yaml
```diff
@@ -13,6 +13,7 @@ jobs:
           - "3.8"
           - "3.9"
           - "3.10"
+          - "3.11"
```
ela-kotulska-frequenz, line 16: Please update release notes.

.github/workflows/ci.yaml
shsms: it will use python 3.11 to build the wheel,  but it will still be the same code that has been tested against all versions from 3.8.

And the wheel file will work in all pythons because of this: https://github.com/frequenz-floss/frequenz-sdk-python/blob/82fc84d1054188eb83a26205b927dc9462ddc270/pyproject.toml#L26

.github/workflows/ci.yaml
shsms: done

ela-kotulska-frequenz (approved):

PR #164, opened 2023-01-17
Add 3-phase current formulas

shsms:
This PR introduces some new types for streaming 3-phase data:
- `Sample3Phase` :: for holding a 3-phase measurement.
- `FormulaChannel3Phase/FormulaReceiver3Phase` :: composable channel/receiver types for streaming `Sample3Phase` objects.
- `FormulaEngine3Phase` :: for combining `Sample` objects from `FormulaEngine` into `Sample3Phase` objects.
- `HigherOrderFormulaBuilder3Phase` :: For composing `FormulaReceiver3Phase` and generating new `FormulaEngine3Phase` from them.

It also adds formula generators for 3-phase `grid_current` and `ev_charger_current`,  and exposes them through the `LogicalMeter`.

### DIFF
--- a/src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
+++ b/src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
@@ -8,10 +8,11 @@ from __future__ import annotations
 import asyncio
 import logging
 import weakref
+from abc import ABC
 from collections import deque
 from datetime import datetime
 from math import isinf, isnan
-from typing import Dict, List, Optional, Set, Tuple
+from typing import Dict, Generic, List, Optional, Set, Tuple, Type, TypeVar
 from uuid import UUID, uuid4
 
 from frequenz.channels import Broadcast, Receiver
@@ -310,11 +311,24 @@ class FormulaBuilder:
         return FormulaEngine(self._name, self._steps, self._metric_fetchers)
 
 
-class FormulaChannel(Broadcast[Sample]):
+_GenericSample = TypeVar("_GenericSample")
+_GenericEngine = TypeVar("_GenericEngine")
+_GenericFormulaChannel = TypeVar("_GenericFormulaChannel")
+_GenericFormulaReceiver = TypeVar("_GenericFormulaReceiver")
+_GenericHOFormulaBuilder = TypeVar("_GenericHOFormulaBuilder")
+
+
+class _BaseFormulaChannel(
+    Generic[_GenericSample, _GenericEngine],
+    Broadcast[_GenericSample],
+    ABC,
+):
     """A broadcast channel implementation for use with formulas."""
 
+    ReceiverType: Type[FormulaReceiver]
+
     def __init__(
-        self, name: str, engine: FormulaEngine, resend_latest: bool = False
+        self, name: str, engine: _GenericEngine, resend_latest: bool = False
     ) -> None:
         """Create a `FormulaChannel` instance.
 
@@ -328,7 +342,7 @@ class FormulaChannel(Broadcast[Sample]):
         super().__init__(name, resend_latest)
 
     @property
-    def engine(self) -> FormulaEngine:
+    def engine(self) -> _GenericEngine:
         """Return the formula engine attached to the channel.
 
         Returns:
@@ -338,7 +352,7 @@ class FormulaChannel(Broadcast[Sample]):
 
     def new_receiver(
         self, name: Optional[str] = None, maxsize: int = 50
-    ) -> FormulaReceiver:
+    ) -> _GenericFormulaReceiver:
         """Create a new FormulaReceiver for the channel.
 
         This implementation is similar to `Broadcast.new_receiver()`, except that it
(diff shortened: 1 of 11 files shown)


src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
```diff
@@ -8,10 +8,11 @@
 import asyncio
```
ela-kotulska-frequenz, line 8: Would it be ok to separate this file (_formula_engine.py) into 3 files?
* _base_formula_engine.py
* _formula_engine.py
* _3Phase_formula_engine.py

Generics can stay. I am telling only about separating this file .

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
```diff
@@ -222,6 +222,72 @@ def new_receiver(
...
+                await sender.send(msg)
+
+    async def _stop(self) -> None:
+        """Stop a running formula engine."""
+        if self._task is None:
+            return
+        await cancel_and_await(self._task)
+
+    def new_receiver(
+        self, name: Optional[str] = None, max_size: int = 50
```
ela-kotulska-frequenz, line 274: why max_size: int = 50?
receiver has default max_size = 10.

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
shsms: Unfortunately,  to do this,  we'll have to do some sacrifices to type safety,  or else, there will be cyclic imports.  

Specifically,  the generic types `_GenericEngine`, etc will have to be unbounded,  because otherwise the base classes file will have to import each engine implementation separately.

So we'll have `_GenericEngine = TypeVar("_GenericEngine")` instead of `= TypeVar("_GenericEngine", FormulaEngine, FormulaEngine3Phase)`
and `ReceiverType: Type[Any]` instead of `: Type[FormulaReceiver | FormulaReceiver3Phase]`.

I think that's tighter constraints and better type checks are more valuable than smaller files.  But I guess it is a tough decision.

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
shsms: No,  receivers have default 50:  https://frequenz-floss.github.io/frequenz-channels-python/v0.11/reference/frequenz/channels/#frequenz.channels._broadcast.Broadcast.new_receiver

ela-kotulska-frequenz: Thanks!
Then we should not override it, here. Yes?

shsms: I don't know, if people want to, they can override, right?

or are you talking about changing the default? my understanding is that 50 has been a good default for our use cases.  do you think there is a reason to change it?

matthias-wende-frequenz: Maybe you should store the default somewhere, like e.g. `DEFAULT_CHANNEL_SIZE` and use it everywhere.

matthias-wende-frequenz: or as an alternative, check here if it has been set by the user and if not don't pass it to the Channels `new_receiver()` function.

ela-kotulska-frequenz: I mean that here you override the default Receiver's max_size.
So if anytime in the future receivers default size change, then it will change anywhere, except for this place.
Why you do this?

If you need receivers default max_size, then you could do:
`def new_receiver(self, name: Optional[str] = None, max_size: Optional[int] = None) `

src/frequenz/sdk/timeseries/logical_meter/_formula_generators/_grid_current_formula.py
```diff
@@ -0,0 +1,97 @@
...
+                ).new_receiver(),
+            ),
+        )
+
+    async def _gen_phase_formula(
+        self,
+        grid_successors: Set[Component],
+        metric_id: ComponentMetricId,
+    ) -> FormulaEngine:
+        builder = self._get_builder("grid-current", metric_id)
```
ela-kotulska-frequenz, line 69: Just curious:
How is it possible that this method doesn't need description? :)

src/frequenz/sdk/timeseries/logical_meter/_formula_generators/_ev_charger_current_formula.py
```diff
@@ -0,0 +1,91 @@
...
+        )
+
+    async def _gen_phase_formula(
+        self,
+        ev_chargers: List[Component],
+        metric_id: ComponentMetricId,
+    ) -> FormulaEngine:
+        builder = self._get_builder("ev-current", metric_id)
+
+        # generate a formula that just adds values from all commponents that are
```
ela-kotulska-frequenz: Comment seems to be wrong.
`generate a formula that just adds values from the ev_chargers`
typo `commponents`

src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py
shsms: Why should we expect that their buffer sizes should be coupled?  `FormulaReceiver`s are very specific to a single use case and with them,  we can fine tune the default more easily.

Broadcast `Receiver`s are more general purpose and we chose 50 randomly and it seems to work for most cases.  But by decoupling them,  we get the freedom to change the general purpose `Receiver`'s buffer size without affecting the `FormulaReceiver`.

src/frequenz/sdk/timeseries/logical_meter/_formula_generators/_grid_current_formula.py
shsms: our tools allow for private methods to not have a docstring. :-)

src/frequenz/sdk/timeseries/logical_meter/_formula_generators/_ev_charger_current_formula.py
shsms: fixed.

ela-kotulska-frequenz (dismissed):

shsms:
Rebased to latest head.

ela-kotulska-frequenz (approved):

PR #167, opened 2023-01-18
Serializing Ringbuffer

### LINKED ISSUE #132: Handle data from disk
### What's needed?

We want an simple and efficient way to handle dumping and loading data from disk.

### Proposed solution

Add dumping the full buffer to disk and reload the buffer from disk to the [ring buffer](https://github.com/frequenz-floss/frequenz-sdk-python/pull/139).

The data should be stored when the application is terminated and on a regular basis, e.g. every 15 minutes.

### Open Questions

We are certainly creating gaps of missing data and it remains an open question how to handle this data. The simplest approach is to fill missing data with NaN, but other approaches like e.g. back filling are among the other options.

### Use cases

Let's consider an application that creates one feature from timeseries data. Usually we would think of creating a buffer, e.g. using a [ring buffer](https://github.com/frequenz-floss/frequenz-sdk-python/issues/113), that stores data for a certain time window, e.g. 15 minutes or 28 days.

When the application restarts, if there is no data on disk, it would take the time of the window size to rebuild this buffer and hence create this feature.

Instead of simply dumping the raw buffer to disk we could:

* Use the raw data and send it to the sdk's data pipeline,
* Write app that continuously stores data from the logical meter as files.

### Additional context

_No response_

mathias-baumann-frequenz:
based on #139

### DIFF
--- a/benchmarks/timeseries/serializable_ringbuffer.py
+++ b/benchmarks/timeseries/serializable_ringbuffer.py
@@ -1,0 +1,81 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Benchmarks the `SerializableRingbuffer` class."""
+
+from __future__ import annotations
+
+import fnmatch
+import os
+import time
+from datetime import datetime, timedelta
+from typing import Any
+
+import numpy as np
+
+from frequenz.sdk.timeseries import Sample
+from frequenz.sdk.timeseries._serializable_ringbuffer import SerializableRingBuffer
+
+FILE_NAME = "ringbuffer.pkl"
+FIVE_MINUTES = timedelta(minutes=5)
+
+# Size of the ringbuffer to dump/load
+SIZE = 4000_000
+# Number of iterations to run the benchmark
+ITERATIONS = 100
+
+
+def delete_files_with_prefix(prefix: str) -> None:
+    """Delete all files starting with the given prefix.
+
+    Args:
+        prefix: Prefix of the files to delete
+    """
+    for file in os.listdir():
+        if fnmatch.fnmatch(file, prefix + "*"):
+            os.remove(file)
+
+
+def benchmark_serialization(
+    ringbuffer: SerializableRingBuffer[Any], iterations: int
+) -> float:
+    """Benchmark the given buffer `iteration` times.
+
+    Args:
+        ringbuffer: Ringbuffer to benchmark to serialize.
+        iterations: amount of iterations to run.
+    """
+    total = 0.0
+    for _ in range(iterations):
+        start = time.time()
+        ringbuffer.dump()
+        SerializableRingBuffer.load(FILE_NAME)
+        end = time.time()
+        total += end - start
+        delete_files_with_prefix(FILE_NAME)
+
+    return total / iterations
+
+
+def main() -> None:
+    """Run Benchmark."""
+    ringbuffer = SerializableRingBuffer(
+        np.arange(0, SIZE, dtype=np.float64), timedelta(minutes=5), FILE_NAME
+    )
+
+    print("size:", SIZE)
+    print("iterations:", ITERATIONS)
+
+    for i in range(0, SIZE, 10000):
+        ringbuffer.update(
+            Sample(datetime.fromtimestamp(200 + i * FIVE_MINUTES.total_seconds()), i)
+        )
+
+    print(
+        "Avg time for Pickle dump/load:  "
+        f"{benchmark_serialization(ringbuffer, ITERATIONS)}s"
+    )
+
+
+if __name__ == "__main__":
+    main()
--- a/src/frequenz/sdk/timeseries/_ringbuffer.py
+++ b/src/frequenz/sdk/timeseries/_ringbuffer.py
@@ -130,7 +130,7 @@ class OrderedRingBuffer(Generic[FloatArray]):
         # Don't add outdated entries
         if timestamp < self._datetime_oldest and self._datetime_oldest != datetime.max:
             raise IndexError(
-                f"Timestamp too old (cut-off is at {self._datetime_oldest})."
+                f"Timestamp {timestamp} too old (cut-off is at {self._datetime_oldest})."
             )
 
         # Update timestamps
--- a/src/frequenz/sdk/timeseries/_ringbuffer.py
+++ b/src/frequenz/sdk/timeseries/_ringbuffer.py
@@ -14,7 +14,7 @@ from typing import Any, Generic, List, TypeVar, overload
 import numpy as np
 import numpy.typing as npt
 
-from frequenz.sdk.timeseries import Sample
+from . import Sample
 
 FloatArray = TypeVar("FloatArray", List[float], npt.NDArray[np.float64])
 
--- a/src/frequenz/sdk/timeseries/_ringbuffer.py
+++ b/src/frequenz/sdk/timeseries/_ringbuffer.py
@@ -205,7 +205,7 @@ class OrderedRingBuffer(Generic[FloatArray]):
                 copy of the data.
 
         Raises:
-            IndexError: when requesting a window with invalid timestamps.
+            IndexError: When requesting a window with invalid timestamps.
 
         Returns:
             The requested window
--- a/src/frequenz/sdk/timeseries/_ringbuffer.py
+++ b/src/frequenz/sdk/timeseries/_ringbuffer.py
@@ -5,17 +5,19 @@
 
 from __future__ import annotations
 
+from collections.abc import Iterable
 from copy import deepcopy
 from dataclasses import dataclass
 from datetime import datetime, timedelta
-from typing import Any, Generic, List, Sequence, TypeVar
+from typing import Any, Generic, List, TypeVar, overload
 
 import numpy as np
+import numpy.typing as npt
 from numpy.lib import math
 
 from frequenz.sdk.timeseries import Sample
 
-Container = TypeVar("Container", list, np.ndarray)
+FloatArray = TypeVar("FloatArray", List[float], npt.NDArray[np.float64])
 
 
 @dataclass
@@ -42,12 +44,12 @@ class Gap:
         return False
 
 
-class OrderedRingBuffer(Generic[Container]):
+class OrderedRingBuffer(Generic[FloatArray]):
     """Time aware ringbuffer that keeps its entries sorted by time."""
 
     def __init__(
         self,
-        buffer: Container,
+        buffer: FloatArray,
         sampling_period: timedelta,
         time_index_alignment: datetime = datetime(1, 1, 1),
     ) -> None:
@@ -179,7 +181,7 @@ class OrderedRingBuffer(Generic[Container]):
 
     def window(
         self, start: datetime, end: datetime, force_copy: bool = False
-    ) -> Container:
+    ) -> FloatArray:
         """Request a view on the data between start timestamp and end timestamp.
 
         If the data is not used immediately it could be overwritten.
@@ -228,14 +230,13 @@ class OrderedRingBuffer(Generic[Container]):
                     window = np.concatenate((window, self._buffer[0:end_index]))
             return window
 
-        def in_gap(gap) -> bool:
-            return gap.contains(start) or gap.contains(end)
-
         # Return a copy if there are none-values in the data
-        if force_copy or any(map(in_gap, self._gaps)):
-            return deepcopy(self._buffer[start_index:end_index])
+        if force_copy or any(
+            map(lambda gap: gap.contains(start) or gap.contains(end), self._gaps)
+        ):
+            return deepcopy(self[start_index:end_index])
 
-        return self._buffer[start_index:end_index]
+        return self[start_index:end_index]
 
     def is_missing(self, timestamp: datetime) -> bool:
         """Check if the given timestamp falls within a gap.
@@ -386,7 +387,33 @@ class OrderedRingBuffer(Generic[Container]):
         """
         return index % self.maxlen
 
-    def __setitem__(self, index_or_slice: int | slice, value: float) -> None:
+    @overload
+    def __setitem__(self, index_or_slice: slice, value: Iterable[float]) -> None:
+        """Set values at the request slice positions.
+
+        No wrapping of the index will be done.
+        Create a feature request if you require this function.
+
+        Args:
+            index_or_slice: Slice specification of the requested data.
+            value: Sequence of value to set at the given range.
+        """
+
+    @overload
+    def __setitem__(self, index_or_slice: int, value: float) -> None:
+        """Set value at requested index.
+
+        No wrapping of the index will be done.
+        Create a feature request if you require this function.
+
+        Args:
+            index_or_slice: Index of the data.
+            value: Value to set at the given position.
+        """
+
+    def __setitem__(
+        self, index_or_slice: int | slice, value: float | Iterable[float]
+    ) -> None:
         """Set item or slice at requested position.
 
         No wrapping of the index will be done.
@@ -398,7 +425,29 @@ class OrderedRingBuffer(Generic[Container]):
         """
         self._buffer.__setitem__(index_or_slice, value)
 
-    def __getitem__(self, index_or_slice: int | slice) -> float | Sequence[float]:
+    @overload
+    def __getitem__(self, index_or_slice: int) -> float:
+        """Get item at requested position.
+
+        No wrapping of the index will be done.
+        Create a feature request if you require this function.
+
+        Args:
+            index_or_slice: Index of the requested data.
+        """
+
+    @overload
+    def __getitem__(self, index_or_slice: slice) -> FloatArray:
+        """Get the data described by the given slice.
+
+        No wrapping of the index will be done.
+        Create a feature request if you require this function.
+
+        Args:
+            index_or_slice: Slice specification of where the requested data is.
+        """
+
+    def __getitem__(self, index_or_slice: int | slice) -> float | FloatArray:
         """Get item or slice at requested position.
 
         No wrapping of the index will be done.
--- a/src/frequenz/sdk/timeseries/_ringbuffer.py
+++ b/src/frequenz/sdk/timeseries/_ringbuffer.py
@@ -13,7 +13,6 @@ from typing import Any, Generic, List, TypeVar, overload
 
 import numpy as np
 import numpy.typing as npt
-from numpy.lib import math
 
 from frequenz.sdk.timeseries import Sample
 
@@ -140,7 +139,7 @@ class OrderedRingBuffer(Generic[FloatArray]):
         self._datetime_oldest = self._datetime_newest - self._time_range
 
         # Update data
-        value: float = math.nan if sample.value is None else sample.value
+        value: float = np.nan if sample.value is None else sample.value
         self._buffer[self.datetime_to_index(timestamp)] = value
 
         self._update_gaps(timestamp, prev_newest, sample.value is None)
--- a/src/frequenz/sdk/timeseries/_serializable_ringbuffer.py
+++ b/src/frequenz/sdk/timeseries/_serializable_ringbuffer.py
@@ -1,0 +1,85 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Ringbuffer implementation with serialization support."""
+
+# For use of the class type hint inside the class itself.
+from __future__ import annotations
+
+import pickle
+from datetime import datetime, timedelta
+from os.path import exists
+
+from ._ringbuffer import FloatArray, OrderedRingBuffer
+
+# Version of the file dumping/loading format
+FILE_FORMAT_VERSION: int = 1
+
+
+class SerializableRingBuffer(OrderedRingBuffer[FloatArray]):
+    """Sorted ringbuffer with serialization support."""
+
+    # pylint: disable=too-many-arguments
+    def __init__(
+        self,
+        buffer: FloatArray,
+        sampling_period: timedelta,
+        path: str,
+        time_index_alignment: datetime = datetime(1, 1, 1),
+    ) -> None:
+        """Initialize the time aware ringbuffer.
+
+        Args:
+            buffer: Instance of a buffer container to use internally.
+            sampling_period: Timedelta of the desired resampling period.
+            path: Path to where the data should be saved to and loaded from.
+            time_index_alignment: Arbitrary point in time used to align
+                timestamped data with the index position in the buffer.
+                Used to make the data stored in the buffer align with the
+                beginning and end of the buffer borders.
+                For example, if the `time_index_alignment` is set to
+                "0001-01-01 12:00:00", and the `sampling_period` is set to
+                1 hour and the length of the buffer is 24, then the data
+                stored in the buffer could correspond to the time range from
+                "2022-01-01 12:00:00" to "2022-01-02 12:00:00" (date chosen
+                arbitrarily here).
+        """
+        super().__init__(buffer, sampling_period, time_index_alignment)
+        self._path = path
+        self._file_format_version = FILE_FORMAT_VERSION
+
+    def dump(self) -> None:
+        """Dump data to disk.
+
+        Raises:
+            I/O related exceptions when the file cannot be written.
+        """
+        with open(self._path, mode="wb+") as fileobj:
+            pickle.dump(self, fileobj)
+
+    @classmethod
+    def load(cls, path: str) -> SerializableRingBuffer[FloatArray] | None:
+        """Load data from disk.
+
+        Args:
+            path: Path to the file where the data is stored.
+
+        Raises:
+            I/O related exceptions when file exists but isn't accessable for any
+            reason.
+
+        Returns:
+            `None` when the file doesn't exist, otherwise an instance of the
+            `SerializableRingBuffer` class, loaded from disk.
+        """
+        if not exists(path):
+            return None
+
+        with open(path, mode="rb") as fileobj:
+            instance: SerializableRingBuffer = pickle.load(fileobj)
+            instance._path = path  # pylint: disable=protected-access
+            # Set latest file format version for next time it dumps.
+            # pylint: disable=protected-access
+            instance._file_format_version = FILE_FORMAT_VERSION
+
+            return instance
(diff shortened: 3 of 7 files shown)


mathias-baumann-frequenz:
Apart from the DCO problem which I currently don't know how to solve, this PR is ready for feedback.

src/frequenz/sdk/timeseries/_serialized_ringbuffer.py
```diff
@@ -0,0 +1,77 @@
...
+
+    # pylint: disable=too-many-arguments
+    def __init__(
+        self,
+        buffer: Container,
+        sampling_period: timedelta,
+        path: str,
+        time_index_alignment: datetime = datetime(1, 1, 1),
+        load_from_disk: bool = True,
+    ) -> None:
```
ela-kotulska-frequenz: 1) Maybe instead of `load_from_disk: bool = True`  we could check if the data in the path are in `self._time_range` from OrderedRingBuffer?

2) Would we create gap for time between  `_datetime_oldest` and `datetime.now()` ?

mathias-baumann-frequenz: no gap would be created, only when you push a new value with a timestamp in the future (future from the ringbuffers perspective, that is)

src/frequenz/sdk/timeseries/_serialized_ringbuffer.py
```diff
@@ -0,0 +1,77 @@
...
+    "_datetime_newest",
+    "_buffer",
+    "_gaps",
+    "_sampling_period",
+    "_time_range",
+    "_time_index_alignment",
+]
+
+
+class SerializedRingbuffer(OrderedRingBuffer[T, Container]):
```
daniel-zullo-frequenz: Should this class be named `SerializedRingBuffer` instead? (Capital Buffer)

src/frequenz/sdk/timeseries/_serialized_ringbuffer.py
```diff
@@ -0,0 +1,77 @@
...
+        path: str,
+        time_index_alignment: datetime = datetime(1, 1, 1),
+        load_from_disk: bool = True,
+    ) -> None:
+        """Initialize the time aware ringbuffer.
+
+        Args:
+            buffer: Instance of a buffer container to use internally.
+            sampling_period: Timedelta of the desired resampling period.
+            path: Path to where the data should be saved.
```
daniel-zullo-frequenz: is `path` only used to dump/save data? if not, then it should be updated

tests/timeseries/serialized_ringbuffer.py
```diff
@@ -0,0 +1,125 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Tests for the `SerializedRingbuffer` class."""
```
daniel-zullo-frequenz: `SerializedRingBuffer` if my related comment is taken

src/frequenz/sdk/timeseries/_serialized_ringbuffer.py
```diff
@@ -0,0 +1,77 @@
...
+
+ATTRIBUTES_TO_PICKLE = [
+    "_datetime_oldest",
+    "_datetime_newest",
+    "_buffer",
+    "_gaps",
+    "_sampling_period",
+    "_time_range",
+    "_time_index_alignment",
+]
```
daniel-zullo-frequenz: Should `ATTRIBUTES_TO_PICKLE` be a tuple rather than a list?

tests/timeseries/serialized_ringbuffer.py
```diff
@@ -0,0 +1,125 @@
...
+    buffer._time_range = timedelta(seconds=200000)
+    # pylint: disable=protected-access
+    buffer._sampling_period = ONE_MINUTE
+    # pylint: disable=protected-access
+    buffer._time_index_alignment = datetime.min
+
+    # Load old data
+    buffer.load()
+
+    assert list(buffer[:]) == list(before[:])
```
daniel-zullo-frequenz: I wonder if `assert np.array_equal(buffer[:], before[:])` should be more suitable to compare avoiding the conversion to a list

src/frequenz/sdk/timeseries/_serialized_ringbuffer.py
```diff
@@ -0,0 +1,77 @@
...
+        time_index_alignment: datetime = datetime(1, 1, 1),
+        load_from_disk: bool = True,
+    ) -> None:
+        """Initialize the time aware ringbuffer.
+
+        Args:
+            buffer: Instance of a buffer container to use internally.
+            sampling_period: Timedelta of the desired resampling period.
+            path: Path to where the data should be saved.
+            time_index_alignment: Arbitary point in time used to align
```
daniel-zullo-frequenz: typo `s/Arbitary/Arbitrary` ?

src/frequenz/sdk/timeseries/_serialized_ringbuffer.py
```diff
@@ -0,0 +1,77 @@
...
+                "0001-01-01 12:00:00", and the `sampling_period` is set to
+                1 hour and the length of the buffer is 24, then the data
+                stored in the buffer could correspond to the time range from
+                "2022-01-01 12:00:00" to "2022-01-02 12:00:00" (date chosen
+                arbitrarily here).
+            load_from_disk: Default True, if True, will load the data from disk
+                upon start.
+        """
+        super().__init__(buffer, sampling_period, time_index_alignment)
+        self._path = path
```
daniel-zullo-frequenz: Should the `path` be checked? (path exists or write permission)
My concern here is, for instance, `load()` relies on the path to load data. I also wonder if it would be more convenient to pass to load() and dump() as a parameter instead

tests/timeseries/serialized_ringbuffer.py
mathias-baumann-frequenz: The buffer works with both `list` and `np.array`, so your solution would work only for the latter while this one, while maybe not the fastest, will work for all. As it is just a test, I valued readability over speed here

src/frequenz/sdk/timeseries/_serialized_ringbuffer.py
daniel-zullo-frequenz: hmm on the other hand it could create inconsistencies between loaded and dumped data if the path is passed as a parameter to the functions

tests/timeseries/serialized_ringbuffer.py
daniel-zullo-frequenz: ok, makes sense

daniel-zullo-frequenz (dismissed):
All my previous comments have been addressed and LGTM

src/frequenz/sdk/timeseries/_ringbuffer.py
```diff
@@ -225,17 +227,24 @@ def window(
...
-
-        return self._buffer[start_index:end_index]
+        if force_copy or any(
+            map(lambda gap: gap.contains(start) or gap.contains(end), self._gaps)
+        ):
+            # Make type system happy
+            ret = deepcopy(self[start_index:end_index])
+            if not isinstance(ret, float):
+                return ret
+            raise TypeError("Type error while copying data.")
```
leandro-lucarella-frequenz: This looks odd. Can you clarify:

* Why you need to make the type system happy and how? Is really the `deepcopy()` for that?
* Why only a `float` is a type error? How is it that `deepcopy()` can return a `float` here?

src/frequenz/sdk/timeseries/_ringbuffer.py
```diff
@@ -225,17 +227,24 @@ def window(
...
+
+        # Make type system happy
+        ret = self[start_index:end_index]
+        if not isinstance(ret, float):
+            return ret
+        raise TypeError("Type error while copying data.")
```
leandro-lucarella-frequenz: Same as above.

src/frequenz/sdk/timeseries/_serialized_ringbuffer.py
```diff
@@ -0,0 +1,78 @@
...
+        if load_from_disk and exists(self._path):
+            self.load()
+
+    def dump(self) -> None:
+        """Dump data to disk."""
+        with open(self._path, mode="wb+") as fileobj:
+            pickle.dump(
+                tuple(getattr(self, attr) for attr in VARS_TO_SERIALIZE),
+                fileobj,
+            )
```
leandro-lucarella-frequenz: I think the correct way to control what to pickle is to use [`__getstate__()`](https://docs.python.org/3/library/pickle.html#object.__getstate__). Not a 100% sure, but I'm almost sure there was a magic method to control this.

leandro-lucarella-frequenz (commented):
I would name it `SerializableRingBuffer` instead.

Since we have more ring buffers now, it might be worth adding a `frequenz.timeseries.ringbuffer` package with all the ring buffers. So in terms of fs:

```
frequenz/timeseries/ringbuffer/
frequenz/timeseries/ringbuffer/__init__.py (importing the other private modules)
frequenz/timeseries/ringbuffer/_ordered.py
frequenz/timeseries/ringbuffer/_serializable.py
```

Also, you are planning to add a dump version number, right?

src/frequenz/sdk/timeseries/_ringbuffer.py
```diff
@@ -5,17 +5,19 @@
...
 from datetime import datetime, timedelta
-from typing import Any, Generic, List, Sequence, TypeVar
+from typing import Any, Generic, List, TypeVar
 
-import numpy as np
+import numpy.typing as npt
+from numpy import concatenate, float64
 from numpy.lib import math
 
 from frequenz.sdk.timeseries import Sample
```
shsms: could you make this use relative paths for sdk components:

``` suggestion
from . import Sample
```

src/frequenz/sdk/timeseries/_serialized_ringbuffer.py
```diff
@@ -0,0 +1,78 @@
...
+
+import pickle
+from datetime import datetime, timedelta
+from os.path import exists
+from typing import List, TypeVar
+
+import numpy as np
+import numpy.typing as npt
+
+from frequenz.sdk.timeseries._ringbuffer import OrderedRingBuffer
```
shsms: and here too:
```suggestion
from ._ringbuffer import OrderedRingBuffer
```

src/frequenz/sdk/timeseries/_ringbuffer.py
mathias-baumann-frequenz: No, again, `deepcopy` is to return a copy of the requested buffer that can be modified without modifying the actual ringbuffer memory. No relation to types here (it existed before as well)

in vim I use pyright which complained about those lines a lot even though mypy and pylint didn't complain, but those errors also propagated into other parts and there mypy/pylint would complain sometimes, so I rather fixed it at the source.

`deepcopy` can return a float because if you pass a single-integer to `__getitem__` you get a single float value back, which in this case, is impossible as we pass a range. The typesystem doesn't know that though.

src/frequenz/sdk/timeseries/_ringbuffer.py
mathias-baumann-frequenz: Same explanation ;)

src/frequenz/sdk/timeseries/_serialized_ringbuffer.py
mathias-baumann-frequenz: Ah good to know, I'll look into that

src/frequenz/sdk/timeseries/_ringbuffer.py
mathias-baumann-frequenz: I didn't know we can do that. sure, will update

mathias-baumann-frequenz:
> Also, you are planning to add a dump version number, right?

Forgot about that one, thanks for reminding.

src/frequenz/sdk/timeseries/_ringbuffer.py
leandro-lucarella-frequenz: > No, again, deepcopy is to return a copy of the requested buffer that can be modified without modifying the actual ringbuffer memory. No relation to types here (it existed before as well)

:+1:

> in vim I use pyright which complained about those lines a lot even though mypy and pylint didn't complain, but those errors also propagated into other parts and there mypy/pylint would complain sometimes, so I rather fixed it at the source.
>
> deepcopy can return a float because if you pass a single-integer to __getitem__ you get a single float value back, which in this case, is impossible as we pass a range. The typesystem doesn't know that though.

Ah, I see, then the correct way to type-hint this is to add `@overload`s for `__getitem__`, this tells the type system which combination of arguments and return types are valid. This way you won't have to do the runtime check yourself and the type system can make sure your types will be OK at runtime.

You can [see an example](https://github.com/leandro-lucarella-frequenz/frequenz-sdk-python/blob/55916591a2659690b88639fa05684e5d8b8136e3/src/frequenz/sdk/timeseries/_resampling/_buffer.py#L135-L143) in my failed attempt to make a more efficient ring buffer for the resampler:

```py
    @overload
    def __getitem__(self, index: int) -> T:
        ...


    @overload
    def __getitem__(self, index: slice) -> Sequence[T]:
        ...


    def __getitem__(self, index: int | slice) -> T | Sequence[T]:
```

src/frequenz/sdk/timeseries/_ringbuffer.py
leandro-lucarella-frequenz: Same reply :D

Feel free to close this conversation if you'll remember to fix both sites :)

benchmarks/timeseries/serializable_ringbuffer.py
```diff
@@ -0,0 +1,82 @@
...
+        SerializableRingBuffer.load(FILE_NAME)
+        end = time.time()
+        total += end - start
+        delete_files_with_prefix(FILE_NAME)
+
+    return total / iterations
+
+
+SIZE = 4000_000
+ITERATIONS = 100
```
christianparpart: Maybe you want to group all the global constants together on top _and_ have each of these also well documented? `SIZE` and `ITERATIONS` also sounds pretty generic, which makes it hard to figure out the meaning just be looking at them. :)

benchmarks/timeseries/serializable_ringbuffer.py
```diff
@@ -0,0 +1,82 @@
...
+    for i in range(0, SIZE, 10000):
+        ringbuffer.update(
+            Sample(datetime.fromtimestamp(200 + i * FIVE_MINUTES.total_seconds()), i)
+        )
+
+    print(
+        "Avg time for Pickle dump/load:  ",
+        benchmark_serialization(ringbuffer, ITERATIONS),
+        "s",
+    )
```
christianparpart, line 77: why not a format-string?

tests/timeseries/test_serializable_ringbuffer.py
```diff
@@ -0,0 +1,107 @@
...
+    assert dumped._gaps == loaded._gaps
+    # pylint: disable=protected-access
+    assert dumped._time_range == loaded._time_range
+    # pylint: disable=protected-access
+    assert dumped._sampling_period == loaded._sampling_period
+    # pylint: disable=protected-access
+    assert dumped._time_index_alignment == loaded._time_index_alignment
+
+
+def test_load_dump_short(tmp_path_factory: Any) -> None:
```
christianparpart: Why `Any` and not something more specific?

tests/timeseries/test_serializable_ringbuffer.py
```diff
@@ -0,0 +1,107 @@
...
+        SerializableRingBuffer(
+            np.empty(shape=(24 * int(FIVE_MINUTES.total_seconds()),), dtype=np.float64),
+            FIVE_MINUTES,
+            f"{tmpdir}/test_array.bin",
+            datetime(2, 2, 2),
+        )
+    )
+
+
+def test_load_dump(tmp_path_factory: Any) -> None:
```
christianparpart: same question here. :)

benchmarks/timeseries/serializable_ringbuffer.py
mathias-baumann-frequenz: I mean, in the context of a benchmark and a ringbuffer I feel those are _very_ unambigious in what they could mean...

tests/timeseries/test_serializable_ringbuffer.py
mathias-baumann-frequenz: Because it was faster and is just a test, not production code, so I didn't invest the time to find the right import for the typehint..

benchmarks/timeseries/serializable_ringbuffer.py
mathias-baumann-frequenz: Does it make much difference? I think this part was written by GPT ;)

leandro-lucarella-frequenz: Slap GPT until it uses f-strings.

src/frequenz/sdk/timeseries/_serializable_ringbuffer.py
```diff
@@ -0,0 +1,91 @@
...
+from datetime import datetime, timedelta
+from os.path import exists
+from typing import List, TypeVar
+
+import numpy as np
+import numpy.typing as npt
+
+from ._ringbuffer import OrderedRingBuffer
+
+Container = TypeVar("Container", List[float], npt.NDArray[np.float64])
```
shsms: why not just:

``` suggestion
from ._ringbuffer import OrderedRingBuffer, Container
```

src/frequenz/sdk/timeseries/_ringbuffer.py
```diff
@@ -5,17 +5,19 @@
...
 from copy import deepcopy
 from dataclasses import dataclass
 from datetime import datetime, timedelta
-from typing import Any, Generic, List, Sequence, TypeVar
+from typing import Any, Generic, List, TypeVar, overload
 
-import numpy as np
+import numpy.typing as npt
+from numpy import concatenate, float64
 from numpy.lib import math
```
leandro-lucarella-frequenz: I usually don't like much import aliases, but `import numpy as np` (and `import numpy.typing as npt` too I think) is extremely widely used, and being a huge library that uses common names like `concatenate` or `math`.

I'm slightly inclined to keep using `import numpy as np` to make it more clear in the code which parts are using `numpy` and which aren't.

As for `from numpy.lib import math`, can't you use [`np.nan`](https://numpy.org/doc/stable/reference/constants.html#numpy.nan)? I can't even find `numpy.lib.math.nan` in the docs or source code :thinking:

benchmarks/timeseries/serializable_ringbuffer.py
shsms: also good reason to rewrite,  to keep all the gpt code out.

src/frequenz/sdk/timeseries/_ringbuffer.py
```diff
@@ -5,17 +5,19 @@
...
 
-from frequenz.sdk.timeseries import Sample
+from . import Sample
 
-Container = TypeVar("Container", list, np.ndarray)
+Container = TypeVar("Container", List[float], npt.NDArray[float64])
```
leandro-lucarella-frequenz: Also `Container` is well established name in Python (`collections.abc.Container`) so maybe consider using something more specific (maybe `FloatArray`?) or just make it private if it's only a hack for type checking and doesn't need to be imported (`_Container`).

src/frequenz/sdk/timeseries/_serializable_ringbuffer.py
```diff
@@ -0,0 +1,91 @@
...
+        """
+        if not exists(path):
+            return None
+
+        with open(path, mode="rb") as fileobj:
+            instance: SerializableRingBuffer = pickle.load(fileobj)
+            instance._path = path  # pylint: disable=protected-access
+            # Set latest file format version for next time it dumps.
+            # pylint: disable=protected-access
+            instance._file_format_version = FILE_FORMAT_VERSION
```
leandro-lucarella-frequenz, line 83: I see one issue with this approach: We are stuck with pickle, or at least to update to another format, we'll need to first try to deserialize using pickle and if that fails try to deserialize in the new format. Since performance is an important thing in this loading and dumping, I wonder if we shouldn't have more flexibility in terms of changing the underlying serialization format.

Maybe it is enough if `load()` looks into the file extension (or magic number a la `file` command) to see which type of file is and deserialize accordingly, and then we always store the file format version as an attribute.

I'm fine to leave it as it is in the PR and address it later, no need to delay this merge because of this.

<details>
<summary>Click to see ideas that are not really that good I thought before thinking about the simple approach with the extension, but that I didn't want to completely remove in case we find some other issue with the file extension/type approach.</summary>

We could:

1. Add a header to the pickle file with the format version (a bit annoying because for quick debugging it won't be trivial to deserialize the dump using standard tools)
2. Encode the version in the filename (the API is a bit dirtier, as the user should only pass the base name, or we should trust the user to pass a properly encoded file name)

1 feels more robust to me, but 2 feels like it could be more practical in real life because of the easy debugging. It could be a heuristic, load can see if the file ends with, for example `.v1` and then interpret that this dump is version 1, and use the latest version if there is no info (or maybe add more heuristics like the mentioned above to just try to deserialize and fall back to another format if it fails).

You would probably need to append the version to the `path` passed in the constructor too I guess, because asking the user to select the format doesn't look like it makes sense.

</details>

leandro-lucarella-frequenz (commented):
The first commit looks like it could use some splitting (there seem to be some docs and error description improvements there), and maybe a bit more explaining in the commit message why those changes were necessary or what are they good for, as there seem to be a few unrelated changes (like the `0` to `0.0`, the function -> lambda, overloads, import changes, self.buffer[] vs self[]).

src/frequenz/sdk/timeseries/_serializable_ringbuffer.py
mathias-baumann-frequenz: I wasn't sure a specific `TypeVar` instance can or should be reused for more than one class

benchmarks/timeseries/serializable_ringbuffer.py
mathias-baumann-frequenz: I disagree. If GPT generates specific code to my instructions with context I have given then I don't see any danger of any copyright related code, certainly not for snippets like this and of this size.

mathias-baumann-frequenz: And

```
    print(
        "Avg time for Pickle dump/load:  "
        f"{benchmark_serialization(ringbuffer, ITERATIONS)}s"
    )
```

Does not look specifically more readable to me...

src/frequenz/sdk/timeseries/_serializable_ringbuffer.py
mathias-baumann-frequenz: While performance is important, I think it's not really an issue because:

* we're loading 4 million entries in 0.01 second, we are very fast.
* Pickle would just fail pretty early if it's not a pickle format
* we can try loading the new format first instead of v1-pickle and optimially that format allows us to check the version using the first few bytes.
* It would be a temporary state until we update all files and then we can just remove the old version.

src/frequenz/sdk/timeseries/_ringbuffer.py
mathias-baumann-frequenz: I think the nan was simply an auto-import

mathias-baumann-frequenz:
> as there seem to be a few unrelated changes (like the 0 to 0.0,

Why do you specficially think this type-change is unrelated in a commit titled "handle types better" though? :P

> the function -> lambda, overloads,

Same here, that is a direct consequence/enhancement of the changed type-handling, no it even _is_ the changed typehandling really. Fits exactly with the commit title.

> import changes, 

Well, only the np change is unrelated I suppose. The rest is strongly coupled with the "handling types better" topic.

> self.buffer[] vs self[]).

Same here, with the better type handling, we can use a shorter notation.


That leaves a minor str change in the error description and in the docu that are unrelated.
If you want I can split those two out.. everything else is very much related to the commit title.

the np change is reverted anyway based on the feedback here.

I can modify the message to include a bit more details..

benchmarks/timeseries/serializable_ringbuffer.py
leandro-lucarella-frequenz: But IMHO:
```py
elapsed_secs = benchmark_serialization(ringbuffer, ITERATIONS)
print(f"Avg time for Pickle dump/load: {elapsed_secs}s")
```
Does look more readable. That said, I don't think this minor formatting issue is worth the time of the discussion (except for the more general GPT part maybe). If you are very convinced this is good enough, leave it as is.

About if to include or not code produced by GPT I won't give an opinion because it is a complex topic that is ongoing and nobody has a clear answer to that question yet. To be on the safe side I would also try to keep non-trivial code generated by GPT out.

tests/timeseries/test_serializable_ringbuffer.py
leandro-lucarella-frequenz: Ah, yeah, those magical arguments passed by `pytest` are really convenient but the fact that you don't need to import anything to use them makes it quite hard to figure out the type.

src/frequenz/sdk/timeseries/_serializable_ringbuffer.py
leandro-lucarella-frequenz: For me it is also weird, and when a `TypeVar` is trivial, I prefer to define one per module (you can reuse it all you want though). But I guess when you restrict the types it can match like this it might make sense to re-use it, specially if both modules are tied together already.

leandro-lucarella-frequenz (approved):
> Why do you specficially think this type-change is unrelated in a commit titled "handle types better" though? :P

What I mean is it makes harder to review and understand the changes, I really wondered why using `0.0` over `0` was needed or better.

For reviewing purposes I would have been ideal that every of this changes are in a separate commit:

* Specify the element type for the container as well.
* Rename the typevar container to FloatArray
* Add overloads for set/get items
* Use shorter notation now that set/get works type-compatible
* Use numpy.typing to start pulling out the full library

(of course now is not worth splitting further and I understand sometimes changes are interrelated and hard to split, so all good, just clarifying why I asked for this :)

>  If you want I can split those two out.. everything else is very much related to the commit title.

This reminds me of early DMD commits saying "More stuff" or "Improve backend" if you were lucky. Of course all changes are enclosed in the commit message, but that doesn't help the reviewing process at all. I guess you get my point :)

Again I understand it is a balance, and sometimes is not worth the extra work of splitting if changes were done all together, and for me the important thing is to think about **what makes reviewing easier**.

For example in this case sneaking the updated docs and error string in the same commit is fine, even maybe the relative import. They are all trivial one liners. But the improvement of the type system and how all that changes are interrelated (or why they were done) are harder to understand in one single commit (and in that case, adding more noise with unrelated import/docs/error updates makes it even more confusing).

PR #168, opened 2023-01-18
Move code in `util` package to `_internal`

shsms:
There was just a single function `cancel_and_await` that was only for internal use.


matthias-wende-frequenz (dismissed):

src/frequenz/sdk/_internal/__init__.py
```diff
@@ -3,7 +3,8 @@
 
 """Utility types and functions for internal use."""
 
+from .asyncio import cancel_and_await
 from .singleton_meta import SingletonMeta
 
 # Explicitly declare the public API.
-__all__ = ["SingletonMeta"]
+__all__ = ["SingletonMeta", "cancel_and_await"]
```
ela-kotulska-frequenz: This module has generic name `_internal`. 

Because of that separation into submodule `asyncio` is fine.

shsms: it is in a submodule already.  are you saying we shouldn't export reexport the function from `_internal`?

shsms: I've updated the PR with that assumption.

ela-kotulska-frequenz (approved):

src/frequenz/sdk/_internal/__init__.py
ela-kotulska-frequenz: Yes that what I meant :)

leandro-lucarella-frequenz:
The idea was that this was public as a way to provide SDK users with useful tools, I really don't see why we have to keep this internal. For example `await_and_cancel()` could be useful for the channels repo. Of course we can't use it there because it would be a circular dependency, but I thought we might want to extract the `util` package and make it a separate library like we did with channels.

For example, I'm looking into https://github.com/frequenz-floss/frequenz-sdk-python/discussions/197, and that's another clear candidate for this, and can be helpful for channels too, as we also have this kind of objects there (`Merge`, `MergeNamed` and `Select`).

If there is no big opposition, I will move the package back to being public, and add the *background service* class in `util.asyncio` too, I think it fits there.

Issue #90, opened 2022-11-29
Rename `.microgrid.Microgrid` to `MicrogridConnectionManager`


leandro-lucarella-frequenz:
### What's needed?

We need a better name for `Microgrid` (previously `MicrogridApi`). Even when users shouldn't use the type directly, we need a less vague name to talk about it internally.

### Proposed solution

@sahas-subramanian-frequenz suggested to name it `MicrogridConnectionManager`, as at the end it is managing the connection to the microgrid (updating the connection if there are changes in the connection parameters and refreshing the component graph too).

We should make sure the documentation makes it clear that users are not supposed to instantiate this class themselves and update it if that's not the case.

### Alternatives and workarounds

An alternative would be to just split the class in 2 and have a `MicrogridApiManager` and a `ComponentGraphManager`, so the name is less weird as it not grouping 2 things that are not fully related (the component graph can change even if the connection parameters don't change, so `MicrogridConnectionManager` is still not 100% accurate).

### Additional context

Even when the users should use the type, we need to keep the type public so it appears in the documentation, so users can know what they can do with the object returned by `microgrid.get()`.

PR #169, opened 2023-01-19
Add mock microgrid implementation

ela-kotulska-frequenz:


tests/utils/mock_microgrid.py
```diff
@@ -0,0 +1,207 @@
...
+from frequenz.sdk.microgrid.client import Connection
+from frequenz.sdk.microgrid.component import (
+    BatteryData,
+    Component,
+    ComponentCategory,
+    InverterData,
+)
+
+
+class MockMicrogridClient:
```
shsms, line 24: Could you add a method `initialize` that creates a wrapper `Microgrid` and puts it in `sdk.microgrid._microgrid._MICROGRID`,  so that actors using `get()` will get the mock as well?

(I think I'll pick up https://github.com/frequenz-floss/frequenz-sdk-python/issues/90 immediately after this PR.  I don't want to have to type that ridiculous name ever again 😆 )

tests/utils/mock_microgrid.py
```diff
@@ -0,0 +1,207 @@
...
+            data: Data to be send
+
+        Returns:
+            Whether data was send, or not.
+        """
+        cid = data.component_id
+        if isinstance(data, BatteryData):
+            return await self._battery_data_senders[cid].send(data)
+
+        return await self._inverter_data_senders[cid].send(data)
```
shsms: could you change to:
``` python
if BatteryData:
    send
elif InverterData:
    send
else:
    raise
```

and update function signature to `data: ComponentData`?  Then we won't have a large param type as we add more types.

shsms (commented):
This is very useful, thanks Ela.  Just a couple of comments below.

tests/utils/mock_microgrid.py
ela-kotulska-frequenz: Good idea!
Done but in a little different way. (last commit)
Because Microgrid has `async def _initialize` method. And I can't force MagicMock to return async from one of its arguments....
I am sure it is possible and I can do it if necessary. But mine solution should be enough.
Please let me know if that works for you :)

tests/utils/mock_microgrid.py
ela-kotulska-frequenz: Done (last commit)

tests/utils/mock_microgrid.py
shsms: Oh I didn't realize that you already have a `Microgrid` instance.

I'm starting to think that it is not a good idea at all to put private (or even public) members inside double quotes,  because they will make it impossible for our tooling to detect bugs,  when these mocked names change.

Can you instead just assign `..._MICROGRID = self.mock_microgrid` inside the new `initialize` method you added?  And maybe a `uninitialize` method,  in `__del__` of MockMicrogridClient to set it to None.

Or you can remove the `initialize` method all together.  Now that I know there is a `Microgrid` instance, I can assign it to `_MICROGRID` directly before my tests as well.

ela-kotulska-frequenz: > Can you instead just assign ..._MICROGRID = self.mock_microgrid inside the new initialize method you added

Done ( in a minute )

> And maybe a uninitialize method, in __del__ of MockMicrogridClient to set it to None.

I think it is not needed. I don't override it. I mock it with pytest mocker. So it won't be problem in the next test.

> Or you can remove the initialize method all together.

Actually I think it is good idea to have it here. User can choose to not call it. And it simplifies test a little.

shsms: > Can you instead just assign ..._MICROGRID = self.mock_microgrid inside the new initialize method you added

when I said assign,  I meant directly assign,  with an `=` assignment operator,  and not with a mocker. :-)  If you directly assign,  you don't have to put code inside double quotes,  and our tooling can identify it as such.

shsms: didn't know that the mocker will make sure the specified variable names exist.  that was my only concern,  so please ignore what I was saying above. :-)

sahas-subramanian-frequenz (approved):

PR #171, opened 2023-01-23
Battery status with channel communication

ela-kotulska-frequenz:


src/frequenz/sdk/actor/power_distributing/_batteries_status.py
```diff
@@ -101,29 +98,25 @@ def get_working_batteries(self, battery_ids: Set[int]) -> Set[int]:
         )
         return uncertain
 
-    def update_last_request_status(self, result: Result):
+    def update_request_response(
+        self, succeed_batteries: Set[int], failed_batteries: Set[int]
+    ):
```
shsms: because this is not connected to a `Result` object anymore,  it can be renamed.  Possibly to `update_status`, `update_battery_status`, etc?  This is already inside `AllBatteryStatus`,  so maybe just `_update_from_ids` might also work.  Also, does this need to be a public method?

ela-kotulska-frequenz:
Unit tests are waiting for: https://github.com/frequenz-floss/frequenz-channels-python/pull/68

src/frequenz/sdk/microgrid/_battery/_status.py
```diff
@@ -33,17 +33,20 @@
 class BatteryStatus(Enum):
     """Tells if battery is can be used."""
 
-    # Component it is not working.
     NOT_WORKING = 0
-    # Component should work, but it failed in last request. It is blocked for few
-    # seconds and it is not recommended to use it until it is necessary.
+    """Component it is not working and should not be used"""
```
shsms:
``` suggestion
    """Component is not working and should not be used."""
```

src/frequenz/sdk/microgrid/_battery/_status.py
```diff
@@ -33,17 +33,20 @@
...
+
     UNCERTAIN = 1
-    # Component is working
+    """Component should work, but it failed in last request. It is blocked for few
+    seconds and it is not recommended to use it until it is necessary.
+    """
```
shsms:
``` suggestion
    """Component should work, but it failed in last request. It is blocked for few
    seconds and it is not recommended to use it unless it is necessary.
    """
```

src/frequenz/sdk/microgrid/_battery/_status.py
```diff
@@ -45,10 +40,25 @@ class BatteryStatus(Enum):
...
+
+T = TypeVar("T")
+
+
 @dataclass
 class _ComponentData(Generic[T]):
     component_id: int
-    receiver: Peekable[T]
+    last_msg_timestamp: datetime = datetime.now(tz=timezone.utc)
+    last_msg_correct: bool = False
```
shsms: I guess you can now rename this class to `_ComponentStreamStatus`?  Also it is not generic anymore.

shsms (commented):
I've checked the first 4 commits later (and one thing with the big commit).

I'll continue tomorrow.

src/frequenz/sdk/actor/power_distributing/_batteries_status.py
```diff
@@ -6,33 +6,76 @@
...
+        Args:
+            batteries: Set of batteries
+
+        Returns:
+            Subset with working batteries.
+        """
+        working = self.working & batteries
+        if len(working) > 0:
+            return working
+        return self.uncertain & batteries
```
shsms: I guess you are already changing these into explicit function calls to `intersection`

src/frequenz/sdk/actor/power_distributing/_batteries_status.py
```diff
@@ -6,33 +6,76 @@
...
+        self.battery_id: int = battery_id
+
+        self.name: str = f"battery-{self.battery_id}-status"
+        channel = Broadcast[BatteryStatus](self.name)
+
+        # We need only latest status, so maxsize=1
+        receiver_name = f"{self.name}-receiver"
+        self.receiver = channel.new_receiver(name=receiver_name, maxsize=1)
+
+        self.sender = channel.new_sender()
```
shsms: Based on how this is getting used, I think this should be a function instead,  with the signature:

``` python
def _create_battery_status_channel(
    battery_id: int,
) -> tuple[Sender[BatteryStatus], Receiver[BatteryStatus]]:
```

Having a class to own receivers that are accessed from somewhere else is a pattern that can be misused.  We should try to make sure that `Receiver` objects are always locally owned,  and not through a proxy, but directly.

Also,  I think it is best to avoid naming classes as `*Channel` or `*Receiver`, etc. in the sdk, unless they are actual channels or receivers deriving from the channels package.

src/frequenz/sdk/microgrid/_battery/_status.py
```diff
@@ -45,10 +40,25 @@ class BatteryStatus(Enum):
     """Component is working"""
 
 
+@dataclass
+class RequestResult:
```
shsms: Does this have to go to `microgrid/_battery`?  I think this is specific to the power distributor.

src/frequenz/sdk/microgrid/_battery/_status.py
```diff
@@ -113,12 +123,10 @@ def is_blocked(self) -> bool:
         return self.blocked_until > datetime.now(tz=timezone.utc)
 
 
-class BatteryStatusTracker(AsyncConstructible):
+class BatteryStatusTracker:
```
shsms: This too,  I think should sit outside the `microgrid` package.  The `microgrid` package is supposed to be a thin wrapper providing access to data from the microgrid and for sending commands to the microgrid.

But this class is really stateful tracking logic, which should ideally be part of a higher level package,  possibly the `actor/power_distributing`.   I see that this code was already merged,  so I can move out later also, if you prefer.

src/frequenz/sdk/microgrid/_battery/_status.py
```diff
@@ -113,12 +123,10 @@ def is_blocked(self) -> bool:
...
     """Class for tracking if battery is working.
 
-    To create an instance of this class you should use `async_new` class method.
-    Standard constructor (__init__) is not supported and using it will raise
-    `NotSyncConstructible` error.
+    Status in send only when it change.
```
shsms:
``` suggestion
    Status updates are sent out only when there is a status change.
```

src/frequenz/sdk/microgrid/_battery/_status.py
```diff
@@ -137,18 +145,17 @@ class BatteryStatusTracker(AsyncConstructible):
...
-    _blocking_status: _BlockingStatus
-    _battery: _ComponentData[BatteryData]
-    _inverter: _ComponentData[InverterData]
-
-    @classmethod
-    async def async_new(
-        cls, battery_id: int, max_data_age_sec: float, max_blocking_duration_sec: float
-    ) -> BatteryStatusTracker:
+    # Instance attributes defined outside __init__
+    _select: Optional[Select]
```
shsms: You don't need this line.  You can just do this inside the `__init__` method, like you're already doing for other variables:

``` python
self._select: Optional[Select] = None
```

src/frequenz/sdk/actor/power_distributing/_batteries_status.py
```diff
@@ -6,33 +6,76 @@
...
+
+
+class BatteriesStatus:
```
shsms: I guess you're going to rename this to `AllBatteryStatus` as discussed, and the file accordingly.

src/frequenz/sdk/actor/power_distributing/_batteries_status.py
```diff
@@ -46,77 +89,98 @@ async def async_new(
...
+        # working.
+        self._current_status = Status(working=set(), uncertain=set())
+
+        # Channel for sending results of requests to the batteries
+        request_result_channel = Broadcast[RequestResult]("battery_request_status")
+        self._request_result_sender = request_result_channel.new_sender()
+
+        self._batteries: Dict[str, BatteryStatusTracker] = {}
+
+        # Receivers are needed to create Select object
```
shsms:
``` suggestion
        # Receivers for individual battery statuses are needed to create a `MergeNamed` object
```

src/frequenz/sdk/microgrid/_battery/_status.py
shsms: can we also rename this to `SetPowerResult` or something?  `RequestResult` sounds too confusing.

src/frequenz/sdk/actor/power_distributing/_batteries_status.py
```diff
@@ -46,77 +89,98 @@ async def async_new(
...
 
-        Args:
-            battery_ids: batteries ids
+    async def _run(self) -> None:
+        """Start tracking batteries status."""
+        while True:
+            try:
+                await self._update_status(self._battery_status_channel)
+            except Exception as err:  # pylint: disable=broad-except
+                _logger.error("BatteriesStatus failed with error: %s. Restart.", err)
```
shsms:
``` suggestion
                _logger.error("BatteriesStatus failed with error: %s. Restarting.", err)

File: src/frequenz/sdk/actor/power_distributing/_batteries_status.py [hunk 14]
```diff
@@ -46,77 +89,98 @@ async def async_new(
...
 
-        Raises:
-            KeyError: If any battery in the given batteries is not in the pool.
+    async def _update_status(self, status_channel: MergeNamed[BatteryStatus]) -> None:
```
shsms: maybe more readable to decompose to a `Receiver`, like below,  but it is ok to keep it as it is also.
``` suggestion
    async def _update_status(self, status_channel: Receiver[tuple[str, BatteryStatus]]) -> None:
```

src/frequenz/sdk/microgrid/_battery/_status.py
```diff
@@ -197,54 +201,120 @@ def battery_id(self) -> int:
         """
         return self._battery.component_id
 
-    def get_status(self) -> BatteryStatus:
-        """Return status of the battery.
+    async def stop(self) -> None:
+        """Stop tracking battery status."""
+        if self._select is not None:
```
shsms: you should probably also `cancel_and_await(self._task)` here, before killing the `Select`.

src/frequenz/sdk/actor/power_distributing/_batteries_status.py
ela-kotulska-frequenz: Done. I decided on : `update_status`

src/frequenz/sdk/microgrid/_battery/_status.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/microgrid/_battery/_status.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/actor/power_distributing/_batteries_status.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/actor/power_distributing/_batteries_status.py
ela-kotulska-frequenz: I changed it to dataclass, renamed to _BatteryStatusChannelHelper, and I described that the channels has only 1 receiver.

It is just internal, helper class. And it is very simple, so I decided to leave it as class.

Returning tuple doesn't work, because I also need channel name as string. So I would return Tuple with 3 elements and that is a lot.
```
def _create_battery_status_channel(
    battery_id: int,
) -> tuple[Sender[BatteryStatus], Receiver[BatteryStatus]]:
```

src/frequenz/sdk/actor/power_distributing/_batteries_status.py
shsms: does this have to be a public method?

src/frequenz/sdk/actor/power_distributing/_batteries_status.py
shsms: I'd say it would be more readable if this gets inlined,  but I guess it is just a matter of style.

src/frequenz/sdk/actor/power_distributing/_batteries_status.py
ela-kotulska-frequenz: Yes, because PowerDistributingActor actor uses it.

src/frequenz/sdk/microgrid/_battery/_status.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/microgrid/_battery/_status.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/actor/power_distributing/_batteries_status.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/actor/power_distributing/_batteries_status.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/actor/power_distributing/_batteries_status.py
ela-kotulska-frequenz: I would prefer to leave it `MergeNamed`.
It is clear and shorter.

src/frequenz/sdk/microgrid/_battery/_status.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/actor/power_distributing/_batteries_status.py
shsms: btw,  why not just use `self._battery_status_channel` and not pass anything at all?

src/frequenz/sdk/microgrid/_battery/_status.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/microgrid/_battery/_status.py
ela-kotulska-frequenz: I moved whole file to frequez.sdk.actor.power_distributing.
Renamed RequestResult to SetPowerResult.

src/frequenz/sdk/microgrid/_battery/_status.py
ela-kotulska-frequenz: Moved to `actor/power_distributing`

src/frequenz/sdk/actor/power_distributing/_batteries_status.py
ela-kotulska-frequenz: I renamed to `BatteryPoolStatus` once again. 
Because it has battery_ids as argument, so it can be created for pool, too.
Also `BatteryPoolStatus` sounds better after all.

sahas-subramanian-frequenz (approved):

PR #180, opened 2023-01-25
Use MockMicrogridClient instead of a MockMicrogridServer in LogicalMeter tests

### LINKED ISSUE #177: Replace mocked microgrid server with the new microgrid client mock in logical meter tests
### What's needed?

The current tests for the logical meter create a mock server and connect to it for data.  This is causing a lot of spurious test failures because of timing/sync issues with the mock server.

This can be bypassed by using the mock client introduced in https://github.com/frequenz-floss/frequenz-sdk-python/pull/169.

### Proposed solution

_No response_

### Use cases

_No response_

_No response_

### Additional context

_No response_

shsms:
Closes #177

### DIFF
--- a/tests/timeseries/test_logical_meter.py
+++ b/tests/timeseries/test_logical_meter.py
@@ -6,6 +6,8 @@
 from __future__ import annotations
 
 from datetime import datetime
+from math import isclose
+
 from frequenz.channels import Receiver, Sender
 from pytest_mock import MockerFixture
 
@@ -53,6 +55,13 @@ class TestLogicalMeter:
             metric_id,
         )
         # pylint: enable=protected-access
+
+    def _equal_float_lists(self, list1: list[float], list2: list[float]) -> bool:
+        return (
+            len(list1) > 0
+            and len(list1) == len(list2)
+            and all(isclose(v1, v2) for v1, v2 in zip(list1, list2))
+        )
 
     async def _synchronize_receivers(
         self,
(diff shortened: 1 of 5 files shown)


tests/timeseries/test_logical_meter.py
```diff
@@ -68,7 +68,7 @@ async def _synchronize_receivers(
         receivers: list[FormulaReceiver | FormulaReceiver3Phase | Receiver[Sample]],
     ) -> None:
         by_ts: dict[
-            datetime, List[FormulaReceiver | FormulaReceiver3Phase | Receiver[Sample]]
+            datetime, list[FormulaReceiver | FormulaReceiver3Phase | Receiver[Sample]]
         ] = {}
         for recv in receivers:
```
ela-kotulska-frequenz, line 73: This commit should probably be squash with previous one.
Otherwise tests don't pass.

shsms: Because we have merge commits now,  this is technically not a problem anymore.  And I've been trying to not follow that rule too strictly,  especially when commits are too long already,  or when it helps with sequential readability.

The previous 3-phase PR also has this problem for example,  with two large commits,  and I think I've done this in other PRs also.

The main reason we were doing this was so that we can use `git bisect` to identify when issues were introduced.  And when we were using rebase merges earlier,  we needed all commits to pass tests,  else it wasn't possible to bisect.

But now,  we have merge commits,  and we can bisect just the merge commits.  So we might as well prioritize readability in individual commits.

shsms: (assuming I understand your comment correctly. please correct me if you're talking about something else)

ela-kotulska-frequenz (approved):

tests/timeseries/test_logical_meter.py
ela-kotulska-frequenz: Ok, I didn't know. Thanks.

PR #186, opened 2023-01-27
Use `Sample` type with the RingBuffer

### LINKED ISSUE #165: Use Sample type in Ringbuffer
### What's needed?

The `OrderedRingBuffer`s update method takes a timestamp and a value as parameters.
The default use case for this class is to store data coming from the logical meter, which sends data of type `Sample` which members are a timestamp and a value.

### Proposed solution

Move he ringbuffer to `timeseries` and use the `Sample` class.

### Use cases

_No response_

Move the `Sample` class from the `timestamp` path to e.g. `util` and use it in the RingBuffer.

### Additional context

_No response_

mathias-baumann-frequenz:
fixes #165

### DIFF
--- a/benchmarks/timeseries/benchmark_ringbuffer.py
+++ b/benchmarks/timeseries/benchmark_ringbuffer.py
@@ -25,6 +25,7 @@ def fill_buffer(
     """Fill the given buffer up to the given amount of days, one sample per minute."""
     random.seed(0)
     basetime = datetime(2022, 1, 1)
+    print("..filling", end="", flush=True)
 
     for day in range(days):
         # Push in random order
@@ -35,11 +36,8 @@ def fill_buffer(
             )
 
 
-def test_days(days: int, buffer: OrderedRingBuffer[int, Any]) -> None:
-    """Fills a buffer completely up and then gets the data for each of the 29 days."""
-    print(".", end="", flush=True)
-
-    fill_buffer(days, buffer, int)
+def test_days(days: int, buffer: OrderedRingBuffer[T, Any]) -> None:
+    """Gets the data for each of the 29 days."""
 
     basetime = datetime(2022, 1, 1)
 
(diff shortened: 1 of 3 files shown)


benchmarks/timeseries/benchmark_ringbuffer.py
```diff
@@ -117,25 +116,26 @@ def main() -> None:
...
+    Array: ..........
+    List:  ..........
     Filling 29 days and running average & mean on every day:
-    Array: 0.09842290654996759 seconds
-    List:  0.1316629376997298 seconds
-    Diff:  -0.03324003114976222
+            Array: 7.6797407130012285 seconds
+            List:  8.115826181398006 seconds
+            Diff:  -0.43608546839677764
+
```
shsms: Why do you think this is?  I can't believe creating a new `Sample` object for every message is making the benchmark 70-80x longer.

Maybe we should also try benchmarking the filling times and average/mean times separately?

benchmarks/timeseries/benchmark_ringbuffer.py
```diff
@@ -30,12 +29,14 @@ def fill_buffer(
         # Push in random order
         for i in random.sample(range(MINUTES_IN_A_DAY), MINUTES_IN_A_DAY):
             buffer.update(
-                basetime + timedelta(days=day, minutes=i, seconds=i % 3),
-                element_type(i),
+                Sample(
+                    basetime + timedelta(days=day, minutes=i, seconds=i % 3),
+                    element_type(i),
```
shsms: I guess only float will be supported from now on,  so we don't need the element_type.

benchmarks/timeseries/benchmark_ringbuffer.py
mathias-baumann-frequenz: Yah, the previous tests were faster due to missing window not down completely or properly I think, so I don't think we can compare that

src/frequenz/sdk/timeseries/_ringbuffer.py
```diff
@@ -105,28 +106,25 @@ def maxlen(self) -> int:
         """
         return len(self._buffer)
 
-    def update(self, timestamp: datetime, value: T, missing: bool = False) -> None:
+    def update(self, sample: Sample) -> None:
```
matthias-wende-frequenz, line 109: Why are you removing the missing flag. That change seems unrelated.

matthias-wende-frequenz: Actually I liked that missing flag, because then the user can decide what's defined as missing. The NaN check is what we have in the `MovingWindow`.

src/frequenz/sdk/timeseries/_ringbuffer.py
```diff
@@ -140,9 +138,10 @@ def update(self, timestamp: datetime, value: T, missing: bool = False) -> None:
         self._datetime_oldest = self._datetime_newest - self._time_range
 
         # Update data
+        value: float = math.nan if sample.value is None else sample.value
         self._buffer[self.datetime_to_index(timestamp)] = value
 
-        self._update_gaps(timestamp, prev_newest, missing)
+        self._update_gaps(timestamp, prev_newest, sample.value is None)
```
matthias-wende-frequenz, line 144: We should check if it's nan, since that's what the logical meter returns. The resampler though returns None. I think that's another argument to stick to the missing semantic.

matthias-wende-frequenz (commented):
I feel we shouldn't remove the flag to indicate a missing sample.

src/frequenz/sdk/timeseries/_ringbuffer.py
mathias-baumann-frequenz: Because `Sample` has an `optional`

src/frequenz/sdk/timeseries/_ringbuffer.py
shsms: The LogicalMeter returns `None`, not `NaN`.  It uses `NaN`s just internally.

matthias-wende-frequenz (approved):

PR #190, opened 2023-01-31
Add a Moving Window Class

### LINKED ISSUE #113: Define a module for windowing streaming data
### What's needed?

We need to access a window of data that is received from the `LogicalMeter` and can be used by downstream users, usually trained machine learning models.

This should use a ring buffer as an underlying data structure

The anticipated feature are

* updates the data regularly
* can access data by time windows
* can create an **ordered copy** of a slice of the ring buffer as numpy array
* OPTIONAL: can return a **resampled** ordered copy of a slice of the ring buffer

The last feature is marked as optional because it's possible to create two windows with the same data with different sampling rates without that feature.

Additional Features are

* aggregations/operations on windows like mean and standard deviation shared with the [TimeseriesAvg](https://github.com/frequenz-floss/frequenz-sdk-python/issues/120) (for further ideas [see also](https://nightlies.apache.org/flink/flink-docs-master/docs/dev/datastream/operators/windows/)

### Proposed solution

Essentially this is the user facing interface to the ring buffer taking care of dealing with all kinds time related and providing bounded and ordered data batches to the user when requested.

Furthermore it's supposed the update the ring buffer on new incoming samples from the `LogicalMeter`.

### Use cases

This will be used in the forecasting pipeline as the central data storage for

* storing time series data that can be piped into ml predictions,
* calculating features as
	* daily averages
	* energy consumption/production in a certain time window
	* mean and standard deviation
	* polynomial coefficients (for power curves)

This could, instead of being a wrapper around, as well directly end up in the ring buffer.

In the past pandas dataframes have been used. It turned out as too inefficient due to

* memory allocation when operating (e.g. (joining, resampling,...),
* creation of new dataframes on new incoming data.

matthias-wende-frequenz:
A data window that moves with the latest data points of a data stream.

When dealing with machine learning models, often the input features
are derived from timeseries data or are timeseries itself.

In order to obtain a timeseries as a data batch from streamed data
one can define a window that is essentially.

This approach creates a window that moves with the data flow by
utilizing a ringbuffer such that copying of buffer and therefore
memory allocations are minimized.

### DIFF
--- a/src/frequenz/sdk/timeseries/_moving_window.py
+++ b/src/frequenz/sdk/timeseries/_moving_window.py
@@ -1,0 +1,189 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""A data window that moves with the latest datapoints of a data stream."""
+
+from __future__ import annotations
+
+import asyncio
+import logging
+from collections.abc import Sequence
+from datetime import datetime, timedelta
+
+import numpy as np
+from frequenz.channels import Receiver
+from numpy.typing import ArrayLike
+
+from .._internal.asyncio import cancel_and_await
+from . import Sample
+from ._ringbuffer import OrderedRingBuffer
+
+log = logging.getLogger(__name__)
+
+
+class MovingWindow(Sequence):
+    """
+    A data window that moves with the latest datapoints of a data stream.
+
+    After initialization the `MovingWindow` can be accessed by an integer
+    index or a timestamp. A sub window can be accessed by using a slice of integers
+    integers or timestamps.
+
+    Note that a numpy ndarray is returned and thus users can use
+    numpys operations directly on a window.
+
+    The window uses an ringbuffer for storage and the first element is aligned to
+    a fixed defined point in time. Since the moving nature of the window, the
+    date of the first and the last element are constantly changing and therefore
+    the point in time that defines the alignment can be outside of the time window.
+    Modulo arithmetic is used to move the `window_alignment` timestamp into the
+    latest window.
+    If for example the `window_alignment` parameter is set to `datetime(1, 1, 1)`
+    and the window size is bigger than one day then the first element will always
+    be aligned to the midnight. For further information see also the
+    [`OrderedRingBuffer`][frequenz.sdk.timeseries._ringbuffer.OrderedRingBuffer]
+    documentation.
+
+
+    **Example1** (calculating the mean of a time interval):
+
+    ```
+    window = MovingWindow(size=100, resampled_data_recv=resampled_data_recv)
+
+    time_start = datetime.now()
+    time_end = time_start + timedelta(minutes=5)
+
+    # ... wait for 5 minutes until the buffer is filled
+    await asyncio.sleep(5)
+
+    # return an numpy array from the window
+    a = window[time_start:time_end]
+    # and use it to for example calculate the mean
+    mean = a.mean()
+    '''
+
+    **Example2** (create a polars data frame from a `MovingWindow`):
+
+    ```
+    import polars as pl
+
+    # create a window that stores two days of data
+    # starting at 1.1.23 with samplerate=1
+    window = MovingWindow(size = (60 * 60 * 24 * 2), sample_receiver)
+
+    # wait for one full day until the buffer is filled
+    asyncio.sleep(60*60*24)
+
+    # create a polars series with one full day of data
+    time_start = datetime(2023, 1, 1)
+    time_end = datetime(2023, 1, 2)
+    s = pl.Series("Jan_1", mv[time_start:time_end])
+    ```
+    """
+
+    def __init__(
+        self,
+        size: int,
+        resampled_data_recv: Receiver[Sample],
+        sampling_period: timedelta,
+        window_alignment: datetime = datetime(1, 1, 1),
+    ) -> None:
+        """
+        Initialize the MovingWindow.
+
+        This method creates the underlying ringbuffer and starts a
+        new task that updates the ringbuffer with new incoming samples.
+        The task stops running only if the channel receiver is closed.
+
+        Args:
+            size: The number of elements that are stored.
+            resampled_data_recv: A receiver that delivers samples with a
+                given sampling period.
+            sampling_period: The sampling period.
+            window_alignment: A datetime object that defines a point in time to which
+                the window is aligned to modulo window size.
+                (default is midnight 01.01.01)
+                For further information, consult the class level documentation.
+
+        Raises:
+            asyncio.CancelledError: when the task gets cancelled.
+        """
+        self._resampled_data_recv = resampled_data_recv
+        self._buffer = OrderedRingBuffer(
+            np.empty(shape=size, dtype=float),
+            sampling_period=sampling_period,
+            time_index_alignment=window_alignment,
+        )
+        self._copy_buffer = False
+        self._update_window_task: asyncio.Task = asyncio.create_task(self._run_impl())
+        log.debug("Cancelling MovingWindow task: %s", __name__)
+
+    async def _run_impl(self) -> None:
+        """Awaits samples from the receiver and updates the underlying ringbuffer."""
+        try:
+            async for sample in self._resampled_data_recv:
+                log.debug("Received new sample: %s", sample)
+                self._buffer.update(sample)
+        except asyncio.CancelledError:
+            log.info("MovingWindow task has been cancelled.")
+            return
+
+        log.error("Channel has been closed")
+
+    async def stop(self) -> None:
+        """Cancel the running task and stop the MovingWindow."""
+        await cancel_and_await(self._update_window_task)
+
+    def __len__(self) -> int:
+        """
+        Return the size of the `MovingWindow`s underlying buffer.
+
+        Returns:
+            The size of the `MovingWindow`.
+        """
+        return len(self._buffer)
+
+    def __getitem__(self, key: int | datetime | slice) -> float | ArrayLike:
+        """
+        Return a sub window of the `MovingWindow`.
+
+        The `MovingWindow` is accessed either by timestamp or by index
+        or by a slice of timestamps or integers.
+
+        * If the key is an integer, the float value of that key
+          at the given position is returned.
+        * If the key is a datetime object, the float value of that key
+          that corresponds to the timestamp is returned.
+        * If the key is a slice of timestamps or integers, an ndarray is returned,
+          where the bounds correspond to the slice bounds.
+          Note that a half open interval, which is open at the end, is returned.
+
+        Args:
+            key: Either an integer or a timestamp or a slice of timestamps or integers.
+
+        Raises:
+            IndexError: when requesting an out of range timestamp or index
+            TypeError: when the key is not a datetime or slice object.
+
+        Returns:
+            A float if the key is a number or a timestamp.
+            an numpy array if the key is a slice.
+        """
+        if isinstance(key, slice):
+            log.debug("Returning slice for [%s:%s].", key.start, key.stop)
+            # we are doing runtime typechecks since there is no abstract slice type yet
+            # see also (https://peps.python.org/pep-0696)
+            if isinstance(key.start, datetime) and isinstance(key.stop, datetime):
+                return self._buffer.window(key.start, key.stop, self._copy_buffer)
+            if isinstance(key.start, int) and isinstance(key.stop, int):
+                return self._buffer[key]
+        elif isinstance(key, datetime):
+            log.debug("Returning value at time %s ", key)
+            return self._buffer[self._buffer.datetime_to_index(key)]
+        elif isinstance(key, int):
+            return self._buffer[key]
+
+        raise TypeError(
+            "Key has to be either a timestamp or an integer "
+            "or a slice of timestamps or integers"
+        )
--- a/tests/timeseries/test_moving_window.py
+++ b/tests/timeseries/test_moving_window.py
@@ -1,0 +1,77 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Tests for the moving window."""
+
+import asyncio
+from datetime import datetime, timedelta
+from typing import Sequence, Tuple
+
+import numpy as np
+from frequenz.channels import Broadcast, Sender
+
+from frequenz.sdk.timeseries import Sample
+from frequenz.sdk.timeseries._moving_window import MovingWindow
+
+
+async def push_lm_data(sender: Sender[Sample], test_seq: Sequence[float]) -> None:
+    """
+    Push data in the passed sender to mock `LogicalMeter` behaviour.
+    Starting with the First of January 2023.
+
+    Args:
+        sender: Sender for pushing resampled samples to the `MovingWindow`.
+        test_seq: The Sequence that is pushed into the `MovingWindow`.
+    """
+    start_ts: datetime = datetime(2023, 1, 1)
+    for i, j in zip(test_seq, range(0, len(test_seq))):
+        timestamp = start_ts + timedelta(seconds=j)
+        await sender.send(Sample(timestamp, float(i)))
+
+    await asyncio.sleep(0.0)
+
+
+def init_moving_window(shape: int) -> Tuple[MovingWindow, Sender[Sample]]:
+    """
+    Initialize the moving window with given shape
+
+    Args:
+        shape: The size of the `MovingWindow`
+
+    Returns:
+        tuple[MovingWindow, Sender[Sample]]: A pair of sender and `MovingWindow`.
+    """
+    lm_chan = Broadcast[Sample]("lm_net_power")
+    lm_tx = lm_chan.new_sender()
+    window = MovingWindow(shape, lm_chan.new_receiver(), timedelta(seconds=1))
+    return window, lm_tx
+
+
+async def test_access_window_by_index() -> None:
+    """Test indexing a window by integer index"""
+    window, sender = init_moving_window(1)
+    await push_lm_data(sender, [1])
+    assert np.array_equal(window[0], 1.0)
+
+
+async def test_access_window_by_timestamp() -> None:
+    """Test indexing a window by timestamp"""
+    window, sender = init_moving_window(1)
+    await push_lm_data(sender, [1])
+    assert np.array_equal(window[datetime(2023, 1, 1)], 1.0)
+
+
+async def test_access_window_by_int_slice() -> None:
+    """Test accessing a subwindow with an integer slice"""
+    window, sender = init_moving_window(5)
+    await push_lm_data(sender, range(0, 5))
+    assert np.array_equal(window[3:5], np.array([3.0, 4.0]))
+
+
+async def test_access_window_by_ts_slice() -> None:
+    """Test accessing a subwindow with a timestamp slice"""
+    window, sender = init_moving_window(5)
+    await push_lm_data(sender, range(0, 5))
+    time_start = datetime(2023, 1, 1) + timedelta(seconds=3)
+    time_end = time_start + timedelta(seconds=2)
+    assert np.array_equal(window[time_start:time_end], np.array([3.0, 4.0]))  # type: ignore
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -14,6 +14,7 @@
 
 * A new class `OrderedRingBuffer` is now available, providing a sorted ring buffer of datetime-value pairs with tracking of any values that have not yet been written.
 * Add logical meter formula for EV power.
+* A `MovingWindow` class has been added that consumes a data stream from a logical meter and updates an `OrderedRingBuffer`.
 
 ## Bug Fixes
 
--- a/src/frequenz/sdk/timeseries/_ringbuffer.py
+++ b/src/frequenz/sdk/timeseries/_ringbuffer.py
@@ -28,7 +28,7 @@ class Gap:
     """End of the range, exclusive."""
 
     def contains(self, timestamp: datetime):
-        """Check if a given timestamp is inside this  gap.
+        """Check if a given timestamp is inside this gap.
 
         Args:
             timestamp: Timestamp to check.
@@ -55,8 +55,8 @@ class OrderedRingBuffer(Generic[Container]):
 
         Args:
             buffer: Instance of a buffer container to use internally.
-            sampling_period: Timedelta of the desired resampling period.
-            time_index_alignment: Arbitary point in time used to align
+            sampling_period: Timedelta of the desired sampling period.
+            time_index_alignment: Arbitrary point in time used to align
                 timestamped data with the index position in the buffer.
                 Used to make the data stored in the buffer align with the
                 beginning and end of the buffer borders.


src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -0,0 +1,182 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class MovingWindow(Sequence):
+    """
+    A data window that moves with the latest datapoints of a data stream.
+
+    After inititialization the `MovingWindow` can be accessed by an integer
+    index or a timestamp. A sub window can be accessed by using a slice ofintegers
```
mathias-baumann-frequenz: `ofintegers` => `of integers`

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -0,0 +1,182 @@
...
+    integers or timestamps.
+
+    Note that a numpy ndarrays is returned and thus users can use
```
mathias-baumann-frequenz: `ndarrays` -> `ndarray`

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -0,0 +1,182 @@
...
+        self,
+        size: int,
+        resampled_data_recv: Receiver[Sample],
+        sampling_period: timedelta,
+        window_alignment: datetime = datetime(1, 1, 1),
+    ) -> None:
+        """
+        Initialize the MovingWindow.
+
+        This method starts creates the underlying ringbuffer and starts a
```
mathias-baumann-frequenz: `This method starts creates the` -> `This method creates the`

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -0,0 +1,182 @@
...
+        new task that updates the ringbuffer with new incoming samples.
+        The task stops running only if the channel receiver is closed.
+
+        Args:
+            size: The number of elements that are stored.
+            resampled_data_recv: A receiver that delivers samples with a
+                given sampling_period_s
```
mathias-baumann-frequenz: should this still say `sampling_period_s` (with the `_s` suffix?)

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -0,0 +1,182 @@
...
+            sampling_period: The sampling period.
+            window_alignment: A datatime object that defines a point in time to which
+                the window is aligned too modulo window size.
```
mathias-baumann-frequenz: `too` -> `to`

mathias-baumann-frequenz (dismissed):

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -0,0 +1,182 @@
...
+        self._buffer = OrderedRingBuffer(
+            np.empty(shape=size, dtype=float),
+            sampling_period=sampling_period,
+            time_index_alignment=window_alignment,
+        )
+        self._copy_buffer = False
+        try:
+            _t = asyncio.create_task(self._run_impl())
+            # make the linter happy
+            assert isinstance(_t, asyncio.Task)
```
daniel-zullo-frequenz: I wonder if `_t: asyncio.Task = asyncio.create_task(self._run_impl())` would work and make the linter happy :thinking:

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -0,0 +1,182 @@
...
+from frequenz.sdk.timeseries._ringbuffer import OrderedRingBuffer
+
+log = logging.getLogger(__name__)
+
+
+class MovingWindow(Sequence):
+    """
+    A data window that moves with the latest datapoints of a data stream.
+
+    After inititialization the `MovingWindow` can be accessed by an integer
```
daniel-zullo-frequenz: s/inititialization/initialization

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -0,0 +1,182 @@
...
+        The `MovingWindow` is accessed either by timestamp or by index
+        or by a slice of timestamps or integers.
+
+        * If the key is an integer, the float value of that key
+          at the given position is returned.
+        * If the key is a datetime object, the float value of that key
+          that corresponds to the timestamp is returned.
+        * If the key is a slice of timestamps or integers, an ndarray is returned,
+          where the bounds correspond to the slice bounds.
+          Note that a half open intervall, which is open at the end, is returned.
```
daniel-zullo-frequenz: s/intervall/interval

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -0,0 +1,182 @@
...
+
+        Args:
+            key: Either an integer or a timestam or a timestamps or integers.
```
daniel-zullo-frequenz: There is typo in `timestam` and timestamp itself might be duplicated

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -0,0 +1,182 @@
...
+    async def _run_impl(self) -> None:
+        """Awaits samples from the receiver and updates the underlying ringbuffer."""
+        async for sample in self._resampled_data_recv:
+            log.debug("Received new sample: %s", sample)
+            log.debug("Updating ring buffer with latest sample")
+
+            # missing values coming from the logical meter are marked as NaN
+            missing = False
+            assert isinstance(sample.value, float)
+            if isnan(sample.value):
```
daniel-zullo-frequenz: what about `if isinstance(sample.value, float) and isnan(sample.value):` and remove the assert above?

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -0,0 +1,182 @@
...
+        if isinstance(key, slice):
+            log.debug("Returning slice for [%s:%s].", key.start, key.stop)
+            # we are doing runtime typechecks since there is no abstract slice type yet
+            # see also (https://peps.python.org/pep-0696)
+            if isinstance(key.start, datetime) and isinstance(key.stop, datetime):
+                return self._buffer.window(key.start, key.stop, self._copy_buffer)
+            if isinstance(key.start, int) and isinstance(key.stop, int):
+                window = self._buffer[key]
+                # we need to assert here to make the type checker happy
+                assert isinstance(window, np.ndarray)
```
daniel-zullo-frequenz: Also I wonder if `window: np.ndarray = self._buffer[key]` would work and make the linter happy

tests/timeseries/test_moving_window.py
```diff
@@ -0,0 +1,77 @@
...
+
+async def test_access_window_by_int_slice() -> None:
+    """Test accessing a subwindow with an integer slice"""
+    window, sender = init_moving_window(5)
+    await push_lm_data(sender, range(0, 5))
+    assert np.array_equal(window[3:5], np.array([3.0, 4.0]))
+
+
+async def test_access_window_by_ts_slice() -> None:
+    """Test accessing a subwindow with an integer slice"""
```
daniel-zullo-frequenz: timestamp slice instead?

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -0,0 +1,182 @@
...
+        This method creates the underlying ringbuffer and starts a
+        new task that updates the ringbuffer with new incoming samples.
+        The task stops running only if the channel receiver is closed.
+
+        Args:
+            size: The number of elements that are stored.
+            resampled_data_recv: A receiver that delivers samples with a
+                given sampling period.
+            sampling_period: The sampling period.
+            window_alignment: A datatime object that defines a point in time to which
```
daniel-zullo-frequenz: s/datatime/datetime ?

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -0,0 +1,182 @@
...
+    def __len__(self) -> int:
+        """
+        Return the size of the `MovingWindow`s underlying buffer.
+
+        Returns:
+            The size of the `MovingWindow`.
+        """
+        return len(self._buffer)
+
+    def __getitem__(self, key: (int | datetime) | slice) -> float | ArrayLike:
```
daniel-zullo-frequenz: You might need to use `Union` here instead as only parenthesis are supported in the latest versions of python/typing

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -0,0 +1,182 @@
...
+                return self._buffer.window(key.start, key.stop, self._copy_buffer)
+            if isinstance(key.start, int) and isinstance(key.stop, int):
+                window = self._buffer[key]
+                # we need to assert here to make the type checker happy
+                assert isinstance(window, np.ndarray)
+                return window
+        elif isinstance(key, datetime):
+            log.debug("Returning value at time %s ", key)
+            value = self._buffer[self._buffer.datetime_to_index(key)]
+            return value
```
daniel-zullo-frequenz: here you might want to return directly `self._buffer[self._buffer.datetime_to_index(key)]` instead to avoid declaring `value`

src/frequenz/sdk/timeseries/_moving_window.py
daniel-zullo-frequenz: actually it should be fine to return `self._buffer[key]` directly and remove the assert and `window` declaration

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: Unfortunately not. The linters problem is, that the variable is not accessed, not the type.

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: `... a slice of ...` is missing ;)

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: No I would prefer the above, because the check if the value is float is again a linter thing. But this logic is moved to the ringbuffer anyways and will be deleted with the next rebase.

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: you are right, that seems to work

matthias-wende-frequenz: It works since I changed the return type from `float | np.ndarray` to `float | ArrayLike` some days ago.

src/frequenz/sdk/timeseries/_moving_window.py
shsms: You can use `from __future__ import annotations` to get this to work in python 3.8.

But why not just have individual methods like `get_time_window`, `get_index`, etc?  then we won't have to do all that runtime type-checking.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -0,0 +1,177 @@
...
+        """
+        self._resampled_data_recv = resampled_data_recv
+        self._buffer = OrderedRingBuffer(
+            np.empty(shape=size, dtype=float),
+            sampling_period=sampling_period,
+            time_index_alignment=window_alignment,
+        )
+        self._copy_buffer = False
+        try:
+            _t: asyncio.Task = asyncio.create_task(self._run_impl())
```
shsms: You might have to put this in `self._task`.  `_t` goes out of scope as soon as `__init__` ends.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -0,0 +1,177 @@
...
+
+        Args:
+            key: Either an integer or a timestamp or a slice of timestamps or integers.
+
+        Raises:
+            IndexError: when requesting an out of range timestamp or index
+            TypeError: when the key is not a datetime or slice object.
+
+        Returns:
+            Union(float, ArrayLike): A float of the key is a number or a timestamp.
```
shsms: We are not specifying types for return values in docstrings anymore.

src/frequenz/sdk/timeseries/_moving_window.py
daniel-zullo-frequenz: oh I see. Then the suggestion from Sahas about making it an attribute of the class would do the trick (but also keeping the state of a variable which is not used is not a good practice :thinking:)
Anyway it's fine with me either way

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: you are right and good catch :)

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: I guess I sacrificed speed for usability. But I'd say it's fine in that case. Accessing such a window shouldn't happen too often.

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: with that change we also don't need the assert any longer :partying_face: .

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: > You can use from __future__ import annotations to get this to work in python 3.8.

That doesn't seem to work unfortunately. Will do the long version.

src/frequenz/sdk/timeseries/_moving_window.py
daniel-zullo-frequenz: oh I see, then it is fine as it is...

src/frequenz/sdk/timeseries/_moving_window.py
shsms: I wonder why.  We have that kind of notation in the sdk already,  and with `from __future__ import annotations` it works just fine.

- https://github.com/frequenz-floss/frequenz-sdk-python/blob/63411e694935b4380719b85e1405bfaf8c1990b5/src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py#L424
- https://github.com/frequenz-floss/frequenz-sdk-python/blob/63411e694935b4380719b85e1405bfaf8c1990b5/src/frequenz/sdk/timeseries/logical_meter/_formula_engine.py#L542

shsms: btw,  in `(int | datetime) | slice`, do you need the brackets? can you try without those?

leandro-lucarella-frequenz:
There seems to be a typo in the commit message:

> In order to obtain a timeseries as a data batch from streamed data
> one can define a window that is essentially.

Maybe just "essential"?

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -0,0 +1,170 @@
...
+    index or a timestamp. A sub window can be accessed by using a slice of integers
+    integers or timestamps.
+
+    Note that a numpy ndarray is returned and thus users can use
+    numpys operations directly on a window.
+
+    **Example1** (calculating the mean of a time interval):
+
+    ```
+    window = MovingWindow(size = 100, resampled_data_recv = resampled_data_recv)
```
leandro-lucarella-frequenz: Silly and this shows that maybe we also need to run `black` on the documentation code samples too :laughing: , but it should be:

```suggestion
    window = MovingWindow(size=100, resampled_data_recv=resampled_data_recv)
```

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -0,0 +1,170 @@
...
+        The task stops running only if the channel receiver is closed.
+
+        Args:
+            size: The number of elements that are stored.
+            resampled_data_recv: A receiver that delivers samples with a
+                given sampling period.
+            sampling_period: The sampling period.
+            window_alignment: A datetime object that defines a point in time to which
+                the window is aligned to modulo window size.
+                (default is midnight 01.01.01)
```
leandro-lucarella-frequenz, line 105: This is an odd default. Also, it is not entirely clear what this means to me from the docs, maybe it would be good to put some example and clarifying why this default makes sense, as it doesn't look obvious to me.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -0,0 +1,170 @@
...
+        self._buffer = OrderedRingBuffer(
+            np.empty(shape=size, dtype=float),
+            sampling_period=sampling_period,
+            time_index_alignment=window_alignment,
+        )
+        self._copy_buffer = False
+        try:
+            self._t: asyncio.Task = asyncio.create_task(self._run_impl())
+        except asyncio.CancelledError:
+            log.debug("Cancelling MovingWindow task: %s", __name__)
```
leandro-lucarella-frequenz: Are you sure `create_task()` can potentially raise a `CancelledError` already? I would be surprised if it does. It is certainly not mentioned in the [docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task).

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -0,0 +1,170 @@
...
+    def __len__(self) -> int:
+        """
+        Return the size of the `MovingWindow`s underlying buffer.
+
+        Returns:
+            The size of the `MovingWindow`.
+        """
+        return len(self._buffer)
+
+    def __getitem__(self, key: Union[int, datetime, slice]) -> Union[float, ArrayLike]:
```
leandro-lucarella-frequenz:
```suggestion
    def __getitem__(self, key: int | datetime | slice) -> float | ArrayLike:
```

leandro-lucarella-frequenz (commented):
I probably would add a way to stop the task too, even when stuff might be supposed to run forever, in a SDK I think we shouldn't assume that much. I think in general we are starting to be too far on the side of the "there is no need for clean up, nobody will ever need it" until someone does, and we are in a big mess and it's very hard to add it (remember DMD? :trollface: ).

src/frequenz/sdk/timeseries/_moving_window.py
leandro-lucarella-frequenz: No, brackets are not needed.

leandro-lucarella-frequenz: > That doesn't seem to work unfortunately. Will do the long version.

Maybe it doesn't work because of the brackets? It's already used in several parts of the SDK, so it does work :)

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: That is inherited from the ring buffer, where we all had understanding trouble :). Will try enhance the docs.

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: No that was thrown by the `run_impl` but I removed it during my last "improvement".
But that didn't make much sense anyway, so I'll remove it.

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: That's what I did at first, but it doesn't work with python3.8.

matthias-wende-frequenz:
> I probably would add a way to stop the task too, even when stuff might be supposed to run forever, in a SDK I think we shouldn't assume that much. I think in general we are starting to be too far on the side of the "there is no need for clean up, nobody will ever need it" until someone does, and we are in a big mess and it's very hard to add it (remember DMD? :trollface: ).

done :partying_face:

src/frequenz/sdk/timeseries/_moving_window.py
leandro-lucarella-frequenz: ~Maybe forward to the `RingBuffer` docs if it is already well explained in there.~

src/frequenz/sdk/timeseries/_moving_window.py
leandro-lucarella-frequenz: But we are using it!

https://github.com/frequenz-floss/frequenz-sdk-python/blob/v0.x.x/src/frequenz/sdk/timeseries/_ringbuffer.py#L389

Are you sure you tried it after you added the `from __future__ import annoations`?

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -0,0 +1,185 @@
...
+    numpys operations directly on a window.
+
+    The window uses an ringbuffer for storage and the first element is aligned to
+    a fixed defined point in time. Since the moving nature of the window, the
+    date of the first and the last element are constantly changing and therefore
+    the point in time that defines the alignment can be outside of the time window.
+    Modulo arithmetic is used to move the `window_alignment` timestamp into the
+    latest window.
+    If for example the `window_alignment` parameter is set to `datetime(1, 1, 1)`
+    and the window size is bigger then one day then the first element will always
```
leandro-lucarella-frequenz:
```suggestion
    and the window size is bigger than one day then the first element will always
```

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: I added an example to the class documentation and referred to that doc there :)

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: You are right it works.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -0,0 +1,185 @@
...
+    be aligned to the midnight. For further information see also the
+    `OrderedRingBuffer` documentation.
```
leandro-lucarella-frequenz:
```suggestion
    [`OrderedRingBuffer`][frequenz.sdk.timeseries._ringbuffer.OrderedRingBuffer] documentation.
```

It is a bit ugly to add cross-links, but in this case it might be worth it.

PS: We probably need to expose the `OrderedRingBuffer` class publicly (i.e. either make the module `frequenz.sdk.timeseries.ringbuffer` (no leading `_`) or do an import in `frequenz/sdk/timeseries/__init__.py`. I think the former is the best approach, to avoid potential dependency problems and also in case we add more ring buffer implementations.

leandro-lucarella-frequenz (commented):
Only some minor typo (and the union) things, but LGTM.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -0,0 +1,185 @@
...
+        log.debug("Cancelling MovingWindow task: %s", __name__)
+
+    async def _run_impl(self) -> None:
+        """Awaits samples from the receiver and updates the underlying ringbuffer."""
+        try:
+            async for sample in self._resampled_data_recv:
+                log.debug("Received new sample: %s", sample)
+                self._buffer.update(sample)
+        except asyncio.CancelledError:
+            log.info("MovingWindow task has been cancelled.")
```
leandro-lucarella-frequenz, line 128: Another silly one:

```suggestion
            log.info("MovingWindow task has been cancelled.")
            return
```

To avoid logging a spurious "channel closed" error.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -0,0 +1,185 @@
...
+
+        log.error("Channel has been closed")
+
+    def stop(self) -> None:
+        """Cancel the running task and stops the MovingWindow."""
+        self.update_window_task.cancel()
```
leandro-lucarella-frequenz: Sorry I didn't see the task is actually a public member, in that case I don't think it is necessary to expose a `stop()`, sorry! If you keep it it should probably be `async` and await on `join()` too to make it really useful, but I'm fine (and maybe it is best) to just expose the task.

This is actually something that we should agree on because it comes up each time we create tasks, right now we have no unified approach to this and it is starting to be a mess IMHO.

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: Ohh damn. I even spotted that one and then forgot to fix :face_with_spiral_eyes:.

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: It wasn't public and you are right that's something we should agree on. I'm in favor of having a method with a common name like stop.

matthias-wende-frequenz: I'll make the task private again and leave the stop method.

src/frequenz/sdk/timeseries/_moving_window.py
leandro-lucarella-frequenz: Nope, it looks like it does :grimacing: 

```console
$ cat src/frequenz/sdk/test.py

def union(x: int | float | str) -> float | int | str:
    return x
$ nox -R -s mypy -- src/frequenz/sdk/test.py
nox > Running session mypy
nox > Re-using existing virtual environment at .nox/mypy.
nox > mypy --install-types --namespace-packages --non-interactive --explicit-package-bases --strict src/frequenz/sdk/test.py
Success: no issues found in 1 source file
nox > Session mypy was successful.
```

Just to be sure:
```console
$ cd .nox/mypy/
$ . bin/activate
$ python --version
Python 3.8.13
```

src/frequenz/sdk/timeseries/_moving_window.py
leandro-lucarella-frequenz: :+1: you might want to use the convenience `frequenz.sdk._internal.asyncio.cancel_and_await()` function.

matthias-wende-frequenz: > If you keep it it should probably be async and await on join() too to make it really useful

~Can you please elaborate? What should be awaited and why does awaiting it make it more useful?~
Okay I'll use that function :)

leandro-lucarella-frequenz: Sure. So I might mention things that you already know or are obvious, please bare with me :)

`cancel()` is sync, so it just tells the task to stop the next time it can (usually when it is `await`ing something).

If you need to wait for the task to finish (for example, before exiting a `test_xxx` function you want all the tasks to finish to avoid errors about tasks still running), you need to call `join()` (this is also true for normal termination, not just cancellation).

If the `update_window_task` member is made private again, then the user have no way to do the joining via a public interface.

So I see 2 approaches. Mimic the `Task` object and make the `MovingWindow` class provide a sync `cancel()` plus async `join()` or simplify it and use an async `stop()` that does both.

I'm not really convinced about what's best but I have the feeling that, even when it might be more work, going for a `Task`-compatible interface will be more future proof and more familiar to Python users. Maybe we could provide the extra `stop()` on top of it for convenience (maybe it can be a simple function like `cancel_and_await()`).

I wonder if we should add a `Task` object to the SDK from which any other objects that spawn tasks (like actors) should inherit. It could also provide much of the boilerplate too, like we could just have an internal list of tasks and the `Task.cancel()` will automatically call `cancel()` on all of them, and so on with the other methods.

OK, I diverged too much here, I'll probably create a new discussion from this comment :D

matthias-wende-frequenz: > OK, I diverged too much here, I'll probably create a new discussion from this comment :D

yes please

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -9,12 +9,12 @@
...
 from datetime import datetime, timedelta
-from typing import Union
 
 import numpy as np
 from frequenz.channels import Receiver
 from numpy.typing import ArrayLike
 
+from frequenz.sdk._internal.asyncio import cancel_and_await
 from frequenz.sdk.timeseries import Sample
 from frequenz.sdk.timeseries._ringbuffer import OrderedRingBuffer
```
shsms: We're using relative paths in the sdk,  so these should be 

``` suggestion
from .._internal.asyncio import cancel_and_await
from . import Sample
from ._ringbuffer import OrderedRingBuffer
```

src/frequenz/sdk/timeseries/_moving_window.py
leandro-lucarella-frequenz: https://github.com/frequenz-floss/frequenz-sdk-python/discussions/197

leandro-lucarella-frequenz (approved):

PR #194, opened 2023-02-04
Add an `EVChargerPool` implementation

shsms:
Currently just has a method for streaming state changes for ev chargers.

The ev charger power and current formulas will get moved from the logical meter to here, in a subsequent PR.

### DIFF
--- a/src/frequenz/sdk/microgrid/component/__init__.py
+++ b/src/frequenz/sdk/microgrid/component/__init__.py
@@ -14,7 +14,7 @@ from ._component_data import (
     InverterData,
     MeterData,
 )
-from ._component_states import EVChargerCableState
+from ._component_states import EVChargerCableState, EVChargerComponentState
 
 __all__ = [
     "BatteryData",
@@ -23,6 +23,7 @@ __all__ = [
     "ComponentCategory",
     "ComponentMetricId",
     "EVChargerCableState",
+    "EVChargerComponentState",
     "EVChargerData",
     "InverterData",
     "InverterType",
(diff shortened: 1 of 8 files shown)


leandro-lucarella-frequenz (commented):
LGTM in general, I only have 2 general design concerns, which are not blockers to merge this, because they have been popping up for some time now and I think we never reach a conclusion:

1. I would put this in the `microgrid` package, as it uses the `microgrid` and component graph. I think the `timeseries` package should be agnostic of the `microgrid` API. Otherwise we could also run into dependency issues, like it happened in the past. But I see the logical meter is there and uses the microgrid API already, so I'm not sure, maybe I'm not remembering properly. At some point we should have some policy of which sub-packages can depend on which other sub-packages to avoid dependency problems.

2. I see `EVChargerPool` creates its own task and `states()` returns a `Receiver`. This looks similar to the dilemma we had with the resampler, if using a receiver is best or if just making the pool async iterable, and let the user create a task if they need one instead.

  I see the `FormulaEngine` does something similar but I see it as a different beast as it process many potentially unrelated formulas. For the EV charge pool states, for me it would be more natural to just return an async iterator instead of a receiver.

As these are all pending discussions, for me it is also OK to merge as is and work it out in the future.

leandro-lucarella-frequenz (approved):

shsms:
> For the EV charge pool states, for me it would be more natural to just return an async iterator instead of a receiver.

we were discussing about this and realized that while composing,  it would be ideal to keep the *pools and logical meter outside of any actors and send receivers to the actors.  And then have a wrapper class that will start the data source/resampling actors and create connected logical meter and *pools, and that give receivers out, which can be sent to our actors.  That helps remove a lot of boiler plate as well, and allows actors to share *pool outputs.

I thought there was an issue, but can't find it.  We can discuss it again during the sdk meeting.

leandro-lucarella-frequenz:
I got a little bit lost, so yeah, probably better to discuss in a meeting, ideally with some diagram involved :D

leandro-lucarella-frequenz:
Ah, we forgot about the RELEASE_NOTES entry.

PR #199, opened 2023-02-09
Move dev dependencies to pyproject.toml

leandro-lucarella-frequenz:
With this we have only one central place where dependencies are declared, so we can make sure we have a consistent view on the dependencies.

As a side effect, the `formatting` session will install some extra unneded dependencies (the package dependencies), but this allow us to pin the tools versions and be sure everyone will use the same version.

A new alias is also added, so users wanting to setup a local development enviroment can now just do a `pip install .[dev]` and that's it.

The CONTRIBUTING guide is updated to reflect this.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -85,7 +85,7 @@ jobs:
       - name: Install build dependencies
         run: |
           python -m pip install -U pip
-          python -m pip install .[docs]
+          python -m pip install .[docs-gen]
 
       - name: Generate the documentation
         env:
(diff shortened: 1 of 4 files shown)


leandro-lucarella-frequenz:
This is something I had in mind for a long while now, but never got into actually do it. What changed recently to finally do it is:

* Issues with hidden `black` updates that broke the CI, maybe we want to start considering pinning the dev tools versions too.
* Start using `nox` in other projects, which will inherit the issues from the `noxfile` in the SDK.

sahas-subramanian-frequenz (approved):
This looks neat!  I guess we just have to document somewhere for actor developers to know where the stuff in the square-brackets is defined,  when they're trying to use the sdk's noxfile as inspiration when migrating from tox to nox, for example.

Maybe in the hello-world-actor.

PR #203, opened 2023-02-12
Add COMPONENT_STATE_DISCHARGING as valid state for the inverter

ela-kotulska-frequenz:
DISCHARGING state was missing by mistake and this caused the power distributor to error out if the inverter is already discharging.


noxfile.py
```diff
@@ -146,7 +146,13 @@ def ci_checks_max(session: nox.Session) -> None:
         session: the nox session.
     """
     session.install(
-        ".[docs]", "pylint", "nox", *PYTEST_DEPS, *FMT_DEPS, *DOCSTRING_DEPS, *MYPY_DEPS
+        ".[docs]",
+        "pylint==2.15",
```
leandro-lucarella-frequenz, line 150: Just a heads up: in v0.x.x this pinning will have to be moved to `pyproject.toml`.

pyproject.toml
```diff
@@ -27,9 +27,9 @@ requires-python = ">= 3.8, < 4"
 dependencies = [
     "frequenz-api-microgrid >= 0.11.0, < 0.12.0",
     "frequenz-channels >= 0.12.0, < 0.13.0",
-    "google-api-python-client >= 2.71, < 3",
-    "grpcio >= 1.51.1, < 2",
-    "grpcio-tools >= 1.51.1, < 2",
+    "google-api-python-client >= 2.71, < 2.73",
+    "grpcio == 1.51.1",
+    "grpcio-tools == 1.51.1",
```
leandro-lucarella-frequenz, line 32: In the last commit freezing more versions you don't mention why it is needed. Also, does it make sense to open issues about the errors you found without limiting the versions like this?

leandro-lucarella-frequenz (approved):
Just minor comments main about what to do in the current v0.x.x with the new version restrictions.

leandro-lucarella-frequenz:
Ah, I see you finally fixed this in v0.x.x in #202, so I guess this won't be merged, right? So closing.

PR #205, opened 2023-02-15
Add BatteryPool implementation for aggregating battery-inverter data

### LINKED ISSUE #140: Implement BatteryPool
### What's needed?

User should be able to subscribe for data from some pool of Batteries.
BatteryPool should communicate with user using channels. This approach would be consisntent with LogicanMeter and EvChargerPool.

## What data are needed:
All these values will be packed in the class and send with one channel. We use DataSourceActor for IT!
* **total_capacity**: sum(battery.capacity for each working battery in battery pool)
* **max_capacity**: sum(battery.capacity * battery.max_soc for each working battery in battery pool)
* **min_capacity**:  sum(battery.capacity * battery.min_soc for each working battery in battery pool)
* **max_soc**: sum(battery.capacity*battery.max_soc for each working battery in battery pool)/total_capacity
* **min_soc**: sum(battery.capacity * battery.min_soc for each working battery in battery pool)/total_capacity 

All the value below will have their own channel.
* **average_soc**: sum(battery.capacity * battery.soc  for each working battery in battery pool)/total_capacity 
* **active_power** sum(battery.adjacent_inverter.active_power for each battery in the battery pool) - even if it is broken we need this information.

* TUPLE **power_supply_bound** : sum( min( battery.power_upper_bound, battery.adjacent_inverter.power_upper_bound ) for each working battery in battery pool )
* TUPLE **power_consumption_bound** : sum( max( battery.power_lower_bound, battery.adjacent_inverter.power_lower_bound ) for each working battery in battery pool )

### How frequently we should send data:
* Capacity, SoC bounds - when they change. These don't change frequently.
* SoP - This can change frequently. However it is not needed for any calculation. It is needed only before making charge/discharge decision.
    *  We could send new value only when it change. However this value are not needed frequently and we would compute it frequently. It seems to be lots of unnecessary work.
    *  User can ask for this value every-time when he needs it. But then we have no channel communication. It won't be possible to move BatteryPool to separate process. (**No** : Implementation should be consistent with Logical Meter. So we should use channels.)
* SoC - when it change
* active power - with given resampling frequency.

### Proposed solution

BatteryPool will use:
* BatteryPoolStatus - to know what batteries are working
* ComponentMetricResamplingActor - to subscribe for the metric data and receive them with some frequency. 
    *  Interpolation methods for the resampler:
          * Capacity, SoC bounds : Return last value -  This remains constant for most of the time, so returning last value seems to be correct.
          * SoP, SoC : 
               * Return last value  (forward filling) 
               * Linear interpolation. . Both SoC and SoP don't change the direction frequently. They either increases or decreases for some time. (or remain constant). If we find linear function based on first and last value in Resampler buffer, then we can predict the next value. How long time window we should store in ResamplerBuffer? After X seconds of inactivity battery/inverter will be considered as broken. (BatteryPoolStatus holds X argument). Time window could be X, too. Then we can interpolate as long as battery last message is considered as *relevant*.
          * active_power - average

### Open questions ###
* Different method to predict SoC, SoP.
* How frequently we should send SoP - it can change frequently. We could send it in second basis and user can use peekable feature to get the latest value.

### Use cases

_No response_

_No response_

### Additional context

_No response_

ela-kotulska-frequenz:
BatteryPool implementation for aggregating battery-inverter metrics into higher level metrics

### DIFF
--- a/examples/battery_pool.py
+++ b/examples/battery_pool.py
@@ -1,0 +1,92 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Script with an example how to use BatteryPool."""
+
+from __future__ import annotations
+
+import asyncio
+import logging
+from datetime import timedelta
+from typing import Any, Dict
+
+from frequenz.channels import Broadcast, Receiver
+from frequenz.channels.util import MergeNamed
+
+from frequenz.sdk import microgrid
+from frequenz.sdk.actor import (
+    ChannelRegistry,
+    ComponentMetricRequest,
+    DataSourcingActor,
+)
+from frequenz.sdk.actor.power_distributing import PowerDistributingActor
+from frequenz.sdk.actor.power_distributing._battery_pool_status import BatteryStatus
+from frequenz.sdk.microgrid.component import ComponentCategory
+from frequenz.sdk.timeseries.battery_pool import BatteryPool
+
+HOST = "microgrid.sandbox.api.frequenz.io"  # it should be the host name.
+PORT = 61060
+
+
+def create_battery_pool() -> BatteryPool:
+    """Create battery pool.
+
+    It needs many instance to be created before.
+
+    Returns:
+        BatteryPool instance ready to use.
+    """
+    channel_registry = ChannelRegistry(name="data-registry")
+
+    # Create a channels for sending/receiving subscription requests
+    data_source_request_channel = Broadcast[ComponentMetricRequest](
+        "data-source", resend_latest=True
+    )
+
+    # Instantiate a data sourcing actor
+    _ = DataSourcingActor(
+        request_receiver=data_source_request_channel.new_receiver(
+            "data_sourcing_receiver"
+        ),
+        registry=channel_registry,
+    )
+
+    battery_status_channel = Broadcast[BatteryStatus]("batteries-status")
+    _ = PowerDistributingActor(
+        users_channels={},
+        battery_status_sender=battery_status_channel.new_sender(),
+    )
+
+    batteries = microgrid.get().component_graph.components(
+        component_category={ComponentCategory.BATTERY}
+    )
+
+    return BatteryPool(
+        batteries_id=set(battery.component_id for battery in batteries),
+        batteries_status_receiver=battery_status_channel.new_receiver(
+            name="battery_pool", maxsize=1
+        ),
+        min_update_interval=timedelta(seconds=0.2),
+    )
+
+
+async def main() -> None:
+    """Create the battery pool, activate all formulas and listen for any update."""
+    logging.basicConfig(
+        level=logging.DEBUG, format="%(asctime)s %(name)s %(levelname)s:%(message)s"
+    )
+    await microgrid.initialize(host=HOST, port=PORT)
+
+    battery_pool = create_battery_pool()
+    receivers: Dict[str, Receiver[Any]] = {
+        "soc": await battery_pool.soc(maxsize=1),
+        "capacity": await battery_pool.capacity(maxsize=1),
+        "power_bounds": await battery_pool.power_bounds(maxsize=1),
+    }
+
+    merged_channel = MergeNamed[Any](**receivers)
+    async for metric_name, metric in merged_channel:
+        print(f"Received new {metric_name}: {metric}")
+
+
+asyncio.run(main())
--- a/src/frequenz/sdk/actor/power_distributing/_battery_pool_status.py
+++ b/src/frequenz/sdk/actor/power_distributing/_battery_pool_status.py
@@ -9,7 +9,7 @@ import logging
 from dataclasses import dataclass
 from typing import Dict, Set
 
-from frequenz.channels import Broadcast, Receiver
+from frequenz.channels import Broadcast, Receiver, Sender
 from frequenz.channels.util import MergeNamed
 
 from ..._internal.asyncio import cancel_and_await
@@ -72,6 +72,7 @@ class BatteryPoolStatus:
     def __init__(
         self,
         battery_ids: Set[int],
+        battery_status_sender: Sender[BatteryStatus],
         max_data_age_sec: float,
         max_blocking_duration_sec: float,
     ) -> None:
@@ -79,6 +80,8 @@ class BatteryPoolStatus:
 
         Args:
             battery_ids: set of batteries ids that should be stored in pool.
+            battery_status_sender: The sender used for sending the status of the
+                batteries in the pool.
             max_data_age_sec: If component stopped sending data, then
                 this is the maximum time when its last message should be considered as
                 valid. After that time, component won't be used until it starts sending
@@ -121,7 +124,7 @@ class BatteryPoolStatus:
             **receivers,
         )
 
-        self._task = asyncio.create_task(self._run())
+        self._task = asyncio.create_task(self._run(battery_status_sender))
 
     async def join(self) -> None:
         """Await for the battery pool, and return when the task completes.
@@ -146,23 +149,30 @@ class BatteryPoolStatus:
         )
         await self._battery_status_channel.stop()
 
-    async def _run(self) -> None:
-        """Start tracking batteries status."""
+    async def _run(self, battery_status_sender: Sender[BatteryStatus]) -> None:
+        """Start tracking batteries status.
+
+        Args:
+            battery_status_sender: The sender used for sending the status of the
+                batteries in the pool.
+        """
         while True:
             try:
-                await self._update_status(self._battery_status_channel)
+                await self._update_status(battery_status_sender)
             except Exception as err:  # pylint: disable=broad-except
                 _logger.error(
                     "BatteryPoolStatus failed with error: %s. Restarting.", err
                 )
 
-    async def _update_status(self, status_channel: MergeNamed[Status]) -> None:
+    async def _update_status(
+        self, battery_status_sender: Sender[BatteryStatus]
+    ) -> None:
         """Wait for any battery to change status and update status.
 
         Args:
-            status_channel: Receivers packed in Select object.
+            battery_status_sender: Sender to send the current status of the batteries.
         """
-        async for channel_name, status in status_channel:
+        async for channel_name, status in self._battery_status_channel:
             battery_id = self._batteries[channel_name].battery_id
             if status == Status.WORKING:
                 self._current_status.working.add(battery_id)
@@ -174,7 +184,7 @@ class BatteryPoolStatus:
                 self._current_status.working.discard(battery_id)
                 self._current_status.uncertain.discard(battery_id)
 
-            # In the future here we should send status to the subscribed actors
+            await battery_status_sender.send(self._current_status)
 
     async def update_status(
         self, succeed_batteries: Set[int], failed_batteries: Set[int]
--- a/src/frequenz/sdk/microgrid/client/_client.py
+++ b/src/frequenz/sdk/microgrid/client/_client.py
@@ -25,6 +25,7 @@ from frequenz.api.microgrid.microgrid_pb2_grpc import MicrogridStub
 from frequenz.channels import Broadcast, Receiver, Sender
 from google.protobuf.empty_pb2 import Empty  # pylint: disable=no-name-in-module
 
+from ..._internal._constants import RECEIVER_MAX_SIZE
 from ..component import (
     BatteryData,
     Component,
@@ -90,6 +91,7 @@ class MicrogridApiClient(ABC):
     async def meter_data(
         self,
         component_id: int,
+        maxsize: int = RECEIVER_MAX_SIZE,
     ) -> Receiver[MeterData]:
         """Return a channel receiver that provides a `MeterData` stream.
 
@@ -99,6 +101,7 @@ class MicrogridApiClient(ABC):
 
         Args:
             component_id: id of the meter to get data for.
+            maxsize: Size of the receiver's buffer.
 
         Returns:
             A channel receiver that provides realtime meter data.
@@ -108,6 +111,7 @@ class MicrogridApiClient(ABC):
     async def battery_data(
         self,
         component_id: int,
+        maxsize: int = RECEIVER_MAX_SIZE,
     ) -> Receiver[BatteryData]:
         """Return a channel receiver that provides a `BatteryData` stream.
 
@@ -117,6 +121,7 @@ class MicrogridApiClient(ABC):
 
         Args:
             component_id: id of the battery to get data for.
+            maxsize: Size of the receiver's buffer.
 
         Returns:
             A channel receiver that provides realtime battery data.
@@ -126,6 +131,7 @@ class MicrogridApiClient(ABC):
     async def inverter_data(
         self,
         component_id: int,
+        maxsize: int = RECEIVER_MAX_SIZE,
     ) -> Receiver[InverterData]:
         """Return a channel receiver that provides an `InverterData` stream.
 
@@ -135,6 +141,7 @@ class MicrogridApiClient(ABC):
 
         Args:
             component_id: id of the inverter to get data for.
+            maxsize: Size of the receiver's buffer.
 
         Returns:
             A channel receiver that provides realtime inverter data.
@@ -144,6 +151,7 @@ class MicrogridApiClient(ABC):
     async def ev_charger_data(
         self,
         component_id: int,
+        maxsize: int = RECEIVER_MAX_SIZE,
     ) -> Receiver[EVChargerData]:
         """Return a channel receiver that provides an `EvChargeData` stream.
 
@@ -153,6 +161,7 @@ class MicrogridApiClient(ABC):
 
         Args:
             component_id: id of the ev charger to get data for.
+            maxsize: Size of the receiver's buffer.
 
         Returns:
             A channel receiver that provides realtime ev charger data.
@@ -440,6 +449,7 @@ class MicrogridGrpcClient(MicrogridApiClient):
     async def meter_data(
         self,
         component_id: int,
+        maxsize: int = RECEIVER_MAX_SIZE,
     ) -> Receiver[MeterData]:
         """Return a channel receiver that provides a `MeterData` stream.
 
@@ -452,6 +462,7 @@ class MicrogridGrpcClient(MicrogridApiClient):
 
         Args:
             component_id: id of the meter to get data for.
+            maxsize: Size of the receiver's buffer.
 
         Returns:
             A channel receiver that provides realtime meter data.
@@ -463,11 +474,12 @@ class MicrogridGrpcClient(MicrogridApiClient):
         return self._get_component_data_channel(
             component_id,
             MeterData.from_proto,
-        ).new_receiver()
+        ).new_receiver(maxsize=maxsize)
 
     async def battery_data(
         self,
         component_id: int,
+        maxsize: int = RECEIVER_MAX_SIZE,
     ) -> Receiver[BatteryData]:
         """Return a channel receiver that provides a `BatteryData` stream.
 
@@ -480,6 +492,7 @@ class MicrogridGrpcClient(MicrogridApiClient):
 
         Args:
             component_id: id of the battery to get data for.
+            maxsize: Size of the receiver's buffer.
 
         Returns:
             A channel receiver that provides realtime battery data.
@@ -491,11 +504,12 @@ class MicrogridGrpcClient(MicrogridApiClient):
         return self._get_component_data_channel(
             component_id,
             BatteryData.from_proto,
-        ).new_receiver()
+        ).new_receiver(maxsize=maxsize)
 
     async def inverter_data(
         self,
         component_id: int,
+        maxsize: int = RECEIVER_MAX_SIZE,
     ) -> Receiver[InverterData]:
         """Return a channel receiver that provides an `InverterData` stream.
 
@@ -508,6 +522,7 @@ class MicrogridGrpcClient(MicrogridApiClient):
 
         Args:
             component_id: id of the inverter to get data for.
+            maxsize: Size of the receiver's buffer.
 
         Returns:
             A channel receiver that provides realtime inverter data.
@@ -519,11 +534,12 @@ class MicrogridGrpcClient(MicrogridApiClient):
         return self._get_component_data_channel(
             component_id,
             InverterData.from_proto,
-        ).new_receiver()
+        ).new_receiver(maxsize=maxsize)
 
     async def ev_charger_data(
         self,
         component_id: int,
+        maxsize: int = RECEIVER_MAX_SIZE,
     ) -> Receiver[EVChargerData]:
         """Return a channel receiver that provides an `EvChargeData` stream.
 
@@ -536,6 +552,7 @@ class MicrogridGrpcClient(MicrogridApiClient):
 
         Args:
             component_id: id of the ev charger to get data for.
+            maxsize: Size of the receiver's buffer.
 
         Returns:
             A channel receiver that provides realtime ev charger data.
@@ -547,7 +564,7 @@ class MicrogridGrpcClient(MicrogridApiClient):
         return self._get_component_data_channel(
             component_id,
             EVChargerData.from_proto,
-        ).new_receiver()
+        ).new_receiver(maxsize=maxsize)
 
     async def set_power(self, component_id: int, power_w: int) -> Empty:
         """Send request to the Microgrid to set power for component.
--- a/src/frequenz/sdk/timeseries/battery_pool/__init__.py
+++ b/src/frequenz/sdk/timeseries/battery_pool/__init__.py
@@ -1,0 +1,15 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Manage a pool of batteries."""
+
+from ._result_types import Bound, CapacityMetrics, PowerMetrics, SoCMetrics
+from .battery_pool import BatteryPool
+
+__all__ = [
+    "BatteryPool",
+    "PowerMetrics",
+    "SoCMetrics",
+    "CapacityMetrics",
+    "Bound",
+]
--- a/src/frequenz/sdk/timeseries/battery_pool/_component_metric_fetcher.py
+++ b/src/frequenz/sdk/timeseries/battery_pool/_component_metric_fetcher.py
@@ -1,0 +1,267 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Module to define how to subscribe and fetch component data."""
+
+from __future__ import annotations
+
+import asyncio
+import logging
+import math
+from abc import ABC, abstractmethod
+from datetime import datetime, timezone
+from typing import Any, Generic, Iterable, Optional, Set, TypeVar
+
+from frequenz.channels import ChannelClosedError, Receiver
+
+from ..._internal._constants import MAX_BATTERY_DATA_AGE_SEC
+from ..._internal.asyncio import AsyncConstructible
+from ...actor._data_sourcing.microgrid_api_source import (
+    _BatteryDataMethods,
+    _InverterDataMethods,
+)
+from ...microgrid import get as get_microgrid
+from ...microgrid.component import (
+    BatteryData,
+    ComponentCategory,
+    ComponentData,
+    ComponentMetricId,
+    InverterData,
+)
+from ._component_metrics import ComponentMetricsData
+
+_logger = logging.getLogger(__name__)
+
+T = TypeVar("T", bound=ComponentData)
+
+
+class ComponentMetricFetcher(AsyncConstructible, ABC):
+    """Define how to subscribe for and fetch the component metrics data."""
+
+    _component_id: int
+    _metrics: Iterable[ComponentMetricId]
+
+    @classmethod
+    async def async_new(
+        cls, component_id: int, metrics: Iterable[ComponentMetricId]
+    ) -> Self:  # type: ignore[name-defined] # pylint: disable=undefined-variable
+        """Create an instance of this class.
+
+        Subscribe for the given component metrics and return them if method
+        `fetch_next` is called.
+
+        Args:
+            component_id: component id
+            metrics: metrics that should be fetched from this component.
+
+        Returns:
+            This class instance.
+        """
+        self: ComponentMetricFetcher = ComponentMetricFetcher.__new__(cls)
+        self._component_id = component_id
+        self._metrics = metrics
+        return self
+
+    @abstractmethod
+    async def fetch_next(self) -> Optional[ComponentMetricsData]:
+        """Fetch metrics for this component."""
+
+
+class LatestMetricsFetcher(ComponentMetricFetcher, Generic[T], ABC):
+    """Subscribe for the latest component data and extract the needed metrics."""
+
+    _receiver: Receiver[T]
+    _max_waiting_time: float
+
+    @classmethod
+    async def async_new(
+        cls,
+        component_id: int,
+        metrics: Iterable[ComponentMetricId],
+    ) -> Self:  # type: ignore[name-defined] # pylint: disable=undefined-variable:
+        """Create instance of this class.
+
+        Subscribe for the requested component data and fetch only the latest component
+        metrics.
+
+        Args:
+            component_id: component id
+            metrics: metrics
+
+        Raises:
+            ValueError: If any requested metric id is not supported.
+
+        Returns:
+            This class instance
+        """
+        self: LatestMetricsFetcher[T] = await super().async_new(component_id, metrics)
+
+        for metric in metrics:
+            # pylint: disable=protected-access
+            if metric not in self._supported_metrics():
+                category = self._component_category()
+                raise ValueError(f"Metric {metric} not supported for {category}")
+
+        # pylint: disable=protected-access
+        self._receiver = await self._subscribe()
+        self._max_waiting_time = MAX_BATTERY_DATA_AGE_SEC
+        return self
+
+    async def fetch_next(self) -> Optional[ComponentMetricsData]:
+        """Fetch the latest component metrics.
+
+        Returns:
+            Component metrics data.
+            None if the channel was closed and fetching next element is impossible.
+        """
+        try:
+            data = await asyncio.wait_for(
+                self._receiver.receive(), self._max_waiting_time
+            )
+
+        except ChannelClosedError:
+            _logger.exception(
+                "Channel for component %d was closed.", self._component_id
+            )
+            return None
+        except asyncio.TimeoutError:
+            # Next time wait infinitely until we receive any message.
+            _logger.debug("Component %d stopped sending data.", self._component_id)
+            return ComponentMetricsData(
+                self._component_id, datetime.now(tz=timezone.utc), {}
+            )
+
+        self._max_waiting_time = MAX_BATTERY_DATA_AGE_SEC
+        metrics = {}
+        for mid in self._metrics:
+            value = self._extract_metric(data, mid)
+            # There is no guarantee that all fields in component message are populated
+            if not math.isnan(value):
+                metrics[mid] = value
+
+        return ComponentMetricsData(self._component_id, data.timestamp, metrics)
+
+    @abstractmethod
+    def _extract_metric(self, data: T, mid: ComponentMetricId) -> float:
+        ...
+
+    @abstractmethod
+    def _supported_metrics(self) -> Set[ComponentMetricId]:
+        ...
+
+    @abstractmethod
+    def _component_category(self) -> ComponentCategory:
+        ...
+
+    @abstractmethod
+    async def _subscribe(self) -> Receiver[Any]:
+        """Subscribe for this component data.
+
+        Size of the receiver buffer should should be 1 to make sure we receive only
+        the latest component data.
+
+        Returns:
+            Receiver for this component metrics.
+        """
+
+
+class LatestBatteryMetricsFetcher(LatestMetricsFetcher[BatteryData]):
+    """Subscribe for the latest battery data using MicrogridApiClient."""
+
+    @classmethod
+    async def async_new(
+        cls,
+        component_id: int,
+        metrics: Iterable[ComponentMetricId],
+    ) -> LatestBatteryMetricsFetcher:
+        """Create instance of this class.
+
+        Subscribe for the requested component data and fetch only the latest component
+        metrics.
+
+        Args:
+            component_id: component id
+            metrics: metrics
+
+        Raises:
+            ValueError: If any requested metric id is not supported.
+
+        Returns:
+            This class instance
+        """
+        self: LatestBatteryMetricsFetcher = await super().async_new(
+            component_id, metrics
+        )
+        return self
+
+    def _supported_metrics(self) -> Set[ComponentMetricId]:
+        return set(_BatteryDataMethods.keys())
+
+    def _extract_metric(self, data: BatteryData, mid: ComponentMetricId) -> float:
+        return _BatteryDataMethods[mid](data)
+
+    async def _subscribe(self) -> Receiver[BatteryData]:
+        """Subscribe for this component data.
+
+        Size of the receiver buffer should should be 1 to make sure we receive only
+        the latest component data.
+
+        Returns:
+            Receiver for this component metrics.
+        """
+        api = get_microgrid().api_client
+        return await api.battery_data(self._component_id, maxsize=1)
+
+    def _component_category(self) -> ComponentCategory:
+        return ComponentCategory.BATTERY
+
+
+class LatestInverterMetricsFetcher(LatestMetricsFetcher[InverterData]):
+    """Subscribe for the latest inverter data using MicrogridApiClient."""
+
+    @classmethod
+    async def async_new(
+        cls,
+        component_id: int,
+        metrics: Iterable[ComponentMetricId],
+    ) -> LatestInverterMetricsFetcher:
+        """Create instance of this class.
+
+        Subscribe for the requested component data and fetch only the latest component
+        metrics.
+
+        Args:
+            component_id: component id
+            metrics: metrics
+
+        Raises:
+            ValueError: If any requested metric id is not supported.
+
+        Returns:
+            This class instance
+        """
+        self: LatestInverterMetricsFetcher = await super().async_new(
+            component_id, metrics
+        )
+        return self
+
+    def _supported_metrics(self) -> Set[ComponentMetricId]:
+        return set(_InverterDataMethods.keys())
+
+    def _extract_metric(self, data: InverterData, mid: ComponentMetricId) -> float:
+        return _InverterDataMethods[mid](data)
+
+    async def _subscribe(self) -> Receiver[InverterData]:
+        """Subscribe for this component data.
+
+        Size of the receiver buffer should should be 1 to make sure we receive only
+        the latest component data.
+
+        Returns:
+            Receiver for this component metrics.
+        """
+        api = get_microgrid().api_client
+        return await api.inverter_data(self._component_id, maxsize=1)
+
+    def _component_category(self) -> ComponentCategory:
+        return ComponentCategory.INVERTER
--- a/src/frequenz/sdk/timeseries/battery_pool/_component_metrics.py
+++ b/src/frequenz/sdk/timeseries/battery_pool/_component_metrics.py
@@ -1,0 +1,83 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Class that stores values of the component metrics."""
+
+from __future__ import annotations
+
+from collections.abc import Mapping
+from datetime import datetime
+
+from ...microgrid.component import ComponentMetricId
+
+
+class ComponentMetricsData:
+    """Store values of the component metrics."""
+
+    def __init__(
+        self,
+        component_id: int,
+        timestamp: datetime,
+        metrics: Mapping[ComponentMetricId, float],
+    ) -> None:
+        """Create class instance.
+
+        Args:
+            component_id: component id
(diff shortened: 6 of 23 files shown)


src/frequenz/sdk/microgrid/client/_client.py
```diff
@@ -80,6 +80,7 @@ async def connections(
     async def meter_data(
         self,
         component_id: int,
+        maxsize: Optional[int] = None,
```
shsms: I feel like making it optional means that there is no max size,  like it will never overflow.  And it makes the default value opaque, and it is not a one step to find the correct default value.

What do you think about making it a constant `RECEIVER_MAX_SIZE = 50` in this file instead?  I'll do that for the formula receiver too.  That way we have the flexibility to change the default if necessary, just for this case.

Maybe we really should just export a constant from the channels repo,  but we can make an issue for that.

src/frequenz/sdk/actor/power_distributing/_battery_pool_status.py
```diff
@@ -121,7 +124,7 @@ def __init__(
             **receivers,
         )
 
-        self._task = asyncio.create_task(self._run())
+        self._task = asyncio.create_task(self._run(battery_status_sender))
```
shsms, line 127: why not an instance variable instead?  then you don't have to document it 3 times,  or pass it around, to two more functions.

ela-kotulska-frequenz:
`active_power` is calculated for working batteries only. It should be computed for all batteries instead.
`active_power` is calculated using logical meter, now.
I will remove it from this PR and add it in the later one. It will use formula engines. (like in ev pool)

src/frequenz/sdk/microgrid/client/_client.py
ela-kotulska-frequenz: Done

examples/battery_pool.py
```diff
@@ -0,0 +1,114 @@
...
+from frequenz.sdk.actor.power_distributing import PowerDistributingActor
+from frequenz.sdk.actor.power_distributing._battery_pool_status import BatteryStatus
+from frequenz.sdk.microgrid.component import ComponentCategory
+from frequenz.sdk.timeseries.battery_pool import BatteryPool, ResamplerArgs
+
+HOST = "microgrid.sandbox.api.frequenz.io"  # it should be the host name.
+PORT = 61060
+
+
+def create_battery_pool() -> BatteryPool:
```
leandro-lucarella-frequenz, line 31: This is a good summary of what needs to be put in a SDK initialization function provided by the SDK :D

src/frequenz/sdk/actor/power_distributing/_battery_pool_status.py
```diff
@@ -72,13 +72,16 @@ class BatteryPoolStatus:
...
+        battery_status_sender: Sender[BatteryStatus],
         max_data_age_sec: float,
         max_blocking_duration_sec: float,
     ) -> None:
         """Create BatteryPoolStatus instance.
 
         Args:
             battery_ids: set of batteries ids that should be stored in pool.
+            battery_status_sender: Channel for sending information which batteries are
+                working.
```
leandro-lucarella-frequenz: This is not really a channel... Maybe:

```suggestion
            battery_status_sender: The sender used for sending the status of the
                batteries in the pool.
```

src/frequenz/sdk/microgrid/component/battery_utils.py
```diff
@@ -0,0 +1,60 @@
...
+            None,
+        )
+    except KeyError as err:
+        _logger.warning(err)
+        inverter_id = None
+
+    return inverter_id
+
+
+def battery_inverter_mapping(batteries: Set[int]) -> Dict[int, int]:
```
leandro-lucarella-frequenz: If we emit the log message in `find_adjacent()` this function is reduced to:

```py
bat_inv_map = {bat_id: graph.find_adjacent(bat_id, ComponentCategory.INVERTER)
    for battery_id in batteries}:
```

I'm not sure we need a function for this, unless it is something that will be used very often in many places.

If we want to keep it as a function, I think it could also be generalized as a `ComponentGraph` method, so users that want exactly what this function does could do (I would default to raise an exception if a mapping is missing, as is the safest choice):

```py
bat_inv_map = component_graph.map_adjacents(battery_ids, ComponentCategory.INVERTER,
    ignore_missing=True)
```

src/frequenz/sdk/microgrid/component/battery_utils.py
```diff
@@ -0,0 +1,60 @@
...
+import logging
+from typing import Dict, Optional, Set
+
+from .. import get as get_microgrid
+from . import ComponentCategory
+
+_logger = logging.getLogger(__name__)
+
+
+def find_adjacent_inverter_id(battery_id: int) -> Optional[int]:
```
leandro-lucarella-frequenz: I wonder if this doesn't belong to the `ComponentGraph`. It could be generalized to find adjacent inverters to any component (maybe it can be parametrized to find any type of component adjacent to other type of componet). Not sure if we have use case for something this generic yet but it sounds like it might be a good utility to have?

Like:

```py
inverters = component_graph.find_adjacent(battery_id, ComponentCategory.INVERTER)
```

src/frequenz/sdk/microgrid/component/battery_utils.py
```diff
@@ -0,0 +1,60 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Util functions for battery."""
```
leandro-lucarella-frequenz: This module sounds a bit weird for me, it feels like if it is something that it is used only for batteries and frequently by SDK users, it should be in the `BatteryPool` somehow (as we want that to be our interface to batteries, right)?

Otherwise it seems like they could be utility functions from the component graph. Se the comments on the specific functions.

src/frequenz/sdk/microgrid/component/battery_utils.py
```diff
@@ -0,0 +1,60 @@
...
+        inverter_id = next(
+            (
+                comp.component_id
+                for comp in graph.predecessors(battery_id)
+                if comp.category == ComponentCategory.INVERTER
+            ),
+            None,
+        )
+    except KeyError as err:
+        _logger.warning(err)
```
leandro-lucarella-frequenz: Do we really want to log an error each time someone tries to find an inverter for a battery and it's not there? This error is also duplicated when the `battery_inverter_mapping()` function is used.

Also, if we keep it, I would write a more descriptive error instead of printing the `err` object raw.

BTW, is a microgrid in a valid state if it has batteries without inverters? Because if it's not maybe this should be an error instead.

src/frequenz/sdk/timeseries/battery_pool/__init__.py
```diff
@@ -0,0 +1,10 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Module to aggregate data from pool of the batteries."""
```
leandro-lucarella-frequenz: I would skip the `Module to`:

```suggestion
"""Aggregates data from a pool of batteries."""
```
or

```suggestion
"""Manage a pool of batteries."""
```

If we'll provide also controlling functions with the battery pool (still not sure if charge/discharge will be done via the battery pool or power manager).

src/frequenz/sdk/timeseries/battery_pool/_component_metric_fetcher.py
```diff
@@ -0,0 +1,305 @@
...
+_logger = logging.getLogger(__name__)
+
+
+class ComponentMetricFetcher(ABC):
+    """Define how to subscribe for and fetch the component metrics data."""
+
+    def __init__(
+        self,
+        component_id: int,
+        metrics: List[ComponentMetricId],
```
leandro-lucarella-frequenz: Super minor, but if you are copying anyway, maybe you can require just a `Iterable[ComponentMetricId]`, so you can pass iterators or generators too.

src/frequenz/sdk/timeseries/battery_pool/_component_metric_fetcher.py
```diff
@@ -0,0 +1,305 @@
...
+            metrics: Metrics needed for the given component
+
+        Raises:
+            ValueError: If any metrics are not supported.
+        """
+        super().__init__(component_id, metrics)
+        if any(map(lambda mid: mid not in _BatteryDataMethods, metrics)):
+            raise ValueError(
+                "Some battery metrics are not supported, supported: "
+                f"{str(_BatteryDataMethods.keys())}, requested: {str(metrics)}"
```
leandro-lucarella-frequenz:
```suggestion
                f"Some battery metrics are not supported: {list(set(metrics) - _BatteryDataMethods.keys())}"
                f" (supported metrics are: {list(_BatteryDataMethods.keys())}"
```

src/frequenz/sdk/timeseries/battery_pool/_component_metric_fetcher.py
```diff
@@ -0,0 +1,305 @@
...
+            metrics: Metrics needed for the given component
+
+        Raises:
+            ValueError: If any metrics are not supported.
+        """
+        super().__init__(component_id, metrics)
+        if any(map(lambda mid: mid not in _InverterDataMethods, metrics)):
+            raise ValueError(
+                "Some battery metrics are not supported, supported: "
+                f"{str(_InverterDataMethods.keys())}, requested: {str(metrics)}"
```
leandro-lucarella-frequenz: Same

src/frequenz/sdk/timeseries/battery_pool/_component_metric_fetcher.py
```diff
@@ -0,0 +1,305 @@
...
+            return None
+
+        metrics = {}
+        for mid in self._metrics:
+            metrics[mid] = _InverterDataMethods[mid](data)
+        return ComponentMetricsData(self._component_id, data.timestamp, metrics)
+
+
+@dataclass
+class ResamplerArgs:
```
leandro-lucarella-frequenz: I was a bit confused by this class, because is not really arguments for the resampler. Is more like a bag of stuff that is used in connection to make requests to the resampler actor. Any reason against just passing these separately to the `ResampledComponentMetricFetcher` instead? They seem to be used only in `_subscribe_for_metric()` anyway.

In the long term, I think we need to simplify the interface to subscribe to the resampler, there is a lot of boilerplate going on here.

src/frequenz/sdk/timeseries/battery_pool/_component_metric_fetcher.py
```diff
@@ -0,0 +1,305 @@
...
+    """The unique namespace to allow reuse of streams in the data pipeline."""
+
+    resampler_subscription_sender: Sender[ComponentMetricRequest]
+    """Sender channel to send the metric subscription request."""
+
+    channel_registry: ChannelRegistry
+    """Channel registry instance to get the receiver for the resampling channel."""
+
+
+class ResampledComponentMetricFetcher(ComponentMetricFetcher):
```
leandro-lucarella-frequenz: This is somehow analogous to the `LogicalMeter`, right?

src/frequenz/sdk/timeseries/battery_pool/_methods.py
```diff
@@ -0,0 +1,413 @@
...
+    ResamplerArgs,
+)
+
+_logger = logging.getLogger(__name__)
+
+# Constant needed in the code
+_MIN_TIMESTAMP = datetime.min.replace(tzinfo=timezone.utc)
+
+
+class BaseMethod(ABC, Generic[T]):
```
leandro-lucarella-frequenz: Maybe `AggregateMethod` or `SendMethod` whatever this is? Without reading the code I have no idea if this is a method for sending, aggregating or what :)

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
```diff
@@ -0,0 +1,475 @@
...
+        ]
+
+    @classmethod
+    def name(cls) -> str:
+        """Return name of the calculator.
+
+        Returns:
+            Name of the calculator
+        """
+        return "Soc_Capacity"
```
leandro-lucarella-frequenz: If this is only for debugging, maybe you can move it to `MetricCalculator` and use an implementation like `return cls.__name__`.

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
```diff
@@ -0,0 +1,475 @@
...
+                component_id,
+            )
+            return True
+
+        return False
+
+
+# Soc and capacity are different kind of metrics.
+# It makes sens to separate them with `_`.
+class SoC_Capacity_Calculator(
```
leandro-lucarella-frequenz: Why underscore in class names here?

src/frequenz/sdk/timeseries/battery_pool/_result_types.py
```diff
@@ -0,0 +1,181 @@
...
+    total_capacity = sum(battery.capacity for battery in working_batteries)
+    soc_lower_bound = capacity_upper_bound/total_capacity
+    ```
+    """
+
+
+# Soc and capacity are different kind of metrics.
+# It makes sens to separate them with `_`.
+@dataclass
+class SoC_Capacity_Metrics:  # pylint: disable=invalid-name
```
leandro-lucarella-frequenz: I don't get this. Why can't we use `SoCCapacityMetrics`?

I'd prefer to avoid at all cost diverging from the class naming convention.

If this refers to a group of different metrics, and this is basically all the metrics a battery pool will have, can't it be `BatteryPoolMetrics`? If is not for the whole thing maybe `SocCapacityCombinedMetrics`?

src/frequenz/sdk/timeseries/battery_pool/_result_types.py
```diff
@@ -0,0 +1,181 @@
...
+    """Sec metrics."""
+
+    capacity_metrics: CapacityMetrics
+    """Capacity metrics."""
+
+
+# Soc and capacity are different kind of metrics.
+# It makes sens to separate them with `_`.
+@dataclass
+class SoC_Capacity:  # pylint: disable=invalid-name
```
leandro-lucarella-frequenz: Same here. Maybe `SocCapacityCombined`?

src/frequenz/sdk/timeseries/battery_pool/battery_pool.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+class BatteryPool:
+    """Calculate high level metrics for a pool of the batteries.
+
+    BatterPool provided methods for fetching different battery and inverter values
+    from microgrid and aggregate then in a high level metrics.
+    Depending from the metrics the results are send in two ways:
+        * when the metric change
+        * with the resampling frequency (specified in the resampler actor outside)
```
leandro-lucarella-frequenz: I think this won't render nicely if it has the extra indent, also a new line makes it more clear.

```suggestion

    * when the metric change
    * with the resampling frequency (specified in the resampler actor outside)
```

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
```diff
@@ -0,0 +1,475 @@
...
+
+        if len(self._bat_inv_map) == 0:
+            raise ValueError(
+                f"Formula {PowerBoundsCalculator.name} requires batteries "
+                "with adjacent inverter",
+            )
+
+        if len(batteries) != len(self._bat_inv_map):
+            _logger.warning(
+                "Some batteries hasn't adjacent inverter. Use %s out of %s",
```
leandro-lucarella-frequenz: This is a 3rd warning for the missing inverter, you **really hate** missing inverters :laughing:

src/frequenz/sdk/timeseries/battery_pool/battery_pool.py
```diff
@@ -0,0 +1,142 @@
...
+            )
+        }
+
+    async def _update_battery_status(self, receiver: Receiver[BatteryStatus]):
+        async for status in receiver:
+            self._working_batteries = status.get_working_batteries(self._batteries)
+            for item in self._active_methods.values():
+                item.update_working_batteries(self._working_batteries)
+
+    async def soc_capacity(
```
leandro-lucarella-frequenz: This goes in general for the common interface for stuff returning new receivers (i.e. logical meter too). I think we should be more explicit about a **new** receiver being created, otherwise people might think they can just call `soc_capacity()` whenever they need a receiver. It will make the method names a bit uglier, but I think is more clear.

1. `p.new_soc_capacity_receiver()`?
2. Maybe return a channel(or wrapper) instead, so users do `p.soc_capacity.new_receiver()`?
3. `p.new_receiver(SOC_CAPACITY)`?

Can be left for a future PR, but I would like to address it soon, before we keep adding methods like these.

/cc @matthias-wende-frequenz

leandro-lucarella-frequenz (commented):
I did a "quick" (with big fat quotes) look and didn't review all the algorithmic stuff and focused more on structure and such.

I know it is hard for big features, but if this could have been done in smaller PRs, I think I would have made reviewing much easier.

src/frequenz/sdk/timeseries/battery_pool/battery_pool.py
leandro-lucarella-frequenz: I was thinking that with approach 2 we could add different ways to retrieve the data:

```py
current = await p.soc_capacity.current_value  # Queries the microgrid API (if it's even possible :D)
last = p.soc_capacity.last_value  # If we have a channel getting this, we could peek into it
resampled_receiver = p.soc_capacity.new_receiver()
change_receiver = p.soc_capacity.new_changes_receiver()
```

This would also remove the inconsistent interface we have right now, where `soc_capacity()` could be sending messages to the channel on every resample period, or on change, is not explicit, and it would be inconsistent for different metrics, which makes me uneasy.

We could also have `new_receiver(only_changes=True)` for example, but if we don't want to support getting just a resampled receiver, we need to assert `only_changes` is `True` at runtime, and we can´t use the type system to help us find this at "compile-time".

examples/battery_pool.py
ela-kotulska-frequenz: Yes :D How soon we will have time for this :)

src/frequenz/sdk/timeseries/battery_pool/battery_pool.py
leandro-lucarella-frequenz: https://github.com/frequenz-floss/frequenz-sdk-python/discussions/215

src/frequenz/sdk/actor/power_distributing/_battery_pool_status.py
ela-kotulska-frequenz: Because it is not needed as an instance variable. 
There is already lots of instance variables.

src/frequenz/sdk/microgrid/component/battery_utils.py
ela-kotulska-frequenz: Actually component graph has now methods to find successors/predecessors.
And I prefer successor/predecessor instead of `adjacent` because it is more error resilient.

src/frequenz/sdk/actor/power_distributing/_battery_pool_status.py

src/frequenz/sdk/microgrid/component/battery_utils.py
ela-kotulska-frequenz: Component graph has already method to find successors and predecessors.
They are well defined and I won't add unknown `adjacent` there.
This method is highly related with battery because we are sure inverter is direct predecessor of the battery.

I moved battery_inverter_mapping to the battery_pool module. It is now private.

src/frequenz/sdk/microgrid/component/battery_utils.py
ela-kotulska-frequenz: Module removed. 
Method moved to the BatteryPool.

src/frequenz/sdk/microgrid/component/battery_utils.py
ela-kotulska-frequenz: > Do we really want to log an error each time someone tries to find an inverter for a battery and it's not there? This error is also duplicated when the battery_inverter_mapping() function is used.

This error tells there is no component with this id in component graph. So yes it would be nice to print it :D

> Also, if we keep it, I would write a more descriptive error instead of printing the err object raw.

It is already descriptive enough in the ComponentGraph.

> BTW, is a microgrid in a valid state if it has batteries without inverters? Because if it's not maybe this should be an error instead.

It is valid state to have inverter without battery. Few months ago we added it.
I think component graph allows for battery without inverter, so it is valid state, too. (I can check it if necessary) 
Anyway these requirements can change in the future, so I prefer to check it here. It is one-time check and cost me nothing.

src/frequenz/sdk/timeseries/battery_pool/__init__.py
ela-kotulska-frequenz: Done, changed to `Manage a pool of batteries.`
Because it will probably charge/discharge batteries in the future.

src/frequenz/sdk/timeseries/battery_pool/_component_metric_fetcher.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/timeseries/battery_pool/_component_metric_fetcher.py
ela-kotulska-frequenz: I simplified whole log message

src/frequenz/sdk/timeseries/battery_pool/_component_metric_fetcher.py
ela-kotulska-frequenz: I simplified whole log message

src/frequenz/sdk/timeseries/battery_pool/_component_metric_fetcher.py
ela-kotulska-frequenz: Removed.
Resampler was needed for `active power` only , but we will use FormulaEngine for `active power`.

src/frequenz/sdk/timeseries/battery_pool/_component_metric_fetcher.py
ela-kotulska-frequenz: Yes, but was formula was computed for  working  components only.
In the end it turns out we need to compute active_power for all batteries in pool. (independently if they are working or not).
So I removed this class and I will use FormulaEngine for active power in future (next PR)

src/frequenz/sdk/timeseries/battery_pool/_methods.py
ela-kotulska-frequenz: Done : changed to `AggregateMethod`

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
ela-kotulska-frequenz: No, I use is outside in battery_pool module to create unique name for the active methods

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
ela-kotulska-frequenz: I separated it as we discussed on the meeting

src/frequenz/sdk/timeseries/battery_pool/_result_types.py
ela-kotulska-frequenz: I separated it as we discussed on the meeting

src/frequenz/sdk/timeseries/battery_pool/_result_types.py
ela-kotulska-frequenz: I separated it as we discussed on the meeting

src/frequenz/sdk/timeseries/battery_pool/battery_pool.py
ela-kotulska-frequenz: Removed. We have only metrics sending on update in this PR.

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
ela-kotulska-frequenz: I don't hate them :D  

It is just still unknown how much inverters battery can have.
First we could have ONLY one inverter per one battery.
Now it can have ZERO or ONE inverter per one battery
In the future we plan to have [0-inf] inverters per one battery. 

But you are right . There are to much logs.  I simplified them

ela-kotulska-frequenz:
I have to fix commit messages and add unit tests.
But implementation is ready for the review

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
```diff
@@ -0,0 +1,478 @@
...
+from ...microgrid.component import ComponentCategory, ComponentMetricId
+from ...timeseries import Sample
+from ._component_metrics import ComponentMetricsData
+from ._result_types import Bound, CapacityMetrics, PowerBoundsMetrics, SoCMetrics
+
+_logger = logging.getLogger(__name__)
+_MIN_TIMESTAMP = datetime.min.replace(tzinfo=timezone.utc)
+
+
+def battery_inverter_mapping(batteries: Iterable[int]) -> Dict[int, int]:
```
leandro-lucarella-frequenz: Super nitpick, but if this is only used in this module I would make it even *more private*.
```suggestion
def _battery_inverter_mapping(batteries: Iterable[int]) -> Dict[int, int]:
```

Or just move it as a private method of `PowerBoundsCalculator` to make the scope even more limited if it's just an utility function used only by that class.

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
```diff
@@ -0,0 +1,478 @@
...
+    Returns:
+        Mapping between battery and adjacent inverter.
+    """
+    graph = microgrid.get().component_graph
+    bat_inv_map: Dict[int, int] = {}
+    for battery_id in batteries:
+        try:
+            predecessors = graph.predecessors(battery_id)
+        except KeyError as err:
+            _logger.error(str(err))
```
leandro-lucarella-frequenz:
```suggestion
        except KeyError:
            _logger.exception("Can't find predecessors for battery with ID %s", battery_id)
```

src/frequenz/sdk/microgrid/component/battery_utils.py
leandro-lucarella-frequenz: Yeah, what I am more concerned about is *false-positive* errors. I.e. if it is normal for a battery not to have an inverter, when we shouldn't print a warning or an error, because then we get the error logs littered with stuff that it's just normal and then it is really easy to miss real errors in the noise.

src/frequenz/sdk/timeseries/battery_pool/_component_metric_fetcher.py
```diff
@@ -0,0 +1,248 @@
...
+        Returns:
+            This class instance
+        """
+        self: LatestMetricsFetcher[T] = await super().async_new(component_id, metrics)
+
+        for metric in metrics:
+            # pylint: disable=protected-access
+            if metric not in self._supported_metrics():
+                category = self._component_category()
+                raise ValueError(f"Metric {metric} not supported fro the {category}")
```
leandro-lucarella-frequenz: Small typo + minor improvement suggestion:

```suggestion
                raise ValueError(f"Metric {metric} not supported for {category}")
```

src/frequenz/sdk/timeseries/battery_pool/_component_metric_fetcher.py
```diff
@@ -0,0 +1,248 @@
...
+            None if the channel was closed and fetching next element is impossible.
+        """
+        try:
+            data = await self._receiver.receive()
+        except ChannelClosedError as err:
+            _logger.error(
+                "Channel for component %d was closed. Err: %s",
+                self._component_id,
+                str(err),
+            )
```
leandro-lucarella-frequenz:
```suggestion
        except ChannelClosedError:
            _logger.exception(
                "Channel for component %d was closed.",
                self._component_id,
            )
```

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
leandro-lucarella-frequenz: OK, couldn't this just be `Capacity` now?

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
```diff
@@ -0,0 +1,478 @@
...
+
+            result.total_capacity += capacity
+            result.bound.upper += capacity * soc_upper_bound
+            result.bound.lower += capacity * soc_lower_bound
+
+        return None if result.timestamp == _MIN_TIMESTAMP else result
+
+
+class SoCCalculator(MetricCalculator[SoCMetrics]):
+    """Define how to calculate Capacity metrics."""
```
leandro-lucarella-frequenz: Shouldn't this be `SoC` instead of `Capacity`?

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
```diff
@@ -0,0 +1,478 @@
...
+        ]
+
+    @classmethod
+    def name(cls) -> str:
+        """Return name of the calculator.
+
+        Returns:
+            Name of the calculator
+        """
+        return "Soc_Capacity"
```
leandro-lucarella-frequenz: And this `"SoC"` too?

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
```diff
@@ -0,0 +1,478 @@
...
+        ]
+
+    @classmethod
+    def name(cls) -> str:
+        """Return name of the calculator.
+
+        Returns:
+            Name of the calculator
+        """
+        return "PowerBoundsMetrics"
```
leandro-lucarella-frequenz: Maybe `"PowerBounds"` without `Metrics` to be consistent with other names?

src/frequenz/sdk/timeseries/battery_pool/_component_metric_fetcher.py
```diff
@@ -0,0 +1,248 @@
...
+    def _extract_metric(self, data: T, mid: ComponentMetricId) -> float:
+        pass
+
+    @abstractmethod
+    def _supported_metrics(self) -> Set[ComponentMetricId]:
+        pass
+
+    @abstractmethod
+    def _component_category(self) -> ComponentCategory:
+        pass
```
leandro-lucarella-frequenz: Funny that you don't get a complain from `mypy` here. Usually `pass` means that you have an implementation that does nothing, so it should complain about not returning a `float`. For abstract methods you should use `...` instead.
```suggestion
    @abstractmethod
    def _extract_metric(self, data: T, mid: ComponentMetricId) -> float:
        ...

    @abstractmethod
    def _supported_metrics(self) -> Set[ComponentMetricId]:
        ...

    @abstractmethod
    def _component_category(self) -> ComponentCategory:
        ...
```

src/frequenz/sdk/timeseries/battery_pool/_component_metrics.py
```diff
@@ -0,0 +1,89 @@
...
+
+        Returns:
+            True if two objects are equal, false otherwise.
+        """
+        if not isinstance(other, ComponentMetricsData):
+            _logger.error(
+                "Comparing %s with %s not supported",
+                type(ComponentMetricsData),
+                type(other),
+            )
```
leandro-lucarella-frequenz: I find this odd, usually in Python it is valid to compare 2 arbitrary types, it will just return `False` if they are not equal. Do we really want to print an error here? I would avoid it unless there is a very good reason.

src/frequenz/sdk/timeseries/battery_pool/_methods.py
```diff
@@ -0,0 +1,265 @@
...
+    """
+
+    _working_batteries: Set[int]
+    _metric_calculator: MetricCalculator[T]
+
+    _result_channel: Broadcast[T]
+    _cached_metrics: Dict[int, ComponentMetricsData]
+
+    _update_task: asyncio.Task[None]
+    _send_task: asyncio.Task[None]
```
leandro-lucarella-frequenz: Just out of curiosity, is this intentional or was this a `AsyncConstructible` class before and is a left over? I remember us talking in a meeting about the possibility to declare types in the class instead of the `__init__()`, but I'm not sure if we agreed on doing so, one problem was that documentation needs to be in the `__init__()` to render nicely, so I think we should still prefer `__init__()` even if we don't really need to add documentation for private members.

We can still have a block defining types at the beginning of the `__init__()`, for example:

```py
    def __init__(...) -> None:
        _working_batteries: Set[int]
        """The set of batteries that are currently working."""

        _metric_calculator: MetricCalculator[T]
        """The metric calculator used to..."""

        # Codes with assignments goes here.
```

src/frequenz/sdk/timeseries/battery_pool/_methods.py
```diff
@@ -0,0 +1,265 @@
...
+    async def stop(self) -> None:
+        """Stop the all pending async task."""
+        await asyncio.gather(
+            *[cancel_and_await(task) for task in self._pending_data_fetchers],
+            return_exceptions=True,
+        )
+        await asyncio.gather(
+            *[cancel_and_await(self._send_task), cancel_and_await(self._update_task)],
+            return_exceptions=True,
+        )
```
leandro-lucarella-frequenz: So you are just silently swallowing exceptions? Maybe at least print a log if there are any?

In any case all this should be handled by the `BackgroundService` class (if I ever find some time to take care of it :grimacing: ).

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
```diff
@@ -0,0 +1,478 @@
...
+    @abstractmethod
+    def name(cls) -> str:
+        """Return name of the formula.
+
+        Returns:
+            Name of the formula
+        """
+
+    @property
+    def batteries(self) -> Set[int]:
```
leandro-lucarella-frequenz, line 91: Warning: This is using `typing.Set` which is analogous to `builtins.set`, which means the user will be able to mutate the set, which is probably not what you want since you made it a property. If you use [`collections.abc.Set`](https://docs.python.org/3.11/library/collections.abc.html#collections.abc.Set) instead, then is an interface for a immutable set, so the user won't be allowed to change it (if they run `mypy` that is, of course at runtime they can do whatever they want :D)

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
```diff
@@ -0,0 +1,478 @@
...
+        batteries that can be used in the calculator.
+
+        Returns:
+            Set of batteries that should be used.
+        """
+        return self._batteries
+
+    @property
+    @abstractmethod
+    def battery_metrics(self) -> Dict[int, List[ComponentMetricId]]:
```
leandro-lucarella-frequenz: Maybe for this too, not sure how is it intended to be used, but maybe [`collections.abc.Mapping`](https://docs.python.org/3.11/library/collections.abc.html#collections.abc.Mapping) and [`collections.abc.Sequence`](https://docs.python.org/3.11/library/collections.abc.html#collections.abc.Sequence).

Etc...

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
```diff
@@ -0,0 +1,478 @@
...
+
+            # All metrics are related so if any is missing then we skip the component.
+            if None in {capacity, soc_lower_bound, soc_upper_bound}:
+                continue
+
+            result.total_capacity += capacity
+            result.bound.upper += capacity * soc_upper_bound
+            result.bound.lower += capacity * soc_lower_bound
+
+        return None if result.timestamp == _MIN_TIMESTAMP else result
```
leandro-lucarella-frequenz, line 233: I wonder how expensive are the calculations above, and if it wouldn't make sense to first do a pass over the timestamps to see if it makes sense to calculate anything at all.

I guess now you are optimizing for the case where everything is working normally and what I propose will be optimizing for the case where data is missing, so maybe is not worth it, but sharing it just in case. :)

src/frequenz/sdk/timeseries/battery_pool/_result_types.py
```diff
@@ -0,0 +1,131 @@
...
+    total_capacity = sum(battery.capacity for battery in working_batteries)
+    
+    bound.lower = capacity_lower_bound/total_capacity
+    bound.upper = capacity_upper_bound/total_capacity
+
+    """
+
+
+@dataclass
+class PowerBoundsMetrics:
```
leandro-lucarella-frequenz: I wonder if it wouldn't be nicer for the user if this is called `PowerMetrics` instead, since then you have `bound` as memebers so it gets a bit redundant to say `PowerBoundsMetrics.supply_bound` for example. Also not sure if this could ever happen, but maybe in the future we need to add more power related metrics?

leandro-lucarella-frequenz (commented):
Most very minor comments. I love how such an amount of complexity gets hidden in a very simple and clean interface :tada:

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
ela-kotulska-frequenz: _logger.exception will print whole traceback, too. It is ok to us it for logging really important error messages.
This message is not so important and it doesn't need traceback.

src/frequenz/sdk/microgrid/component/battery_utils.py
ela-kotulska-frequenz: True, thanks for noting that!
I will change it to `info` log :)

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
ela-kotulska-frequenz: hmm good point, however it is extremely rare that all components of certain type are broken.
It would add extra complexity. And I would (most probably) have to check it anyway because mypy /pylint doesn't see that it is not None. :(

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
ela-kotulska-frequenz: Done, thanks!

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/timeseries/battery_pool/_component_metric_fetcher.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/timeseries/battery_pool/_component_metric_fetcher.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/timeseries/battery_pool/_component_metric_fetcher.py
ela-kotulska-frequenz: Thanks, fixed.

src/frequenz/sdk/timeseries/battery_pool/_component_metrics.py
ela-kotulska-frequenz: Removed

src/frequenz/sdk/timeseries/battery_pool/_methods.py
ela-kotulska-frequenz: Yes, it was AsyncConstructible before.
But it is not anymore so I removed it.

src/frequenz/sdk/timeseries/battery_pool/_methods.py
ela-kotulska-frequenz: It should not raise exception, because exception is catched in the `cancel_and_await`. 
If for some reason it will raise and exception then something is really really wrong and should not be catched.
So I removed `return_exceptions=True`, thanks

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
ela-kotulska-frequenz: Done, thanks!

src/frequenz/sdk/timeseries/battery_pool/_result_types.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
ela-kotulska-frequenz: I forgot, about this comment. I will fix it tomorrow.

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
ela-kotulska-frequenz: Unfortunately it doesn't work for python3.8 :( 
```
_____ ERROR collecting tests/timeseries/_battery_pool/test_battery_pool.py _____
tests/timeseries/_battery_pool/test_battery_pool.py:22: in <module>
    from frequenz.sdk.timeseries.battery_pool import (
src/frequenz/sdk/timeseries/battery_pool/__init__.py:7: in <module>
    from .battery_pool import BatteryPool
src/frequenz/sdk/timeseries/battery_pool/battery_pool.py:23: in <module>
    class BatteryPool:
src/frequenz/sdk/timeseries/battery_pool/battery_pool.py:40: in BatteryPool
    batteries_id: Set[int] | None = None,
E   TypeError: 'ABCMeta' object is not subscriptable
```

ela-kotulska-frequenz: I can surround it with ditto mark ( batteries_id: "Set[int]" | None = None, )
But doing it everywhere seems to much.
We can change it when we delete support for python3.8.

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
leandro-lucarella-frequenz: I can't see that, in the current state I see the name is still public but now it's being used somewhere else (`_methods.py`), so all good anyway. Feel free to resolved this one!

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
leandro-lucarella-frequenz: In my experience it is always nice to have a backtrace in case you need to investigate something a bit more, which is really hard when problems happen in production and you can't reproduce it easily afterwards. I'm concern about thinking that this error could only happen when a very obvious condition happens too, so normally it should be obvious to figure out why this happen, but what if this is not true for some obscure case we didn't foresee?

So my question is, do you **really** want to avoid printing the backtrace? What would be the disadvantage or doing so?

src/frequenz/sdk/actor/power_distributing/_battery_pool_status.py
leandro-lucarella-frequenz: For what's worth, I agree with Sahas that it would make sense to make it an instance variable, it seems like something very inherent to the class. If at some point you would want to update the sender, it won't be possible if you pass it around. But for me is minor and I'm fine to leave it as it is too if you feel it shouldn't be an instance variable, in the extremely rare event that we want to give the option to replace the sender in the future we can change it then.

leandro-lucarella-frequenz (dismissed):
I think this is good enough to give an approval, it is a quite big PR and there are only very minor things left and I think we can discuss it forever, but it is important to get it in.

For me there is only https://github.com/frequenz-floss/frequenz-sdk-python/pull/205#discussion_r1126277745 left, but I consider it also minor, I just don't see a reason not to go with the traceback and get all the information we can when there is a problem :)

I leave it to you @ela-kotulska-frequenz to decide if to merge it or wait for Sahas to confirm he's fine too.

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
ela-kotulska-frequenz: I remember it was hard for me to read logs from async application where you have traceback for every single (not very important) error message.
At some point people will learn to ignore such traceback in logs.

I would assume logging.exception is for very important exceptions where this traceback is really needed.

ela-kotulska-frequenz:
Only rebased so I merge without re-aproval

ela-kotulska-frequenz:
> to decide if to merge it or wait for Sahas to confirm he's fine too.

We can make fixes when Sahas come back .
Also giving someone so long PR immediately after holiday is to cruel for me :D

PR #208, opened 2023-02-16
Rename microgrid.Microgrid to microgrid.ConnectionManager and its Insecure-companion

### LINKED ISSUE #90: Rename `.microgrid.Microgrid` to `MicrogridConnectionManager`

christianparpart:
Closes #90

Mind, I did not rename from `microgrid.Microgrid` to `microgrid.MicrogridConnectionManager` but to `microgrid.ConnectionManager` as this reads way less repeatitive. Also, this "type is not meant to be used by users directly". :)

### DIFF
--- a/src/frequenz/sdk/microgrid/__init__.py
+++ b/src/frequenz/sdk/microgrid/__init__.py
@@ -9,11 +9,11 @@ for monitoring and adjusting the state of a microgrid.
 
 from . import client, component
 from ._graph import ComponentGraph
-from ._microgrid import Microgrid, get, initialize
+from ._microgrid import ConnectionManager, get, initialize
 
 __all__ = [
     "ComponentGraph",
-    "Microgrid",
+    "ConnectionManager",
     "get",
     "initialize",
     "client",
--- a/src/frequenz/sdk/microgrid/_microgrid.py
+++ b/src/frequenz/sdk/microgrid/_microgrid.py
@@ -1,7 +1,7 @@
 # License: MIT
 # Copyright © 2022 Frequenz Energy-as-a-Service GmbH
 
-"""Microgrid singleton abstraction.
+"""Microgrid Connection Manager singleton abstraction.
 
 This module provides a singleton abstraction over the microgrid. The main
 purpose is to provide the connection the microgrid API client and the microgrid
@@ -22,7 +22,7 @@ _DEFAULT_MICROGRID_HOST = "[::1]"
 _DEFAULT_MICROGRID_PORT = 443
 
 
-class Microgrid(ABC):
+class ConnectionManager(ABC):
     """Creates and stores core features."""
 
     def __init__(self, host: str, port: int) -> None:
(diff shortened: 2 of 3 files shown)


ela-kotulska-frequenz (approved):

shsms:
Could you rename the file also to `_connection_manager.py`

leandro-lucarella-frequenz:
> Mind, I did not rename from `microgrid.Microgrid` to `microgrid.MicrogridConnectionManager` but to `microgrid.ConnectionManager` as this reads way less repeatitive. Also, this "type is not meant to be used by users directly". :)

Yeah, we definitely have a style issue here, `microgrid.ConnectionManager` is definitely the way to go if we stick with the coding style we say we are using (https://github.com/frequenz-io/rnd/blob/master/doc/coding-standards/python.md#style says we use https://google.github.io/styleguide/pyguide.html#22-imports, so `from frequenz.sdk import microgrid; microgrid.ConnectionManager`).

Right now I think we are using mostly a style that allows us to import symbols directly, without a namespace (`from frequenz.sdk.microgrid import MicrogridConnectionManager; MicrogridConnectionManager`).

I think we need to decide for one and stick with it (and update the docs if necessary). I prefer to stick to Google style because otherwise we need to add exceptions to Google style as we want to follow it in general. Also is a little less typing (but harder to get context if you don't have the source file name at hand).

leandro-lucarella-frequenz:
* https://github.com/frequenz-floss/frequenz-sdk-python/discussions/209

leandro-lucarella-frequenz (commented):
Just for the records, we can merge this as is as this type is mostly internal.

leandro-lucarella-frequenz (approved):

christianparpart:
> Just for the records, we can merge this as is as this type is mostly internal.

Yes, also, I just did a rename locally and then discarded it again, it's actually cleaner in the caller side the way it is, to be honest. :)

shsms:
looks like we forgot to rename `_microgrid.py` and the `_MICROGRID` object,  so we still have `frequenz.sdk.microgrid._microgrid._MICROGRID` :D

PR #211, opened 2023-02-20
benchmark data sourcing actor

### LINKED ISSUE #198: Benchmarks for the data sourcing actor
### What's needed?

We need benchmarks for the data sourcing actor, where we see how long
it takes to stream a certain number of values for a certain number of
microgrid components.

### Proposed solution

Prereqs:
1. The `MockMicrogrid` package in tests currently [sends out 2000](https://github.com/frequenz-floss/frequenz-sdk-python/blob/2524cb2dddeeed00289e46aab4af8f0544ae9a33/tests/timeseries/mock_microgrid.py#L110)
   values per component_id.  This needs to be configurable.
   
2. The [`start`](https://github.com/frequenz-floss/frequenz-sdk-python/blob/0bcf61cb19ff52e7c64db14acdde11867c61f236/tests/timeseries/mock_microgrid.py#L89) method starts the data sourcing actor and the
   resampling actor, but sends out the channel only for the resampling
   actor.  There should ideally be a `start_mock_client` method that
   just sets up the mock client, which the `start` method calls.  Then
   we would be able just call `start_mock_client` and then create our
   own actor instances.

The data sourcing actor can be benchmarked by a function that takes as
input:
  - number of components
  - the number of messages to be sent out for each battery as inputs

This benchmarking function can be implemented like this:

1. Create a `MockMicrogrid` instance, with the given number of ev
   chargers like [here](https://github.com/frequenz-floss/frequenz-sdk-python/blob/2524cb2dddeeed00289e46aab4af8f0544ae9a33/tests/timeseries/test_logical_meter.py#L96-L99).
2. the `MockMicrogrid` has a default `sample_rate_s` of 0.01, this can
   be set to 0.0 (or almost 0.0 if there are synchronization issues).
3. Create a `DataSourcingActor` instance, and send requests to it for
   each of the [3 current phases](https://github.com/frequenz-floss/frequenz-sdk-python/blob/0bcf61cb19ff52e7c64db14acdde11867c61f236/src/frequenz/sdk/microgrid/component/_component.py#L126-L128),
   for all component ids in `MockMicrogrid.evc_ids`.  Collect the
   corresponding receivers into a `channels.MergeNamed` and count the
   number of messages received from each component_id+metrid_id pair.
4. Confirm the number of messages sent is the number of messages
   received.  Measure the time required to stream the configured
   number of messages.

### Use cases

_No response_

_No response_

### Additional context

_No response_

mathias-baumann-frequenz:
Apart from the benchmark, the following preparations where done:

- Add num_values to have a configurable amount of iterations
- Add start_mock_client method
- MockMicrogrid: Make it work outside of pytest as well

The benchmark is functional, but both the output and how/what exactly we measure
needs work. The main reason for the draft is the mocker problem (see comments in
PR)

### DIFF
--- a/benchmarks/timeseries/benchmark_datasourcing.py
+++ b/benchmarks/timeseries/benchmark_datasourcing.py
@@ -1,0 +1,190 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Benchmark the data sourcing actor.
+
+To be able to access the `tests` package we need to adjust the PYTHONPATH.
+
+Usage:
+
+    PYTHONPATH=. python benchmark_datasourcing.py <num ev chargers> <num messages per battery>
+"""
+import argparse
+import asyncio
+import sys
+import tracemalloc
+from time import perf_counter
+from typing import Any, Tuple
+
+from frequenz.channels import Broadcast, Receiver, ReceiverStoppedError
+
+from frequenz.sdk import microgrid
+from frequenz.sdk.actor import (
+    ChannelRegistry,
+    ComponentMetricRequest,
+    DataSourcingActor,
+)
+from frequenz.sdk.microgrid.component import ComponentMetricId
+
+try:
+    from tests.timeseries.mock_microgrid import MockMicrogrid
+    from tests.utils import MockMicrogridClient
+except ImportError:
+    print(
+        "Error: Unable to import the `tests` package. "
+        "Please make sure that the PYTHONPATH env variable is set to the root of the repository."
+    )
+    sys.exit(1)
+
+COMPONENT_METRIC_IDS = [
+    ComponentMetricId.CURRENT_PHASE_1,
+    ComponentMetricId.CURRENT_PHASE_2,
+    ComponentMetricId.CURRENT_PHASE_3,
+]
+
+
+def enable_mock_client(client: MockMicrogridClient) -> None:
+    """Enable the mock microgrid client.
+
+    Args:
+        client: the mock microgrid client to enable.
+    """
+    # pylint: disable=protected-access
+    microgrid.connection_manager._CONNECTION_MANAGER = client.mock_microgrid
+
+
+# pylint: disable=too-many-locals
+async def benchmark_data_sourcing(
+    num_ev_chargers: int, num_msgs_per_battery: int
+) -> None:
+    """Benchmark the data sourcing actor.
+
+    Benchmark the data sourcing actor by sending out a number of requests and
+    printing out the number of samples sent and the time taken.
+
+    Args:
+        num_ev_chargers: number of EV Chargers to create for the mock microgrid.
+        num_msgs_per_battery: number of messages to send out for each battery.
+    """
+    num_expected_messages = (
+        num_ev_chargers * len(COMPONENT_METRIC_IDS) * num_msgs_per_battery
+    )
+    mock_grid = MockMicrogrid(
+        grid_side_meter=False, num_values=num_msgs_per_battery, sample_rate_s=0.0
+    )
+
+    mock_grid.add_ev_chargers(num_ev_chargers)
+    mock_grid.start_mock_client(enable_mock_client)
+
+    request_channel = Broadcast[ComponentMetricRequest](
+        "DataSourcingActor Request Channel"
+    )
+
+    channel_registry = ChannelRegistry(name="Microgrid Channel Registry")
+    request_receiver = request_channel.new_receiver(
+        "datasourcing-benchmark", maxsize=(num_ev_chargers * len(COMPONENT_METRIC_IDS))
+    )
+    request_sender = request_channel.new_sender()
+
+    consume_tasks = []
+
+    start_time = perf_counter()
+    samples_sent = 0
+
+    async def consume(channel: Receiver[Any]) -> None:
+        while True:
+            try:
+                await channel.ready()
+                channel.consume()
+            except ReceiverStoppedError:
+                return
+
+            nonlocal samples_sent
+            samples_sent += 1
+
+    for evc_id in mock_grid.evc_ids:
+        for component_metric_id in COMPONENT_METRIC_IDS:
+            request = ComponentMetricRequest(
+                "current_phase_requests", evc_id, component_metric_id, None
+            )
+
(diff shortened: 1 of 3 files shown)


tests/utils/mock_microgrid.py
```diff
@@ -71,16 +74,24 @@ def __init__(self, components: Set[Component], connections: Set[Connection]):
...
+        params = ("frequenz.sdk.microgrid._microgrid._MICROGRID", self.mock_microgrid)
         # Mock _MICROGRID, so `get` method return this mocked microgrid.
-        mocker.patch(
-            "frequenz.sdk.microgrid._microgrid._MICROGRID", self.mock_microgrid
-        )
+        if mocker is None:
+            from unittest.mock import patch
+
+            # Not working
+            patch(*params)
```
mathias-baumann-frequenz: This patch call is not working at all.

benchmarks/timeseries/benchmark_datasourcing.py
```diff
@@ -0,0 +1,116 @@
...
+        grid_side_meter=False,
+        num_values=num_msgs_per_battery,
+        sample_rate_s=0.00010,
+    )
+
+    mock_grid.add_ev_chargers(num_evc_chargers)
+
+    client = mock_grid.start_mock_client()
+    # Not working
+    # client = mock_grid.start_mock_client(mock_microgrid)
```
mathias-baumann-frequenz: As the mocker seems to be more like a context object, I thought I try it like this, with the patch decorator at the top, but passing this to the method also didn't work

mathias-baumann-frequenz:
CI will fail until we use the new changes in the channels repo.

benchmarks/timeseries/benchmark_datasourcing.py
```diff
@@ -0,0 +1,170 @@
...
+        num_msgs_per_battery_param,
+        record_allocations,
+    ) = parse_args()
+
+    if record_allocations:
+        tracemalloc.start()
+
+    asyncio.run(
+        benchmark_data_sourcing(num_evc_chargers_pararm, num_msgs_per_battery_param)
+    )
```
daniel-zullo-frequenz, line 176: This is usually not a good practice to run programs/applications
https://github.com/frequenz-io/rnd/blob/master/doc/coding-standards/python.md#defining-applications

benchmarks/timeseries/benchmark_datasourcing.py
```diff
@@ -0,0 +1,170 @@
...
+
+def parse_args() -> Tuple[int, int, bool]:
+    """Parse the command line arguments.
+
+    Returns:
+        A tuple of (num evc chargers, num messages per battery, record allocations).
+    """
+    parser = argparse.ArgumentParser(description="Benchmark the data sourcing actor.")
+    parser.add_argument(
+        "num_evc_chargers",
```
daniel-zullo-frequenz: why `num_evc_chargers` and not `num_ev_chargers`? I personally found a bit redundant the `c` in `evc` but I might be missing something?

benchmarks/timeseries/benchmark_datasourcing.py
```diff
@@ -0,0 +1,170 @@
...
+    num_evc_chargers: int, num_msgs_per_battery: int
+) -> None:
+    """Benchmark the data sourcing actor.
+
+    Args:
+        num_evc_chargers: number of EV Chargers to create for the mock microgrid.
+        num_msgs_per_battery: number of messages to send out for each battery.
+
+    Returns:
+        A tuple of (number of messages sent, time taken to stream the messages).
```
daniel-zullo-frequenz: the return of the function is `None`

benchmarks/timeseries/benchmark_datasourcing.py
```diff
@@ -0,0 +1,185 @@
...
+            async def consume(channel: Receiver[Any]) -> None:
+                while True:
+                    try:
+                        await channel.ready()
+                        channel.consume()
+                    except ReceiverStoppedError:
+                        return
+
+                    nonlocal samples_sent
+                    samples_sent += 1
```
daniel-zullo-frequenz: can this inner function be define outside/above the for-loops instead?

benchmarks/timeseries/benchmark_datasourcing.py
```diff
@@ -0,0 +1,185 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
```
daniel-zullo-frequenz:
```suggestion
# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
```

daniel-zullo-frequenz (commented):
I have a few suggestions to check for

benchmarks/timeseries/benchmark_datasourcing.py
daniel-zullo-frequenz: If this comment applies then the replacement should be done for each definition of the variable

benchmarks/timeseries/benchmark_datasourcing.py
```diff
@@ -0,0 +1,190 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Benchmark the data sourcing actor.
+
+To be able to access the `tests` package we need to adjust the PYTHONPATH.
+
+Usage:
+
+    PYTHONPATH=. python benchmark_datasourcing.py <num evc chargers> <num messages per battery>
```
daniel-zullo-frequenz: s/evc/ev

benchmarks/timeseries/benchmark_datasourcing.py
```diff
@@ -0,0 +1,190 @@
...
+    print(
+        f"Missing per EVC: {(num_expected_messages - samples_sent) / num_ev_chargers}"
+    )
+
+
+def parse_args() -> Tuple[int, int, bool]:
+    """Parse the command line arguments.
+
+    Returns:
+        A tuple of (num evc chargers, num messages per battery, record allocations).
```
daniel-zullo-frequenz: s/evc/ev

daniel-zullo-frequenz (dismissed):
Other than a couple of cosmetic comments, LGTM!

mathias-baumann-frequenz:
Fixed decorative problems

PR #219, opened 2023-02-22
Fix mypy invocation in noxfile

### LINKED ISSUE #218: mypy checks are broken
### What happened?

It turns out `mypy` has not been running on the sdk codebase for the past 3 months due to a bug in the noxfile,  in: https://github.com/frequenz-floss/frequenz-sdk-python/blob/70d9d43f7f37807baf98e6896a330588b24673da/noxfile.py#L194-L198

mypy to run as part of the CI.

### Affected version(s)

_No response_

### Affected part(s)

### Extra information

_No response_

leandro-lucarella-frequenz:
When invoking `mypy`, the `src` directory (`frequenz.sdk` package) isn't being checked because 2 bugs:

* The `src` path is excluded
* A transformation of `src` into a package name was a ill done copy&paste from the channels repository, and said `frequenz.channels` instead of `frequenz.sdk`

This commit fixes this issue by reworking how the conversion between paths and packages needed by `mypy` is done. Now the conversion is more generic and explicit, providing a list of paths and a mapping between paths and packages.

This is a bit more verbose to declare but it can be reused more safely and gives some extra flexibility, which can be handy when we move this to a general tooling repo.

We also use package names for plain python files (like `noxfile.py`), so we don't need to invoke `mypy` twice.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -52,8 +52,24 @@ Usage:
 
 from __future__ import annotations
 
+from typing import Any, Iterable
+
 import nox
 import toml
+
+DEFAULT_PATH_PACKAGES = {
+    "benchmarks": "benchmarks",
+    "docs": "docs",
+    "examples": "examples",
+    "src": "frequenz.sdk",
+    "tests": "tests",
+    "noxfile.py": "noxfile",
+}
+"""A list of path to be used by default and its corresponding package name.
+
+The package name is needed for mypy, as it takes packages when full import
+checking needs to be done.
+"""
 
 
 def min_dependencies() -> list[str]:
@@ -98,14 +114,7 @@ def _source_file_paths(session: nox.Session) -> list[str]:
     """
     if session.posargs:
         return session.posargs
-    return [
-        "benchmarks",
-        "docs",
-        "examples",
-        "src",
-        "tests",
-        "noxfile.py",
-    ]
+    return list(DEFAULT_PATH_PACKAGES.keys())
 
 
 # Run all checks except `ci_checks` by default.  When running locally with just
@@ -169,36 +178,24 @@ def mypy(session: nox.Session, install_deps: bool = True) -> None:
         # fast local tests with `nox -R -e mypy`.
         session.install("-e", ".[mypy]")
 
-    common_args = [
+    def _flatten(iterable: Iterable[Iterable[Any]]) -> Iterable[Any]:
+        return [item for sublist in iterable for item in sublist]
+
+    args = (
+        session.posargs
+        if session.posargs
+        else _flatten(("-p", p) for p in DEFAULT_PATH_PACKAGES.values())
+    )
+
+    session.run(
+        "mypy",
         "--install-types",
         "--namespace-packages",
         "--non-interactive",
         "--explicit-package-bases",
         "--strict",
-    ]
-
-    # If we have session arguments, we just use those...
-    if session.posargs:
-        session.run("mypy", *common_args, *session.posargs)
-        return
-
-    check_paths = _source_file_paths(session)
-    pkg_paths = [
-        path
-        for path in check_paths
-        if not path.startswith("src") and not path.endswith(".py")
-    ]
-    file_paths = [path for path in check_paths if path.endswith(".py")]
-
-    pkg_args = []
-    for pkg in pkg_paths:
-        if pkg == "src":
-            pkg = "frequenz.channels"
-        pkg_args.append("-p")
-        pkg_args.append(pkg)
-
-    session.run("mypy", *common_args, *pkg_args)
-    session.run("mypy", *common_args, *file_paths)
+        *args,
+    )
 
 
 @nox.session
--- a/noxfile.py
+++ b/noxfile.py
@@ -50,13 +50,13 @@ Usage:
         nox -R -e pytest_min -- -s -x tests/timeseries/test_logical_meter.py
 """
 
-from typing import List
+from __future__ import annotations
 
 import nox
 import toml
 
 
-def min_dependencies() -> List[str]:
+def min_dependencies() -> list[str]:
     """Extract the minimum dependencies from pyproject.toml.
 
     Raises:
@@ -74,7 +74,7 @@ def min_dependencies() -> List[str]:
     if not dependencies:
         raise RuntimeError(f"No dependencies found in file: {toml_file.name}")
 
-    min_deps: List[str] = []
+    min_deps: list[str] = []
     for dep in dependencies:
         min_dep = dep.split(",")[0]
         if any(op in min_dep for op in (">=", "==")):
@@ -84,7 +84,7 @@ def min_dependencies() -> List[str]:
     return min_deps
 
 
-def _source_file_paths(session: nox.Session) -> List[str]:
+def _source_file_paths(session: nox.Session) -> list[str]:
     """Return the file paths to run the checks on.
 
     If positional arguments are present in the nox session, we use those as the
--- a/noxfile.py
+++ b/noxfile.py
@@ -140,7 +140,7 @@ def ci_checks_max(session: nox.Session) -> None:
     Args:
         session: the nox session.
     """
-    session.install(".[dev]")
+    session.install("-e", ".[dev]")
 
     formatting(session, False)
     mypy(session, False)
@@ -158,7 +158,7 @@ def formatting(session: nox.Session, install_deps: bool = True) -> None:
         install_deps: True if dependencies should be installed.
     """
     if install_deps:
-        session.install(".[format]")
+        session.install("-e", ".[format]")
 
     paths = _source_file_paths(session)
     session.run("black", "--check", *paths)
@@ -228,7 +228,7 @@ def docstrings(session: nox.Session, install_deps: bool = True) -> None:
         install_deps: True if dependencies should be installed.
     """
     if install_deps:
-        session.install(".[docs-lint]")
+        session.install("-e", ".[docs-lint]")
 
     paths = _source_file_paths(session)
     session.run("pydocstyle", *paths)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,6 +59,7 @@ docs-gen = [
 docs-lint = [
     "pydocstyle",
     "darglint",
+    "tomli",  # Needed by pydocstyle to read pyproject.toml
 ]
 format = [
     "black",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,6 @@ dependencies = [
     "networkx >= 2.8, < 3",
     "pandas >= 1.5.2, < 2",
     "protobuf >= 4.21.6, < 5",
-    "pyarrow >= 10.0.1, < 11",
     "pydantic >= 1.9",
     "sympy >= 1.10.1, < 2",
     "toml >= 0.10",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -139,7 +139,10 @@ required_plugins = [ "pytest-asyncio", "pytest-mock" ]
 [[tool.mypy.overrides]]
 module = [
     "grpc.aio",
-    "grpc.aio.*"
+    "grpc.aio.*",
+    # There is a stubs package available, but it's not working:
+    # https://github.com/eggplants/networkx-stubs/issues/1
+    "networkx",
 ]
 ignore_missing_imports = true
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -136,6 +136,12 @@ src_paths = ["src", "examples", "tests"]
 asyncio_mode = "auto"
 required_plugins = [ "pytest-asyncio", "pytest-mock" ]
 
+[tool.mypy]
+# This is a temporary hack, for details see:
+# * https://github.com/frequenz-floss/frequenz-sdk-python/pull/219
+# * https://github.com/frequenz-floss/frequenz-sdk-python/issues/226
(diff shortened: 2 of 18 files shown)


leandro-lucarella-frequenz:
This is a draft because now `mypy` fails with some errors that were never reported before, I'll fix them at some point if nobody else did (feel free to jump in and push more commits fixing `mypy` issues).

shsms:
I'd say we add exceptions for the affected files in the noxfile and get this in,  so that new PRs get checked.  And fix the broken files in subsequent PRs.

leandro-lucarella-frequenz:
Mmm, not sure about that, I think we at least get some type checking on the affected files indirectly from other packages (like `tests`), if we add exceptions then we don't get any type checking for the affected files anymore. Not sure what is worse. I would try to fix the current issues ASAP and be done with it.

leandro-lucarella-frequenz:
If it's too difficult or takes too long, I agree we can go with the exception option.

leandro-lucarella-frequenz:
I pushed all fixes that don't require a deep knowledge on the affected modules to fix. The only missing fixes are in:

* `src/frequenz/sdk/microgrid/client/_client.py`: @ela-kotulska-frequenz maybe you can have a look at this one?

  * There are only 2 missing errors that I don't know how to fix and look like proper bugs, as they are calls to functions without passing all required arguments.
  * The changes I made also needs a review because there are some fishy things, like most calls being done with `await` to functions that are not `async`.

* `src/frequenz/sdk/timeseries/_formula_engine/*`: This is the package (several modules are affected) with the most errors and which look the harder to fix, because there is a lot of generic code with complicated types, this probably needs to wait until @sahas-subramanian-frequenz is back from holidays as it probably will take too much time for anybody else to fix it. We should probably add these files to some `mypy` exclusion to be able to merge this PR before this is fixed.

leandro-lucarella-frequenz:
I just pushed a commit to exclude the formula engine from `mypy` and created an issue to fix it separately in the future.

* https://github.com/frequenz-floss/frequenz-sdk-python/issues/226

Since this affects the whole code base, I would like to merge it ASAP, hope you can have a look at the microgrid API client next week @ela-kotulska-frequenz

mathias-baumann-frequenz:
Needs rebase

leandro-lucarella-frequenz:
> Needs rebase

Done! I increased the priority to urgent and removed the draft status because if we keep merging PRs we might be as well introducing more (uncaught) typing errors.

Ideally we should merge this before any other PR, although it still misses the fix for the 2 remaining microgrid API errors.

src/frequenz/sdk/timeseries/_ringbuffer.py
```diff
@@ -220,14 +220,17 @@ def window(
...
+                elif isinstance(self._buffer, np.ndarray):
+                    return np.concatenate(
+                        (self._buffer[start_index:], self._buffer[0:end_index])
+                    )
                 else:
-                    window = np.concatenate((window, self._buffer[0:end_index]))
-            return window
+                    raise TypeError(
+                        "Bad type, something seems to have went wrong with type checking"
+                    )
```
daniel-zullo-frequenz: I think this change will need to update the docstring to add `TypeError` in the `Raises` session

noxfile.py
```diff
@@ -50,13 +50,29 @@
...
+    "docs": "docs",
+    "examples": "examples",
+    "src": "frequenz.sdk",
+    "tests": "tests",
+    "noxfile.py": "noxfile",
+}
+"""A list of path to be used by default and its corresponding package name.
+
+The package name is needed for mypy, as it takes packages when full import
+checking needs to be doneg
```
daniel-zullo-frequenz: typo in `doneg`

src/frequenz/sdk/timeseries/_ringbuffer.py
```diff
@@ -422,10 +425,22 @@ def __setitem__(
...
+        # from the @overloads, but it's not currently working without this
+        # hack).
+        # Then we need to ignore a no-untyped-call error, for some reason it
+        # can't get the type for self._buffer.__setitem__()
+        if isinstance(index_or_slice, SupportsIndex) and isinstance(value, float):
+            self._buffer.__setitem__(index_or_slice, value)  # type: ignore[no-untyped-call]
+        elif isinstance(index_or_slice, slice) and isinstance(value, Iterable):
+            self._buffer.__setitem__(index_or_slice, value)  # type: ignore[no-untyped-call]
+        else:
+            raise TypeError("Incompatible input arguments")
```
daniel-zullo-frequenz: `TypeError` might need to be added in the docstring `Raises` session

daniel-zullo-frequenz (commented):
I have a couple of minor comments to check for, LGTM otherwise

noxfile.py
```diff
@@ -182,23 +191,13 @@ def mypy(session: nox.Session, install_deps: bool = True) -> None:
...
+        return [item for sublist in it for item in sublist]
 
-    pkg_args = []
-    for pkg in pkg_paths:
-        if pkg == "src":
-            pkg = "frequenz.channels"
-        pkg_args.append("-p")
-        pkg_args.append(pkg)
+    packages = (DEFAULT_PATH_PACKAGES[i] for i in _source_file_paths(session))
+    pkg_args = _flatten(("-p", p) for p in packages)
```
ela-kotulska-frequenz: Maybe just:
pkg_args = (("-p", p) for p in DEFAULT_PATH_PACKAGES.values())

src/frequenz/sdk/timeseries/_ringbuffer.py
leandro-lucarella-frequenz: I think this is actually a good reason to make it an `assert`, I didn't want to use `assert False`, but it doesn't make sense to document an error that will only happen if type checking is not performed.

src/frequenz/sdk/timeseries/_ringbuffer.py
leandro-lucarella-frequenz: Same as above.

src/frequenz/sdk/timeseries/_ringbuffer.py
daniel-zullo-frequenz: Right, I see your point. Though I don't know if the nox-docstrings checks will complain about it.

matthias-wende-frequenz (dismissed):

src/frequenz/sdk/timeseries/_ringbuffer.py
leandro-lucarella-frequenz: No, it doesn't.

leandro-lucarella-frequenz:
Updated fixing daniz comments and the last microgrid api issues. I also had to change a few commits because `mypy` didn't complain but `pylint` did, or test were failing.

Diff here: https://github.com/frequenz-floss/frequenz-sdk-python/compare/633c36f6e5639dd96534050ad5f89196b373e9f7..0c5316660cff4955722348698c21b883a7a585a1

leandro-lucarella-frequenz:
Mmm, tests are still failing, it seems like the ignore for `formula_engine` is not working, but it is working locally. I need to look into it :unamused:

leandro-lucarella-frequenz:
OK, I think I found the issue and it should work now. It was a problem with the `ci_checks_max` `nox` session, that is only used in the CI. This didn't install the package as editable, so the exclusion rule didn't work as it uses paths, and `src` was not found if the package is installed normally.

leandro-lucarella-frequenz:
OK, new issue comes from adding pyarrow-stubs, it seems it can't find a matching dependency for python 3.10 and 3.11 :unamused:

leandro-lucarella-frequenz:
OK, it is finally passing. I had to remove `pyarrow-stubs` because it wasn't supported in Python 3.11 (and the repo didn't existed anymore, it looked a bit fishy).

src/frequenz/sdk/microgrid/client/_client.py
```diff
@@ -212,7 +212,7 @@ async def components(self) -> Iterable[Component]:
         try:
             component_list = await self.api.ListComponents(
                 microgrid_pb.ComponentFilter(),
-                timeout=DEFAULT_GRPC_CALL_TIMEOUT,
+                timeout=int(DEFAULT_GRPC_CALL_TIMEOUT),
```
ela-kotulska-frequenz: How did you find that timeout is `int`?

I am looking and the definition and it is float:
https://github.com/grpc/grpc/blob/v1.51.3/src/python/grpcio/grpc/aio/_base_channel.py#L31

src/frequenz/sdk/microgrid/client/_client.py
```diff
@@ -266,8 +278,13 @@ async def connections(
...
                 self.components(),
-                self.api.ListConnections(
-                    connection_filter, timeout=int(DEFAULT_GRPC_CALL_TIMEOUT)
+                # grpc.aio is missing types and mypy thinks this is not
+                # awaitable, but it is
+                cast(
+                    Awaitable[Any],
+                    self.api.ListConnections(
+                        connection_filter, timeout=int(DEFAULT_GRPC_CALL_TIMEOUT)
+                    ),
```
ela-kotulska-frequenz: Did you try:
`Awaitable[Any] -> Awaitable[microgrid_pb.ConnectionList]` ?

pyproject.toml
```diff
@@ -136,6 +136,12 @@ src_paths = ["src", "examples", "tests"]
 asyncio_mode = "auto"
 required_plugins = [ "pytest-asyncio", "pytest-mock" ]
 
+[tool.mypy]
+# This is a temporary hack, for details see:
+# * https://github.com/frequenz-floss/frequenz-sdk-python/pull/219
+# * https://github.com/frequenz-floss/frequenz-sdk-python/issues/226
+exclude = ["src/frequenz/sdk/timeseries/_formula_engine/"]
```
ela-kotulska-frequenz, line 142: From type checking of from tests?

leandro-lucarella-frequenz: I don't get this. This is excluding the `_formula_engine` subdirectory because it has a lot of errors that need to be fixed in a separate PR. This is not related to tests.

noxfile.py
leandro-lucarella-frequenz: I don't get this. We need to *flatten* the `("-p", p)` tuple to then pass it as a list of arguments to `mypy`. But I think you found a bug, not sure if this is what you meant. I'm translating the output of `_source_file_paths(session)` but this could give just the positional arguments to `nox -s mypy -- args`, so it will fail in that case :unamused:

src/frequenz/sdk/microgrid/client/_client.py
leandro-lucarella-frequenz: I don't remember exactly, but it was very likely because I was hinted by `mypy`. Maybe it also gets the type info wrong as with the `request_iterator`... :thinking: 

I will try to revert it and see what's complaining about and add a `type: ignore` if it's wrong.

Thanks for checking!

src/frequenz/sdk/microgrid/client/_client.py
leandro-lucarella-frequenz: No, I was a bit lazy to try to figure out what was the correct type (and looking at types for pb is torture :scream:), I just wanted to let `gather` know that it was an awaitable.

I will try, thanks!

src/frequenz/sdk/microgrid/client/_client.py
leandro-lucarella-frequenz:
```
src/frequenz/sdk/microgrid/client/_client.py:227: error: Argument "timeout" to "__call__" of "UnaryUnaryMultiCallable" has incompatible type "float"; expected "Optional[int]"  [arg-type]
```

I'll add an ignore :+1:

src/frequenz/sdk/microgrid/client/_client.py
leandro-lucarella-frequenz: Works! :+1:

leandro-lucarella-frequenz:
Pushed the suggested fixes and also added a commit to remove the `pyarrow` dependency, as it was only used in the modules recently removed.

noxfile.py
leandro-lucarella-frequenz: I fixed this bug but I'm still unsure if this is what you saw that was wrong or it was something else :)

pyproject.toml
ela-kotulska-frequenz: Ok, then it is fine.
I thought you wanted to quiet the tests because they are failing randomly.
But it is definitely better to keep them and quiet mypy in the _formula_engine

noxfile.py
ela-kotulska-frequenz: It looks good now :) Just that you check  session.posargs: in two places (in line 189, for the first time)
```
    # If we have session arguments, we just use those...
    if session.posargs:
        session.run("mypy", *common_args, *session.posargs)
        return'
```

leandro-lucarella-frequenz: 🤦‍♂️

That explains why it actually worked when passing positional args :shrug:

ela-kotulska-frequenz: Hahahaha xD 
You seems to be so tired of this PR, that you started to belief in things that just happen in magic way :D

leandro-lucarella-frequenz: Fixed.

ela-kotulska-frequenz (approved):

noxfile.py
leandro-lucarella-frequenz: > You seems to be so tired of this PR, that you started to belief in things that just happen in magic way :D

YES!

Get off of me!

PR #224, opened 2023-02-24
Update MockMicrogrid for use with datasourcing benchmark

mathias-baumann-frequenz:
Extends the mock microgrid with features required for the datasourcing actor
benchmark which will be added in a later PR

- Add num_values to have a configurable amount of iterations
- MockMicrogrid: Make it work outside of pytest as well
- Add start_mock_client method

### DIFF
--- a/tests/timeseries/mock_microgrid.py
+++ b/tests/timeseries/mock_microgrid.py
@@ -53,8 +53,16 @@ class MockMicrogrid:  # pylint: disable=too-many-instance-attributes
 
     _microgrid: MockMicrogridClient
 
-    def __init__(self, grid_side_meter: bool, sample_rate_s: float = 0.01):
-        """Create a new instance."""
+    def __init__(
+        self, grid_side_meter: bool, num_values: int = 2000, sample_rate_s: float = 0.01
+    ):
+        """Create a new instance.
+
+        Args:
+            grid_side_meter: whether the main meter should be on the grid side or not.
+            num_values: number of values to generate for each component.
+            sample_rate_s: sample rate in seconds.
+        """
         self._components: Set[Component] = set(
             [
                 Component(1, ComponentCategory.GRID),
@@ -64,6 +72,7 @@ class MockMicrogrid:  # pylint: disable=too-many-instance-attributes
         self._connections: Set[Connection] = set([Connection(1, 4)])
         self._id_increment = 0
         self._grid_side_meter = grid_side_meter
+        self._num_values = num_values
         self._sample_rate_s = sample_rate_s
 
         self._connect_to = self.grid_id
@@ -106,7 +115,7 @@ class MockMicrogrid:  # pylint: disable=too-many-instance-attributes
     async def _comp_data_send_task(
         self, comp_id: int, make_comp_data: Callable[[int, datetime], ComponentData]
     ) -> None:
-        for value in range(1, 2000):
+        for value in range(1, self._num_values + 1):
             timestamp = datetime.now(tz=timezone.utc)
             val_to_send = value + int(comp_id / 10)
             # for inverters with component_id > 100, send only half the messages.
--- a/tests/timeseries/mock_microgrid.py
+++ b/tests/timeseries/mock_microgrid.py
@@ -94,6 +94,31 @@ class MockMicrogrid:  # pylint: disable=too-many-instance-attributes
         self._actors: list[typing.Any] = []
         self._start_meter_streaming(4)
 
+    def start_mock_client(
+        self, initialize_cb: Callable[[MockMicrogridClient], None]
+    ) -> MockMicrogridClient:
+        """Set up the mock client.
+
+        Creates the microgrid mock client, initializes it, and starts the streaming
+        tasks.
+
+        For unittests, users should use the `start()` method.
+
+        Args:
+            initialize_cb: callback to initialize the mock client.
+
+        Returns:
+            A MockMicrogridClient instance.
+        """
+        self._microgrid = MockMicrogridClient(self._components, self._connections)
+
+        initialize_cb(self._microgrid)
+
+        self._streaming_tasks = [
+            asyncio.create_task(coro) for coro in self._streaming_coros
+        ]
+        return self._microgrid
+
     async def start(
         self, mocker: MockerFixture
     ) -> Tuple[Broadcast[ComponentMetricRequest], ChannelRegistry]:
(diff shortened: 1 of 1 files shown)


tests/timeseries/mock_microgrid.py
```diff
@@ -56,8 +56,16 @@ class MockMicrogrid:  # pylint: disable=too-many-instance-attributes
...
 
-    def __init__(self, grid_side_meter: bool, sample_rate_s: float = 0.01):
-        """Create a new instance."""
+    def __init__(
+        self, grid_side_meter: bool, num_values: int = 2000, sample_rate_s: float = 0.01
+    ):
+        """Create a new instance.
+
+        Args:
+            grid_side_meter: whether the main meter is connected to the grid or not.
```
mathias-baumann-frequenz: I am not quite sure about this, can someone validate this?

shsms: I think this might be a bit better:

```suggestion
            grid_side_meter: whether the main meter should be on the grid side or not.
```

tests/utils/mock_microgrid.py
```diff
@@ -71,16 +74,21 @@ def __init__(self, components: Set[Component], connections: Set[Connection]):
             id: channel.new_sender() for id, channel in ev_charger_channels.items()
         }
 
-    def initialize(self, mocker: MockerFixture) -> None:
+    def initialize(self, mocker: MockerFixture | None) -> None:
         """Mock `microgrid.get` call to return this mock_microgrid.
 
         Args:
-            mocker: mocker from the current test
+            mocker: mocker from the current test or `None` to use default mocker.
```
shsms: how about:

``` suggestion
            mocker: mocker from the current test, or `None` to initialize in the global context.

File: tests/utils/mock_microgrid.py [hunk 3]
```diff
@@ -71,16 +74,21 @@ def __init__(self, components: Set[Component], connections: Set[Connection]):
...
         """
+        params = ("frequenz.sdk.microgrid._microgrid._MICROGRID", self.mock_microgrid)
         # Mock _MICROGRID, so `get` method return this mocked microgrid.
-        mocker.patch(
-            "frequenz.sdk.microgrid._microgrid._MICROGRID", self.mock_microgrid
-        )
+        if mocker is None:
+            # This is a hack, optimally we should use mocker.patch, but I
+            # couldn't get it to work
```
shsms: I don't think it is so bad as this comment makes it sound :D

tests/utils/mock_microgrid.py
```diff
@@ -71,16 +74,21 @@ def __init__(self, components: Set[Component], connections: Set[Connection]):
...
 
-    def initialize(self, mocker: MockerFixture) -> None:
+    def initialize(self, mocker: MockerFixture | None) -> None:
         """Mock `microgrid.get` call to return this mock_microgrid.
 
         Args:
-            mocker: mocker from the current test
+            mocker: mocker from the current test or `None` to use default mocker.
         """
+        params = ("frequenz.sdk.microgrid._microgrid._MICROGRID", self.mock_microgrid)
```
shsms: Now that there's only one call to `mocker.patch`,  I guess you don't need to separate the args from the call anymore.

leandro-lucarella-frequenz (dismissed):
I think this should really be moved eventually to `frequenz.sdk.util.test`, so even actor developers could use it. For this, I would remove the dependency on `pytest_mock`/`MockerFixture` as I don't think it is really necessary (we should be able to just use `unittest.mock`) and would introduce `pytest` and `pytest_mock` as a dependency for the SDK.

But those are just some random thoughts on where to go in the future, certainly not for the scope of this PR. We even have an issue on this now that I remember: https://github.com/frequenz-floss/frequenz-sdk-python/issues/44

tests/utils/mock_microgrid.py
```diff
@@ -71,16 +74,23 @@ def __init__(self, components: Set[Component], connections: Set[Connection]):
...
-        )
+        if mocker is None:
+            # This is a hack, optimally we should use mocker.patch, but I
+            # couldn't get it to work yet.
+            # pylint: disable=protected-access
+            microgrid._microgrid._MICROGRID = self.mock_microgrid
+        else:
+            mocker.patch(
+                "frequenz.sdk.microgrid._microgrid._MICROGRID", self.mock_microgrid
+            )
```
ela-kotulska-frequenz: Please don't this in that way, because then function behaves completely differently depending from the argument. It is unstable and not safe to use. Also user can think he doesn't need mocker at all.

If mocker is MockerFixture:
After the test, the path: `frequenz.sdk.microgrid._microgrid._MICROGRID` comes to  previous value. And that is clear. And it doesn't mess up in other tests.

That doesn't happen when you do`microgrid._microgrid._MICROGRID = self.mock_microgrid`.
It should not be used in unit tests.


To do this please create separate function in the file you need it.
This function should create MockMicrogridClient but not initialize it. You can initialize it in your own way like this:
```
mock_microgrid = MockMicrogridClient()
microgrid._microgrid._MICROGRID = mock_microgrid.microgrid
``` 

Please note that `MockMicrogridClient.initialize` is an optional method. You don't have to call it if that doesn't work for you :)

tests/timeseries/mock_microgrid.py
```diff
@@ -95,6 +95,26 @@ def __init__(
...
+        Returns:
+            A MockMicrogridClient instance.
+        """
+        self._microgrid = MockMicrogridClient(self._components, self._connections)
+        self._microgrid.initialize(mocker)
+
+        self._streaming_tasks = [
+            asyncio.create_task(coro) for coro in self._streaming_coros
+        ]
+        return self._microgrid
```
ela-kotulska-frequenz: This should be in the separate function (outside mock_microgrid), too.
Or it should be marked as unsafe option to start mock client.

mathias-baumann-frequenz: It's used inside `mock_microgrid` though...

tests/timeseries/mock_microgrid.py
```diff
@@ -95,6 +95,26 @@ def __init__(
         self._actors: list[typing.Any] = []
         self._start_meter_streaming(4)
 
+    def start_mock_client(
```
ela-kotulska-frequenz, line 98: Can it be private method?

tests/timeseries/mock_microgrid.py
ela-kotulska-frequenz: Can it be private method? Or not?

mathias-baumann-frequenz: Well, the benchmark will be using it from the outside. We can add a warning suppression for it..

ela-kotulska-frequenz: It is fine like this.

ela-kotulska-frequenz (dismissed):
Approved.
Please make sure code pass `mypy -p frequenz.sdk` before the merge.
(Or wait for Lucas PR)

tests/timeseries/mock_microgrid.py
ela-kotulska-frequenz: Or you could add comment that user should use `start` method in unit tests.

mathias-baumann-frequenz:
`mypy -p frequenz.sdk` yields a loooot of errors, as far as I can tell most of them not in my changes.

mathias-baumann-frequenz:
I added the comment and rebased, but the mypy command request seems difficult to fullfill. I guess I'll wait for lucas PR then :)

ela-kotulska-frequenz (dismissed):

ela-kotulska-frequenz (dismissed):

ela-kotulska-frequenz (approved):

PR #225, opened 2023-02-24
Add & use close channel method in the registry

### LINKED ISSUE #216: Add close channel feature to Datasourcing Actor
### What's needed?

Extend the Datasourcing actor to automatically close all corresponding sending channels when its input channel closes.

### Proposed solution

The `ChannelRegistry` should expose a method `_close_channel`,  which
  1. calls `close` on the channel
  2. removes the channel from the channel registry,  so that further requests to the registry with the same name would create a new channel.

The `close_channel` method is private, just to signal that it is not to be used lightly, and because we insist on adding an explanation comment, when disabling pylint warnings,  so this method will (hopefully) never be used without an explanation comment.

And the data sourcing actor can call this method with the channel names coming from the original requests.

### Use cases

Main use case is making testing and benchmarking of the actors easier and more precise.

For the production environment there is no extra runtime cost as a result of this feature.

_No response_

### Additional context

_No response_

mathias-baumann-frequenz:
Add all required code to automatically close channels when we're done sending
values in the mock_microgrid.

Required for the datasourcing actor benchmark.

- Add private method to allow closing of channels
- Update mock_microgrid to close channels when done sending.
- Update datasourcing actor to close channels.

### DIFF
--- a/src/frequenz/sdk/actor/_channel_registry.py
+++ b/src/frequenz/sdk/actor/_channel_registry.py
@@ -49,3 +49,15 @@ class ChannelRegistry:
         if key not in self._channels:
             self._channels[key] = Broadcast(f"{self._name}-{key}")
         return self._channels[key].new_receiver()
+
+    async def _close_channel(self, key: str) -> None:
+        """Close a channel with the given key.
+
+        This method is private and should only be used in special cases.
+
+        Args:
+            key: A key to identify the channel.
+        """
+        if key in self._channels:
+            if channel := self._channels.pop(key, None):
+                await channel.close()
--- a/src/frequenz/sdk/actor/_data_sourcing/microgrid_api_source.py
+++ b/src/frequenz/sdk/actor/_data_sourcing/microgrid_api_source.py
@@ -343,7 +343,10 @@ class MicrogridApiSource:
             stream_senders = self._get_metric_senders(
                 category, self._req_streaming_metrics[comp_id]
             )
-        api_data_receiver = self.comp_data_receivers[comp_id]
+        api_data_receiver: Receiver[Any] = self.comp_data_receivers[comp_id]
+
+        senders_done: asyncio.Event = asyncio.Event()
+        pending_messages = 0
 
         def process_msg(data: Any) -> None:
             tasks = []
@@ -351,9 +354,27 @@ class MicrogridApiSource:
                 for sender in senders:
                     tasks.append(sender.send(Sample(data.timestamp, extractor(data))))
             asyncio.gather(*tasks)
+            nonlocal pending_messages
+            pending_messages -= 1
+            if pending_messages == 0:
+                senders_done.set()
 
         async for data in api_data_receiver:
+            pending_messages += 1
+            senders_done.clear()
             process_msg(data)
+
+        while pending_messages > 0:
+            await senders_done.wait()
+
+        await asyncio.gather(
+            *[
+                # pylint: disable=protected-access
+                self._registry._close_channel(r.get_channel_name())
+                for requests in self._req_streaming_metrics[comp_id].values()
+                for r in requests
+            ]
+        )
 
     async def _update_streams(
         self,
--- a/tests/utils/mock_microgrid.py
+++ b/tests/utils/mock_microgrid.py
@@ -49,6 +49,13 @@ class MockMicrogridClient:
         meter_channels = self._create_meter_channels()
         ev_charger_channels = self._create_ev_charger_channels()
 
+        self._all_channels: Dict[int, Broadcast[Any]] = {
+            **bat_channels,
+            **inv_channels,
+            **meter_channels,
+            **ev_charger_channels,
+        }
+
         mock_api = self._create_mock_api(
             bat_channels, inv_channels, meter_channels, ev_charger_channels
         )
@@ -127,6 +134,15 @@ class MockMicrogridClient:
             return await self._ev_charger_data_senders[cid].send(data)
 
         raise RuntimeError(f"{type(data)} is not supported in MockMicrogridClient.")
+
+    async def close_channel(self, cid: int) -> None:
+        """Close channel for given component id.
+
+        Args:
+            cid: Component id
+        """
+        if cid in self._all_channels:
+            await self._all_channels[cid].close()
 
     def _create_battery_channels(self) -> Dict[int, Broadcast[BatteryData]]:
         """Create channels for the batteries.
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -12,6 +12,7 @@
 
 * A new class `SerializableRingbuffer` is now available, extending the `OrderedRingBuffer` class with the ability to load & dump the data to disk.
 * Add the `run(*actors)` function for running and synchronizing the execution of actors. This new function simplifies the way actors are managed on the client side, allowing for a cleaner and more streamlined approach. Users/apps can now run actors simply by calling run(actor1, actor2, actor3...) without the need to manually call join() and deal with linting errors.
+* The datasourcing actor now automatically closes all sending channels when the input channel closes.
 
 ## Bug Fixes
 
--- a/tests/timeseries/mock_microgrid.py
+++ b/tests/timeseries/mock_microgrid.py
@@ -148,6 +148,8 @@ class MockMicrogrid:  # pylint: disable=too-many-instance-attributes
                 await self._microgrid.send(make_comp_data(val_to_send, timestamp))
             await asyncio.sleep(self._sample_rate_s)
 
+        await self._microgrid.close_channel(comp_id)
+
     def _start_meter_streaming(self, meter_id: int) -> None:
         self._streaming_coros.append(
             self._comp_data_send_task(


src/frequenz/sdk/actor/_channel_registry.py
```diff
@@ -49,3 +49,14 @@ def new_receiver(self, key: str) -> Receiver[Any]:
...
+    async def _close_channel(self, key: str) -> None:
+        """Close a channel with the given key.
+
+        This method is private and should only be used in special cases.
+
+        Args:
+            key: A key to identify the channel.
+        """
+        if key in self._channels:
+            await self._channels[key].close()
```
shsms: After closing, this should also remove the channel from the registry, from `self._channels`

src/frequenz/sdk/actor/_channel_registry.py
```diff
@@ -49,3 +49,15 @@ def new_receiver(self, key: str) -> Receiver[Any]:
         if key not in self._channels:
             self._channels[key] = Broadcast(f"{self._name}-{key}")
         return self._channels[key].new_receiver()
+
+    async def _close_channel(self, key: str) -> None:
```
leandro-lucarella-frequenz, line 53: I wouldn't make this private if it is being used by external (non-test) modules.

src/frequenz/sdk/actor/_data_sourcing/microgrid_api_source.py
```diff
@@ -343,18 +343,39 @@ async def _handle_data_stream(
...
+            senders_done.clear()
             process_msg(data)
 
+        while pending_messages > 0:
+            await senders_done.wait()
+
+        for _, requests in self._req_streaming_metrics[comp_id].items():
+            for request in requests:
+                # pylint: disable=protected-access
+                await self._registry._close_channel(request.get_channel_name())
```
leandro-lucarella-frequenz: I don't like how obscure this is getting. 3 nested functions, `nonlocal`s and adding callbacks to a implicitly created task that is left to run wild seems like too much, at rleast for not getting some extra documentation.

You can remove some complexity making at least the done callback sequential (I didn't test it so it might have some bug, but *I think* it should work):

```py
        async def process_msg(data: Any) -> None:
            tasks = []
            for extractor, senders in stream_senders:
                for sender in senders:
                    tasks.append(sender.send(Sample(data.timestamp, extractor(data))))
            await asyncio.gather(*tasks)
            nonlocal pending_messages
            pending_messages -= 1
            if pending_messages == 0:
                senders_done.set()

        async for data in api_data_receiver:
            senders_done.clear()
            pending_messages += 1
            asyncio.create_task(process_msg(data))  # document this is left dangling on purpose

        while pending_messages > 0:
            await senders_done.wait()
```

Ideally we should keep track of running tasks and then wait for all at this point, instead of having a condition and the "global" counter, but I can understand it is some extra work so I would be fine with merging without that.

leandro-lucarella-frequenz: You don't need to use `.items()` if you don't care about the keys. Also, it would be good to make all those close run in parallel:

```suggestion
        await asyncio.gather(*[
            self._registry._close_channel(r.get_channel_name())
            for requests in d for r in requests])
```

src/frequenz/sdk/actor/_channel_registry.py
mathias-baumann-frequenz: @sahas-subramanian-frequenz suggested that so that it will not be readily used as it invalidates all the senders of a channel when run.

The idea was that it should always be exceptional if/when it is used.

src/frequenz/sdk/actor/_data_sourcing/microgrid_api_source.py
mathias-baumann-frequenz: Your code does indeed look nicer. I struggled with this a bit and the intention was to keep working on this bit in future PRs.
I also def. want to try a variant of this where we don't create a new task for every single receiver & sample.

mathias-baumann-frequenz: Hmm if I don't use `items()` it gives me only the keys, at least if I can believe pyright. 

```
   for requests in self._req_streaming_metrics[comp_id]:                        
       for request in requests:
                      ^^^^^^^^ "ComponentMetricId" is not iterable
                               __iter__" method not defined (Pyright reportGeneralTypeIssues)
```

ela-kotulska-frequenz:
Please update release notes. Because this will be change that might influence users.

src/frequenz/sdk/actor/_data_sourcing/microgrid_api_source.py
leandro-lucarella-frequenz: Sorry, for the values you need to use `.values()` :)

src/frequenz/sdk/actor/_channel_registry.py
leandro-lucarella-frequenz: OK, let's leave it like this as this needs further alignment. :+1: 

Probably the whole channel registry should be hidden to the users in the future anyway...

src/frequenz/sdk/actor/_data_sourcing/microgrid_api_source.py
```diff
@@ -343,17 +343,21 @@ async def _handle_data_stream(
...
+        await asyncio.gather(
+            *[
+                # pylint: disable=protected-access
+                self._registry._close_channel(r.get_channel_name())
+                for requests in self._req_streaming_metrics[comp_id].values()
+                for r in requests
+            ]
+        )
 
     async def _update_streams(
```
leandro-lucarella-frequenz: This is definitely simpler but it is changing the semantics and I'm not sue if it is OK.

Before you were sending all samples in the background, and you you are sending them sequentially. So let's say you need to send `sample1` and `sample2`, you are doing effectively:

```py
await send(sample1)
await send(sample2)
```

So now if for some reasons `send(sample1)` gets stuck for 60 seconds, `sample2` will not be send for that amount of time either, nor you'll receive new messages, so the whole data pipeline will get stuck.

I think the background sending is a good idea and probably shouldn't be removed (although also having that unbound is **not** a good idea IMHO, because if there is something funky with sending this will create tasks indefinitely), it is just the implementation was too obscure (probably to get it work quickly and easily).

The part of sending the data for one received sample in parallel is easy, as it is bound anyway:

```py
        async for data in api_data_receiver:
            await asyncio.gather(*[
                sender.send(Sample(data.timestamp, extractor(data)))
                for extractor, senders in stream_senders
                for sender in senders])
```

This at least will do all the `send()` in parallel, but again, if there is only one `send()` that gets blocked, then you'll stop processing new data until that is done, which is not probably what we want.

Again, a more resilient implementation IMHO should have a tasks manager that look at the age of pending sending tasks and if they are too old it cancels them, so they are not dangling forever, but implementing this is a not trivial amount of effort, so I wouldn't do it in this PR.

tests/utils/mock_microgrid.py
```diff
@@ -49,6 +49,12 @@ def __init__(self, components: Set[Component], connections: Set[Connection]):
         meter_channels = self._create_meter_channels()
         ev_charger_channels = self._create_ev_charger_channels()
 
+        self._all_channels: Dict[int, Broadcast[Any]] = {}
+        self._all_channels.update(bat_channels)
+        self._all_channels.update(inv_channels)
+        self._all_channels.update(meter_channels)
+        self._all_channels.update(ev_charger_channels)
```
leandro-lucarella-frequenz: Not sure what happened with my pro-tip (maybe I forgot to submit?) but here it goes again:

```suggestion
        self._all_channels: Dict[int, Broadcast[Any]] = {
		**bat_channels,
		**inv_channels,
		**meter_channels,
		**ev_charger_channels,
	}
```

(just a suggestion)

src/frequenz/sdk/actor/_data_sourcing/microgrid_api_source.py
mathias-baumann-frequenz: Is there an actual scenario where it could get stuck though? 

As far as I know, the receiver queue will just drop the oldest entry, so it shouldn't actually get stuck.

I just looked a bit closer at the underlying mechanisms. The only reason `send()` is `async` is because we do 

```
async with self._chan.recv_cv: # with aquires lock() of Condition
     self._chan.recv_cv.notify_all()
```

and in the receiver we do

```
        while len(self._q) == 0:
            if self._chan.closed:
                return False
            async with self._chan.recv_cv: # Aquire lock
                await self._chan.recv_cv.wait() # release lock and wait()
```

And I am wondering... is that really necessary? Why was a `Condition` used here and not simply an `Event`? (Reminder: A `Condition` is a `lock` + `event`)
I can currently not think of any reason why we would need a `lock` here. All our operations on the queue are atomic (from a task perspective), either we can successfully push/pop or we drop it, there is no waiting for other tasks during that operation.

And while looking at this, I was looking at https://docs.python.org/3/library/asyncio-queue.html as well, this class implements already everything that is re-implemented in `Broadcast` & friends, only they used an `Event` as well, instead of a `Condition`.

So if there isn't a big reason we're using a lock here, then I think we can actually make the whole `send()` method sync instead of async and remove a whole class of possible problems and certainly complexity.

Thoughts?

mathias-baumann-frequenz: And as additional thought: Why not just use the asyncio.Queue class?
The comment in my local file says

```
    """A queue, useful for coordinating producer and consumer coroutines.
```

which sounds exactly what we're doing. Including max len and everything.

mathias-baumann-frequenz: Well, now you did it :P I tried my idea and here is the result: https://github.com/frequenz-floss/frequenz-sdk-python/pull/233

src/frequenz/sdk/actor/_channel_registry.py
```diff
@@ -49,3 +49,15 @@ def new_receiver(self, key: str) -> Receiver[Any]:
...
+        """Close a channel with the given key.
+
+        This method is private and should only be used in special cases.
+
+        Args:
+            key: A key to identify the channel.
+        """
+        if key in self._channels:
+            await self._channels[key].close()
+            del self._channels[key]
```
ela-kotulska-frequenz: In different order, please! :D 

Someone can call this method 2 times, then both task will await  for channel to be closed. The first task will delete channel, the other will raise an exception :D

```
channel = self._channels.pop(key, None)
if channel is not None:
   await channel.close()
```

src/frequenz/sdk/actor/_data_sourcing/microgrid_api_source.py
ela-kotulska-frequenz: > So now if for some reasons send(sample1) gets stuck for 60 seconds, sample2 will not be send for that amount of time either, 

@leandro-lucarella-frequenz  in that case `gather`, won't help very much. 
As you said we will send sample2, but we won't send another message. If we think this is possible, then we should wait with the timeout. (using asyncio.wait)
I agree that background sending is better, but from the other side: Simple await is 2.5 times faster then asyncio.create_task so maybe it is worth? We create HUGE number of this tasks in 1 sec.... (battery is sending data every 0.2 sec and we need task for each metric) 
I guess asyncio.create_task is useful for a little heavier operations, then putting value on the queue.

>  I think we can actually make the whole send() method sync 

Matthias nice catch. But I would wait with decision for Sahas. He might had reasons for doing it like that.

> And as additional thought: Why not just use the asyncio.Queue class?

I think (but I am not sure ) that we made some tests in past and it turns out `asyncio.Queue` are very slow...  From the other side if it will allow send to be sync, then maybe it will improve performance.
I won't find this performance test, @leandro-lucarella-frequenz  do you remember where we store them?

leandro-lucarella-frequenz: > Is there an actual scenario where it could get stuck though?
> 
> As far as I know, the receiver queue will just drop the oldest entry, so it shouldn't actually get stuck.

Yes, the thing is when you `await`, then you lose the CPU, so even if the send doesn't block, the loop might schedule another task that gets a long time to process.

> So if there isn't a big reason we're using a lock here, then I think we can actually make the whole send() method sync instead of async and remove a whole class of possible problems and certainly complexity.

Yes, I think this would be the correct approach. I think we should have a distinction between sync and async sendersin the channels (I don't think we have real async senders for now, but there is an issue about creating a blocking channel, where `send()` will wait if the queue is full)  if we want to give guarantees about some sender not blocking, otherwise you can't really control if the loop will give you back the CPU or not.

> I think (but I am not sure ) that we made some tests in past and it turns out asyncio.Queue are very slow... From the other side if it will allow send to be sync, then maybe it will improve performance.
I won't find this performance test, @leandro-lucarella-frequenz do you remember where we store them?

Nope, I don't remember any performance tests with `queue`.

But I think we all agree we'd like to make `send()` sync where possible, so I think we can close this discussion with the following tasks list:

- [x] Create an issue in the channels repo about having sync and async senders (and make all current senders that can be sync, sync): https://github.com/frequenz-floss/frequenz-channels-python/issues/76
- [ ] Leave this PR implementation as is, as it should work with the current implementation (under certain parameters, like yielding to `send()` won't starve the current task for too long) and future changes
- [ ] Add a comment explaining the current situation (with a link to the channels issue)

Agreed?

src/frequenz/sdk/actor/_channel_registry.py
mathias-baumann-frequenz: Good catch!

leandro-lucarella-frequenz: I don't think calling `close()` on a closed channel raises an exception. At least with `Broadcast` it doesn't look like it does:

https://github.com/frequenz-floss/frequenz-channels-python/blob/81eed2460e18452df03382e8c12b5aae9a583597/src/frequenz/channels/_broadcast.py#L95-L109

But in any case I think your suggestion is good and can be further improved using:

```py
if channel := self._channels.pop(key, None):
   await channel.close()
```

src/frequenz/sdk/actor/_data_sourcing/microgrid_api_source.py
mathias-baumann-frequenz: And what about the idea i showcased in https://github.com/frequenz-floss/frequenz-channels-python/pull/75
And
https://github.com/frequenz-floss/frequenz-sdk-python/pull/233

src/frequenz/sdk/actor/_channel_registry.py
ela-kotulska-frequenz: > I don't think calling close() on a closed channel raises an exception. At least with Broadcast it doesn't look like it does:

@leandro-lucarella-frequenz  ` del self._channels[key]` will raise an exception, not channel.close() :)

src/frequenz/sdk/actor/_data_sourcing/microgrid_api_source.py
ela-kotulska-frequenz: Answered in the PR.

leandro-lucarella-frequenz: Again, I wouldn't block this PR until something is fixed in the channels repo, let's go with the assumption that at some point this send will be sync. I already created the issue in the channels repo.

It would be interesting if you can add timing info here and in the channels issue (https://github.com/frequenz-floss/frequenz-channels-python/issues/76) so we can have a better understanding about how much faster things could be if `send()` is async.

mathias-baumann-frequenz:
I updated the branch with the recent feedback but also removed the controversial commit regarding `send()` in the data sourcing actor and went back to counting the pending tasks before exiting method.

We can still update it later once we decided how exactly

src/frequenz/sdk/actor/_data_sourcing/microgrid_api_source.py
```diff
@@ -343,18 +343,39 @@ async def _handle_data_stream(
...
+            if pending_messages == 0:
+                senders_done.set()
 
         async for data in api_data_receiver:
+            pending_messages += 1
+            senders_done.clear()
             process_msg(data)
 
+        while pending_messages > 0:
+            await senders_done.wait()
```
ela-kotulska-frequenz, line 368: I won't argue but to remove all this complexity we could do: 
```
 async for data in api_data_receiver:
       await process_msg(data)
```

This is also much save because:
* we will send data in the order (everywhere in the code we assume it is correct order)
*  api_data_receiver has maxsize =  50. If we exceed this number, then we start to see logs that messages get lost and we are not processing them in time. 
* In current solution we won't see any logs, we will just produce infinite number of async task, that would block everything.

mathias-baumann-frequenz: you mean we await the `asyncio.gather(*tasks)` call? In that case I don't need all the complexity here.. But didn't we want to wait for sahas before making such changes? :)

mathias-baumann-frequenz: I tried to keep the impact on the code behavior low because of that

leandro-lucarella-frequenz: Sounds good to me. I think whatever we do now is likely to change if we go for a sync `send()`. We are just speculating now.

leandro-lucarella-frequenz (dismissed):
IMHO this is good to go, we can update the sending logic after we figured out the sync `send()`. But @ela-kotulska-frequenz do please raise your voice if you don't agree :)

ela-kotulska-frequenz (dismissed):

mathias-baumann-frequenz:
Rebased on latest changes.

ela-kotulska-frequenz (approved):

PR #231, opened 2023-03-01
Add wrappers for the component data in unit tests

ela-kotulska-frequenz:
In the unit tests we have to mock component data with only few parameters.
Current ComponentData definition doesn't allow for it because
we have to give all class parameters.
We used to avoid that by creating protobuf message with only
necessary data and use it to create ComponentData.
But this approach is complex and not resilient for the changes in the
protobuf api.
This commit wraps ComponentData. Each parameter has default protobuf
value that will be used if no argument was given.

### DIFF
--- a/tests/timeseries/mock_microgrid.py
+++ b/tests/timeseries/mock_microgrid.py
@@ -7,18 +7,10 @@ from __future__ import annotations
 
 import asyncio
 import typing
+from datetime import datetime, timezone
 from typing import Callable, Set, Tuple
 
-from frequenz.api.microgrid import (
-    battery_pb2,
-    common_pb2,
-    ev_charger_pb2,
-    inverter_pb2,
-    meter_pb2,
-    microgrid_pb2,
-)
 from frequenz.channels import Broadcast
-from google.protobuf.timestamp_pb2 import Timestamp  # pylint: disable=no-name-in-module
 from pytest_mock import MockerFixture
 
 from frequenz.sdk import microgrid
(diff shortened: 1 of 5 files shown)


tests/utils/component_data_wrapper.py
```diff
@@ -0,0 +1,148 @@
...
+
+"""Wrappers for the ComponentData.
+
+In the unit tests we have to mock component data with only few parameters.
+Current ComponentData definition doesn't allow for it.
+One way to to achieve that is to create protobuf message and use it to
+create ComponentData. But this approach is complex and not resilient
+for the changes in the protobuf (api) definition.
+This module wraps ComponentData so that only part of it can be initialized.
+It should be used in unit tests only.
```
mathias-baumann-frequenz: I'd suggest a different phrasing/description for this part:


This module allows the easy construction of mock `ComponentData` instances to be used in unit tests.
Usually, all parameters for an instance need to be defined, this module helps by allowing users to only specify the parameters they are interested in for their tests. The rest will be filled with reasonable defaults.
This also abstracts away changes in the protobuf definition and minimizes the places that will need to be updated in such cases.

mathias-baumann-frequenz: My thought was that it currently reads a bit like a "we used to do this, now we can do this" and I figured it would be more suited to have a simple description of the functionality instead

tests/timeseries/mock_microgrid.py
```diff
@@ -123,27 +120,11 @@ def _start_meter_streaming(self, meter_id: int) -> None:
...
-                                    ),
-                                )
-                            )
-                        ),
-                    ),
+                lambda value, ts: MeterDataWrapper(
+                    component_id=meter_id,
+                    timestamp=ts,
+                    active_power=value,
+                    current_per_phase=(value + 100.0, value + 101.0, value + 102.0),
```
mathias-baumann-frequenz, line 127: This is indeed much nicer

tests/utils/component_data_wrapper.py
ela-kotulska-frequenz: Done, thanks a lot!

mathias-baumann-frequenz (approved):

mathias-baumann-frequenz:
ci failure, otherwise lgtm

PR #236, opened 2023-03-06
Make resampling window size constant

### LINKED ISSUE #170: Resampled timestamps don't have constant interval
### What happened?

The resampler uses the trigger times from a `frequenz.channels.Timer` as the timestamp for resampled messages.  But the `Timer` returns the clock time at which it triggered, and not a calculated timestamp that always has a constant interval, adjusting for any delays from the rest of the program, as seen below:

``` python
>>> import asyncio
>>> from frequenz.channels.util import Timer
>>> async def test():
...   async for t in Timer(1.0): print(t)
... 
>>> asyncio.run(test())
2023-01-20 15:43:30.898930+00:00
2023-01-20 15:43:31.900304+00:00
2023-01-20 15:43:32.901836+00:00
2023-01-20 15:43:33.903355+00:00
2023-01-20 15:43:34.904885+00:00
2023-01-20 15:43:35.906479+00:00
```

Timestamps in subsequent messages from the resampler should all have a constant interval, equal to the given resampling interval.

### Affected version(s)

v0.16.0

### Affected part(s)

### Extra information

_No response_

leandro-lucarella-frequenz:
The resampler was using the timer timestamp to do the resampling, but this is wrong because the timer could (and probably always will) fire in the *future* (after the resampling period has passed).

The effect was that samples produced by the resampler were not perfectly spread using an exact resampling period, but they included the error produced by the timer firing

This PR fixes this issue and also fixes another bug found, samples in the future could have been included in a resampling window.  Now we only consider relevant samples that fit to the exact resampling window.

Part of #170.

### DIFF
--- a/src/frequenz/sdk/timeseries/_resampling.py
+++ b/src/frequenz/sdk/timeseries/_resampling.py
@@ -12,7 +12,7 @@ import math
 from bisect import bisect
 from collections import deque
 from dataclasses import dataclass
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from typing import AsyncIterator, Callable, Coroutine, Optional, Sequence
 
 from frequenz.channels.util import Timer
@@ -354,6 +354,18 @@ class Resampler:
         self._timer: Timer = Timer(config.resampling_period_s)
         """The timer to trigger the next resampling."""
 
+        self._window_end: datetime = datetime.now(timezone.utc) + timedelta(
+            seconds=self._config.resampling_period_s
+        )
+        """The time in which the current window ends.
+
+        This is used to make sure every resampling window is generated at
+        precise times. We can't rely on the timer timestamp because timers will
+        never fire at the exact requested time, so if we don't use a precise
+        time for the end of the window, the resampling windows we produce will
+        have different sizes.
+        """
+
     @property
     def config(self) -> ResamplerConfig:
         """Get the resampler configuration.
@@ -435,11 +447,12 @@ class Resampler:
                 timeseries from the resampler before calling this method
                 again).
         """
-        async for timer_timestamp in self._timer:
+        async for _ in self._timer:
             results = await asyncio.gather(
-                *[r.resample(timer_timestamp) for r in self._resamplers.values()],
+                *[r.resample(self._window_end) for r in self._resamplers.values()],
                 return_exceptions=True,
             )
+            self._update_window_end()
             exceptions = {
                 source: results[i]
                 for i, source in enumerate(self._resamplers)
@@ -451,6 +464,11 @@ class Resampler:
                 raise ResamplingError(exceptions)
             if one_shot:
                 break
+
+    def _update_window_end(self) -> None:
+        self._window_end = self._window_end + timedelta(
+            seconds=self._config.resampling_period_s
+        )
 
 
 class _ResamplingHelper:
--- a/src/frequenz/sdk/timeseries/_resampling.py
+++ b/src/frequenz/sdk/timeseries/_resampling.py
@@ -447,12 +447,12 @@ class Resampler:
                 timeseries from the resampler before calling this method
                 again).
         """
-        async for _ in self._timer:
+        async for timer_timestamp in self._timer:
             results = await asyncio.gather(
                 *[r.resample(self._window_end) for r in self._resamplers.values()],
                 return_exceptions=True,
             )
-            self._update_window_end()
+            self._update_window_end(timer_timestamp)
             exceptions = {
                 source: results[i]
                 for i, source in enumerate(self._resamplers)
@@ -465,7 +465,23 @@ class Resampler:
             if one_shot:
                 break
 
-    def _update_window_end(self) -> None:
+    def _update_window_end(self, timer_timestamp: datetime) -> None:
+        # We use abs() here to account for errors in the timer where the timer
+        # fires before its time even when in theory it shouldn't be possible,
+        # but we want to be resilient to timer implementation changes that
+        # could end up in this case, either because there were some time jump
+        # or some rounding error.
+        timer_error_s = abs((timer_timestamp - self._window_end).total_seconds())
+        if timer_error_s > (self._config.resampling_period_s / 10.0):
+            _logger.warning(
+                "The resampling timer fired too late. It should have fired at "
+                "%s, but it fired at %s (%s seconds difference; resampling "
+                "period is %s seconds)",
+                self._window_end,
+                timer_timestamp,
+                timer_error_s,
+                self._config.resampling_period_s,
+            )
         self._window_end = self._window_end + timedelta(
             seconds=self._config.resampling_period_s
         )
--- a/src/frequenz/sdk/timeseries/_resampling.py
+++ b/src/frequenz/sdk/timeseries/_resampling.py
@@ -661,14 +661,15 @@ class _ResamplingHelper:
         # We need to pass a dummy Sample to bisect because it only support
         # specifying a key extraction function in Python 3.10, so we need to
         # compare samples at the moment.
-        cut_index = bisect(self._buffer, Sample(minimum_relevant_timestamp, None))
+        min_index = bisect(self._buffer, Sample(minimum_relevant_timestamp, None))
+        max_index = bisect(self._buffer, Sample(timestamp, None))
         # Using itertools for slicing doesn't look very efficient, but
         # experiements with a custom (ring) buffer that can slice showed that
         # it is not that bad. See:
         # https://github.com/frequenz-floss/frequenz-sdk-python/pull/130
         # So if we need more performance beyond this point, we probably need to
         # resort to some C (or similar) implementation.
-        relevant_samples = list(itertools.islice(self._buffer, cut_index, None))
+        relevant_samples = list(itertools.islice(self._buffer, min_index, max_index))
         value = (
             conf.resampling_function(relevant_samples, conf, props)
             if relevant_samples
--- a/src/frequenz/sdk/timeseries/_resampling.py
+++ b/src/frequenz/sdk/timeseries/_resampling.py
@@ -346,8 +346,13 @@ class Resampler:
             config: The configuration for the resampler.
         """
         self._config = config
+        """The configuration for this resampler."""
+
         self._resamplers: dict[Source, _StreamingHelper] = {}
+        """A mapping between sources and the streaming helper handling that source."""
+
         self._timer: Timer = Timer(config.resampling_period_s)
+        """The timer to trigger the next resampling."""
 
     @property
     def config(self) -> ResamplerConfig:
--- a/tests/timeseries/test_resampling.py
+++ b/tests/timeseries/test_resampling.py
@@ -156,16 +156,14 @@ async def test_helper_buffer_too_big(
 
     _ = helper.resample(datetime.now(timezone.utc))
     # Ignore errors produced by wrongly finalized gRPC server in unrelated tests
-    assert _filter_logs(
+    assert (
+        "frequenz.sdk.timeseries._resampling",
+        logging.ERROR,
+        f"The new buffer length ({DEFAULT_BUFFER_LEN_MAX + 1}) "
+        f"for timeseries test is too big, using {DEFAULT_BUFFER_LEN_MAX} instead",
+    ) in _filter_logs(
         caplog.record_tuples,
-    ) == [
-        (
-            "frequenz.sdk.timeseries._resampling",
-            logging.ERROR,
-            f"The new buffer length ({DEFAULT_BUFFER_LEN_MAX + 1}) "
-            f"for timeseries test is too big, using {DEFAULT_BUFFER_LEN_MAX} instead",
-        )
-    ]
+    )
     # pylint: disable=protected-access
     assert helper._buffer.maxlen == DEFAULT_BUFFER_LEN_MAX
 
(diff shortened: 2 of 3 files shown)


leandro-lucarella-frequenz:
2 comments:

1. I'm not sure if we want the warning about the timer being too delayed, it seemed like a good idea at the time but I'm not 100% sure, it would be good to hear other opinions.
2. I just ran the benchmark with the new linear search, and it is considerably slower than the `bisect`, I thought if we had 2 do 2 bisects it might be slower but now I'm not that sure. I will use 2 bisects and compare. Results:

   ```
    WITH 1 BISECT             WITH LINEAR SEARCH

   resamples=10 samples=10      resamples=10 samples=10
   0.0008012260077521205        0.0009745770003064536
   resamples=10 samples=100     resamples=10 samples=100
   0.006307974021183327         0.00778788000025088
   resamples=10 samples=1000    resamples=10 samples=1000
   0.0604277040110901           0.0797138289999566
   resamples=100 samples=10     resamples=100 samples=10
   0.00808265499654226          0.009330201999546262
   resamples=100 samples=100    resamples=100 samples=100
   0.0624066719901748           0.07925070599958417
   resamples=100 samples=1000   resamples=100 samples=1000
   0.6098103950207587           0.7917559800007439
   resamples=1000 samples=10    resamples=1000 samples=10
   0.07815460601705126          0.09300610200079973
   resamples=1000 samples=100   resamples=1000 samples=100
   0.6357307220168877           0.7950725000009697
   resamples=1000 samples=1000  resamples=1000 samples=1000
   6.42200836900156             8.173040212997876
   ```

src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -452,6 +465,11 @@ async def resample(self, *, one_shot: bool = False) -> None:
             if one_shot:
                 break
 
+    def _update_window_end(self, timer_timestamp: datetime) -> None:
+        self._window_end = self._window_end + timedelta(
+            seconds=self._config.resampling_period_s
+        )
```
mathias-baumann-frequenz, line 487: This method is not using the parameter `timer_timestamp`

src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -437,9 +449,10 @@ async def resample(self, *, one_shot: bool = False) -> None:
         """
         async for timer_timestamp in self._timer:
             results = await asyncio.gather(
-                *[r.resample(timer_timestamp) for r in self._resamplers.values()],
```
mathias-baumann-frequenz, line 435: Nothing in this loop block is using the loop variable `timer_timestamp`

tests/timeseries/test_resampling.py
```diff
@@ -173,6 +173,88 @@ async def test_helper_buffer_too_big(
...
+    config = ResamplerConfig(
+        resampling_period_s=resampling_period_s,
+        max_data_age_in_periods=1.0,
+        resampling_function=resampling_fun_mock,
+        initial_buffer_len=4,
+    )
+    resampler = Resampler(config)
+
+    source_recvr = source_chan.new_receiver()
+    source_sendr = source_chan.new_sender()
```
mathias-baumann-frequenz: No need to change now, but it's a bit weird that the last `e` in both those variables is worth leaving out, might as well type the full name? :D

src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -653,17 +652,23 @@ def resample(self, timestamp: datetime) -> Sample:
...
+        # Do a linear search to find the min and max indexes for samples
+        # relevant to this window.
+        min_index = 0
+        max_index = len(self._buffer)
+        for i, sample in enumerate(self._buffer):
+            if sample.timestamp <= minimum_relevant_timestamp:
+                min_index = i + 1
+            elif sample.timestamp > timestamp:
+                max_index = i
+                break
```
mathias-baumann-frequenz: Not sure, but might the builtin function [`MAX()`](https://docs.python.org/3/library/functions.html#max) and `MIN()` be used here?

src/frequenz/sdk/timeseries/_resampling.py
leandro-lucarella-frequenz: Yeah, it was split commit, it is using it in the next commit, I can resplit.

src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -466,6 +466,17 @@ async def resample(self, *, one_shot: bool = False) -> None:
                 break
 
     def _update_window_end(self, timer_timestamp: datetime) -> None:
+        timer_error_s = abs((timer_timestamp - self._window_end).total_seconds())
```
ela-kotulska-frequenz, line 474: Why abs? Is it possible that timer fires to fast?

src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -653,17 +652,23 @@ def resample(self, timestamp: datetime) -> Sample:
...
         # experiements with a custom (ring) buffer that can slice showed that
         # it is not that bad. See:
         # https://github.com/frequenz-floss/frequenz-sdk-python/pull/130
         # So if we need more performance beyond this point, we probably need to
         # resort to some C (or similar) implementation.
-        relevant_samples = list(itertools.islice(self._buffer, cut_index, None))
+        relevant_samples = list(itertools.islice(self._buffer, min_index, max_index))
         value = (
             conf.resampling_function(relevant_samples, conf, props)
             if relevant_samples
```
ela-kotulska-frequenz, line 675: It is not for this PR, but maybe you should resample only if data are missing in current time window?
If data are correct then you should return the value closest to the end_window timestamp.

Otherwise values won't be accurate.
relevant_samples = [0,0,0,500, 1000, 5000] and  5000 is the current value, then you will return 1083.

So we might react slower on the peaks.

src/frequenz/sdk/timeseries/_resampling.py
ela-kotulska-frequenz: Maybe it would be faster if min_index and max_index were instance variables. And you could update them when you push value to the self._buffer.
Then in this call it would be enough to use/update these indexes. You won't need linear search.

tests/timeseries/test_resampling.py
leandro-lucarella-frequenz: Yeah, I guess it was originally a copy&paste I did from somewhere else and then kept copy&pasting, it is definitely not my style :)

src/frequenz/sdk/timeseries/_resampling.py
leandro-lucarella-frequenz: That would need 2 iterations over the array, and I also need the index, not the value. In any case, I might go back to using `bisect` if the performance is better.

src/frequenz/sdk/timeseries/_resampling.py
leandro-lucarella-frequenz: It shouldn't, but I don't trust it and rather cover for that situation in case the timer implementation ever changes. We know the timer is imprecise, I prefer not to assume it is only for one side. But I can add a comment to make this clear :+1:

src/frequenz/sdk/timeseries/_resampling.py
leandro-lucarella-frequenz: > It is not for this PR, but maybe you should resample only if data are missing in current time window?

No, the resampler should always call the resampling function with all relevant data.

Which resampling function to use (last-value being only one of them). AFAIK for power, we do want to smooth the curve, so we want some sort of average (IIRC about what @matthias-wende-frequenz once told me).

In any case, what to do with the data is the resampling function responsibility.

src/frequenz/sdk/timeseries/_resampling.py
leandro-lucarella-frequenz: Good idea. I will benchmark that too.

src/frequenz/sdk/timeseries/_resampling.py
ela-kotulska-frequenz: > AFAIK for power, we do want to smooth the curve, so we want some sort of average

Hmm I would like to know more about it. @matthias-wende-frequenz  could you explain?
I think we would like to smooth the power we send, not for the power we receive. We would like to receive true values and react fast. For the power we send, we already have power bounds that smooths sending power.

leandro-lucarella-frequenz: Was the off-the-records explanation enough @ela-kotulska-frequenz ? Can we close this as resolved?

src/frequenz/sdk/timeseries/_resampling.py
leandro-lucarella-frequenz: I looked into this and just removing both calculation of min/max (without even adding the tracking of min/max when samples are added) ~doesn't speed things up a lot~, and it needs major refactoring, as now the `_ResamplingHelper` doesn't know anything about what the current window is.

I think if we want to optimize this even further, it should be a separate issue.

leandro-lucarella-frequenz:
Here is the new analysis of performance using different methods. The conclusion for me is that there are no big performance differences between the 3 methods actually doing a search for the case with very few samples (10, 100) which is probably the most realistic example, so I will leave. For cases with many samples then binary search is clearly better.

Also bisecting needs to create a sample for now (until we move to Python 3.10+), not sure if the performance could be even better then.

It might be worth exploring the live updates of the `min`/`max` but I see it as low prio for now and definitely out of the scope of this PR.

```
 NO SEARCH                     1 DUMMY LINEAR SEARCH         2 SMART LINEAR SEARCHES       2 BISECTS
			                                                                   
 resamples=10 samples=10       resamples=10 samples=10       resamples=10 samples=10       resamples=10 samples=10      
 0.0008582130103604868         0.000956905001658015          0.0009785559959709644         0.000999016992864199         
 resamples=10 samples=100      resamples=10 samples=100      resamples=10 samples=100      resamples=10 samples=100     
 0.0068692619970534            0.007905827005743049          0.007647272999747656          0.007172472993261181         
 resamples=10 samples=1000     resamples=10 samples=1000     resamples=10 samples=1000     resamples=10 samples=1000    
 0.06876378500601277           0.07962176299770363           0.07558727399737108           0.06751152400102             
 resamples=100 samples=10      resamples=100 samples=10      resamples=100 samples=10      resamples=100 samples=10     
 0.008302549002110027          0.009530770999845117          0.009258945006877184          0.00964375599869527          
 resamples=100 samples=100     resamples=100 samples=100     resamples=100 samples=100     resamples=100 samples=100    
 0.06984430299780797           0.0807874250021996            0.07607155700679868           0.06979095100541599          
 resamples=100 samples=1000    resamples=100 samples=1000    resamples=100 samples=1000    resamples=100 samples=1000   
 0.6862357860081829            0.7878465870016953            0.7553019879997009            0.6735617559897946           
 resamples=1000 samples=10     resamples=1000 samples=10     resamples=1000 samples=10     resamples=1000 samples=10    
 0.0829010349989403            0.09500528799253516           0.0934388720052084            0.09525918100553099          
 resamples=1000 samples=100    resamples=1000 samples=100    resamples=1000 samples=100    resamples=1000 samples=100   
 0.6967007719940739            0.7917574560124194            0.7657141630043043            0.7278314160066657           
 resamples=1000 samples=1000   resamples=1000 samples=1000   resamples=1000 samples=1000   resamples=1000 samples=1000  
 6.972432320006192             8.14082598099776              8.074481772011495             7.17819636000786             



NO SEARCH

        min_index = 0
        max_index = len(self._buffer)


1 DUMB LIENAR SEARCH

        # Do a linear search to find the min and max indexes for samples
        # relevant to this window.
        min_index = 0
        max_index = len(self._buffer)
        for i, sample in enumerate(self._buffer):
            if sample.timestamp <= minimum_relevant_timestamp:
                min_index = i + 1
            elif sample.timestamp > timestamp:
                max_index = i
                break


2 SMART LINEAR SEARCHES

        # Do a linear search to find the min and max indexes for samples
        # relevant to this window.
        min_index = 0
        max_index = len(self._buffer)
        for i, sample in enumerate(self._buffer):
            if sample.timestamp > minimum_relevant_timestamp:
                break
            min_index = i + 1

        i = len(self._buffer)
        for sample in reversed(self._buffer):
            i -= 1
            if sample.timestamp > timestamp:
                max_index = i
                break


2 BISECTS

        # We need to pass a dummy Sample to bisect because it only support
        # specifying a key extraction function in Python 3.10, so we need to
        # compare samples at the moment.
        min_index = bisect(self._buffer, Sample(minimum_relevant_timestamp, None))
        max_index = bisect(self._buffer, Sample(timestamp, None))
```

src/frequenz/sdk/timeseries/_resampling.py
leandro-lucarella-frequenz: Added a comment clarifying this.

leandro-lucarella-frequenz:
Updated to address the review comments. The new version just use 2 `bisect` because the code is much simpler and it behaves slightly better with larger buffers (and almost the same with small ones).

src/frequenz/sdk/timeseries/_resampling.py
ela-kotulska-frequenz: Yes :)

ela-kotulska-frequenz (dismissed):

tests/timeseries/test_resampling.py
```diff
@@ -890,7 +886,7 @@ def _get_buffer_len(resampler: Resampler, source_recvr: Source) -> int:
 def _filter_logs(
     record_tuples: list[tuple[str, int, str]],
     *,
-    logger_name: str,
+    logger_name: str = "frequenz.sdk.timeseries._resampling",
```
ela-kotulska-frequenz: Maybe we could have _filter_logs in the utils folder?
In separate PR.

ela-kotulska-frequenz (dismissed):

PR #237, opened 2023-03-07
Fix bug in battery status

ela-kotulska-frequenz:
Inverter that was discharging was considered as not working.
If inverter was discharging, then it was impossible to
set any command until it changed state.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ dependencies = [
     "frequenz-api-microgrid >= 0.11.0, < 0.12.0",
     "frequenz-channels >= 0.11.0, < 0.12.0",
     "google-api-python-client >= 2.71, < 3",
-    "grpcio >= 1.51.1, < 2",
+    "grpcio == 1.51.1",
     "grpcio-tools >= 1.51.1, < 2",
     "networkx >= 2.8, < 3",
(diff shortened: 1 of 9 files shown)


pyproject.toml
```diff
@@ -28,7 +28,7 @@ dependencies = [
     "frequenz-api-microgrid >= 0.11.0, < 0.12.0",
     "frequenz-channels >= 0.11.0, < 0.12.0",
     "google-api-python-client >= 2.71, < 3",
-    "grpcio >= 1.51.1, < 2",
+    "grpcio == 1.51.1",
```
leandro-lucarella-frequenz, line 31: Can you maybe split all the pinning (here and in `noxfile.py`) to a separate commit to make it more clear (I guess this is to make the CI pass?).

We probably should consider start pinning the dev dependencies always, so we don't have to do this when we need to release a hot fix for an old version.

leandro-lucarella-frequenz (dismissed):
Just a suggestion, but feel free to merge as is too it is it too much work, it is an old release anyway...

pyproject.toml
ela-kotulska-frequenz: True :) Thanks. I separated commits.

leandro-lucarella-frequenz (approved):

PR #245, opened 2023-03-09
Move component data wrapper to frequenz.sdk.simulate

mathias-baumann-frequenz:
So it can be used in actors as well.


mathias-baumann-frequenz:
I have no idea why those tests fail, I cannot reproduce that locally at all. My local `black` is very happy.

ela-kotulska-frequenz (approved):

leandro-lucarella-frequenz (changes requested):
I really don't like the name, so just blocking this from being merged until we discuss it.

mathias-baumann-frequenz:
I am not using this after all now that we have the battery pool

PR #254, opened 2023-03-13
Change BatteryPool to send metrics only when they change

ela-kotulska-frequenz:
We compared the metric result but with timestamp.
Because timestamps were different comparison was always True. Exclude timestamp to not it in the comparison methods. Remove Sample from the metric result types, because we don't return any Sample metric now and it is not consistent with the rest the result types, now.

### DIFF
--- a/src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
+++ b/src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
@@ -12,7 +12,6 @@ from typing import Generic, Iterable, TypeVar
 
 from ... import microgrid
 from ...microgrid.component import ComponentCategory, ComponentMetricId, InverterType
-from ...timeseries import Sample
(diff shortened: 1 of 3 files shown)


daniel-zullo-frequenz (dismissed):
Code-wise LGTM.

As a footnote I'd like to suggest to split these kind of changes in separate commits for future PRs. IIUC in this case it would be a commit to `Remove sample from the metric` and another commit to `Do not compare timestamp in metrics` (or something like that)

ela-kotulska-frequenz:
> As a footnote I'd like to suggest to split these kind of changes in separate commits for future PRs. IIUC in this case it would be a commit to Remove sample from the metric and another commit to Do not compare timestamp in metrics (or something like that)

But they are related :(  Otherwise I would have to add `field(compare=False)` to Sample definitions. 
That would be breaking change and I would like to avoid it. :)

ela-kotulska-frequenz:
Rebased

PR #255, opened 2023-03-13
Fix resampler behaviour when resampling is to late

### LINKED ISSUE #170: Resampled timestamps don't have constant interval

ela-kotulska-frequenz:
Previous solution used Timer, that was fired after the specific time. Also it didn't take time difference spent on resampling into account when sleeping.
In result the time difference between resampling period was constantly increasing. And there was no option to catch up with resampling.

Fixes #170.

### DIFF
--- a/src/frequenz/sdk/timeseries/_resampling.py
+++ b/src/frequenz/sdk/timeseries/_resampling.py
@@ -429,6 +429,30 @@ class Resampler:
             return False
         return True
 
+    async def _wait_for_next_resampling_period(self) -> None:
+        """Wait for next resampling period.
+
+        If resampling period already started, then return without sleeping.
+        That would allow us to catch up with resampling.
+        Print warning if function woke up to late.
+        """
+        now = datetime.now(tz=timezone.utc)
+        if self._window_end > now:
+            sleep_for = self._window_end - now
+            await asyncio.sleep(sleep_for.total_seconds())
+
+        timer_error_s = (now - self._window_end).total_seconds()
+        if timer_error_s > (self._config.resampling_period_s / 10.0):
+            _logger.warning(
+                "The resampling task woke up too late. Resampling should have started "
+                "at %s, but it started at %s (%s seconds difference; resampling "
+                "period is %s seconds)",
+                self._window_end,
+                now,
+                timer_error_s,
+                self._config.resampling_period_s,
+            )
+
     async def resample(self, *, one_shot: bool = False) -> None:
         """Start resampling all known timeseries.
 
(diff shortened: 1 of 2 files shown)


src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -429,6 +429,30 @@ def remove_timeseries(self, source: Source) -> bool:
...
+        """
+        now = datetime.now(tz=timezone.utc)
+        if self._window_end > now:
+            sleep_for = self._window_end - now
+            await asyncio.sleep(sleep_for.total_seconds())
+
+        timer_error_s = (now - self._window_end).total_seconds()
+        if timer_error_s > (self._config.resampling_period_s / 10.0):
+            _logger.warning(
+                "The resampling started too late. It should have started at "
```
leandro-lucarella-frequenz: I would still mention something more towards the timer drifting here, IMHO is less clear if we just say it *started too late*, why did it start too late?

src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -447,12 +471,14 @@ async def resample(self, *, one_shot: bool = False) -> None:
...
-        async for timer_timestamp in self._timer:
+        while True:
+            await self._wait_for_next_resampling_period()
+
             results = await asyncio.gather(
                 *[r.resample(self._window_end) for r in self._resamplers.values()],
                 return_exceptions=True,
             )
-            self._update_window_end(timer_timestamp)
+            self._update_window_end()
```
leandro-lucarella-frequenz: You could also remove the function if it's just one line now :)

leandro-lucarella-frequenz (dismissed):
This is definitely a small hack to fix the issue, just a couple of very minor comments, but all optional so I'll approve.

I still think we should have a periodic timer in the channels repo and that we should use a monotonic clock for timers, because really weird shit can happen if the machine wall clock is changed otherwise.

If you merge this, could you also add a new issue to remind us to replace this with a proper periodic timer when it's available?

leandro-lucarella-frequenz (dismissed):

ela-kotulska-frequenz:
Fixed to long line in tests.
I got approve before so I will allow myself to merge :)

PR #258, opened 2023-03-15
Use all batteries when none are working in PowerDistributor

ela-kotulska-frequenz:


src/frequenz/sdk/actor/power_distributing/power_distributing.py
```diff
@@ -535,6 +535,24 @@ def _get_components_pairs(
 
         return bat_inv_map, inv_bat_map
 
+    def _get_working_batteries(self, batteries: Set[int]) -> Set[int]:
+        """Get subset with working batteries.
+
+        If there are no working batteries in the given set,
+        then return input set to use them all in case of gap discharge.
```
shsms: I think you should say in the PR description that this is being done for gap discharge,  and keep it generic in the docstring,
```suggestion
        If none of the given batteries are working, then treat all of them
        as working.
```

src/frequenz/sdk/actor/power_distributing/power_distributing.py
```diff
@@ -535,6 +535,24 @@ def _get_components_pairs(
...
+
+        Args:
+            batteries: requested batteries
+
+        Returns:
+            Subset with working batteries.
+            Input set if there are no working batteries in the given set.
```
shsms: That looks like two values are returned.  I think out documentation tool treats that formatting as two return values as well.  Maybe something like this is better,  where you clearly say `or`
```suggestion
            Subset of working batteries, or entire input set if there are no working batteries in
                the given set.
```

shsms (commented):
Could you please use a clear and specific PR title?  Something like "Use all batteries when none are working, in PowerDistributor" maybe?

src/frequenz/sdk/actor/power_distributing/power_distributing.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/actor/power_distributing/power_distributing.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/actor/power_distributing/power_distributing.py
```diff
@@ -538,6 +538,24 @@ def _get_components_pairs(
...
+
+        If none of the given batteries are working, then treat all of them
+        as working.
+
+        Args:
+            batteries: requested batteries
+
+        Returns:
+            Subset with working batteries or input set if none of the given batteries
+            are working.
```
shsms: The second line needs to be indented, to be compatible with out documentation tool.

ela-kotulska-frequenz: Done

leandro-lucarella-frequenz (commented):
I'm confusing by this behaviour, would it be OK for all power distributor users to suddenly think all batteries are fine when there are all actually broken? Shouldn't this be somehow optional behavior?

ela-kotulska-frequenz:
@leandro-lucarella-frequenz 
If all batteries are broken then we probably won't be able to do any operations anyway. But we can keep trying :P So it should not be a problem. 

But I agree. Actors should behave different if all batteries are broken (for example do gap-discharge).
That is why we should probably have extra parameter in the PowerDistributinActor request that tells that user would like to use all batteries (even not working). I described it here: https://github.com/frequenz-floss/frequenz-sdk-python/issues/259  because it might be a problem if we have NaN values and we use cached values.

leandro-lucarella-frequenz:
OK, I'm trying to get my head around this because I don't fully understand the use case. Who's the main user for this at this moment? I guess this PR come from some current need to support this. I would like to have a better understanding of what's needed. If this is required for some current deployment, we can merge and discuss in a separate issue. @sahas-subramanian-frequenz maybe you know more?

ela-kotulska-frequenz:
>  Who's the main user for this at this moment?

Peak shaving application.
If all batteries are broken, then peak-shaving is trying to do gap-discharge. 
Gap-discharge - it tries to discharge with small amount of power, to do everything that is possible to avoid the peak.

ela-kotulska-frequenz:
I am not working since tomorrow.
Please feel free to merge it or start new PR :)

leandro-lucarella-frequenz:
> I am not working since tomorrow.
> Please feel free to merge it or start new PR :)

Yeah, we'll take care of it! Thanks!

shsms:
> I guess this PR come from some current need to support this. I would like to have a better understanding of what's needed. 

Yes,  the peak shaving application has a thing called `gap-discharge` where it starts discharging the batteries slowly when it is missing data.  This is to buy us some time to identify the issue.

The sdk was blocking such commands earlier,  and this PR fixes that.

> If this is required for some current deployment, we can merge and discuss in a separate issue.

I agree,  it is better to take this in, and create a new discussion on how to provide a better interface for such power management edge cases.

leandro-lucarella-frequenz (approved):

PR #263, opened 2023-03-15
Check battery capacity in BatteryStatus

ela-kotulska-frequenz:
Battery with NaN capacity will be considered as not working. It was discussed that capacity is very important metric and we can't guess it in case it is missing.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,7 +6,7 @@
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with --> 
+* Update BatteryStatus to mark battery with unknown capacity as not working (#263)
 
 ## New Features
 
--- a/src/frequenz/sdk/actor/power_distributing/_battery_status.py
+++ b/src/frequenz/sdk/actor/power_distributing/_battery_status.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import math
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
 from enum import Enum
@@ -269,6 +270,7 @@ class BatteryStatusTracker:
                 self._is_message_reliable(msg.inner)
                 and self._is_battery_state_correct(msg.inner)
                 and self._no_critical_error(msg.inner)
+                and self._is_capacity_present(msg.inner)
             )
             self._battery.last_msg_timestamp = msg.inner.timestamp
             self._battery.data_recv_timer.reset()
@@ -351,6 +353,26 @@ class BatteryStatusTracker:
             return Status.UNCERTAIN
 
         return Status.WORKING
+
+    def _is_capacity_present(self, msg: BatteryData) -> bool:
+        """Check whether the battery capacity is NaN or not.
+
+        If battery capacity is missing, then we can't work with it.
+
+        Args:
+            msg: battery message
+
+        Returns:
+            True if battery capacity is present, false otherwise.
+        """
+        if math.isnan(msg.capacity):
+            if self._last_status == Status.WORKING:
+                _logger.warning(
+                    "Battery %d capacity is NaN",
+                    msg.component_id,
+                )
+            return False
+        return True
 
     def _no_critical_error(self, msg: Union[BatteryData, InverterData]) -> bool:
         """Check if battery or inverter message has any critical error.
--- a/tests/actor/test_battery_status.py
+++ b/tests/actor/test_battery_status.py
@@ -37,12 +37,13 @@ from ..utils.component_data_wrapper import BatteryDataWrapper, InverterDataWrapp
 from ..utils.mock_microgrid import MockMicrogridClient
 
 
-def battery_data(
+def battery_data(  # pylint: disable=too-many-arguments
     component_id: int,
     timestamp: Optional[datetime] = None,
     relay_state: BatteryRelayState.ValueType = BatteryRelayState.RELAY_STATE_CLOSED,
     component_state: BatteryState.ValueType = BatteryState.COMPONENT_STATE_CHARGING,
     errors: Optional[Iterable[BatteryError]] = None,
+    capacity: float = 0,
 ) -> BatteryData:
     """Create BatteryData with given arguments.
 
@@ -66,6 +67,7 @@ def battery_data(
 
     return BatteryDataWrapper(
         component_id=component_id,
+        capacity=capacity,
         timestamp=datetime.now(tz=timezone.utc) if timestamp is None else timestamp,
         _relay_state=relay_state,
         _component_state=component_state,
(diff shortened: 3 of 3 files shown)


src/frequenz/sdk/actor/power_distributing/_battery_status.py
```diff
@@ -352,6 +354,25 @@ def _get_current_status(self) -> Status:
...
+
+        Returns:
+            True if battery capacity is NaN, false otherwise.
+        """
+        if math.isnan(msg.capacity) and self._last_status == Status.WORKING:
+            _logger.warning(
+                "Battery %d capacity is NaN",
+                msg.component_id,
+            )
+            return False
```
daniel-zullo-frequenz, line 374: According to the documentation this should return False regardless of the status so I wonder if either the code or the documentation is not  right

src/frequenz/sdk/actor/power_distributing/_battery_status.py
```diff
@@ -352,6 +354,25 @@ def _get_current_status(self) -> Status:
 
         return Status.WORKING
 
+    def _capacity_not_nan(self, msg: BatteryData) -> bool:
+        """Battery capacity should not be guessed.
```
daniel-zullo-frequenz: What about `Check whether the battery capacity is NaN or not` or something around that

src/frequenz/sdk/actor/power_distributing/_battery_status.py
tiyash-basu-frequenz: Is the logic supposed to check
a. if there is indeed a battery data sample, and just the capacity is missing?, OR
b. that the capacity is missing?

The way this method is being used, it suggests option a. So I wonder if the method name should be something like `_is_capacity_present`.

ela-kotulska-frequenz: Done, thanks @daniel-zullo-frequenz  and @tiyash-basu-frequenz

src/frequenz/sdk/actor/power_distributing/_battery_status.py
ela-kotulska-frequenz: Done

RELEASE_NOTES.md
```diff
@@ -6,7 +6,7 @@
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with --> 
+* Update BatteryStatus to mark battery with unknown capacity and not working (#263)
```
leandro-lucarella-frequenz:
```suggestion
* Update BatteryStatus to mark battery with unknown capacity as not working (#263)
```

I guess?

src/frequenz/sdk/actor/power_distributing/_battery_status.py
```diff
@@ -352,6 +354,26 @@ def _get_current_status(self) -> Status:
...
+
+        Returns:
+            True if battery capacity is missing, false otherwise.
+        """
+        if math.isnan(msg.capacity):
+            if self._last_status == Status.WORKING:
+                _logger.warning(
+                    "Battery %d capacity is NaN",
+                    msg.component_id,
+                )
```
leandro-lucarella-frequenz, line 373: How often will this message be produced if a battery capacity is `NaN` for a prolonged period? Only once? Once per resampling period?

Not for this PR, but more like a mental note: the more I think about alerts, the more I'm convinced in the future we need to use counters instead of ERROR/WARNINGs to produce alerts. In this case for example, we probably want to set an alert for `broken_battery_count > 0` and receive it only once, and have it in alerting state until it goes back to `broken_battery_count == 0`.

FYI @matthias-wende-frequenz @thomas-nicolai-frequenz

tests/actor/test_battery_status.py
```diff
@@ -353,6 +355,15 @@ async def test_sync_update_status_with_messages(
         )
         assert tracker._update_status(select) is None  # type: ignore[arg-type]
 
+        # Check if NaN capacity changes the battery status.
+        select = FakeSelect(battery=battery_data(component_id=BATTERY_ID))
```
leandro-lucarella-frequenz, line 359: Maybe you can inherit from the real `Select` to avoid having to put a `type: ignore` everywhere?

leandro-lucarella-frequenz (commented):
Just a minor typo and a question for the future. Other than that, LGTM.

src/frequenz/sdk/actor/power_distributing/_battery_status.py
```diff
@@ -352,6 +354,26 @@ def _get_current_status(self) -> Status:
...
+    def _is_capacity_present(self, msg: BatteryData) -> bool:
+        """Check whether the battery capacity is NaN or not.
+
+        If battery capacity is missing, then we can't work with it.
+
+        Args:
+            msg: battery message
+
+        Returns:
+            True if battery capacity is missing, false otherwise.
```
daniel-zullo-frequenz: Shouldn't it be the other way around? or `True if battery capacity is present(not missing)`?

thomas-nicolai-frequenz: @ela-kotulska-frequenz is on sick leave until further notice so I'm not sure who can give you feedback on that.

src/frequenz/sdk/actor/power_distributing/_battery_status.py
thomas-nicolai-frequenz: I'm not sure if that should be a worry of the SDK of how the alerting will work for now. Feels too much to address right now. Though, it doesn't mean that I disagree but it feels to early to address that problem.

tests/actor/test_battery_status.py
ela-kotulska-frequenz: It was really hard to test the async code so I decided on FakeSelect.
Maybe with async-solipsism it will be ok... I didn't know async-solipsism when I wrote those tests.

Anyway this tests just follows the pattern in others tests in this file. It works and it is efficient so I will leave it.
If for some reason  this pylint error is problematic then we should rewrite all tests in this file.

src/frequenz/sdk/actor/power_distributing/_battery_status.py
ela-kotulska-frequenz: > How often will this message be produced if a battery capacity is NaN

Only once. Only if latest battery status was WORKING, then the log will be printed.
If capacity is NaN, then the status will change to NOT_WORKING and the we won't see the log again.

src/frequenz/sdk/actor/power_distributing/_battery_status.py
ela-kotulska-frequenz: Done

RELEASE_NOTES.md
ela-kotulska-frequenz: Done

daniel-zullo-frequenz (commented):
LGTM

src/frequenz/sdk/actor/power_distributing/_battery_status.py
leandro-lucarella-frequenz: > I'm not sure if that should be a worry of the SDK of how the alerting will work for now. Feels too much to address right now. Though, it doesn't mean that I disagree but it feels to early to address that problem.

Yeah, definitely, thinking about the mid-long term here, didn't mean now. But sometimes when talking about this in the air is not easy to come up with good examples, so I just wanted to point this out.

tests/actor/test_battery_status.py
leandro-lucarella-frequenz: Yeah, I meant only to make `FakeSelect` inherit from `Select`, just that (not to rewrite any tests), but maybe it doesn't simply work if you inherit because some real methods are used then.

leandro-lucarella-frequenz (approved):

PR #266, opened 2023-03-16
Add `component_data` method to EVChargerPool

shsms:
This PR:
1. introduces the `component_data` method to the EVChargerPool, which ties the resampled 3-phase current values for an ev charger with the component state, which comes directly from the microgrid,  streaming objects of type:

   ``` python
   @dataclass(frozen=True)
   class EVChargerData:
       """Data for an EV Charger, including the 3-phase current and the component state."""

       component_id: int
       current: Sample3Phase
       state: EVChargerState
   ```

   This is done so that the state-tracking code can piggy-back on the resampler to detect if a component is missing data, and switch its state to `MISSING` whenever that is the case.

3. adds these methods for `Sample3Phase`: `__iter__, min, max, map`.

### DIFF
--- a/src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool.py
+++ b/src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool.py
@@ -9,6 +9,7 @@ import asyncio
 import logging
 import uuid
 from asyncio import Task
+from collections import abc
 from dataclasses import dataclass
 
 from frequenz.channels import Broadcast, ChannelClosedError, Receiver, Sender
@@ -86,6 +87,15 @@ class EVChargerPool:
             self._channel_registry,
             self._resampler_subscription_sender,
         )
+
+    @property
+    def component_ids(self) -> abc.Set[int]:
+        """Return component IDs of all EV Chargers managed by this EVChargerPool.
+
+        Returns:
+            Set of managed component IDs.
+        """
+        return self._component_ids
 
     async def current(self) -> FormulaReceiver3Phase:
         """Fetch the total current for the EV Chargers in the pool.
--- a/src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool.py
+++ b/src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool.py
@@ -5,26 +5,40 @@
 
 from __future__ import annotations
 
+import asyncio
 import logging
 import uuid
+from asyncio import Task
+from dataclasses import dataclass
 
-from frequenz.channels import Sender
+from frequenz.channels import Broadcast, ChannelClosedError, Receiver, Sender
 
 from ...actor import ChannelRegistry, ComponentMetricRequest
 from ...microgrid import connection_manager
-from ...microgrid.component import ComponentCategory
+from ...microgrid.component import ComponentCategory, ComponentMetricId
+from .. import Sample, Sample3Phase
 from .._formula_engine import FormulaEnginePool, FormulaReceiver, FormulaReceiver3Phase
 from .._formula_engine._formula_generators import (
     EVChargerCurrentFormula,
     EVChargerPowerFormula,
     FormulaGeneratorConfig,
 )
+from ._state_tracker import EVChargerState, StateTracker
 
 logger = logging.getLogger(__name__)
 
 
 class EVChargerPoolError(Exception):
     """An error that occurred in any of the EVChargerPool methods."""
+
+
+@dataclass(frozen=True)
+class EVChargerData:
+    """Data for an EV Charger, including the 3-phase current and the component state."""
+
+    component_id: int
+    current: Sample3Phase
+    state: EVChargerState
 
 
 class EVChargerPool:
@@ -62,6 +76,10 @@ class EVChargerPool:
                     component_category={ComponentCategory.EV_CHARGER}
                 )
             }
+        self._state_tracker: StateTracker | None = None
+        self._status_streams: dict[
+            int, tuple[Task[None], Broadcast[EVChargerData]]
+        ] = {}
         self._namespace: str = f"ev-charger-pool-{uuid.uuid4()}"
         self._formula_pool: FormulaEnginePool = FormulaEnginePool(
             self._namespace,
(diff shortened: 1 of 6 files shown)


shsms:
Rebased on #270, so that I can already start using this branch in the ev actor.  So the first 3 commits may be skipped while reviewing.

src/frequenz/sdk/timeseries/ev_charger_pool/_state_tracker.py
```diff
@@ -110,74 +70,51 @@ def __init__(self, component_ids: set[int]) -> None:
...
             data: component data from the microgrid, for an EV Charger in the pool.
-
-        Returns:
-            A new `EVChargerPoolStates` instance representing all the EV Chargers in
-                the pool, in case there has been a state change for any of the EV
-                Chargers, or `None` otherwise.
         """
         evc_id = data.component_id
         new_state = EVChargerState.from_ev_charger_data(data)
         if evc_id not in self._states or self._states[evc_id] != new_state:
```
ela-kotulska-frequenz: This `if` line can be removed now :) 
```
new_state = EVChargerState.from_ev_charger_data(data)
self._states[evc_id] = new_state
```

shsms: true, done.

src/frequenz/sdk/timeseries/ev_charger_pool/_state_tracker.py
```diff
@@ -59,6 +59,14 @@ def from_ev_charger_data(cls, data: EVChargerData) -> EVChargerState:
             return EVChargerState.EV_PLUGGED
         return EVChargerState.IDLE
 
+    def ev_connected(self) -> bool:
+        """Check whether an EV is connected to the charger.
+
+        Returns:
+            Whether an EV is connected to the charger.
```
daniel-zullo-frequenz, line 66: What about something around `True whether an EV is connected to the charger, False otherwise` instead?

src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool.py
```diff
@@ -87,7 +87,7 @@ def __init__(
             self._resampler_subscription_sender,
         )
 
-    async def total_current(self) -> FormulaReceiver3Phase:
+    async def current(self) -> FormulaReceiver3Phase:
```
daniel-zullo-frequenz, line 100: What is the reason to rename these functions removing the suffix `total_`, it'd be good if this is stated in the commit description. The documentation for these methods don't need to be updated, right?

daniel-zullo-frequenz: oh I've just seen these methods are removed in a separated commit. Then the renaming commit does not make much sense. Anyway resolving this comment...

src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool.py
```diff
@@ -148,3 +166,106 @@ async def power(self, component_id: int) -> FormulaReceiver:
...
+            await resampler_subscribe(ComponentMetricId.CURRENT_PHASE_1),
+            await resampler_subscribe(ComponentMetricId.CURRENT_PHASE_2),
+            await resampler_subscribe(ComponentMetricId.CURRENT_PHASE_3),
+        )
+
+    async def _stream_component_data(
+        self,
+        component_id: int,
+        sender: Sender[EVChargerData],
+    ) -> None:
```
daniel-zullo-frequenz, line 194: you might want to add documentation for this function in the commit the function is added

tests/timeseries/test_ev_charger_pool.py
```diff
@@ -112,3 +116,100 @@ async def test_ev_power(  # pylint: disable=too-many-locals
...
+        assert (
+            status.current.value_p1,
+            status.current.value_p2,
+            status.current.value_p3,
+        ) == (None, None, None)
+        assert status.state == EVChargerState.MISSING
+
+        await send_resampled_current(None, None, None)
+        mockgrid.evc_cable_states[evc_id] = EVChargerCableState.EV_PLUGGED
+        # mockgrid.evc_component_states[0] = EVChargerComponentState.READY
```
daniel-zullo-frequenz: I´d suggest to remove this commented line

src/frequenz/sdk/timeseries/ev_charger_pool/_state_tracker.py
```diff
@@ -110,74 +70,50 @@ def __init__(self, component_ids: set[int]) -> None:
...
+        # starts arriving from the individual components.
+        self._states: dict[int, EVChargerState] = {
+            component_id: EVChargerState.MISSING for component_id in component_ids
+        }
+
+    def get(self, component_id: int) -> EVChargerState:
+        """Return the current state of the EV Charger with the given component ID.
+
+        Args:
+            component_id: id of the ev charger whose state is being fetched.
```
daniel-zullo-frequenz: s/ev/EV

daniel-zullo-frequenz (commented):
I have a  few minor suggestions, LGTM otherwise

src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool.py
```diff
@@ -102,49 +129,105 @@ async def total_power(self) -> FormulaReceiver:
...
         """
-        if component_id not in self._component_ids:
-            raise EVChargerPoolError(
-                f"{component_id=} is not part of the EVChargerPool"
-                f" (with ids={self._component_ids})"
+        if recv := self._status_streams.get(component_id, None):
+            if not recv[0].done():
+                return recv[1].new_receiver()
+            logger.warning("Restarting component_status for id: %s", component_id)
+            output_chan = recv[1]
```
leandro-lucarella-frequenz: I would explode the tuple here to make it easier to read, the type declaration is a bit too far away:
```suggestion
            task, output_chan = recv
            if not task.done():
                return output_chan.new_receiver()
            logger.warning("Restarting component_status for id: %s", component_id)
```

src/frequenz/sdk/timeseries/ev_charger_pool/_state_tracker.py
leandro-lucarella-frequenz: We are using the "Whether xxx" for bools in many places, I think it is clear enough and avoid mentioning `True`/`False` everywhere.

(for more context, we stole this from [dart's style guide](https://dart.dev/guides/language/effective-dart/documentation#prefer-starting-a-boolean-variable-or-property-comment-with-whether-followed-by-a-noun-or-gerund-phrase), I think it is generally applicable to Python too and it makes a very consistent and clear style for documentation and even some method/property naming rules)

Speaking of which.... I think this method should be called `is_ev_connected()` :)

src/frequenz/sdk/timeseries/ev_charger_pool/_state_tracker.py
```diff
@@ -43,54 +43,29 @@ def from_ev_charger_data(cls, data: EVChargerData) -> EVChargerState:
...
-@dataclass(frozen=True)
-class EVChargerPoolStates:
-    """States of all EV Chargers in the pool."""
-
-    _states: dict[int, EVChargerState]
-    _changed_component: Optional[int] = None
-
-    def __iter__(self) -> Iterator[tuple[int, EVChargerState]]:
-        """Iterate over states of all EV Chargers.
+    def ev_connected(self) -> bool:
```
leandro-lucarella-frequenz:
```suggestion
    def is_ev_connected(self) -> bool:
```

src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool.py
```diff
@@ -62,14 +76,27 @@ def __init__(
...
         self._namespace: str = f"ev-charger-pool-{uuid.uuid4()}"
         self._formula_pool: FormulaEnginePool = FormulaEnginePool(
             self._namespace,
             self._channel_registry,
             self._resampler_subscription_sender,
         )
 
-    async def total_current(self) -> FormulaReceiver3Phase:
+    @property
+    def component_ids(self) -> set[int]:
```
leandro-lucarella-frequenz: You probably don't want the client to mutate this, so I suggest returning a [`collections.abc.Set`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Set) instead, which is immutable.

leandro-lucarella-frequenz (commented):
Also just a few minor things from my side :+1:

src/frequenz/sdk/timeseries/ev_charger_pool/_state_tracker.py
daniel-zullo-frequenz: ok, I didn't know about that pattern. However I have to admit this sounds to me like returning a value is optional rather than a boolean flag is returned, but this is probably just me :D

+1 about renaming the function to `is_ev_connected()`

shsms: renamed.

src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool.py
shsms: done

tests/timeseries/test_ev_charger_pool.py
shsms: done

src/frequenz/sdk/timeseries/ev_charger_pool/_state_tracker.py
shsms: fixed

src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool.py
shsms: I'm assuming you just mean the type hint,  so that mypy would catch it when people try to mutate it.  Because the type is an ABC otherwise,  and there's no default read-only implementation.

leandro-lucarella-frequenz: Yes, just the type hint.

And there is a read-only implementation, it is called `frozenset` and it is a built-in too.

shsms: oh true,  I've modified it to return a frozenset too.

shsms: I've updated the type hint to frozenset also,  so it is easier to add type hints for users as well,  because mypy complains about missing `AbstractSet`,  it turns out `Set` is an alias:

```python
    @property
    def component_ids(self) -> frozenset[int]:
        ...
        return frozenset(self._component_ids)

File: src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool.py [hunk 10]
```diff
@@ -62,14 +77,27 @@ def __init__(
...
         self._namespace: str = f"ev-charger-pool-{uuid.uuid4()}"
         self._formula_pool: FormulaEnginePool = FormulaEnginePool(
             self._namespace,
             self._channel_registry,
             self._resampler_subscription_sender,
         )
 
-    async def total_current(self) -> FormulaReceiver3Phase:
+    @property
+    def component_ids(self) -> Set[int]:
```
daniel-zullo-frequenz: this is still returning `Set`, and I understood you wanted to change it to `frozenset`. Have you pushed the changes?

shsms: oops, sorry, just did.

daniel-zullo-frequenz (dismissed):

src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool.py
leandro-lucarella-frequenz: OK, the difference here is you are copying the data (and rebuilding a hash set). Specially being a property, IMHO it should be as light as possible.

I guess it's not a big deal, but IMHO it would be a better balance to just make the type system enforce this.

I just tried this and for me `mypy` doesn't complain about missing `AbstractSet` but it does say that `AbstractSet` doesn't support X instead of `Set`, which is indeed a bit weird, but I'm not sure if weird enough.

```py

from collections.abc import Set


def f() -> Set[int]:
    return {1, 2, 3}


s1: Set[int] = f()
s2: set[int] = f()  # line 11
f().add(4)  # line 12
```

```console
$ mypy t.py
t.py:11: error: Incompatible types in assignment (expression has type "AbstractSet[int]", variable has type "Set[int]")  [assignment]
t.py:12: error: "AbstractSet[int]" has no attribute "add"  [attr-defined]
Found 2 errors in 1 file (checked 1 source file)
```

Anyway, just giving an opinion, whatever you prefer is fine with me.

leandro-lucarella-frequenz (dismissed):
As said in the follow-up comment, I think having the set immutable by the type system is a better approach but not a blocker.

src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool.py
shsms: Line 11's error was confusing for me,  maybe others can figure it out,  or even better,  can just use it directly, without assigning.  

It is not a performance sensitive part, but I agree it doesn't make sense to make a copy with every call.

The other option is to use a frozenset for the internal representation,  because once created,  we don't update it internally either.

I don't know that using `abc.Set` is a good move,  but I've done it anyway,  and can leave it like this until we have more info,  and deal with it separately if necessary.

daniel-zullo-frequenz (approved):

PR #270, opened 2023-03-19
Add a `DataPipeline` implementation

shsms:
... that abstracts away the creation of the data sourcing and resampling actors,  and provides an easy way to access the logical meter or one of the pools.

Currently only provides two methods: `logical_meter` and `ev_charger_pool`.

`battery_pool` probably needs some consensus on how args should be pass and interaction between `PowerDistributingActor`,  which can be done in a subsequent PR.

This allows for the following interface:

``` python
async def run():
    await microgrid.initialize(
        host=HOST, port=PORT, resampler_config=ResamplerConfig(resampling_period_s=1.0)
    )
    grid_power = microgrid.logical_meter().grid_power()
```

This PR also renames:
1. `microgrid._microgrid` to `microgrid.connection_manager`
2. `tests/utils/mock_microgrid.py` to `mock_microgrid_client.py`


shsms:
Was this supposed to have a different name?  Especially because the *Pools are also supposed to send control commands back to the microgrid?

src/frequenz/sdk/timeseries/data_pipeline.py
```diff
@@ -0,0 +1,115 @@
...
+        Returns:
+            An EVChargerPool instance.
+        """
+        if self._ev_charger_pool is None:
+            self._ev_charger_pool = EVChargerPool(
+                channel_registry=self._channel_registry,
+                resampler_subscription_sender=self._resampling_request_sender(),
+                component_ids=self._ev_charger_ids,
+            )
+        return self._ev_charger_pool
```
ela-kotulska-frequenz: We might need only one instance of ChannelRegistry and many instances of EvChargerPool.
Consider to add ev_charger_ids argument to this methods (instead of in constructor): 
```
def ev_charger_pool(self, ev_charger_ids: Set[int]) -> EVChargerPool:
```
And save them by components ids.
So if user will ask for ev_pool:
* with components ids that already was created, then you should return this EvPool.
* If EVPool with this ids does not exist, then class should create this EVPool, save and return.

ela-kotulska-frequenz:
> Was this supposed to have a different name? Especially because the *Pools are also supposed to send control commands back to the microgrid?

Maybe SDKInterface?

src/frequenz/sdk/timeseries/data_pipeline.py
```diff
@@ -0,0 +1,115 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Create, connect and own instances of data pipeline components."""
```
mathias-baumann-frequenz: I feel this description doesn't really tell me anything about what this class can be used for

src/frequenz/sdk/timeseries/data_pipeline.py
```diff
@@ -0,0 +1,115 @@
...
+        self._ev_charger_ids = ev_charger_ids
+
+        self._channel_registry = ChannelRegistry(name="Data Pipeline Registry")
+
+        self._data_sourcing_actor: _ActorInfo | None = None
+        self._resampling_actor: _ActorInfo | None = None
+        self._logical_meter: LogicalMeter | None = None
+        self._ev_charger_pool: EVChargerPool | None = None
+
+    def _data_sourcing_request_sender(self) -> Sender[ComponentMetricRequest]:
```
mathias-baumann-frequenz: I'd suggest to put all private methods to the bottom to see the public relevant methods first

src/frequenz/sdk/timeseries/data_pipeline.py
shsms: done

src/frequenz/sdk/timeseries/data_pipeline.py
shsms: I've added a few more lines,  but this is now also internal.  So end users who don't have enough context will no longer be exposed to it directly.

src/frequenz/sdk/timeseries/data_pipeline.py
shsms: done.

src/frequenz/sdk/microgrid/__init__.py
```diff
@@ -7,15 +7,29 @@
...
-from . import client, component
+from ..actor import ResamplerConfig
+from . import _data_pipeline, client, component, connection_manager
+from ._data_pipeline import ev_charger_pool, logical_meter
 from ._graph import ComponentGraph
-from ._microgrid import ConnectionManager, get, initialize
+
+
+async def initialize(host: str, port: int, resampler_config: ResamplerConfig) -> None:
+    """Initialize the microgrid connection manager and the data pipleine.
```
daniel-zullo-frequenz:
```suggestion
    """Initialize the microgrid connection manager and the data pipeline.
```

src/frequenz/sdk/microgrid/_data_pipeline.py
```diff
@@ -0,0 +1,207 @@
...
+        ComponentMetricsResamplingActor,
+        DataSourcingActor,
+        ResamplerConfig,
+    )
+    from ..timeseries.ev_charger_pool import EVChargerPool
+    from ..timeseries.logical_meter import LogicalMeter
+
+
+@dataclass
+class _ActorInfo:
```
daniel-zullo-frequenz, line 34: you might want to add a brief line of documentation for this class

src/frequenz/sdk/microgrid/_data_pipeline.py
```diff
@@ -0,0 +1,207 @@
...
+
+        if key not in self._ev_charger_pools:
+            self._ev_charger_pools[key] = EVChargerPool(
+                channel_registry=self._channel_registry,
+                resampler_subscription_sender=self._resampling_request_sender(),
+                component_ids=ev_charger_ids,
+            )
+        return self._ev_charger_pools[key]
+
+    def _data_sourcing_request_sender(self) -> Sender[ComponentMetricRequest]:
```
daniel-zullo-frequenz, line 119: documentation is missing for this function

daniel-zullo-frequenz (commented):
I have only a few minor comments to check for and LGTM otherwise

I think these changes need an entry to the release notes

src/frequenz/sdk/microgrid/__init__.py
shsms: wow, fixed. :-)

src/frequenz/sdk/microgrid/_data_pipeline.py
shsms: done.

src/frequenz/sdk/microgrid/_data_pipeline.py
shsms: done, for the other private functions as well.

shsms:
> I think these changes need an entry to the release notes

This is also done.

src/frequenz/sdk/microgrid/_data_pipeline.py
```diff
@@ -0,0 +1,224 @@
...
+        self._resampler_config = resampler_config
+
+        self._channel_registry = ChannelRegistry(name="Data Pipeline Registry")
+
+        self._data_sourcing_actor: _ActorInfo | None = None
+        self._resampling_actor: _ActorInfo | None = None
+        self._logical_meter: "LogicalMeter" | None = None
+        self._ev_charger_pools: dict[frozenset[int], "EVChargerPool"] = {}
+
+    def logical_meter(self) -> LogicalMeter:
```
daniel-zullo-frequenz, line 69: I wonder if this be suitable to make it a @property

daniel-zullo-frequenz (commented):
I only have a question and LGTM otherwise

src/frequenz/sdk/microgrid/_data_pipeline.py
shsms: I don't know, in my opinion, this does actual work, so it is not a property.  Also, it is internal and not the public api, so it doesn't matter.

daniel-zullo-frequenz: I was mainly considering to use this kind of syntax `microgrid.logical_meter.grid_power()` instead but as you have said it doesn't matter as it is internal

daniel-zullo-frequenz (approved):

src/frequenz/sdk/microgrid/_data_pipeline.py
shsms: microgrid.logical_meter() is a separate function,  that calls this class method: https://github.com/frequenz-floss/frequenz-sdk-python/pull/270/files#diff-9f32f637c4ac702aa43914f3ab2949c4b6adef41b482c1d300a4262c06b1f995R191-R199

PR #277, opened 2023-03-29
Various small fixes

### LINKED ISSUE #273: Check all `except CancelledError` that they re-raise the exception
CancelledErrors should be re-raised according to https://docs.python.org/3/library/asyncio-task.html#task-cancellation and we should check our code that it really does that.

mathias-baumann-frequenz:
- Decrease code duplication by using existing method.
- Enhance documentation of power distributing actor method.
- FormularEngine: Re-Raise CancelledError in run method

fixes #273

### DIFF
--- a/src/frequenz/sdk/actor/_decorator.py
+++ b/src/frequenz/sdk/actor/_decorator.py
@@ -14,6 +14,8 @@ import inspect
 import logging
 from typing import Any, Generic, Optional, Type, TypeVar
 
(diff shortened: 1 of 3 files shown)


leandro-lucarella-frequenz (commented):
LGTM but I will let @sahas-subramanian-frequenz to approve, as it touches the formula engine.

sahas-subramanian-frequenz (approved):
LGTM as well,  I guess that was a bug in the formula engine.

PR #299, opened 2023-03-31
Add PeriodicFeatureExtractor

### LINKED ISSUE #120: Add a `PeriodicFeatureExtractor`
### What's needed?

For forecasting we have been using average load and production profiles. 

Given a time series, we want a module that is able to calculate the item wise average of repeating time windows of the same size and the same sample-rate, possibly all having the same distance.

**Example:**

Consider the time series $T = [1,2,3,4,5,6,7,8,9,1]$ and a window a length $2$ with distance $2$ starting at the first position.

Hence we get the sub-series $t_1=[1,2], t_2=[5,6], t_3=[9,1]$ and therefore the average $$t_{avg}=[(1+5+9)/3, (2+6+1)/3] = [5, 3]$$

*Note*: A position could likewise be a timestamp in a usual time-series representation.

### Proposed solution

We want to design such a module in a way such that it

* uses batches of data (most likely cached in memory)
* is designed to work together with the data structure defined [here](https://github.com/frequenz-floss/frequenz-sdk-python/issues/113)
* it should be able to run calculations on demand and memoize the result

### Use cases

It's been used for calculating average load and production profiles certain days.
For example it could just calculate the average load of some consume of the last seven days as well as the average PV production of the last four Tuesdays.

It's possible to run the calculation and update the mean on a data stream whenever a new sample arrives. However we plan on storing batches of data in any case, as this can be used as a feature for ml models, and thus calculating the averages from the batches seems obvious.
Additionally this option hasn't been further explored so far.

### Additional context

We have had a use-case where we where looking at a full days averaged data plus a 15 minute [margin](https://github.com/frequenz-floss/frequenz-sdk-python/issues/120#issuecomment-1359587158). Thus we will need to add a margin for a window size in order to re-engineer this feature.

matthias-wende-frequenz:
A class for creating a profile from periodically occurring windows in a buffer of historical data that can be used for creating features for ml models.

closes #120

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -28,7 +28,13 @@ This release drops support for Python versions older than 3.11.
 
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+* The `MovingWindow` as well as the OrderedRingBuffer are having new public methods that are returning the oldest and newest timestamp of all stored samples.
+
+* The `PeriodicFeatureExtractor` has been added.
+
+  This is a tool to create certain profiles out of periodic reoccurring windows inside a `MovingWindow`.
+
+  As an example one can create a daily profile of specific weekdays which will be returned as numpy arrays.
 
 ## Bug Fixes
 
--- a/benchmarks/timeseries/periodic_feature_extractor.py
+++ b/benchmarks/timeseries/periodic_feature_extractor.py
@@ -1,0 +1,232 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""
+Benchmarks for the PeriodicFeatureExtractor class.
+
+This module contains benchmarks that are comparing
+the performance of a numpy implementation with a python
+implementation.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import logging
+from datetime import datetime, timedelta, timezone
+from functools import partial
+from timeit import timeit
+from typing import List
+
+import numpy as np
+from frequenz.channels import Broadcast
+from numpy.random import default_rng
+from numpy.typing import NDArray
+
+from frequenz.sdk.timeseries import MovingWindow, PeriodicFeatureExtractor, Sample
+
+
+async def init_feature_extractor(period: int) -> PeriodicFeatureExtractor:
+    """Initialize the PeriodicFeatureExtractor class."""
+    # We only need the moving window to initialize the PeriodicFeatureExtractor class.
+    lm_chan = Broadcast[Sample]("lm_net_power")
+    moving_window = MovingWindow(
+        timedelta(seconds=1), lm_chan.new_receiver(), timedelta(seconds=1)
+    )
+
+    await lm_chan.new_sender().send(Sample(datetime.now(tz=timezone.utc), 0))
+
+    # Initialize the PeriodicFeatureExtractor class with a period of period seconds.
+    # This works since the sampling period is set to 1 second.
+    return PeriodicFeatureExtractor(moving_window, timedelta(seconds=period))
+
+
+def _calculate_avg_window(
+    feature_extractor: PeriodicFeatureExtractor,
+    window: NDArray[np.float_],
+    window_size: int,
+) -> NDArray[np.float_]:
+    """
+    Reshapes the window and calculates the average.
+
+    This method calculates the average of a window by averaging over all
+    windows fully inside the passed numpy array having the period
+    `self.period`.
+
+    Args:
+        feature_extractor: The instance of the PeriodicFeatureExtractor to use.
+        window: The window to calculate the average over.
+        window_size: The size of the window to calculate the average over.
+        weights: The weights to use for the average calculation.
+
+    Returns:
+        The averaged window.
+    """
+    reshaped = feature_extractor._reshape_np_array(  # pylint: disable=protected-access
+        window, window_size
+    )
+    # ignoring the type because np.average returns Any
+    return np.average(reshaped[:, :window_size], axis=0)  # type: ignore[no-any-return]
+
+
+def _calculate_avg_window_py(
+    feature_extractor: PeriodicFeatureExtractor,
+    window: NDArray[np.float_],
+    window_size: int,
+    weights: List[float] | None = None,
+) -> NDArray[np.float_]:
+    """
+    Plain python version of the average calculator.
+
+    This method avoids copying in any case but is 15 to 600 slower then the
+    numpy version.
+
+    This method is only used in these benchmarks.
+
+    Args:
+        feature_extractor: The instance of the PeriodicFeatureExtractor to use.
+        window: The window to calculate the average over.
+        window_size: The size of the window to calculate the average over.
+        weights: The weights to use for the average calculation.
+
+    Returns:
+        The averaged window.
+    """
+
+    def _num_windows(
+        window: NDArray[np.float_] | MovingWindow, window_size: int, period: int
+    ) -> int:
+        """
+        Get the number of windows that are fully contained in the MovingWindow.
+
+        This method calculates how often a given window, defined by it size, is
+        fully contained in the MovingWindow at its current state or any numpy
+        ndarray given the period between two window neighbors.
+
+        Args:
+            window: The buffer that is used for the average calculation.
+            window_size: The size of the window in samples.
+
+        Returns:
+            The number of windows that are fully contained in the MovingWindow.
+        """
+        num_windows = len(window) // period
+        if len(window) - num_windows * period >= window_size:
+            num_windows += 1
+
+        return num_windows
+
+    period = feature_extractor._period  # pylint: disable=protected-access
+
+    num_windows = _num_windows(
+        window,
+        window_size,
+        period,
+    )
+
+    res = np.empty(window_size)
+
+    for i in range(window_size):
+        assert num_windows * period - len(window) <= period
+        summe = 0
+        for j in range(num_windows):
+            if weights is None:
+                summe += window[i + (j * period)]
+            else:
+                summe += weights[j] * window[i + (j * period)]
+
+        if not weights:
+            res[i] = summe / num_windows
+        else:
+            res[i] = summe / np.sum(weights)
+
+    return res
+
+
+def run_benchmark(
+    array: NDArray[np.float_],
+    window_size: int,
+    feature_extractor: PeriodicFeatureExtractor,
+) -> None:
+    """Run the benchmark for the given ndarray and window size."""
+
+    def run_avg_np(
+        array: NDArray[np.float_],
+        window_size: int,
+        feature_extractor: PeriodicFeatureExtractor,
+    ) -> None:
+        """
+        Run the FeatureExtractor.
+
+        The return value is discarded such that it can be used by timit.
+
+        Args:
+            a: The array containing all data.
+            window_size: The size of the window.
+            feature_extractor: An instance of the PeriodicFeatureExtractor.
+        """
+        _calculate_avg_window(feature_extractor, array, window_size)
+
+    def run_avg_py(
+        array: NDArray[np.float_],
+        window_size: int,
+        feature_extractor: PeriodicFeatureExtractor,
+    ) -> None:
+        """
+        Run the FeatureExtractor.
+
+        The return value is discarded such that it can be used by timit.
+
+        Args:
+            a: The array containing all data.
+            window_size: The size of the window.
+            feature_extractor: An instance of the PeriodicFeatureExtractor.
+        """
+        _calculate_avg_window_py(feature_extractor, array, window_size)
+
+    time_np = timeit(
+        partial(run_avg_np, array, window_size, feature_extractor), number=10
+    )
+    time_py = timeit(
+        partial(run_avg_py, array, window_size, feature_extractor), number=10
+    )
+    print(time_np)
+    print(time_py)
+    print(f"Numpy is {time_py / time_np} times faster!")
+
+
+DAY_S = 24 * 60 * 60
+
+
+async def main() -> None:
+    """
+    Run the benchmarks.
+
+    The benchmark are comparing the numpy
+    implementation with the python implementation.
+    """
+    # initialize random number generator
+    rng = default_rng()
+
+    # create a random ndarray with 29 days -5 seconds of data
+    days_29_s = 29 * DAY_S
+    feature_extractor = await init_feature_extractor(10)
+    data = rng.standard_normal(days_29_s)
+    run_benchmark(data, 4, feature_extractor)
+
+    days_29_s = 29 * DAY_S + 3
+    data = rng.standard_normal(days_29_s)
+    run_benchmark(data, 4, feature_extractor)
+
+    # create a random ndarray with 29 days +5 seconds of data
+    data = rng.standard_normal(29 * DAY_S + 5)
+
+    feature_extractor = await init_feature_extractor(7 * DAY_S)
+    # TEST one day window and 6 days distance. COPY (Case 3)
+    run_benchmark(data, DAY_S, feature_extractor)
+    # benchmark one day window and 6 days distance. NO COPY (Case 1)
+    run_benchmark(data[: 28 * DAY_S], DAY_S, feature_extractor)
+
+
+logging.basicConfig(level=logging.DEBUG)
+asyncio.run(main())
--- a/src/frequenz/sdk/timeseries/_moving_window.py
+++ b/src/frequenz/sdk/timeseries/_moving_window.py
@@ -140,7 +140,8 @@ class MovingWindow:
             input_sampling_period <= size
         ), "The input sampling period should be equal to or lower than the window size."
 
-        sampling = input_sampling_period
+        self._sampling_period = input_sampling_period
+
         self._resampler: Resampler | None = None
         self._resampler_sender: Sender[Sample] | None = None
         self._resampler_task: asyncio.Task[None] | None = None
@@ -151,15 +152,17 @@ class MovingWindow:
             ), "The resampling period should be equal to or lower than the window size."
 
             self._resampler = Resampler(resampler_config)
-            sampling = resampler_config.resampling_period
+            self._sampling_period = resampler_config.resampling_period
 
         # Sampling period might not fit perfectly into the window size.
-        num_samples = math.ceil(size.total_seconds() / sampling.total_seconds())
+        num_samples = math.ceil(
+            size.total_seconds() / self._sampling_period.total_seconds()
+        )
 
         self._resampled_data_recv = resampled_data_recv
         self._buffer = OrderedRingBuffer(
             np.empty(shape=num_samples, dtype=float),
-            sampling_period=sampling,
+            sampling_period=self._sampling_period,
             align_to=align_to,
         )
 
@@ -169,6 +172,16 @@ class MovingWindow:
         self._update_window_task: asyncio.Task[None] = asyncio.create_task(
             self._run_impl()
         )
+
+    @property
+    def sampling_period(self) -> timedelta:
+        """
+        Return the sampling period of the MovingWindow.
+
+        Returns:
+            The sampling period of the MovingWindow.
+        """
+        return self._sampling_period
 
     async def _run_impl(self) -> None:
         """Awaits samples from the receiver and updates the underlying ring buffer.
--- a/src/frequenz/sdk/timeseries/_moving_window.py
+++ b/src/frequenz/sdk/timeseries/_moving_window.py
@@ -281,7 +281,17 @@ class MovingWindow:
             an numpy array if the key is a slice.
         """
         if isinstance(key, slice):
+            if isinstance(key.start, int) or isinstance(key.stop, int):
+                if key.start is None or key.stop is None:
+                    key = slice(slice(key.start, key.stop).indices(self.__len__()))
+            elif isinstance(key.start, datetime) or isinstance(key.stop, datetime):
+                if key.start is None:
+                    key = slice(self._buffer.time_bound_oldest, key.stop)
+                if key.stop is None:
+                    key = slice(key.start, self._buffer.time_bound_newest)
+
             _logger.debug("Returning slice for [%s:%s].", key.start, key.stop)
+
             # we are doing runtime typechecks since there is no abstract slice type yet
             # see also (https://peps.python.org/pep-0696)
             if isinstance(key.start, datetime) and isinstance(key.stop, datetime):
--- a/src/frequenz/sdk/timeseries/_periodic_feature_extractor.py
+++ b/src/frequenz/sdk/timeseries/_periodic_feature_extractor.py
@@ -1,0 +1,419 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""
+A feature extractor for historical timeseries data.
+
+It creates a profile from periodically occurring windows in a buffer of
+historical data.
+
+The profile is created out of all windows that are fully contained in the
+underlying buffer, e.g. the moving window, with the same start and end time
+modulo a fixed period.
+"""
+
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass
+from datetime import datetime, timedelta
+from typing import List, Tuple
+
+import numpy as np
+from numpy.typing import NDArray
+
+from .._internal._math import is_close_to_zero
+from ._moving_window import MovingWindow
+from ._ringbuffer import OrderedRingBuffer
+
+_logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class RelativePositions:
+    """
+    Holds the relative positions of a window in a buffer.
+
+    When calculating a profile the user has to pass two datetime objects which
+    are determaning a window. It's allowed that this window is not contained in
+    the historical data buffer.
+    In order to calculate the profile the window is shifted to the first
+    position in the buffer where it occurrs first and represented by indexes
+    relative to the oldest sample in the buffer.
+    """
+
+    start: int
+    """The relative start position of the window."""
+    end: int
+    """The relative end position of the window."""
+    next: int
+    """The relative position of the next incoming sample."""
+
+
+class PeriodicFeatureExtractor:
+    """
+    A feature extractor for historical timeseries data.
+
+    This class is creating a profile from periodically occurring windows in a
+    buffer of historical data.
+
+    The profile is created out of all windows that are fully contained in the
+    underlying buffer with the same start and end time modulo a fixed period.
+
+    Consider for example a timeseries $T$ of historical data and sub-series
+    $S_i$ of $T$ all having the same size $l$ and the same fixed distance $p$
+    called period, where period of two sub-windows is defined as the distance
+    of two points at the same position within the sub-windows.
+
+    This class calculates a statistical profile $S$ over all $S_i$, i.e. the
+    value of $S$ at position $i$ is calculated by performing a certain
+    calculation, e.g. an average, over all values of $S_i$ at position $i$.
+
+    Note:
+        The oldest window or the window that is currently overwritten in the
+        `MovingWindow` is not considered in the profile.
+
+    Note:
+        When constructing a `PeriodicFeatureExtractor` object the
+        `MovingWindow` size has to be a integer multiple of the period.
+
+    Example:
+        ```python
+        from datetime import datetime, timedelta, timezone
+
+        moving_window = MovingWindow(
+            size=timedelta(days=35),
+            resampled_data_recv=microgrid.logical_meter().grid_power.new_receiver(),
+            input_sampling_period=timedelta(seconds=1),
+        )
+
+        feature_extractor = PeriodicFeatureExtractor(
+            moving_window = moving_window,
+            period=timedelta(days=7),
+        )
+
+        now = datetime.now(timezone.utc)
+
+        # create a daily weighted average for the next 24h
+        avg_24h = feature_extractor.avg(
+            now,
+            now + timedelta(hours=24),
+            weights=[0.1, 0.2, 0.3, 0.4]
+        )
+
+        # create a daily average for Thursday March 23 2023
+        th_avg_24h = feature_extractor.avg(datetime(2023, 3, 23), datetime(2023, 3, 24))
+        ```
+    """
+
+    def __init__(
+        self,
+        moving_window: MovingWindow,
+        period: timedelta,
+    ) -> None:
+        """
+        Initialize a PeriodicFeatureExtractor object.
+
+        Args:
+            moving_window: The MovingWindow that is used for the average calculation.
+            period: The distance between two succeeding intervals.
+
+        Raises:
+            ValueError: If the MovingWindow size is not a integer multiple of the period.
+        """
+        self._moving_window = moving_window
+
+        self._sampling_period = self._moving_window.sampling_period
+        """The sampling_period as float to use it for indexing of samples."""
+
+        self._period = int(period / self._sampling_period)
+        """Distance between two succeeding intervals in samples."""
+
+        _logger.debug("Initializing PeriodicFeatureExtractor!")
+        _logger.debug("MovingWindow size: %i", len(self._moving_window))
+        _logger.debug(
+            "Period between two succeeding intervals (in samples): %i",
+            self._period,
+        )
+
+        if not len(self._moving_window) % self._period == 0:
+            raise ValueError(
+                "The MovingWindow size is not a integer multiple of the period."
+            )
+
+        if not is_close_to_zero(self._period - period / self._sampling_period):
+            raise ValueError(
+                "The period is not a multiple of the sampling period. "
+                "This might result in unexpected behaviour."
+            )
+
+    @property
+    def _buffer(self) -> OrderedRingBuffer[NDArray[np.float64]]:
+        return self._moving_window._buffer  # pylint: disable=protected-access
+
+    def _timestamp_to_rel_index(self, timestamp: datetime) -> int:
+        """
+        Get the index of a timestamp relative to the oldest sample in the MovingWindow.
+
+        In other word consider an integer axis where the zero is defined as the
+        oldest element in the MovingWindow. This function returns the index of
+        the given timestamp an this axis.
+
+        This method can return negative values.
+
+        Args:
+            timestamp: A timestamp that we want to shift into the window.
+
+        Returns:
+            The index of the timestamp shifted into the MovingWindow.
+        """
+        # align timestamp to the sampling period
+        timestamp = self._buffer.normalize_timestamp(timestamp)
+
+        # distance between the input ts and the ts of oldest known samples (in samples)
+        dist_to_oldest = int(
+            (timestamp - self._buffer.time_bound_oldest) / self._sampling_period
+        )
+
+        _logger.debug("Shifting ts: %s", timestamp)
+        _logger.debug("Oldest timestamp in buffer: %s", self._buffer.time_bound_oldest)
+        _logger.debug("Distance to the oldest sample: %i", dist_to_oldest)
+
+        return dist_to_oldest
+
+    def _reshape_np_array(
+        self, array: NDArray[np.float_], window_size: int
+    ) -> NDArray[np.float_]:
+        """
+        Reshape a numpy array to a 2D array where each row represents a window.
+
+        There are three cases to consider
+
+        1. The array size is a multiple of window_size + period,
+           i.e. num_windows is integer and we can simply reshape.
+        2. The array size is NOT a multiple of window_size + period and
+           it has length n * period + m, where m < window_size.
+        3. The array size is NOT a multiple of window_size + period and
+           it has length n * period + m, where m >= window_size.
+
+        Note that in the current implementation of this class we have the restriction
+        that period is a multiple integer of the size of the MovingWindow and hence
+        only case 1 can occur.
+
+        Args:
+            array: The numpy array to reshape.
+            window_size: The size of the window in samples.
+
+        Returns:
+            The reshaped 2D array.
+
+        Raises:
+            ValueError: If the array is smaller or equal to the given period.
+        """
+        # Not using the num_windows function here because we want to
+        # differentiate between the three cases.
+        if len(array) < self._period:
+            raise ValueError(
+                f"The array (length:{len(array)}) is too small to be reshaped."
+            )
+
+        num_windows = len(array) // self._period
+
+        # Case 1:
+        if len(array) - num_windows * self._period == 0:
+            resized_array = array
+        # Case 2
+        elif len(array) - num_windows * self._period < window_size:
+            resized_array = array[: num_windows * self._period]
+        # Case 3
+        else:
+            num_windows += 1
+            resized_array = np.resize(array, num_windows * self._period)
+
+        return resized_array.reshape(num_windows, self._period)
+
+    def _get_relative_positions(
+        self, start: datetime, window_size: int
+    ) -> RelativePositions:
+        """
+        Return relative positions of the MovingWindow.
+
+        This method calculates the shifted relative positions of the start
+        timestamp, the end timestamps as well as the next position that is
+        overwritten in the ringbuffer.
+        Shifted in that context means that the positions are moved as close
+        assume possible to the oldest sample in the MovingWindow.
+
+        Args:
+            start: The start timestamp of the window.
+            window_size: The size of the window in samples.
+
+        Returns:
+            The relative positions of the start, end and next samples.
+
+        Raises:
+            ValueError: If the start timestamp is after the end timestamp.
+        """
+        # The number of usable windows can change, when the current position of
+        # the ringbuffer is inside one of the windows inside the MovingWindow.
+        # Since this is possible, we assume that one window is always not used
+        # for the average calculation.
+        #
+        # We are ignoring either the window that is currently overwritten if
+        # the current position is inside that window or the window that would
+        # be overwritten next.
+        #
+        # Move the window to its first appereance in the MovingWindow relative
+        # to the oldest sample stored in the MovingWindow.
+        #
+        # In other words the oldest stored sample is considered to have index 0.
+        #
+        # Note that the returned value is a index not a timestamp
+        rel_start_sample = self._timestamp_to_rel_index(start) % self._period
+        rel_end_sample = rel_start_sample + window_size
+
+        # check if the newest time bound, i.e. the sample that is currently written,
+        # is inside the interval
+        rb_current_position = self._buffer.time_bound_newest
+        rel_next_position = (
+            self._timestamp_to_rel_index(rb_current_position) + 1
+        ) % self._period
+        # fix the rel_next_position if modulo period the next position
+        # is smaller then the start sample position
+        if rel_next_position < rel_start_sample:
+            rel_next_position += self._period
+
+        rel_next_position += self._period * (window_size // self._period)
+
+        _logger.debug("current position of the ringbuffer: %s", rb_current_position)
+        _logger.debug("relative start_sample: %s", rel_start_sample)
+        _logger.debug("relative end_sample: %s", rel_end_sample)
+        _logger.debug("relative next_position: %s", rel_next_position)
+
+        return RelativePositions(rel_start_sample, rel_end_sample, rel_next_position)
+
+    def _get_buffer_bounds(
+        self, start: datetime, end: datetime
+    ) -> Tuple[int, int, int]:
+        """
+        Get the bounds of the ringbuffer used for further operations.
+
+        This method uses the given start and end timestamps to calculate the
+        part of the ringbuffer that can be used for further operations, like
+        average or min/max.
+
+        Here we cut out the oldest window or the window that is currently
+        overwritten in the MovingWindow such that it is not considered in any
+        further operation.
+
+        Args:
+            start: The start timestamp of the window.
+            end: The end timestamp of the window.
+
+        Returns:
+            The bounds of the to be used buffer and the window size.
+
+        Raises:
+            ValueError: If the start timestamp is after the end timestamp.
+        """
+        window_size = self._timestamp_to_rel_index(end) - self._timestamp_to_rel_index(
+            start
+        )
+        if window_size <= 0:
+            raise ValueError("Start timestamp must be before end timestamp")
+        if window_size > self._period:
+            raise ValueError(
+                "The window size must be smaller or equal than the period."
+            )
+
+        rel_pos = self._get_relative_positions(start, window_size)
+
+        if window_size > len(self._moving_window):
+            raise ValueError(
+                "The window size must be smaller than the size of the `MovingWindow`"
+            )
+
+        # shifted distance between the next incoming sample and the start of the window
+        dist_to_start = rel_pos.next - rel_pos.start
+
+        # get the start and end position inside the ringbuffer
+        end_pos = (
+            self._timestamp_to_rel_index(self._buffer.time_bound_newest) + 1
+        ) - dist_to_start
+
+        # Note that these check is working since we are using the positions
+        # relative to the oldest sample stored in the MovingWindow.
+        if rel_pos.start <= rel_pos.next < rel_pos.end:
+            # end position is start_position of the window that is currently written
+            # that's how end_pos is currently set
+            _logger.debug("Next sample will be inside the window time interval!")
+        else:
+            _logger.debug("Next sample will be outside the window time interval!")
+            # end position is start_position of the window that
+            # is overwritten next, hence we adding period.
+            end_pos += self._period
+
+        # add the offset to the oldest sample in the ringbuffer and wrap around
+        # to get the start and end positions in the ringbuffer
+        rb_offset = self._buffer.datetime_to_index(self._buffer.time_bound_oldest)
+        start_pos = self._buffer.wrap(end_pos + self._period + rb_offset)
+        end_pos = self._buffer.wrap(end_pos + rb_offset)
+
+        _logger.debug("start_pos in ringbuffer: %s", start_pos)
+        _logger.debug("end_pos in ringbuffer: %s", end_pos)
+
+        return (start_pos, end_pos, window_size)
+
+    def _get_reshaped_np_array(
+        self, start: datetime, end: datetime
+    ) -> Tuple[NDArray[np.float_], int]:
+        """
+        Create a reshaped numpy array from the MovingWindow.
+
+        The reshaped array is a two dimemsional array, where one dimension is
+        the window_size and the other the number of windows returned by the
+        `_get_buffer_bounds` method.
+
+        Args:
+            start: The start timestamp of the window.
+            end: The end timestamp of the window.
+
+        Returns:
+            A tuple containing the reshaped numpy array and the window size.
+        """
+        (start_pos, end_pos, window_size) = self._get_buffer_bounds(start, end)
+
+        if start_pos >= end_pos:
+            window_start = self._buffer[start_pos : len(self._moving_window)]
+            window_end = self._buffer[0:end_pos]
+            # make the linter happy
+            assert isinstance(window_start, np.ndarray)
+            assert isinstance(window_end, np.ndarray)
+            window_array = np.concatenate((window_start, window_end))
+        else:
+            window_array = self._buffer[start_pos:end_pos]
+
+        return (self._reshape_np_array(window_array, window_size), window_size)
+
+    def avg(
+        self, start: datetime, end: datetime, weights: List[float] | None = None
+    ) -> NDArray[np.float_]:
+        """
+        Create the average window out of the window defined by `start` and `end`.
+
+        This method calculates the average of a window by averaging over all
+        windows fully inside the MovingWindow having the period
+        `self.period`.
+
+        Args:
+            start: The start of the window to average over.
+            end: The end of the window to average over.
+            weights: The weights to use for the average calculation (oldest first).
+
+        Returns:
+            The averaged timeseries window.
+        """
+        (reshaped, window_size) = self._get_reshaped_np_array(start, end)
+        return np.average(  # type: ignore[no-any-return]
+            reshaped[:, :window_size], axis=0, weights=weights
+        )
--- a/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
+++ b/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
@@ -152,6 +152,26 @@ class OrderedRingBuffer(Generic[FloatArray]):
 
         self._update_gaps(timestamp, prev_newest, sample.value is None)
 
+    @property
+    def time_bound_oldest(self) -> datetime:
+        """
+        Return the time bounds of the ring buffer.
+
+        Returns:
+            The timestamp of the oldest sample of the ring buffer.
+        """
+        return self._datetime_oldest
+
+    @property
+    def time_bound_newest(self) -> datetime:
+        """
+        Return the time bounds of the ring buffer.
+
+        Returns:
+            The timestamp of the newest sample of the ring buffer.
+        """
+        return self._datetime_newest
+
     def datetime_to_index(
         self, timestamp: datetime, allow_outside_range: bool = False
     ) -> int:
--- a/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
+++ b/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
@@ -130,7 +130,7 @@ class OrderedRingBuffer(Generic[FloatArray]):
             IndexError: When the timestamp to be added is too old.
         """
         # adjust timestamp to be exactly on the sample period time point
-        timestamp = self._normalize_timestamp(sample.timestamp)
+        timestamp = self.normalize_timestamp(sample.timestamp)
 
         # Don't add outdated entries
         if (
@@ -188,7 +188,7 @@ class OrderedRingBuffer(Generic[FloatArray]):
         Returns:
             Index where the value for the given timestamp can be found.
         """
-        timestamp = self._normalize_timestamp(timestamp)
+        timestamp = self.normalize_timestamp(timestamp)
 
         if not allow_outside_range and (
             self._datetime_newest + self._sampling_period < timestamp
@@ -199,7 +199,7 @@ class OrderedRingBuffer(Generic[FloatArray]):
                 f"outside the range [{self._datetime_oldest} - {self._datetime_newest}]"
             )
 
-        return self._wrap(
+        return self.wrap(
             round(
                 (timestamp - self._time_index_alignment).total_seconds()
                 / self._sampling_period.total_seconds()
@@ -376,7 +376,7 @@ class OrderedRingBuffer(Generic[FloatArray]):
             new_gap.start = timestamp + self._sampling_period
             self._gaps.append(new_gap)
 
-    def _normalize_timestamp(self, timestamp: datetime) -> datetime:
+    def normalize_timestamp(self, timestamp: datetime) -> datetime:
         """Normalize the given timestamp to fall exactly on the resampling period.
 
         Args:
@@ -404,7 +404,7 @@ class OrderedRingBuffer(Generic[FloatArray]):
 
         return normalized_timestamp
 
-    def _wrap(self, index: int) -> int:
+    def wrap(self, index: int) -> int:
         """Normalize the given index to fit in the buffer by wrapping it around.
 
         Args:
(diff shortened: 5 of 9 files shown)


matthias-wende-frequenz:
The name should change in order to have a more consistent terminology (in line with MovingWindow). A maybe better name is `AvgWindow`, a tool that creates averaged windows.

@cwasicki, @idlir-shkurti-frequenz, @camille-bouvy-frequenz suggestions?

cwasicki:
Maybe something like `LagWindow`?

On another note, are you planning to support other metrics beyond `average` as well? Because we would also need e.g. `min`, `max`, `std`.

tests/timeseries/test_periodic_feature_extractor.py
```diff
@@ -0,0 +1,137 @@
...
+
+    return PeriodicFeatureExtractor(moving_window=window, period=period)
+
+
+async def init_feature_extractor_no_data(period: int) -> PeriodicFeatureExtractor:
+    """
+    Initialize a PeriodicFeatureExtractor with a `MovingWindow` that contains no data.
+
+    Args:
+        period: List[float]
```
idlir-shkurti-frequenz: Please define what the period is here.

matthias-wende-frequenz: done :)

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,471 @@
...
+            resized_array = array[: num_windows * self._distance]
+        # Case 3
+        else:
+            log.debug("Case 3:")
+            num_windows += 1
+            resized_array = np.resize(array, num_windows * self._distance)
+
+        return resized_array.reshape(num_windows, self._distance)
+
+    def _calculate_avg_window(
```
matthias-wende-frequenz: This method is no longer used except in the benchmark and should be moved to the benchmark.

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,471 @@
...
+        Returns:
+            The averaged window.
+        """
+        reshaped = self._reshape_np_array(window, window_size)
+        # ignoring the type because np.average returns Any
+        return np.average(  # type: ignore[no-any-return]
+            reshaped[:, :window_size], axis=0, weights=weights
+        )
+
+    def _calculate_avg_window_py(
```
matthias-wende-frequenz: This method is no longer used except in the benchmark and should be moved to the benchmark.

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,471 @@
...
+            end: The end timestamp of the window.
+
+        Returns:
+            A tuple containing the reshaped numpy array and the window size.
+        """
+        (start_pos, end_pos, window_size) = self._get_buffer_bounds(start, end)
+
+        if start_pos >= end_pos:
+            window_start = self._moving_window[start_pos:]
+            window_end = self._moving_window[:end_pos]
```
matthias-wende-frequenz: Don't use the moving windows [] operator for data access, as the semantics are planned to change soon. As an alternative we can use the ringbuffer directly as we have the instance as `self._buffer` available.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -204,6 +217,24 @@ async def sink_buffer(sample: Sample) -> None:
...
+        return self._buffer.time_bound_oldest()
+
+    def time_bound_newest(self) -> datetime:
+        """
+        Return the time bounds of the ring buffer.
+
+        Returns:
+            The timestamp of the newest sample of the ring buffer.
+        """
+        return self._buffer.time_bound_newest()
```
leandro-lucarella-frequenz: These are good candidates for properties too:

```suggestion
    @property
    def time_bound_oldest(self) -> datetime:
        """
        Return the time bounds of the ring buffer.

        Returns:
            The timestamp of the oldest sample of the ring buffer.
        """
        return self._buffer.time_bound_oldest()

    @property
    def time_bound_newest(self) -> datetime:
        """
        Return the time bounds of the ring buffer.

        Returns:
            The timestamp of the newest sample of the ring buffer.
        """
        return self._buffer.time_bound_newest()
```

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -152,6 +152,24 @@ def update(self, sample: Sample) -> None:
...
+        return self._datetime_oldest
+
+    def time_bound_newest(self) -> datetime:
+        """
+        Return the time bounds of the ring buffer.
+
+        Returns:
+            The timestamp of the newest sample of the ring buffer.
+        """
+        return self._datetime_newest
```
leandro-lucarella-frequenz: These are good candidates for properties:

```suggestion
    @property
    def time_bound_oldest(self) -> datetime:
        """
        Return the time bounds of the ring buffer.

        Returns:
            The timestamp of the oldest sample of the ring buffer.
        """
        return self._datetime_oldest

    @property
    def time_bound_newest(self) -> datetime:
        """
        Return the time bounds of the ring buffer.

        Returns:
            The timestamp of the newest sample of the ring buffer.
        """
        return self._datetime_newest
```

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -163,6 +166,16 @@ def __init__(  # pylint: disable=too-many-arguments
             self._run_impl()
         )
 
+    @property
+    def sampling_period(self) -> timedelta:
+        """
+        Return the sampling period of the MovingWindow.
+
+        Returns:
+            The sampling period of the MovingWindow.
```
leandro-lucarella-frequenz, line 182: Nitpicking :grimacing: 

```suggestion
        Return the sampling period of the `MovingWindow`.

        Returns:
            The sampling period of the `MovingWindow`.
```

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -261,7 +292,21 @@ def __getitem__(self, key: SupportsIndex | datetime | slice) -> float | ArrayLik
             an numpy array if the key is a slice.
         """
         if isinstance(key, slice):
+            if isinstance(key.start, int) or isinstance(key.stop, int):
+                if key.start is None:
+                    _logger.debug("Setting start to 0")
+                    key = slice(0, key.stop)
+                if key.stop is None:
+                    _logger.debug("Setting stop to last index %s", self.__len__())
+                    key = slice(key.start, self.__len__())
```
leandro-lucarella-frequenz: Not sure if you can apply it here, but there is a fairly undocumented method of `slice` that will convert the slice `None` values to real values if you give it a `len` for the container:

```console
$ pydoc slice.indices |cat -p
Help on method_descriptor in slice:

slice.indices = indices(...)
    S.indices(len) -> (start, stop, stride)
    
    Assuming a sequence of length len, calculate the start and stop
    indices, and the stride length of the extended slice described by
    S. Out of bounds indices are clipped in a manner consistent with the
    handling of normal slices.

```

Some usage examples:

```pycon
>>> slice(5).indices(10)
(0, 5, 1)
>>> slice(20).indices(10)
(0, 10, 1)
>>> slice(1, 10).indices(10)
(1, 10, 1)
>>> slice(1, 15).indices(10)
(1, 10, 1)
>>> slice(1, -1).indices(10)
(1, 9, 1)
>>> slice(0, -1).indices(10)
(0, 9, 1)
>>> slice(-5, -1).indices(10)
(5, 9, 1)
```

It won't probably work for the wrapping around, but it might help to reduce the boilerplate here.

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
leandro-lucarella-frequenz, line 1: If this needs to expose only one class, I would make it a private module (`_periodic_feature_extractor.py`) and expose the class in `frequenz.sdk.timeseries` (`src/frequenz/sdk/timeseries/__init__.py`).

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,443 @@
...
+
+import logging
+import math
+from datetime import datetime, timedelta
+from typing import Any, List, Tuple
+
+import numpy as np
+from numpy.typing import NDArray
+
+from frequenz.sdk.timeseries._moving_window import MovingWindow
```
leandro-lucarella-frequenz:
```suggestion
from ._moving_window import MovingWindow
```

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,443 @@
...
+
+log = logging.getLogger(__name__)
```
leandro-lucarella-frequenz: We had some PR recently changing all logger instances to `_logger` :)

```suggestion
_logger = logging.getLogger(__name__)
```

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,443 @@
...
+Consider a timeseries $T$ of historical data and
+subseries $S_i$ of $T$ all having the same size $l$.
+and the same fixed distance $p$ called period,
+where period of two subwindows is defined as the distance
+of two points at the same position within the subwindows.
+
+This class calculates a statistical profile S over all $S_i$,
+i.e. the value of S at position i is calculated by performing
+a certian calculation, e.g. an average, over all values of
+$S_i$ at position i.
```
leandro-lucarella-frequenz: Some format improvements. And I replaced the `S` in `profile S over` and `value of S at position` to `$S$`, not sure if that's correct.

```suggestion
Example:

    Consider a timeseries $T$ of historical data and
    sub-series $S_i$ of $T$ all having the same size $l$.
    and the same fixed distance $p$ called period,
    where period of two sub-windows is defined as the distance
    of two points at the same position within the sub-windows.

    This class calculates a statistical profile $S$ over all $S_i$,
    i.e. the value of $S$ at position i is calculated by performing
    a certain calculation, e.g. an average, over all values of
    $S_i$ at position i.
```

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,443 @@
...
+
+    This class is creating a profile from periodically occurring
+    windows in a buffer of historical data.
+
+    The profile is created out of all windows that are fully
+    contained in the underlying buffer with the same
+    start and end time modulo a fixed period.
+
+    Note that the oldest window or the window that is currently overwritten in
+    the MovingWindow is not considered in the profile.
```
leandro-lucarella-frequenz:
```suggestion
    the `MovingWindow` is not considered in the profile.
```

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,443 @@
...
+
+    Note further that when constructing a PeriodicFeatureExtractor object
```
leandro-lucarella-frequenz:
```suggestion
    Note further that when constructing a `PeriodicFeatureExtractor` object
```

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,443 @@
...
+    the period.
+
+    Example:
+        ```python
+            feature_extractor = PeriodicFeatureExtractor(
+            moving_window = moving_window, # 28 days moving window
+            distance=timedelta(days=7),
+        )
+
+        now = datetime.now(tzinfo=timezone.utc)
```
leandro-lucarella-frequenz: This is just a FYI because I've seen many people use `datetime.now(tzinfo=...)`. You don't actually need to use a keyword argument here, you can just write `datetime.now(timezone.utc)`. Since the value already says `timezone` in it, I think it is pretty clear that is the timezone we are passing (and the type checking will make sure the type is right also).

I'm fine with leaving the keyword argument, again, just a FYI in case it wasn't clear.

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,443 @@
...
+
+        # create a daily weighted average for the next 24h
+        avg_24h = feature_extractor.avg(
+            now, now + timedelta(hours=24),
+            weights=[0.1, 0.2, 0.3, 0.4]
+        )
+
+        # create a daily average for Thursday March 23 2023
+        th_avg_24h = feature_extractor(datetime(2023, 03, 23), datetime(2023, 03, 24))
```
leandro-lucarella-frequenz:
```suggestion
        th_avg_24h = feature_extractor.avg(datetime(2023, 03, 23), datetime(2023, 03, 24))
```

?

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,443 @@
...
+class PeriodicFeatureExtractor:
+    """
+    A tool that extracts a timeseries feature out of historical data.
+
+    This class is creating a profile from periodically occurring
+    windows in a buffer of historical data.
+
+    The profile is created out of all windows that are fully
+    contained in the underlying buffer with the same
+    start and end time modulo a fixed period.
```
sahas-subramanian-frequenz: where do the start and end time come from?  is it from the MovingWindow?  Because the text doesn't say MovingWindow, but "underlying buffer".

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,443 @@
...
+
+    Note that the oldest window or the window that is currently overwritten in
+    the MovingWindow is not considered in the profile.
+
+    Note further that when constructing a PeriodicFeatureExtractor object
+    the `MovingWindow` size has to be a integer multiple of
+    the period.
```
sahas-subramanian-frequenz: There are a lot of inconsistently in paragraph formatting.  Sometimes they are very narrow,  sometimes too wide.  Could you please look into tools that would keep them consistent?

Vim appears to have features for automatically filling paragraphs:  https://stackoverflow.com/questions/9509321/what-is-the-equivalent-of-emacss-fill-paragraph-in-macvim

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,443 @@
...
+
+        log.debug("len(moving_window): %i", len(self._moving_window))
+        log.debug("distance: %i", self._distance)
+
+        if not len(self._moving_window) % self._distance == 0:
+            raise ValueError(
+                "The MovingWindow size is not a integer multiple of the distance."
+            )
+
+        if self._distance - period.total_seconds() / self._sampling_period != 0:
```
leandro-lucarella-frequenz: Maybe use `math.isclose()` here, or maybe better the alias for zero that we talked about it #337 ?

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,443 @@
...
+            period: The distance between two succeeding intervals.
+
+        Raises:
+            ValueError: If the MovingWindow size is not a integer multiple of the period.
+        """
+        self._moving_window = moving_window
+        self._buffer = moving_window._buffer
+
+        # internally we are working with indexes of samples
+        self._sampling_period = self._moving_window.sampling_period.total_seconds()
```
leandro-lucarella-frequenz: Any reason to use a `float` here, for what I've seen, all operations you do are also supported by `timedelta` (even modulo), so you could just as well use `timedelta` internally, right?

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,443 @@
...
+
+    Note further that when constructing a PeriodicFeatureExtractor object
+    the `MovingWindow` size has to be a integer multiple of
+    the period.
+
+    Example:
+        ```python
+            feature_extractor = PeriodicFeatureExtractor(
+            moving_window = moving_window, # 28 days moving window
+            distance=timedelta(days=7),
```
sahas-subramanian-frequenz:
``` suggestion
            moving_window=moving_window,  # 28 days moving window
            period=timedelta(days=7),

File: src/frequenz/sdk/timeseries/periodic_feature_extractor.py [hunk 21]
```diff
@@ -0,0 +1,443 @@
...
+
+        Args:
+            moving_window: The MovingWindow that is used for the average calculation.
+            period: The distance between two succeeding intervals.
+
+        Raises:
+            ValueError: If the MovingWindow size is not a integer multiple of the period.
+        """
+        self._moving_window = moving_window
+        self._buffer = moving_window._buffer
```
leandro-lucarella-frequenz: I would probably also add a public `property` to access the underlying buffer of a `MovingWindow` if there is a case for it.

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,443 @@
...
+            distance=timedelta(days=7),
+        )
+
+        now = datetime.now(tzinfo=timezone.utc)
+
+        # create a daily weighted average for the next 24h
+        avg_24h = feature_extractor.avg(
+            now, now + timedelta(hours=24),
+            weights=[0.1, 0.2, 0.3, 0.4]
+        )
```
sahas-subramanian-frequenz: just some black-like formatting for the examples
``` suggestion
        avg_24h = feature_extractor.avg(
            now, 
            now+timedelta(hours=24),
            weights=[0.1, 0.2, 0.3, 0.4],
        )

File: src/frequenz/sdk/timeseries/periodic_feature_extractor.py [hunk 23]
```diff
@@ -0,0 +1,443 @@
...
+            ValueError: If the MovingWindow size is not a integer multiple of the period.
+        """
+        self._moving_window = moving_window
+        self._buffer = moving_window._buffer
+
+        # internally we are working with indexes of samples
+        self._sampling_period = self._moving_window.sampling_period.total_seconds()
+
+        # distance between two succeeding intervals in samples
+        self._distance = int(period.total_seconds() / self._sampling_period)
```
sahas-subramanian-frequenz: It is nice to document variables using docstrings, because then lsp can display it when we lookup/highlight instances of the variable:

``` suggestion
        self._distance = int(period.total_seconds() / self._sampling_period)
        """Distance between two succeeding intervals in samples."""

File: src/frequenz/sdk/timeseries/periodic_feature_extractor.py [hunk 24]
```diff
@@ -0,0 +1,443 @@
...
+    ) -> NDArray[np.float_]:
+        """
+        Plain python version of the average calculator.
+
+        This method avoids copying in any case but is 15 to 600
+        slower then the numpy version.
+
+        This method is only used in benchmarks and unit tests.
+
+        ONLY USE THIS IF YOU NEED TO STRICTLY AVOID COPIES!
```
leandro-lucarella-frequenz: You can use special blocks to highlight this, for example:

```suggestion
        Danger:
            This method avoids copying in any case but is 15 to 600
            slower then the numpy version.

            This method is only used in benchmarks and unit tests.

            **ONLY USE THIS IF YOU NEED TO STRICTLY AVOID COPIES!**
```

(you can use `Warning:` too)

Here is the output:

image

image

The only downside is right now this is rendered collapsed, not expanded, but there is an [issue](https://github.com/mkdocstrings/mkdocstrings/issues/468) in `mkdocstrings` to fix this (and I think it should be a pretty easy so I guess it should come eventually).

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,443 @@
...
+            window_array = np.concatenate((window_start, window_end))
+        else:
+            # ignoring the type check because we know that the returned type is an ndarray
+            window_array = self._moving_window[start_pos:end_pos]  # type: ignore[assignment]
+
+        return (self._reshape_np_array(window_array, window_size), window_size)
+
+    def avg(
+        self, start: datetime, end: datetime, weights: List[float] | None = None
+    ) -> NDArray[Any]:
```
leandro-lucarella-frequenz: Shouldn't this return `NDArray[np.float_]`? I think it's not that terrible to do some type ignore in the internal methods but I would try to keep the public interface as strictly typed as possible.

benchmarks/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,213 @@
...
+    Args:
+        feature_extractor: The instance of the PeriodicFeatureExtractor to use.
+        window: The window to calculate the average over.
+        window_size: The size of the window to calculate the average over.
+        weights: The weights to use for the average calculation.
+
+    Returns:
+        The averaged window.
+    """
+    reshaped = feature_extractor._reshape_np_array(  # pylint: disable=W0212
```
leandro-lucarella-frequenz: Can you maybe replace the `pylint` numeric codes with the text representation (in this case `W0212` -> `protected-access`)? It is much harder to know what's being ignored otherwise.

benchmarks/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,213 @@
...
+        window, window_size
+    )  # pylint: disable=W0212
```
leandro-lucarella-frequenz: Also do you really need this twice? :thinking:

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,443 @@
...
+        if not len(self._moving_window) % self._distance == 0:
+            raise ValueError(
+                "The MovingWindow size is not a integer multiple of the distance."
+            )
+
+        if self._distance - period.total_seconds() / self._sampling_period != 0:
+            log.warning(
+                "The distance is not a multiple of the sampling period. "
+                "This might result in unexpected behaviour."
+            )
```
sahas-subramanian-frequenz: why not raise here too?  are there any use cases where this needs to be supported?

RELEASE_NOTES.md
```diff
@@ -10,7 +10,18 @@
...
+
+* The `OrderedRingBuffer`s `wrap` and `normalize_timestamp` are now public.
+
+* The `PeriodicFeatureExtractor` has been added.
+
+  This is a tool to create certain profiles out of periodic reoccurring windows
+  inside a `MovingWindow`.
+
+  As an example one can create a daily profile of specific weekdays which will
+  be returned as numpy arrays.
```
leandro-lucarella-frequenz: Reformatted because of how notes are rendered in releases (GitHub keep the end of lines) and removed references to `OrderedRingBuffer` as it is not public (so users don't care about changes in it):

```suggestion
* The `MovingWindow` has new public methods that are returning the oldest and newest timestamp of all stored samples.

* The `PeriodicFeatureExtractor` has been added.

  This is a tool to create certain profiles out of periodic reoccurring windows inside a `MovingWindow`.

  As an example one can create a daily profile of specific weekdays which will be returned as numpy arrays.
```

leandro-lucarella-frequenz (commented):
OK, here is my review.  They are all small style-ish things, so LGTM in general.

I didn't get into the math to be honest because is not my strongest point, maybe someone from the @frequenz-floss/datasci-team can have a look at that. I also only looked at the benchmark and tests superficially, maybe someone else can review those in more detail, otherwise I can give it another go if you think it is important to review.

There is also a commit using an outdated class name: `Add tests for the PeriodicWindowProfiler`

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
leandro-lucarella-frequenz: I just noticed this is private, so it won't be actually rendered, but in any case I think is good to always write docs as if they were public, in case sometime we decide to make it public, as it happened in this PR :D

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,443 @@
...
+        Returns:
+            The index of the timestamp in the MovingWindow.
+        """
+        return self._buffer.datetime_to_index(timestamp)
+
+    def _timestamp_to_rel_index(self, timestamp: datetime) -> int:
+        """
+        Get the index of a timestamp relative to the oldest sample in the MovingWindow.
+
+        In other word consider an integer axis where the zero is defined
```
sahas-subramanian-frequenz:
``` suggestion
        In other words, consider an integer axis where the zero is defined

File: src/frequenz/sdk/timeseries/periodic_feature_extractor.py [hunk 31]
```diff
@@ -0,0 +1,443 @@
...
+        Args:
+            timestamp: A timestamp that we want to shift into the window.
+
+        Returns:
+            The index of the timestamp shifted into the MovingWindow.
+        """
+        # align timestamp to the sampling period
+        timestamp = self._buffer.normalize_timestamp(timestamp)
+
+        # distance between the input ts and the ts of oldest know samples (in samples)
```
sahas-subramanian-frequenz:
``` suggestion
        # distance between the input ts and the ts of oldest known samples (in samples)

File: src/frequenz/sdk/timeseries/periodic_feature_extractor.py [hunk 32]
```diff
@@ -0,0 +1,443 @@
...
+        )
+        log.debug("Distance to the oldest sample: %i", dist_to_oldest)
+
+        return dist_to_oldest
+
+    def _num_windows(
+        self, window: NDArray[np.float_] | MovingWindow, window_size: int
+    ) -> int:
+        """
+        Get the number of windows that are fully contained in the MovingWindow.
```
sahas-subramanian-frequenz: looking at the code,  it looks like this returns the number of windows that "can" be contained in the MovingWindow, no?  There are no windows when no data has been inserted, for example.

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,443 @@
...
+
+        This just calls the ring buffers datetime_to_index method.
+
+        Args:
+            timestamp: The timestamp to convert.
+
+        Returns:
+            The index of the timestamp in the MovingWindow.
+        """
+        return self._buffer.datetime_to_index(timestamp)
```
sahas-subramanian-frequenz: this function is just getting forwarded, to `_buffer.datetime_to_index`, and is just getting called once.  Maybe it can be dropped.

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,443 @@
...
+        Args:
+            window: The buffer that is used for the average calculation.
+            window_size: The size of the window in samples.
+
+        Returns:
+            The number of windows that are fully contained in the MovingWindow.
+        """
+        num_windows = math.floor(len(window) / self._distance)
+        if len(window) - num_windows * self._distance >= window_size:
+            num_windows += 1
```
sahas-subramanian-frequenz: maybe we need a comment that says `len(window)` returns the number of samples, not the size of the buffer.  Also an explanation for what this means: `len(window) - num_windows * self._distance >= window_size`

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,443 @@
...
+        return resized_array.reshape(num_windows, self._distance)
+
+    def _calculate_avg_window_py(
+        self,
+        window: NDArray[np.float_],
+        window_size: int,
+        weights: List[float] | None = None,
+    ) -> NDArray[np.float_]:
+        """
+        Plain python version of the average calculator.
```
sahas-subramanian-frequenz: to make it sound more imperative.

``` suggestion
        """
        Calculate average using a plain Python implementation.

File: src/frequenz/sdk/timeseries/periodic_feature_extractor.py [hunk 36]
```diff
@@ -0,0 +1,443 @@
...
+        if len(array) < self._distance:
+            raise ValueError(
+                f"The array (length:{len(array)}) is too small to be reshaped."
+            )
+
+        num_windows = math.floor(len(array) / self._distance)
+
+        # Case 1:
+        if len(array) - num_windows * self._distance == 0:
+            log.debug("Case 1:")
```
sahas-subramanian-frequenz: Not sure these log lines are helpful in a large app, without any additional info.

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,443 @@
...
+        self,
+        window: NDArray[np.float_],
+        window_size: int,
+        weights: List[float] | None = None,
+    ) -> NDArray[np.float_]:
+        """
+        Plain python version of the average calculator.
+
+        This method avoids copying in any case but is 15 to 600
+        slower then the numpy version.
```
sahas-subramanian-frequenz:
``` suggestion
        This method avoids copying in all cases, but is 15 to 600 times
        slower than the numpy version.

File: src/frequenz/sdk/timeseries/periodic_feature_extractor.py (hunk 24 above)
Comment (sahas-subramanian-frequenz) [reply src/frequenz/sdk/timeseries/periodic_feature_extractor.py]: The `Note` and `Note further` in the class docstring can probably use this pattern as well?

File: src/frequenz/sdk/timeseries/periodic_feature_extractor.py [hunk 38]
```diff
@@ -0,0 +1,443 @@
...
+        num_windows = self._num_windows(window, window_size)
+
+        if len(window) - num_windows * self._distance >= window_size:
+            num_windows += 1
+
+        res = np.empty(window_size)
+
+        for i in range(window_size):
+            assert num_windows * (self._distance) - len(window) <= self._distance
+            summe = 0
```
sahas-subramanian-frequenz: :de:

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,443 @@
...
+            for j in range(num_windows):
+                if weights is None:
+                    summe += window[i + (j * (self._distance))]
+                else:
+                    summe += weights[j] * window[i + (j * (self._distance))]
+
+            if weights is None:
+                res[i] = summe / num_windows
+            else:
+                res[i] = summe / np.sum(weights)
```
sahas-subramanian-frequenz: You also need to check `weights` is an empty list, before sending to `np.sum`,  you can change the check such that it would pass if weights is None, an empty list or set etc.

``` suggestion
            if not weights:
                res[i] = summe / num_windows
            else:
                res[i] = summe / np.sum(weights)

File: src/frequenz/sdk/timeseries/periodic_feature_extractor.py [hunk 40]
```diff
@@ -0,0 +1,443 @@
...
+        # that would be overwritten next.
+
+        # Move the window to its first appereance in the MovingWindow
+        # relative to the oldest sample stored in the MovingWindow.
+        # In other words the oldest stored sample is considered to have index 0.
+        # Note that the returned value is a index not a timestamp
+        rel_start_sample = self._timestamp_to_rel_index(start) % self._distance
+        rel_end_sample = self._timestamp_to_rel_index(end) % self._distance
+        # fix the position if modulo period the end_sample
+        # is smaler then the start sample position
```
sahas-subramanian-frequenz: what does "if modulo period the end_sample" mean?

``` suggestion
        # fix the position if modulo period the end_sample
        # is smaller then the start sample position

File: benchmarks/timeseries/periodic_feature_extractor.py [hunk 41]
```diff
@@ -0,0 +1,213 @@
...
+    window_size: int,
+    weights: List[float] | None = None,
+) -> NDArray[np.float_]:
+    """
+    Plain python version of the average calculator.
+
+    This method avoids copying in any case but is 15 to 600
+    slower then the numpy version.
+
+    This method is only used in these benchmarks.
```
sahas-subramanian-frequenz: There's an implementation of this method in the `PeriodicFeatureExtractor` class,  with the same code.  Can we drop this function in the benchmark file and use the one from the class?  If not,  maybe useful to document why not.

tests/timeseries/test_periodic_feature_extractor.py
```diff
@@ -0,0 +1,137 @@
...
+
+async def init_feature_extractor(
+    data: List[float], period: timedelta
+) -> PeriodicFeatureExtractor:
+    """
+    Initialize a PeriodicFeatureExtractor with a `MovingWindow` that contains the data.
+
+    Parameters:
+        data: The data that is pushed into the moving window.
+        period: The distance between two successive windows.
```
sahas-subramanian-frequenz:
``` suggestion
    Args:
        data: The data that is pushed into the moving window.
        period: The distance between two successive windows.

**Review by sahas-subramanian-frequenz — COMMENTED** [review]
Many commit messages still say "TimeseriesAvg" and "PeriodicWindowProfiler" instead of "PeriodicFeatureExtractor"

File: src/frequenz/sdk/timeseries/periodic_feature_extractor.py (hunk 24 above)
Comment (leandro-lucarella-frequenz) [reply src/frequenz/sdk/timeseries/periodic_feature_extractor.py]: I missed those, but yeah, there is a `Note:` *admonition* too. You can see the full list here: https://squidfunk.github.io/mkdocs-material/reference/admonitions/#supported-types

File: src/frequenz/sdk/timeseries/periodic_feature_extractor.py (hunk 22 above)
Comment (leandro-lucarella-frequenz) [reply src/frequenz/sdk/timeseries/periodic_feature_extractor.py]: @mathias-baumann-frequenz maybe we should also run black, isort, pylint, etc. to all code examples in the future too :scream: (in a far future, not for now for sure)

File: src/frequenz/sdk/timeseries/periodic_feature_extractor.py (hunk 28 above)
Comment (leandro-lucarella-frequenz) [reply src/frequenz/sdk/timeseries/periodic_feature_extractor.py]: +1, I actually thought the same when reading this but assumed there was. And if there is, it would be good to state it explicitly in a comment, so people looking at the code in the future don't ask themselves the same question.

File: tests/timeseries/test_periodic_feature_extractor.py [hunk 43]
```diff
@@ -0,0 +1,137 @@
...
+    feature_extractor = await init_feature_extractor(data, timedelta(seconds=3))
+    assert np.allclose(feature_extractor.avg(start, end), [5 / 3, 4 / 3])
+
+    feature_extractor = await init_feature_extractor(data, timedelta(seconds=4))
+    assert np.allclose(feature_extractor.avg(start, end), [1, 2])
+
+    data: List[float] = [1, 2, 2.5, 1, 1, 1, 2, 2, 1, 1, 1, 1, 1, 1, 1]  # type: ignore[no-redef]
+
+    feature_extractor = await init_feature_extractor(data, timedelta(seconds=5))
+    assert np.allclose(feature_extractor.avg(start, end), [1.5, 1.5])
```
cwasicki, line 121: Added more tests in #375

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,443 @@
...
+        Create the average window out of the window defined by `start` and `end`.
+
+        This method calculates the average of a window by averaging over
+        all windows fully inside the MovingWindow having the distance
+        `self.distance`.
+
+        Args:
+            start: The start of the window to average over.
+            end: The end of the window to average over.
+            weights: The weights to use for the average calculation.
```
cwasicki: From this it's not clear in which order the weights are applied (i.e. oldest first)

benchmarks/timeseries/periodic_feature_extractor.py
matthias-wende-frequenz: Ups. That should not be in the `PeriodicFeatureExtractor` any longer. I'll will remove from there.

benchmarks/timeseries/periodic_feature_extractor.py
matthias-wende-frequenz: Just making sure pylint really listens! :D

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
leandro-lucarella-frequenz: And now admonitions are expanded by default! :tada:

https://github.com/mkdocstrings/mkdocstrings/issues/468#issuecomment-1537489843

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: Unfortunately that works only for integers, but at least we can reduce the code by two lines.

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
matthias-wende-frequenz: Visual Basic style!

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
matthias-wende-frequenz: it means the `rel_next_position % self.period`

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
matthias-wende-frequenz: yes underlaying buffer is the moving window. this class can (in theory) work with a numpy array too.

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
matthias-wende-frequenz: yes it's oldest first as dictated by numpy :). Will add a comment

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
matthias-wende-frequenz: ~All the internal operations are based on indexes and not on times any longer.~
You are right. I'll change it :)

src/frequenz/sdk/timeseries/_moving_window.py
leandro-lucarella-frequenz: Amazing! :stuck_out_tongue:

src/conftest.py
```diff
@@ -122,6 +122,8 @@ def evaluate(self, example: Example) -> None | str:
         import_header.append(
             path_to_import_statement(Path(os.path.relpath(example.path)))
         )
+        # Add microgrid as default import
+        import_header.append("from frequenz.sdk import microgrid")
```
leandro-lucarella-frequenz: I'm not sure if it is a good idea to add this, is just one extra line for examples and it makes examples more complete.

leandro-lucarella-frequenz (commented):
Just one more small comment, I started to review again but then noticed a new review wasn't requested yet so not sure if you are still working on the PR... Please request a review again when it is ready for a new round.

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,392 @@
...
+        Raises:
+            ValueError: If the MovingWindow size is not a integer multiple of the period.
+        """
+        self._moving_window = moving_window
+        self._buffer = moving_window._buffer
+
+        self._sampling_period = self._moving_window.sampling_period.total_seconds()
+        """The sampling_period as integer to use it for indexing of samples."""
+
+        self._period = int(period.total_seconds() / self._sampling_period)
```
cwasicki: Does that work correctly if period is not a multiple of sampling period?

cwasicki: oh, I see you check this below. For unexpected behavior I would raise an exception or do you see valid use-cases for this?

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,392 @@
...
+        period between two window neighbors.
+
+        Args:
+            window: The buffer that is used for the average calculation.
+            window_size: The size of the window in samples.
+
+        Returns:
+            The number of windows that are fully contained in the MovingWindow.
+        """
+        num_windows = math.floor(len(window) / self._period)
```
cwasicki: Could be `//` instead of floor (given it's positive)

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,392 @@
...
+        if len(window) - num_windows * self._period >= window_size:
```
cwasicki: When can this happen?

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,392 @@
...
+    ) -> NDArray[np.float_]:
+        """
+        Reshape a numpy array to a 2D array where each row represents a window.
+
+        There are three cases to consider
+
+        1. the array size is a multiple of window_size + period,
+           i.e. num_windows is integer and we can simply reshape
+        2. the array size is NOT a multiple of window_size + period and
+           it has length n * (window_size + period) + m, where m < window_size
```
cwasicki: `window_size+period` -> `period`?

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,392 @@
...
+        3. the array size is NOT a multiple of window_size + period and
+           it has length n * (window_size + period) + m, where m >= window_size
```
cwasicki: Can this happen?

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,392 @@
...
+
+        Returns:
+            The index of the timestamp shifted into the MovingWindow.
+        """
+        # align timestamp to the sampling period
+        timestamp = self._buffer.normalize_timestamp(timestamp)
+
+        # distance between the input ts and the ts of oldest know samples (in samples)
+        dist_to_oldest = int(
+            abs((timestamp - self._moving_window.time_bound_oldest).total_seconds())
```
matthias-wende-frequenz: Using abs is a bug. Using modulo arithmetic alone is enough and thus it's fine to return a negative value here.

src/frequenz/sdk/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,437 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""
+A class that extracts features out of historical data.
+
+It creats a profile from periodically occurring windows in a buffer of
```
shsms:
```suggestion
It creates a profile from periodically occurring windows in a buffer of

File: src/frequenz/sdk/timeseries/periodic_feature_extractor.py (hunk 48 above)
Comment (matthias-wende-frequenz) [reply src/frequenz/sdk/timeseries/periodic_feature_extractor.py]: This can't happen any longer with the current restrictions. This function was actually meant to work in a more generic situation. I'll comment that code out and leave it was a comment for future reference.

Comment (matthias-wende-frequenz) [reply src/frequenz/sdk/timeseries/periodic_feature_extractor.py]: I moved that function to the benchmarks as at hasn't been used in the code at all.

File: src/frequenz/sdk/timeseries/periodic_feature_extractor.py (hunk 50 above)
Comment (matthias-wende-frequenz) [reply src/frequenz/sdk/timeseries/periodic_feature_extractor.py]: only case 1 can happen in the current implementation. But afair all three cases are covered in the benchmarks or the tests. I'll leave it for future work, when we remove the condition and add a comment.

File: src/frequenz/sdk/timeseries/periodic_feature_extractor.py (hunk 46 above)
Comment (matthias-wende-frequenz) [reply src/frequenz/sdk/timeseries/periodic_feature_extractor.py]: I didn't want to add unnecessary restrictions and I haven't thought it fully through. I suggest we address this by adding tests in a follow up PR.

**Review by shsms — DISMISSED** [review]
🎉

File: src/frequenz/sdk/timeseries/_periodic_feature_extractor.py [hunk 53]
```diff
@@ -0,0 +1,414 @@
...
+
+import logging
+from dataclasses import dataclass
+from datetime import datetime, timedelta
+from typing import List, Tuple
+
+import numpy as np
+from numpy.typing import NDArray
+
+from frequenz.sdk._internal._math import is_close_to_zero
```
leandro-lucarella-frequenz:
```suggestion
from .._internal._math import is_close_to_zero
```

RELEASE_NOTES.md
```diff
@@ -28,7 +28,15 @@ This release drops support for Python versions older than 3.11.
 
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+* The `MovingWindow` as well as the OrderedRingBuffer are having new public methods that are returning the oldest and newest timestamp of all stored samples.
+
+* The `OrderedRingBuffer`s `wrap` and `normalize_timestamp` are now public.
```
leandro-lucarella-frequenz: This is not being exposed, so technically it shouldn't be in the release notes, as users are not supposed to use it directly.

src/frequenz/sdk/timeseries/_periodic_feature_extractor.py
```diff
@@ -0,0 +1,414 @@
...
+        if not len(self._moving_window) % self._period == 0:
+            raise ValueError(
+                "The MovingWindow size is not a integer multiple of the period."
+            )
+
+        if not is_close_to_zero(self._period - period / self._sampling_period):
+            _logger.warning(
+                "The period is not a multiple of the sampling period. "
+                "This might result in unexpected behaviour."
+            )
```
leandro-lucarella-frequenz, line 148: There was a pending question about if we want to allow this at all or not (i.e. raising an exception or leaving it only as a warning).

tests/timeseries/test_moving_window.py
```diff
@@ -133,3 +143,6 @@ async def test_resampling_window(fake_time: time_machine.Coordinates) -> None:
     assert len(window) == window_size / output_sampling
     for value in window:  # type: ignore
         assert 4.9 < value < 5.1
+
+
+asyncio.run(test_access_window_by_int_slice())
```
leandro-lucarella-frequenz: Why this? `pytest` should properly run `async` tests in a async context automatically.

```suggestion
```

tests/timeseries/test_moving_window.py
```diff
@@ -14,6 +14,7 @@
 from frequenz.channels import Broadcast, Sender
 
 from frequenz.sdk.timeseries import Sample
+from frequenz.sdk.timeseries._base_types import UNIX_EPOCH
 from frequenz.sdk.timeseries._moving_window import MovingWindow
 from frequenz.sdk.timeseries._resampling import ResamplerConfig
```
leandro-lucarella-frequenz: Test should try to use public imports when possible. I think all symbols in here are available simply in `frequenz.timeseries`.
```suggestion
from frequenz.sdk.timeseries import Sample, UNIX_EPOCH, MovingWindow, ResamplerConfig
```

src/frequenz/sdk/timeseries/_periodic_feature_extractor.py
```diff
@@ -0,0 +1,414 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""
+A class that extracts features out of historical data.
```
leandro-lucarella-frequenz: I would remove the "a class that", maybe something like:

```suggestion
An extractor to get features out of historical data.
```

Also, please bear in mind that this module is **private**, so all this documentation won't be shown in the rendered documentation in the website. I'd suggest moving all the important documentation to the class instead.

src/frequenz/sdk/timeseries/_periodic_feature_extractor.py
```diff
@@ -0,0 +1,414 @@
...
+from ._moving_window import MovingWindow
+from ._ringbuffer import OrderedRingBuffer
+
+_logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class RelativePositions:
+    """
+    A dataclass that holds the relative positions of a window in a buffer.
```
leandro-lucarella-frequenz: Same here, maybe:

```suggestion
    The relative positions of a window in a buffer.
```

BTW, positions of what? By reading the class doc for me it is not clear what this is actually for.

src/frequenz/sdk/timeseries/_periodic_feature_extractor.py
```diff
@@ -0,0 +1,414 @@
...
+
+    The positions are relative to the oldest sample in the buffer.
+    """
+
+    start: int
+    end: int
+    next: int
```
leandro-lucarella-frequenz: Ideally these should be documented too.

```suggestion
    start: int
    """Whatever."""

    end: int
    """Whatever."""

    next: int
    """Whatever."""
```

src/frequenz/sdk/timeseries/_periodic_feature_extractor.py
```diff
@@ -0,0 +1,414 @@
...
+
+        if start_pos >= end_pos:
+            window_start = self._buffer[start_pos : len(self._moving_window)]
+            window_end = self._buffer[0:end_pos]
+            # make the linter happy
+            assert isinstance(window_start, np.ndarray)
+            assert isinstance(window_end, np.ndarray)
+            window_array = np.concatenate((window_start, window_end))
+        else:
+            # ignoring the type check because we know that the returned type is an ndarray
```
leandro-lucarella-frequenz: Is this a left-over of a previous version? There is not type ignore here...

src/frequenz/sdk/timeseries/_periodic_feature_extractor.py
```diff
@@ -0,0 +1,414 @@
...
+        Args:
+            start: The start of the window to average over.
+            end: The end of the window to average over.
+            weights: The weights to use for the average calculation (oldest first).
+
+        Returns:
+            The averaged timeseries window.
+        """
+        (reshaped, window_size) = self._get_reshaped_np_array(start, end)
+        return np.average(  # type: ignore[no-any-return]
```
leandro-lucarella-frequenz, line 417: Maybe it was for this line?

src/frequenz/sdk/timeseries/_periodic_feature_extractor.py
```diff
@@ -0,0 +1,414 @@
...
+
+        if not is_close_to_zero(self._period - period / self._sampling_period):
+            _logger.warning(
+                "The period is not a multiple of the sampling period. "
+                "This might result in unexpected behaviour."
+            )
+
+    @property
+    def _buffer(self) -> OrderedRingBuffer[NDArray[np.float64]]:
+        return self._moving_window._buffer  # pylint: disable=protected-access
```
leandro-lucarella-frequenz, line 152: Any particular reason why not just make the `MovingWindow` buffer a public property too?

tests/timeseries/test_periodic_feature_extractor.py
leandro-lucarella-frequenz, line 1: Not for this PR, but as a general thing. I think we should start mocking more in **unit tests**.  These tests are not testing the feature extractor in isolation, they are also testing the moving window, and the moving window is also testing the ring buffer and on top of that, it is using a channel, and senders and receivers, so there is quite a bit going on here.

This tend to make tests much slower and sometimes more complicated, or duplicate a lot of testing from the used classes.  Also, it happened to me that working on some component suddenly test for a completely different component start to fail, which is confusing and distracting.

I think having **integration tests** is also important, but I feel we are mixing them too much, so in the future I'd try to make a more clear boundary between unit tests and integration tests, specially as we build higher and higher abstractions the problem intensifies if we keep testing the whole stack when testing the higher level abstractions in unit tests.

tests/timeseries/test_periodic_feature_extractor.py
```diff
@@ -0,0 +1,140 @@
...
+
+    Args:
+        data: The data that is pushed into the moving window.
+        period: The distance between two successive windows.
+
+    Returns:
+        PeriodicFeatureExtractor
+    """
+    window, sender = init_moving_window(timedelta(seconds=len(data)))
+    await push_lm_data(sender, data)
```
leandro-lucarella-frequenz: Just for the records, I had to go look into the moving window test to see what "lm" stands for, it would be more clear to call it `push_logical_meter_data()` :)

tests/timeseries/test_periodic_feature_extractor.py
```diff
@@ -0,0 +1,140 @@
...
+"""Tests for the timeseries averager."""
+
+from datetime import datetime, timedelta, timezone
+from typing import List
+
+import numpy as np
+from frequenz.channels import Broadcast
+
+from frequenz.sdk.timeseries import MovingWindow, PeriodicFeatureExtractor, Sample
+from frequenz.sdk.timeseries._base_types import UNIX_EPOCH
```
leandro-lucarella-frequenz:
```suggestion
from frequenz.sdk.timeseries import MovingWindow, PeriodicFeatureExtractor, Sample, UNIX_EPOCH
```

tests/timeseries/test_periodic_feature_extractor.py
```diff
@@ -0,0 +1,140 @@
...
+        period: The distance between two successive windows.
+
+    Returns:
+        PeriodicFeatureExtractor
+    """
+    # We only need the moving window to initialize the PeriodicFeatureExtractor class.
+    lm_chan = Broadcast[Sample]("lm_net_power")
+    # monkey patch the MovingWindow class to have a fixed size such that
+    # it can be initialized with a distance bigger then the window size.
+    MovingWindow.__len__ = lambda self: period  # type: ignore[method-assign]
```
leandro-lucarella-frequenz: You should use `with mock.patch(...)` or the `mocker` *fixture* so the `MovingWindow.__len__` is restored after the test, otherwise you can run to really weird shit.

tests/timeseries/test_periodic_feature_extractor.py
```diff
@@ -0,0 +1,140 @@
...
+
+async def test_profiler_calculate_np() -> None:
+    """
+    Test the calculation of the average using a numpy array and compare it
+    against the pure python method with the same functionality.
+    """
+    data = np.array([2, 2.5, 1, 1, 1, 2])
+    feature_extractor = await init_feature_extractor_no_data(4)
+    window_size = 2
+    reshaped = feature_extractor._reshape_np_array(  # pylint: disable=W0212
```
leandro-lucarella-frequenz: Please use the symbolic `pylint` check names (`protected-access` in this case)

tests/timeseries/test_periodic_feature_extractor.py
```diff
@@ -0,0 +1,140 @@
...
+        data, window_size
+    )  # pylint: disable=W0212
```
leandro-lucarella-frequenz: Also do they really need to be specified twice?

leandro-lucarella-frequenz (commented):
Quite a few comments but most of them super minor. The only 2 comments that are important to me are:

1. Patching of `MovingWindow.__len__` (and not restoring it)
2. The `asyncio.run()` in the test

And there is also the unit tests doing to little mocking, which is quite important IMHO, but again, I don't expect that to be changed in this PR as it is a much bigger topic and we shouldn't block this PR until we address it.

RELEASE_NOTES.md
matthias-wende-frequenz: True

src/frequenz/sdk/timeseries/_periodic_feature_extractor.py
matthias-wende-frequenz: To be honest I haven't fully thought these cases through but I think it'll work. Since I didn't want to add any unnecessary restrictions I left the warning and decided to add follow up unit tests to get some more insights.

tests/timeseries/test_moving_window.py
matthias-wende-frequenz: removed. good catch!

tests/timeseries/test_periodic_feature_extractor.py
matthias-wende-frequenz: I thought I changed all these comments.

tests/timeseries/test_periodic_feature_extractor.py
matthias-wende-frequenz: no they don't. Fixed!

src/frequenz/sdk/timeseries/_periodic_feature_extractor.py
cwasicki: I think if we don't know how it behaves we should raise, we can still change it when we have more insights.

matthias-wende-frequenz: Fair enough! Will change to raise :).

src/frequenz/sdk/timeseries/_periodic_feature_extractor.py
matthias-wende-frequenz: I've reworked the documentation by adding more context.

src/frequenz/sdk/timeseries/_periodic_feature_extractor.py
matthias-wende-frequenz: Done but I changed Whatever to something meaningful :p.

src/frequenz/sdk/timeseries/_periodic_feature_extractor.py
matthias-wende-frequenz: probably

src/frequenz/sdk/timeseries/_periodic_feature_extractor.py
matthias-wende-frequenz: probably not :D

src/frequenz/sdk/timeseries/_periodic_feature_extractor.py
matthias-wende-frequenz: You mean in the `MovingWindow`? I'd rather prefer to keep it hidden for users.

tests/timeseries/test_periodic_feature_extractor.py
matthias-wende-frequenz: Actually in that case the patching is not needed any longer, since the semantics of period have changed.

tests/timeseries/test_periodic_feature_extractor.py
matthias-wende-frequenz: Agreed!

src/frequenz/sdk/timeseries/_periodic_feature_extractor.py
```diff
@@ -0,0 +1,419 @@
...
+    position in the buffer where it occurrs first and represented by indexes
+    relative to the oldest sample in the buffer.
+    """
+
+    start: int
+    """The relative start position of the window."""
+    end: int
+    """The relative end position of the window."""
+    next: int
+    """The relative position of the next incoming sample sample."""
```
leandro-lucarella-frequenz:
```suggestion
    """The relative position of the next incoming sample."""
```

benchmarks/timeseries/periodic_feature_extractor.py
```diff
@@ -0,0 +1,235 @@
...
+from frequenz.sdk.timeseries import MovingWindow, PeriodicFeatureExtractor, Sample
+
+
+async def init_feature_extractor(period: int) -> PeriodicFeatureExtractor:
+    """Initialize the PeriodicFeatureExtractor class."""
+    # We only need the moving window to initialize the PeriodicFeatureExtractor class.
+    lm_chan = Broadcast[Sample]("lm_net_power")
+    # monkey patch the MovingWindow class to have a fixed size such that
+    # it can be initialized with a period bigger then the window size.
+    MovingWindow.__len__ = lambda self: period  # type: ignore[method-assign]
```
leandro-lucarella-frequenz: The patching is still there!

leandro-lucarella-frequenz (commented):
Maybe you forgot to push, but the patching of the `MovingWindow.__len__` is still there. And a new typo. Other than that, LGTM!

src/frequenz/sdk/timeseries/_periodic_feature_extractor.py
matthias-wende-frequenz: That's intended! Having it twice makes a bigger impact :D.

benchmarks/timeseries/periodic_feature_extractor.py
matthias-wende-frequenz: weird I can't find it

matthias-wende-frequenz: ah damn found it. That was done in 2 places.

shsms (approved):

shsms (approved):

PR #305, opened 2023-04-04
Refactor power distributor's `Result`s

### LINKED ISSUE #275: Add a more descriptive `__str__()` method to the power distributor's `Result`
### What's needed?

When there is a failure to send a command to the power distributor it is common to log the result if it fails. With the current `Result.__str__()` implementation, that's not very descriptive though.

Here is an example:

| ERROR:Could not charge/discharge batteries. Result: <frequenz.sdk.actor.power_distributing.result.PartialFailure object at 0x7fe2b859b3a0>

### Proposed solution

We should implement a more descriptive `__str__()` method for `Result` and its sub-classes, showing also the most important attributes of that particular class (if not all).

Maybe even convert it to a `dataclass`. `Result` is a `ABC` but it doesn't have any `abstractmethod`, so there is not much point in being a `ABC`. If we make it a `dataclass` we'll get more descriptive `str` without any effort.

### Use cases

* Better logging.

_No response_

### Additional context

We started monitoring error logs and the example above popped out.

leandro-lucarella-frequenz:
We make `Result` and its subclasses dataclasses so they have a proper `__str__()` method that shows the details.

We also make some property renaming to make it more consistent and improve documentation.

Fixes #275.

### DIFF
--- a/src/frequenz/sdk/actor/power_distributing/request.py
+++ b/src/frequenz/sdk/actor/power_distributing/request.py
@@ -12,10 +12,10 @@ class Request:
     """Request to set power to the `PowerDistributingActor`."""
 
     power: int
-    """The amount of power to be set."""
+    """The requested power in watts."""
 
     batteries: set[int]
-    """The set of batteries to use when requesting the power to be set."""
+    """The component ids of the batteries to be used for this request."""
 
     request_timeout_sec: float = 5.0
     """The maximum amount of time to wait for the request to be fulfilled."""
@@ -24,8 +24,8 @@ class Request:
     """Whether to adjust the power to match the bounds.
 
     If `True`, the power will be adjusted (lowered) to match the bounds, so
-    only the decreased power will be set.
+    only the reduced power will be set.
 
     If `False` and the power is outside the batteries' bounds, the request will
-    fail and replied to with an `OutOfBound` result.
+    fail and be replied to with an `OutOfBound` result.
     """
--- a/src/frequenz/sdk/actor/power_distributing/request.py
+++ b/src/frequenz/sdk/actor/power_distributing/request.py
@@ -2,22 +2,30 @@
 # Copyright © 2022 Frequenz Energy-as-a-Service GmbH
 """Definition of the user request."""
 
-from dataclasses import dataclass
-from typing import Set
+from __future__ import annotations
+
+import dataclasses
 
 
-@dataclass
+@dataclasses.dataclass
 class Request:
-    """Request from the user."""
+    """Request to set power to the `PowerDistributingActor`."""
 
-    # How much power to set
     power: int
-    # In which batteries the power should be set
-    batteries: Set[int]
-    # Timeout for the server to respond on the requests.
+    """The amount of power to be set."""
+
+    batteries: set[int]
+    """The set of batteries to use when requesting the power to be set."""
+
     request_timeout_sec: float = 5.0
-    # If True and requested power value is above upper bound, then the power will be
-    # decreased to match the bounds. Only the decreased power will be set.
-    # If False and the requested power is above upper bound, then request won't
-    # be processed. result.OutOfBound message will be send back to the user.
+    """The maximum amount of time to wait for the request to be fulfilled."""
+
     adjust_power: bool = True
+    """Whether to adjust the power to match the bounds.
+
+    If `True`, the power will be adjusted (lowered) to match the bounds, so
+    only the decreased power will be set.
+
+    If `False` and the power is outside the batteries' bounds, the request will
+    fail and replied to with an `OutOfBound` result.
+    """
--- a/src/frequenz/sdk/actor/power_distributing/result.py
+++ b/src/frequenz/sdk/actor/power_distributing/result.py
@@ -27,8 +27,8 @@ class Success(Result):
     def __init__(
         self,
         request: Request,
-        succeed_power: int,
-        used_batteries: Set[int],
+        succeeded_power: int,
+        succeeded_batteries: Set[int],
         excess_power: int,
     ) -> None:
         """Create class instance.
@@ -42,8 +42,8 @@ class Success(Result):
                 because it was outside available power bounds.
         """
         super().__init__(request)
-        self.succeed_power: int = succeed_power
-        self.used_batteries: Set[int] = used_batteries
+        self.succeed_power: int = succeeded_power
+        self.used_batteries: Set[int] = succeeded_batteries
         self.excess_power: int = excess_power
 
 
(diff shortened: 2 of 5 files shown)


leandro-lucarella-frequenz:
Please have a look at the commit messages for more info.

src/frequenz/sdk/actor/power_distributing/request.py
```diff
@@ -2,22 +2,30 @@
...
 class Request:
-    """Request from the user."""
+    """Request to set power to the `PowerDistributingActor`."""
 
-    # How much power to set
     power: int
-    # In which batteries the power should be set
-    batteries: Set[int]
-    # Timeout for the server to respond on the requests.
+    """The amount of power to be set."""
```
shsms: What do you think about:
``` suggestion
    """The requested power in watts."""

File: src/frequenz/sdk/actor/power_distributing/request.py [hunk 2]
```diff
@@ -2,22 +2,30 @@
...
+
+    batteries: set[int]
+    """The set of batteries to use when requesting the power to be set."""
```
shsms:
``` suggestion
    """The component ids of the batteries to be used for this request."""

File: src/frequenz/sdk/actor/power_distributing/request.py [hunk 3]
```diff
@@ -2,22 +2,30 @@
...
-    # decreased to match the bounds. Only the decreased power will be set.
-    # If False and the requested power is above upper bound, then request won't
-    # be processed. result.OutOfBound message will be send back to the user.
+    """The maximum amount of time to wait for the request to be fulfilled."""
+
     adjust_power: bool = True
+    """Whether to adjust the power to match the bounds.
+
+    If `True`, the power will be adjusted (lowered) to match the bounds, so
+    only the decreased power will be set.
```
shsms: Maybe reduced works better?

``` suggestion
    If `True`, the power will be adjusted (lowered) to match the bounds, so
    only the reduced power will be set.

File: src/frequenz/sdk/actor/power_distributing/result.py [hunk 4]
```diff
@@ -3,122 +3,92 @@
...
+
+    failed_power: int
+    """The part of the requested power that failed to be set."""
+
+    failed_batteries: set[int]
+    """The subset of batteries for which the request failed."""
+
+
+@dataclasses.dataclass
+class Error(Result, _BaseResultMixin):
```
shsms: `Result` already derives from `_BaseResultMixin`,  do we need to derive from it separately also?

shsms:
Btw,  weren't you also going to add a `__str__` or `__repr__` method to these types, so we get better logs?

src/frequenz/sdk/actor/power_distributing/request.py
leandro-lucarella-frequenz: Thanks, I didn't know the unit and thought of finding out later and then forgot!

src/frequenz/sdk/actor/power_distributing/result.py
leandro-lucarella-frequenz: I guess not, it was probably an artifact from my attempt to use `Result = Success | ...`. I can remove it for now.

leandro-lucarella-frequenz: Done and updated the PR.

leandro-lucarella-frequenz:
> Btw, weren't you also going to add a __str__ or __repr__ method to these types, so we get better logs?

dataclass takes care of it :sunglasses: 

```pycon
>>> from frequenz.sdk.actor.power_distributing.result import OutOfBound, Success
>>> oob = OutOfBound('test', 'test') # We don't need to care about types here :D
>>> str(oob)
"OutOfBound(request='test', bound='test')"
>>> repr(oob)
"OutOfBound(request='test', bound='test')"
>>> suc = Success(1, 2, 3, 4)
>>> str(suc)
'Success(succeeded_power=1, succeeded_batteries=2, excess_power=3, request=4)'
>>> repr(suc)
'Success(succeeded_power=1, succeeded_batteries=2, excess_power=3, request=4)'
```

leandro-lucarella-frequenz:
I just realized that this is wrong:

```pycon
>>> str(suc)
'Success(succeeded_power=1, succeeded_batteries=2, excess_power=3, request=4)'
```

`request` should be the first argument, not the fourth. Test passed anyway because the only 2 wrong were `Success` and `PartialFailure` but both are constructed using keyword arguments anyway.

I actually wanted to make the result dataclasses `kw_only` but sadly it is not supported in Python 3.8, so we'll have to wait for that.

Anyway, I updated the PR to always use keyword arguments for all classes, just in case, and to fix the attribute order.

leandro-lucarella-frequenz:
Also as an extra clarification, this problem would have been caught by the type system if `Success` and `PartialFailure` were not constructed using keyword arguments.

src/frequenz/sdk/actor/power_distributing/request.py
```diff
@@ -2,22 +2,30 @@
...
+    """The maximum amount of time to wait for the request to be fulfilled."""
+
     adjust_power: bool = True
+    """Whether to adjust the power to match the bounds.
+
+    If `True`, the power will be adjusted (lowered) to match the bounds, so
+    only the reduced power will be set.
+
+    If `False` and the power is outside the batteries' bounds, the request will
+    fail and replied to with an `OutOfBound` result.
```
daniel-zullo-frequenz: maybe `reply` instead of replied

src/frequenz/sdk/actor/power_distributing/result.py
```diff
@@ -3,122 +3,100 @@
...
-        Args:
-            request: The user's request to which this message responds.
-            msg: Error message explaining why error happened.
-        """
-        super().__init__(request)
-        self.msg: str = msg
 
+@dataclasses.dataclass
+class Success(_BaseSuccessMixin, Result):  # Order matters here. See above.
+    """Result returned when setting the power succeed for all batteries."""
```
daniel-zullo-frequenz:
```suggestion
    """Result returned when setting the power has succeeded for all batteries."""
```

daniel-zullo-frequenz (dismissed):
I only have a couple of optional cosmetic comments

src/frequenz/sdk/actor/power_distributing/request.py
leandro-lucarella-frequenz: I don't think using `reply` is correct here. I think I'm missing a "be": "the request will fail and be replied to with an `OutOfBound` result."

But maybe a native English speaker can solve this mistery? :D

src/frequenz/sdk/actor/power_distributing/result.py
leandro-lucarella-frequenz: Here I wanted to put "setting the power succeeded for all batteries."

leandro-lucarella-frequenz:
Updated taking into account @daniel-zullo-frequenz comments.

daniel-zullo-frequenz (approved):

src/frequenz/sdk/actor/power_distributing/request.py
daniel-zullo-frequenz: yeah `be replied` sounds right to me

PR #326, opened 2023-04-12
Make interface for serializing a ringbuffer more flexible.

mathias-baumann-frequenz:


src/frequenz/sdk/timeseries/_ringbuffer_serialization.py
leandro-lucarella-frequenz, line 1: A few comments about code structure. If you need several files to implement the ring buffer, then I would go for a package for it:

```
.../timeseries/_ringbuffer/__init__.py       <-- import the "public interface"
                           buffer.py         <-- the implementation
                           serialization.py  <-- the serialization code
```

I would probably also include the `load()` and `dump()` functions as "public interface" (I use quotes because it is still an internal module, but it would be the internal public interface), so you can do:

```py
from frequenz.sdk.timeseries import _ringbuffer

b = _ringbuffer.load(...)
_ringbuffer.dump(b)
```

leandro-lucarella-frequenz (approved):
Beside the small comment about structure, it does make a lot of sense not having a sub-class just to implement load/dump and just be able to serialize any ring buffer.

The only thing that looks weird to me is that now the version number and the class are dissociated (or not so easily linked). Each time the class changes its internal representation, the version number should be changed too, right?

I wonder if there is some way to tied both of them to avoid issues in the future where we change the internal representation of the ring buffer and forget to change the version number.

On a second thought, I guess this problem existed with the previous approach too.

PR #327, opened 2023-04-13
Make FormulaEngine instances composable, instead of FormulaReceivers

### LINKED ISSUE #226: Fix type checking in the `timeseries._formula_engine` package
This is the package (several modules are affected) with the most type checking errors and which look the harder to fix, because there is a lot of generic code with complicated types.

_Originally posted by @leandro-lucarella-frequenz in https://github.com/frequenz-floss/frequenz-sdk-python/issues/219#issuecomment-1444080540_

### LINKED ISSUE #314: Use a better name for methods creating a new receiver
### What's needed?

Right now the `LogicalMeter`, `BatteryPool`, and `EVChargerPool` name functions returning a receiver after the data that comes from the receiver.

Examples:

```py
grid_power_receiver = logical_meter.grid_power()
power_bounds_receiver = battery_pool.power_bounds()
```

This can be a bit obscure for people that are new to programming with channels, senders and receivers in general and specifically new to the SDK.

People might think that the method (which sounds like a regular getter) might be idempotent so calling:

```py
grid_power = await logical_meter.grid_power().receive()
# do something
grid_power = await logical_meter.grid_power().receive()
```

is OK, while in reality is creating multiple receivers, which adds overhead to the channel processing. Also if the channel has the option to always return the last value, it will receive a repeated value if there was nothing new between the first `receive()` and the second.

### Proposed solution

Make the metric attributes an object that can be asked for a new receiver instead.

In the example above, this would be:

```py
grid_power_receiver = logical_meter.grid_power.new_receiver()
power_bounds_receiver = battery_pool.power_bounds.new_receiver()
```

This makes it much more obvious that a new receiver is being created, so this code will look wrong more naturally:

```py
grid_power = await logical_meter.grid_power.new_receiver().receive()
# do something
grid_power = await logical_meter.grid_power.new_receiver().receive()
```

### Use cases

_No response_

Some alternatives were discussed in:

* https://github.com/frequenz-floss/frequenz-sdk-python/discussions/215

### Extensions

Using this scheme also enables the possibility to add other ways to query a metric in the future. For example there could be a way to query the current or last value:

```py
current = await p.soc_capacity.current_value()  # Queries the microgrid API (if it's even possible :D)
last = p.soc_capacity.last_value  # If we have a channel getting this, we could peek into it
```

Also we could use it to compose metrics, for example:

```py
composed_receiver = (
    logical_meter.grid_power + battery_pool.active_power
    ).new_receiver()
```

### Notes

* If getting the wrapper is not trivial, then it should be a function call instead of a property:
  ```py
  grid_power_receiver = logical_meter.grid_power().new_receiver()
  ```
  But we should try to cache these wrappers to allow for a property syntax if possible.

sahas-subramanian-frequenz:
This PR changes how formulas are composed -
  - receivers from formulas are no longer composable.
  - formula composition is now done by composing FormulaEngine instances.
  - Automatic formulas from the logical meter and *pools, are now properties, and return `FormulaEngine` instances, which can be composed further, or can provide a receiver to fetch values.

  ``` python
  grid_power_receiver = microgrid.logical_meter().grid_power.new_receiver()

  self._inverter_power = (
      microgrid.logical_meter().pv_power
      + microgrid.logical_meter().battery_power
  ).build("inverter_power")

  inverter_power_receiver = self._inverter_power.new_receiver()
  ```

The type hint issues in the formula engine are also resolved.

Based on #336.


src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
```diff
@@ -436,21 +437,21 @@ def __init__(
...
+    def engine(self) -> FormulaEngine | FormulaEngine3Phase:
         """Return the formula engine attached to the channel.
 
         Returns:
             A FormulaEngine instance.
         """
         return self._engine
 
-    def new_receiver(
+    def new_receiver(  # type: ignore[override]
```
mathias-baumann-frequenz: Why is this necessary?

sahas-subramanian-frequenz: Because `_BaseFormulaChannel` inherits `Broadcast`,  we should inherit the type of its methods as well,  i.e. this `new_receiver` should return `Receiver[Sample]`, etc.  But we are returning `FormulaReceiver` which derives from `Receiver[Sample]`,  but is still a different type with additional methods.  and mypy doesn't like this.

sahas-subramanian-frequenz:
Trying to get https://github.com/frequenz-floss/frequenz-sdk-python/issues/314 into this PR as well,  draft until then.

sahas-subramanian-frequenz:
This is ready for review again.

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
sahas-subramanian-frequenz: This is no longer relevant,  because `FormulaReceiver` are gone now.  FormulaEngines directly return broadcast receivers.

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
```diff
@@ -473,76 +495,85 @@ def push_average(self, metrics: List[Tuple[str, Receiver[Sample], bool]]) -> Non
...
         Args:
-            recv: A first input stream to create a builder with, so that python
+            engine: A first input stream to create a builder with, so that python
                 operators `+, -, *, /` can be used directly on newly created instances.
         """
-        self._steps: deque[tuple[TokenType, _GenericFormulaReceiver | str]] = deque()
-        self._steps.append((TokenType.COMPONENT_METRIC, recv.clone()))
-        recv._deactivate()  # pylint: disable=protected-access
-        self._engine = None
+        self._engine: FormulaEngine | FormulaEngine3Phase
```
daniel-zullo-frequenz: is it really needed to keep the state for this attribute?

daniel-zullo-frequenz (commented):
I only have a comment to check for

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
sahas-subramanian-frequenz: true,  it is not necessary.  removed.

daniel-zullo-frequenz (approved):

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine_pool.py
```diff
@@ -73,7 +68,7 @@ async def from_string(
         """
         channel_key = formula + component_metric_id.value
         if channel_key in self._engines:
-            return self._engines[channel_key].new_receiver()
+            return self._engines[channel_key]  # type: ignore
```
leandro-lucarella-frequenz, line 71: Why does this requires a type ignore?

src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool.py
```diff
@@ -104,40 +104,47 @@ def component_ids(self) -> abc.Set[int]:
...
+            A FormulaEngine that will calculate and stream the total current of all EV
+                Chargers.
         """
-        return await self._formula_pool.from_generator(
+        return self._formula_pool.from_generator(
             "ev_charger_total_current",
             EVChargerCurrentFormula,
             FormulaGeneratorConfig(component_ids=self._component_ids),
-        )
+        )  # type: ignore[return-value]
```
leandro-lucarella-frequenz, line 125: Also here or world be nice to clarify went the ignore is needed.

src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool.py
```diff
@@ -104,40 +104,47 @@ def component_ids(self) -> abc.Set[int]:
...
+            A FormulaEngine that will calculate and stream the total power of all EV
+                Chargers.
         """
-        return await self._formula_pool.from_generator(
+        return self._formula_pool.from_generator(
             "ev_charger_total_power",
             EVChargerPowerFormula,
             FormulaGeneratorConfig(component_ids=self._component_ids),
-        )
+        )  # type: ignore[return-value]
```
leandro-lucarella-frequenz, line 145: And here, etc.

leandro-lucarella-frequenz (commented):
Amazing to see so many async, await and new_receiver() go! 👋

Issue #191, opened 2023-01-31
Configuration management in the sdk


ela-kotulska-frequenz:
### What's needed?

This is still in the discussion! 

SDK and each actor has some config variables. This config should control how the application behaves.
All config should be stored in single file. This file should be loaded at startup. If file change, sdk should read updates, validate them and send to the subscribed actors. We should assume that all actors uses `marshmallow` to create configs.

### Proposed solution

Using [marshmallow](https://pypi.org/project/marshmallow/) (+ [apispec](https://pypi.org/project/apispec/) + [marshmallow_dataclass](https://pypi.org/project/marshmallow-dataclass/)). It is very convenient for validation, which will be necessary too, but also for documentation. We can then even export it automatically to [openAPI](https://github.com/OAI/OpenAPI-Specification) and then render it nicely with swagger.

It is very likely OpenAPI should have good support in the UI part, so they can directly use the same config specification we produce.

Quick example of how an actor or user should declare the config (producing OpenAPI docs too):

```py
import json
from typing import Optional
from dataclasses import field

import marshmallow_dataclass
import marshmallow.validate
from apispec import APISpec
from apispec.ext.marshmallow import MarshmallowPlugin


# Create an APISpec
spec = APISpec(
    title="Swagger Example",
    version="1.0.0",
    openapi_version="3.0.2",
    plugins=[MarshmallowPlugin()],
)


@marshmallow_dataclass.dataclass
class PeakShavingConfig:
    target_kw: float = field(default=5.0, metadata={  # First metadata is for marshmallow
        "validate": marshmallow.validate.Range(min=0.0, max=10.0),
        "metadata": {"description": "Peak shaving target (in kW)"},  # Second metadata is for apispec/OpenAPI
        })


@marshmallow_dataclass.dataclass
class EvChargingConfig:
    max_power_w: float = field(metadata={  # No default -> required
        "validate": marshmallow.validate.Range(min=0.0, max=10.0),
        "metadata": {"description": "Maximum allowed power for the site at the grid connection (in Watt)"},
        })
    min_power_w: Optional[float] = field(metadata={  # No default but optional -> NOT required
        "validate": marshmallow.validate.Range(min=0.0, max=10.0),
        "metadata": {"description": "Minimum allowed power for the site at the grid connection (in Watt)"},
        })
    data_gathering_duration_seconds: float = field(default=5.0, metadata={
        "validate": marshmallow.validate.Range(min=0.0, max=10.0),
[...]

PR #332, opened 2023-04-14
Add function to parse duration strings.

mathias-baumann-frequenz:

### DIFF
--- a/src/frequenz/sdk/actor/__init__.py
+++ b/src/frequenz/sdk/actor/__init__.py
@@ -5,7 +5,7 @@
 
 from ..timeseries._resampling import ResamplerConfig
 from ._channel_registry import ChannelRegistry
-from ._config_managing import ConfigManagingActor
+from ._config_managing import ConfigManagingActor, parse_duration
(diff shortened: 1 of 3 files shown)


leandro-lucarella-frequenz (commented):
I don't think we should merge this. The idea is to use marshmallow to validate configs, so we should go directly for that. I would just use the [`TimeDelta`](https://marshmallow.readthedocs.io/en/stable/marshmallow.fields.html#marshmallow.fields.TimeDelta) validator (that just converts an `int` or `float`(interpreted as seconds) to `timedelta`), as this config should be written by a machine in most of the cases.

We could think about adding a custom validator to accept a more human friendly syntax if needed (but in that case I would use [`isodate`](https://pypi.org/project/isodate/) under the hood, using [ISO duration format](https://en.wikipedia.org/wiki/ISO_8601#Durations), which is basically the same you wrote but with some weird extra characters, but at least it is ISO :laughing: ).

leandro-lucarella-frequenz:
For more context:

* https://github.com/frequenz-floss/frequenz-sdk-python/issues/191

PR #251, opened 2023-03-10
Increase sampling rate in the flaky formula test

ela-kotulska-frequenz:
.. so it fails less frequently
This tests fails randomly and should be fixed soon.

Issue #249, opened 2023-03-10
Some Formula tests are flaky


ela-kotulska-frequenz:
### What happened?

```
    async def test_3_phase_formulas(self, mocker: MockerFixture) -> None:
        """Test 3 phase formulas current formulas and their composition."""
        mockgrid = MockMicrogrid(grid_side_meter=False)
        mockgrid.add_batteries(3)
        mockgrid.add_ev_chargers(1)
        request_chan, channel_registry = await mockgrid.start(mocker)
        logical_meter = LogicalMeter(
            channel_registry,
            request_chan.new_sender(),
            microgrid.get().component_graph,
        )
        grid_current_recv = await logical_meter.grid_current()
        pool = EVChargerPool(channel_registry, request_chan.new_sender())
        ev_current_recv = await pool.total_current()
    
        engine = (grid_current_recv.clone() - ev_current_recv.clone()).build(
            "net_current"
        )
        net_current_recv = engine.new_receiver()
    
        await synchronize_receivers(
            [grid_current_recv, ev_current_recv, net_current_recv]
        )
    
        for _ in range(10):
            grid_amps = await grid_current_recv.receive()
            ev_amps = await ev_current_recv.receive()
            net_amps = await net_current_recv.receive()
    
>           assert grid_amps.value_p1 is not None and grid_amps.value_p1 > 0.0
E           assert (None is not None)
E            +  where None = Sample3Phase(timestamp=datetime.datetime(2023, 3, 10, 7, 9, 39, 78987, tzinfo=datetime.timezone.utc), value_p1=None, value_p2=None, value_p3=None).value_p1
```

### What did you expect instead?

We should fix them

### Affected part(s)

Unit, integration and performance tests (part:tests)

ela-kotulska-frequenz:
Increased the sampling period rate so it fails less frequently: https://github.com/frequenz-floss/frequenz-sdk-python/pull/251

mathias-baumann-frequenz:
This test is also flaky: 

```
=================================== FAILURES ===================================
_____________________ TestEVChargerPool.test_state_updates _____________________

self = <tests.timeseries.test_ev_charger_pool.TestEVChargerPool object at 0x7ff695e1c990>
mocker = <pytest_mock.plugin.MockerFixture object at 0x7ff68f98a510>

    async def test_state_updates(self, mocker: MockerFixture) -> None:
        """Test ev charger state updates are visible."""
    
[...]

PR #335, opened 2023-04-14
Add an option to align resampling windows

### LINKED ISSUE #328: Resampler should align timestamps to Unix Epoch
### What's needed?

Currently if the resampling actor is started with period 2sec, at 10:10:01.2,  then the messages out of the resampling actor would have timestamps:
 - 10:10:03.2
 - 10:10:05.2
 - 10:10:07.2

This means that if the resampling actor restarts,  timestamps will go out of sync.  They should instead be:
 - 10:10:02
 - 10:10:04
 - 10:10:06

### Proposed solution

The resampling actor should align timestamps to unix epoch, and not to actor start time, so that timestamps between restarts will still be compatible.

### Use cases

This is necessary 
- when dumping/loading data from the ring buffer.
- for usecases such as the 24hr forecast or peak shaving where we need the features to be aligned to unix epoch, specifically for 15min windows, i.e. 11:00, 11:15, 11:30, 11:45, 12:00, ... (copied from @idlir-shkurti-frequenz's comment below)

_No response_

### Additional context

_No response_

leandro-lucarella-frequenz:
The `ResamplerConfig` now has a new option `align_to` to align the resampling windows to some arbitrary time. This means that the timestamps of the produced samples will be a multiple of the resampling period beginning at the `align_to` time.

By default `align_to` is the Unix epoch, and if it is `None` it will have the current behaviour of aligning timestamps to the time when the resampler is started.

Fixes #328.

### DIFF
--- a/src/frequenz/sdk/timeseries/__init__.py
+++ b/src/frequenz/sdk/timeseries/__init__.py
@@ -6,8 +6,35 @@ Handling of timeseries streams.
 
 A timeseries is a stream (normally an async iterator) of
 [`Sample`][frequenz.sdk.timeseries.Sample]s.
+
+# Periodicity and alignment
+
+All the data produced by this package is always periodic and aligned to the
+`UNIX_EPOCH` (by default).
+
+Classes normally take a (re)sampling period as and argument and, optionally, an
+`align_to` argument.
+
+This means timestamps are always separated exaclty by a period, and that this
+timestamp falls always at multiples of the period, starting at the `align_to`.
+
+This ensures that the data is predictable and consistent among restarts.
+
+Example:
+    If we have a period of 10 seconds, and are aligning to the UNIX
+    epoch. Assuming the following timeline starts in 1970-01-01 00:00:00
+    UTC and our current `now` is 1970-01-01 00:00:32 UTC, then the next
+    timestamp will be at 1970-01-01 00:00:40 UTC:
+
+    ```
+    align_to = 1970-01-01 00:00:00         next event = 1970-01-01 00:00:40
+    |                                       |
+    |---------|---------|---------|-|-------|---------|---------|---------|
+    0        10        20        30 |      40        50        60        70
+                                   now = 1970-01-01 00:00:32
+    ```
 """
 
-from ._base_types import Sample, Sample3Phase
+from ._base_types import UNIX_EPOCH, Sample, Sample3Phase
 
-__all__ = ["Sample", "Sample3Phase"]
+__all__ = ["Sample", "Sample3Phase", "UNIX_EPOCH"]
--- a/src/frequenz/sdk/timeseries/_resampling.py
+++ b/src/frequenz/sdk/timeseries/_resampling.py
@@ -17,6 +17,7 @@ from typing import AsyncIterator, Callable, Coroutine, Optional, Sequence
 
 from .._internal.asyncio import cancel_and_await
 from . import Sample
+from ._base_types import UNIX_EPOCH
 
 _logger = logging.getLogger(__name__)
 
@@ -191,6 +192,17 @@ class ResamplerConfig:
     It must be at bigger than `warn_buffer_len`.
     """
 
+    align_to: datetime | None = UNIX_EPOCH
+    """The time to align the resampling period to.
+
+    The resampling period will be aligned to this time, so the first resampled
+    sample will be at the first multiple of `resampling_period` starting from
+    `align_to`. It must be an aware datetime and can be in the future too.
+
+    If `align_to` is `None`, the resampling period will be aligned to the
+    time the resampler is created.
+    """
+
     def __post_init__(self) -> None:
         """Check that config values are valid.
 
@@ -230,6 +242,10 @@ class ResamplerConfig:
                 "initial_buffer_len (%s) is bigger than warn_buffer_len (%s)",
                 self.initial_buffer_len,
                 self.warn_buffer_len,
+            )
+        if self.align_to is not None and self.align_to.tzinfo is None:
+            raise ValueError(
+                f"align_to ({self.align_to}) should be a timezone aware datetime"
             )
 
 
@@ -348,9 +364,7 @@ class Resampler:
         self._resamplers: dict[Source, _StreamingHelper] = {}
         """A mapping between sources and the streaming helper handling that source."""
 
-        self._window_end: datetime = (
-            datetime.now(timezone.utc) + self._config.resampling_period
-        )
+        self._window_end: datetime = self._calculate_window_end()
         """The time in which the current window ends.
 
         This is used to make sure every resampling window is generated at
@@ -358,6 +372,9 @@ class Resampler:
         never fire at the exact requested time, so if we don't use a precise
         time for the end of the window, the resampling windows we produce will
         have different sizes.
+
+        The window end will also be aligned to the `config.align_to` time, so
+        the window end is deterministic.
         """
 
     @property
@@ -490,6 +507,29 @@ class Resampler:
                 timer_error,
                 self._config.resampling_period,
             )
+
+    def _calculate_window_end(self) -> datetime:
+        """Calculate the end of the current resampling window.
+
+        The calculated resampling window end is a multiple of
+        `self._config.resampling_period` starting at `self._config.align_to`.
+
+        if `self._config.align_to` is `None`, the current time is used.
+
+        Returns:
+            The end of the current resampling window aligned to
+                `self._config.align_to`.
+        """
+        now = datetime.now(timezone.utc)
+        period = self._config.resampling_period
+        align_to = self._config.align_to
+
+        if align_to is None:
+            return now + period
+
+        elapsed = (now - align_to) % period
+
+        return now + period - elapsed
 
 
 class _ResamplingHelper:
--- a/tests/timeseries/mock_microgrid.py
+++ b/tests/timeseries/mock_microgrid.py
@@ -119,7 +119,14 @@ class MockMicrogrid:  # pylint: disable=too-many-instance-attributes
 
         # pylint: disable=protected-access
         _data_pipeline._DATA_PIPELINE = _data_pipeline._DataPipeline(
-            ResamplerConfig(resampling_period=timedelta(seconds=self._sample_rate_s))
+            ResamplerConfig(
+                resampling_period=timedelta(seconds=self._sample_rate_s),
+                # Align to the time the resampler is created to avoid flakiness
+                # in the tests, it seems test using the mock microgrid assume
+                # that the resampling window is aligned to the start of the
+                # test.
+                align_to=None,
+            )
         )
         # pylint: enable=protected-access
 
--- a/tests/timeseries/test_resampling.py
+++ b/tests/timeseries/test_resampling.py
@@ -34,7 +34,8 @@ from frequenz.sdk.timeseries._resampling import (
 
 from ..utils import a_sequence
 
-# pylint: disable=too-many-locals,redefined-outer-name
+# We relax some pylint checks as for tests they don't make a lot of sense.
+# pylint: disable=too-many-lines,disable=too-many-locals,redefined-outer-name
 
 
 # Setting 'autouse' has no effect as this method replaces the event loop for all tests in the file.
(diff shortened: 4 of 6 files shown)


leandro-lucarella-frequenz:
Creating it as an early draft so I can already get feedback while working on the TODO. I did some local tests and it seems to be working fine, so the implementation itself (the last commit) is ready for a review.

leandro-lucarella-frequenz:
I was just looking at the moving window and ring buffer to see if the alignment functions can be unified, but it doesn't look trivial, as the moving window just forwards the alignment problem to the ring buffer and the ring buffer has other issues to deal with, like wrapping around, so it needs more operations related to the alignment.

But I think we could unify the API at least:

* The ring buffer uses `time_index_alignment`
* The moving window uses `window_alignment`
* The resampler uses `align_to`

@matthias-wende-frequenz @mathias-baumann-frequenz do you think it makes sense to unify everything to use `align_to`? The other too are too specific and don't play well in other components. I think this value needs good documentation because it is not trivial to understand, so I guess using a more general keyword argument name is not a big loss, at least compared to the added clarity if we use the same term for aligning different components. At the end the concept is the same, right?

We should also probably unify the default too. The resampler uses the epoch (I defined a `UNIX_EPOCH` *constant*) while the other 2 use `datetime(1, 1, 1, tzinfo=timezone.utc)`. Maybe we should have a `frequenz.sdk.timeseries.alignment` module defining the generic alignment functions and constants?

leandro-lucarella-frequenz:
The test `tests/timeseries/_formula_engine/test_formula_composition.py::TestFormulaComposition::test_formula_composition` is definitely flaky, testing it locally and in the CI it sometimes passes and it sometimes doesn't. I tried increasing the `sampling_rate` of the `MicrogridMock` (as done in #251) but it doesn't seem to help. Also I'm not sure why this commit makes the test flakyness to trigger.

We should try to find a final solution for this:

* #249

sahas-subramanian-frequenz:
This particular test already has a (temporary) fix in `main`,  can you rebase?

leandro-lucarella-frequenz:
> This particular test already has a (temporary) fix in `main`, can you rebase?

I will, but these failing tests are rebased on top of all relevant changes already (only the bump on the `frequenz-floss/setup-git-user` action dependency is missing).

Also locally I tried increasing the sampling period even up to `0.25` and it was still failing very often. Didn't tried with higher values though, but that already looks excessive.

leandro-lucarella-frequenz:
I just tried and not even using 1 second removes the flakiness :thinking: :scream:

sahas-subramanian-frequenz:
> I just tried and not even using 1 second removes the flakeyness 

how about locally?

src/frequenz/sdk/timeseries/__init__.py
```diff
@@ -6,8 +6,35 @@
...
+    UTC and our current `now` is 1970-01-01 00:00:32 UTC, then the next
+    timestamp will be at 1970-01-01 00:00:40 UTC:
+
+    ```
+    align_to = 1970-01-01 00:00:00         next event = 1970-01-01 00:00:40
+    |                                       |
+    |---------|---------|---------|-|-------|---------|---------|---------|
+    0        10        20        30 |      40        50        60        70
+                                   now = 1970-01-01 00:00:32
+    ```
```
leandro-lucarella-frequenz: This is jumping a bit ahead and should be true after #343 is done (I'm already working on it). I couldn't find a better place to put this generic documentation on, and the `UNIX_EPOCH`. An `alignment` module doesn't make a lot of sense, as what the ring buffer needs to do is pretty different from what the resampler needs to do, and at the end the alignment itself is just a simple math operation: `now + period - (now - align_to) % period`, a bit too simple to deserve its own module.

leandro-lucarella-frequenz:
Remove draft status, as the only pending thing is the flaky tests.

leandro-lucarella-frequenz:
> how about locally?

Yes, I tried it locally.

leandro-lucarella-frequenz:
I wonder if now this is flaky based on if the now is aligned to the epoch or not. Since I'm not familiar with the `MockMicrogrid` and how it interacts with `sdk.micrigrid`, I'm not sure if this could be the issue. I could try to align to `now` in the tests to see if it fixes it.

leandro-lucarella-frequenz:
OK, this seems to remove the flakyness, so I guess that was the reason:

```diff
diff --git a/tests/timeseries/mock_microgrid.py b/tests/timeseries/mock_microgrid.py
index 7a6133a..7531b08 100644
--- a/tests/timeseries/mock_microgrid.py
+++ b/tests/timeseries/mock_microgrid.py
@@ -119,7 +119,9 @@ class MockMicrogrid:  # pylint: disable=too-many-instance-attributes
 
         # pylint: disable=protected-access
         _data_pipeline._DATA_PIPELINE = _data_pipeline._DataPipeline(
-            ResamplerConfig(resampling_period=timedelta(seconds=self._sample_rate_s))
+            ResamplerConfig(
+                resampling_period=timedelta(seconds=self._sample_rate_s), align_to=None
+            )
         )
         # pylint: enable=protected-access
 
```

Now the question is, should I add this change or should I adapt the tests to work with alignment? I think it would be too big of a change for this PR, specially because I think the most sensible way to make this work would be to mock time, otherwise is hard to align the tests to the resampler alignment... :roll_eyes:

leandro-lucarella-frequenz:
I just pushed those changes to check if now test pass more reliably.

leandro-lucarella-frequenz:
It seems to be working now.

src/frequenz/sdk/timeseries/__init__.py
```diff
@@ -6,8 +6,35 @@
...
+All the data produced by this package is always periodic and aligned to the
+`UNIX_EPOCH` (by default).
+
+Classes normally take a (re)sampling period as and argument and, optionally, an
+`align_to` argument.
+
+This means timestamps are always separated exaclty by a period, and that this
+timestamp falls always at multiples of the period, starting at the `align_to`.
+
+This ensures that that the data is predictable and consistent among restarts.
```
mathias-baumann-frequenz: that that

src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -191,6 +192,17 @@ class ResamplerConfig:
...
 
+    align_to: datetime | None = UNIX_EPOCH
+    """The time to align the resampling period to.
+
+    The resampling period will be aligned to this time, so the first resampled
+    sample will be at the first multiple of `resampling_period` starting from
+    `align_to`. It must be an aware datetime and can be in the future too.
+
+    If `align_to` is `None`, the resampling period will be aligned to the
+    time the resampler is created.
```
mathias-baumann-frequenz, line 203: Is that reasonable with `None`? I can't think of any scenario where we would want to use the `time the resampler is created` Any other behavior seems more useful to me than that

src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -231,6 +243,8 @@ def __post_init__(self) -> None:
                 self.initial_buffer_len,
                 self.warn_buffer_len,
             )
+        if self.align_to is not None and self.align_to.tzinfo is None:
+            raise ValueError(f"align_to ({self.align_to}) should be an aware datetime")
```
mathias-baumann-frequenz: Might be useful in other places too to have that. Maybe make it a function in the future? Not in this PR.

tests/timeseries/mock_microgrid.py
```diff
@@ -119,7 +119,12 @@ async def start(self, mocker: MockerFixture) -> None:
 
         # pylint: disable=protected-access
         _data_pipeline._DATA_PIPELINE = _data_pipeline._DataPipeline(
-            ResamplerConfig(resampling_period=timedelta(seconds=self._sample_rate_s))
+            ResamplerConfig(
+                resampling_period=timedelta(seconds=self._sample_rate_s),
+                # Align to the time the resampler is created to avoid flakybess
+                # in the tests
```
mathias-baumann-frequenz: Having a value in the tests that is different every time _reduces_ flakyness??

Also you wrote `flakybess`.

tests/timeseries/test_resampling.py
```diff
@@ -34,7 +34,8 @@
 
 from ..utils import a_sequence
 
-# pylint: disable=too-many-locals,redefined-outer-name
+# We relax some pylint checks as for tests don't make a lot of sense.
```
mathias-baumann-frequenz: *they

src/frequenz/sdk/timeseries/__init__.py
leandro-lucarella-frequenz: that that that!

src/frequenz/sdk/timeseries/_resampling.py
leandro-lucarella-frequenz: I think it is a reasonable semantic for `None`, yes. If it is a useful semantic or not is another issue. I don't see any harm in providing the option (please note that `None` is **not** the default), and we are using it to make some old tests pass that otherwise will have to be rewritten/fixed (which would be good to know anyway, but a bit out of the scope of this PR).

src/frequenz/sdk/timeseries/_resampling.py
leandro-lucarella-frequenz: Maybe, I'm not sure, if the idea is to rely on `pylint` for this, it might be redundant at the end. To be honest I wasn't sure if I should include it not, it looks like it might be overkill if we start littering all the code with checks for datetime awareness. It is really a shame this can't be checked by the type checker. Sometimes I think it might make sense to have a sealed `AwareDatetime` that can only be converted to a naive `datetime` by calling a conversion function that takes a target `timezone` (pretty much like `str` -> `bytes` needs a target `encoding`).

I'm really undecided about this, also while converting `int` in seconds to `timedelta` I was thinking it might be really useful to have a `seconds(x)` alias for `timedelta(seconds=x)`. Or to have some sort of `now()` that produces a timezone aware UTC now.

But then I think it is just saving a few characters and it does make the code a bit less obvious for the casual reader (anyone familiar with python but not with our code knows what `timedelta(seconds=x)` is, but `seconds(x)` is an `int`? a `float`? a `timedelta`?)

Anyway, this is material for a discussion, so if you want to continue talking about it, I suggest creating a **Discussion** topic.

As for the PR, I'm fine with both leaving the check or remove it.

tests/timeseries/mock_microgrid.py
leandro-lucarella-frequenz: Yes. I guess the test is assuming the resampling happens relative to the time the test starts. Will fix the typo though.

leandro-lucarella-frequenz:
Updated with some fixes for comments made by @mathias-baumann-frequenz

src/frequenz/sdk/timeseries/_resampling.py
sahas-subramanian-frequenz:
``` suggestion
            raise ValueError(f"align_to ({self.align_to}) should be a timezone aware datetime")

File: tests/timeseries/mock_microgrid.py [hunk 7]
```diff
@@ -119,7 +119,14 @@ async def start(self, mocker: MockerFixture) -> None:
 
         # pylint: disable=protected-access
         _data_pipeline._DATA_PIPELINE = _data_pipeline._DataPipeline(
-            ResamplerConfig(resampling_period=timedelta(seconds=self._sample_rate_s))
+            ResamplerConfig(
+                resampling_period=timedelta(seconds=self._sample_rate_s),
+                # Align to the time the resampler is created to avoid flakyness
```
mathias-baumann-frequenz: I think it's actually spelled "flakiness" :sweat_smile:

sahas-subramanian-frequenz:
```suggestion
                # Align to the time the resampler is created to avoid flakiness

sahas-subramanian-frequenz (dismissed):
Just a couple of doc improvements,  looks great otherwise!

leandro-lucarella-frequenz:
I fixed the remaining typos as suggested and enabled auto-merge.

sahas-subramanian-frequenz (approved):

PR #336, opened 2023-04-15
Update `BatteryPool` tests, examples to use `microgrid.battery_pool()`

sahas-subramanian-frequenz:
... instead of creating their own `BatteryPool` instances.


tests/timeseries/_battery_pool/test_battery_pool.py
```diff
@@ -130,27 +131,37 @@ async def setup_all_batteries(mocker: MockerFixture) -> AsyncIterator[SetupArgs]
...
     yield args
-    await asyncio.gather(*[sdk.stop(), battery_pool.stop(), streamer.stop()])
+    await asyncio.gather(
+        *[
+            microgrid._data_pipeline._DATA_PIPELINE._stop(),
+            battery_pool.stop(),
+            streamer.stop(),
+        ]
+    )
+    # pylint: enable=protected-access
```
mathias-baumann-frequenz, line 164: Oh, I didn't know you can write it after, too

sahas-subramanian-frequenz: yes,  that functions like a block,  you `disable` first,  then do all the pylint violations,  then `enable` back, so that the rest of the code isn't affected.

src/frequenz/sdk/microgrid/_data_pipeline.py
```diff
@@ -170,7 +170,9 @@ def battery_pool(
 
         if key not in self._battery_pools:
             self._battery_pools[key] = BatteryPool(
-                batteries_status_receiver=self._battery_status_channel.new_receiver(),
+                batteries_status_receiver=self._battery_status_channel.new_receiver(
+                    maxsize=1
+                ),
```
daniel-zullo-frequenz, line 177: I guess this is done because it only needs to keep the latest update, right? Otherwise please clarify it in the commit message of this change

examples/battery_pool.py
```diff
@@ -81,7 +28,7 @@ async def main() -> None:
         host=HOST, port=PORT, resampler_config=ResamplerConfig(resampling_period_s=1.0)
     )
 
-    battery_pool = create_battery_pool()
+    battery_pool = microgrid.battery_pool()
     receivers: Dict[str, Receiver[Any]] = {
```
daniel-zullo-frequenz, line 32: Have you run the example with these changes? just checking...

daniel-zullo-frequenz (dismissed):

src/frequenz/sdk/microgrid/_data_pipeline.py
sahas-subramanian-frequenz: yup, it is mentioned in the BatteryPool documentation: https://github.com/frequenz-floss/frequenz-sdk-python/blob/c565ddf9e2e240c55e2873eb351daf4589e262db/src/frequenz/sdk/timeseries/battery_pool/battery_pool.py#L42-L43

Just realized we're doing only one of those.  I will do the other one also in the same commit.

sahas-subramanian-frequenz: updated.

examples/battery_pool.py
sahas-subramanian-frequenz: yes,  it is able to use the battery_pool instance from the microgrid module.

daniel-zullo-frequenz (approved):

PR #338, opened 2023-04-16
Add `power` formula to `BatteryPool`

### LINKED ISSUE #260: Add `active power` to the BatteryPool
### What's needed?

acitive_power = sum(battery.adjacent_inverter.active_power for each battery in the battery pool).
Even is battery is considered as broken, its active power should be included in this formula. (active power of the inverter adjacent to this battery).
If active_power is NaN or None, then we should replace it with 0.

Because we can include broken batteries, then it  can be implemented with formula engine (in a way EVPool is doing it).

### Proposed solution

_No response_

### Use cases

_No response_

_No response_

### Additional context

_No response_

sahas-subramanian-frequenz:
And remove `battery_power` from the `LogicalMeter`.

Based on #327

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,8 +6,7 @@
 
 ## Upgrading
 
-+ Battery power is no longer available through the `LogicalMeter`, but
-  through the `BatteryPool` (#338)
+* Battery power is no longer available through the `LogicalMeter`, but through the `BatteryPool` (#338)
 
   ``` python
   battery_power_receiver = microgrid.battery_pool().power.new_receiver()
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,6 +6,13 @@
 
 ## Upgrading
 
++ Battery power is no longer available through the `LogicalMeter`, but
+  through the `BatteryPool` (#338)
+
+  ``` python
+  battery_power_receiver = microgrid.battery_pool().power.new_receiver()
+  ```
+
 + Formulas composition has changed (#327) -
   - receivers from formulas are no longer composable.
   - formula composition is now done by composing FormulaEngine instances.
@@ -18,7 +25,7 @@
 
   self._inverter_power = (
       microgrid.logical_meter().pv_power
-      + microgrid.logical_meter().battery_power
+      + microgrid.battery_pool().power
   ).build("inverter_power")
 
   inverter_power_receiver = self._inverter_power.new_receiver()
--- a/src/frequenz/sdk/microgrid/_data_pipeline.py
+++ b/src/frequenz/sdk/microgrid/_data_pipeline.py
@@ -176,6 +176,8 @@ class _DataPipeline:
 
         if key not in self._battery_pools:
             self._battery_pools[key] = BatteryPool(
+                channel_registry=self._channel_registry,
+                resampler_subscription_sender=self._resampling_request_sender(),
                 batteries_status_receiver=self._battery_status_channel.new_receiver(
                     maxsize=1
                 ),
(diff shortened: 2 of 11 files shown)


src/frequenz/sdk/microgrid/_data_pipeline.py
```diff
@@ -10,12 +10,18 @@
...
 import typing
 from collections import abc
 from dataclasses import dataclass
 
 from frequenz.channels import Bidirectional, Broadcast, Sender
 
+from . import connection_manager
+from .component import ComponentCategory
+
+logger = logging.getLogger(__name__)
```
daniel-zullo-frequenz, line 23: you might want to define it as `_logger` instead

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
```diff
@@ -389,262 +495,105 @@ def push_average(self, metrics: List[Tuple[str, Receiver[Sample], bool]]) -> Non
...
         Args:
-            recv: A first input stream to create a builder with, so that python
+            engine: A first input stream to create a builder with, so that python
                 operators `+, -, *, /` can be used directly on newly created instances.
         """
-        self._steps: deque[tuple[TokenType, _GenericFormulaReceiver | str]] = deque()
-        self._steps.append((TokenType.COMPONENT_METRIC, recv.clone()))
-        recv._deactivate()  # pylint: disable=protected-access
-        self._engine = None
+        self._engine: FormulaEngine | FormulaEngine3Phase
```
daniel-zullo-frequenz: Should `self._engine` also be None and set to None?

daniel-zullo-frequenz: Actually I didn't spot any reason in the changes to keep this attribute  `self._engine`. Can you please double-check if it is still needed?

src/frequenz/sdk/microgrid/_data_pipeline.py
sahas-subramanian-frequenz: Renaming all custom variables in a separate PR: #347

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
daniel-zullo-frequenz: already addressed

RELEASE_NOTES.md
```diff
@@ -6,6 +6,13 @@
 
 ## Upgrading
 
++ Battery power is no longer available through the `LogicalMeter`, but
+  through the `BatteryPool` (#338)
```
leandro-lucarella-frequenz:
```suggestion
* Battery power is no longer available through the `LogicalMeter`, but through the `BatteryPool` (#338)
```

(we are using mostly `*` in the file so far, and when the release notes are copied to the GitHub release, the markdown parsing is strange and keeps the new-lines, which makes it look weird)

leandro-lucarella-frequenz (commented):
Other than the super minimal formatting issue, LGTM!

RELEASE_NOTES.md
sahas-subramanian-frequenz: I don't like the idea of using `*` everywhere,  for second level also.  I like `+, -, *` in that order, to make it more readable.

I also think the Github line folding behaviour for release notes is intentional,  and it should be folded there to make the release notes easily readable, instead of having to read from left end of screen to the right.

But I guess "usually" is a good motivator as well.  So changed.

leandro-lucarella-frequenz: I'm fine with using different chars for different levels, is just the inconsistency at the same level that makes it harder to read for me.

About the line length, maybe you are right, but I think it is a mistake, because the rendered should be able to decide which length is the most appropriate, if you have a very narrow screen, like a cell phone, then the lines will be automatically and artificially wrapped, making it very hard to read.

leandro-lucarella-frequenz (approved):

PR #346, opened 2023-04-19
Mock resampled streams in test_formula_composition

sahas-subramanian-frequenz:
... to decouple it from the resampler.  This needs to be generalized
and used in all the tests in a subsequent PR.

### DIFF
--- a/tests/timeseries/_formula_engine/test_formula_composition.py
+++ b/tests/timeseries/_formula_engine/test_formula_composition.py
@@ -3,13 +3,18 @@
 
 """Tests for formula composition."""
 
+from __future__ import annotations
 
+from datetime import datetime
 from math import isclose
+from typing import Any, Dict, List
 
+from frequenz.channels import Broadcast, Receiver, Sender
 from pytest_mock import MockerFixture
 
 from frequenz.sdk import microgrid
 from frequenz.sdk.microgrid.component import ComponentMetricId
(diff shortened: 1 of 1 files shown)


leandro-lucarella-frequenz (commented):
LGTM. I wonder if this is something that can be added to the `MockMicrogrid`.

If we can't and we need to change all the tests, I think we might better try rethink how mocking is being done.

I have the feeling we have a lot of complexity in tests because we are mocking *too deep*. I think we should not mock the whole microgrid, but just the parts that interact with what we are trying to test.

One way to do this is to start with a test where the `microgrid` object is just a *sealed* `Mock` object, so any use of any attributes of microgrid will produce a test failure. Then you for each attribute accesses, you mock only that attribute. That should make the test more self contained, simpler to understand and less flaky.

I know I'm talking from a very theoretical standpoint here, so maybe is not that easy for this case in particular, but it's something to think about. Sometimes I think we are closer to integration tests than to unit tests (of course this doesn't mean we shouldn't do integration tests, but I think being in the middle is what makes things a bit complicated).

sahas-subramanian-frequenz:
> I wonder if this is something that can be added to the MockMicrogrid.

yes,  that's what I meant by "generalized", should have been more specific.  This PR can't be merged anymore,  but because the urgency has gone away as well,  I will do it in a separate PR next week.

PR #350, opened 2023-04-20
Update generated formulas to follow the new naming scheme

### LINKED ISSUE #340: Update generated formulas to use new naming scheme
Updated proposal:
* Terminology follows `{component}_{quantity}` and `{component}_{consumption,production}_{quantity}`.
* Components are: `grid`, `battery`, `consumer` and individual active components `pv`, `chp`, `wind`.

 
Using the example of power:
* **grid_power**: Can be positive (consumption from the grid) or negative (production into the grid).
* **{battery,ev}_power**: Can be positive (charge battery) or negative (discharge battery). Equivalently for EV charging station.
* **{pv,chp,wind}_power**: Active components, the power from local (renewable) generation. Negative if electric power is generated, otherwise zero or positive in case of self-consumption of the generator. (alternative terms: generation, supply, source, active)
* **consumer_power**: This aggregates the remaining parts not covered by the above metrics. Under normal circumstances this is expected to correspond to the gross consumption of the site excluding active parts and battery. The implementation should still support negative values to also handle exotic site topologies (e.g. unidentified generators) and sudden short-term effects. (alternative terms: load, demand, sink, passive)

In addition to that, for each of the above there will two additional metrics:
 * **{component}_consumption_power**: Positive power value if `{component}_power` is positive, otherwise zero.
  * **{component}_production_power**:  Positive power value if `{component}_power` is negative, otherwise zero.

Remarks:
* Gross consumption: Consumption before accounting for any local generation from solar, wind or CHP.
* Net consumption/load: This term traditionally refers to the difference between the gross electricity consumption and the local generation (like PV production). This is the electricity consumption that needs to be met by the battery or from the main grid. This means `consumer_power + {pv,chp,wind}_power` (note that latter is negative when electricity is produced). Since this term caused confusion in the past, it's not included in this proposal. 
* Residual consumption/load: In microgrid context sometimes used as the remaining difference between the net consumption and the battery power, i.e. what we define as grid power.   
* Consumer: Term deliberately chosen to indicate that this component is expected to predominantly consume electric power. Under normal circumstances there should be no active sources in this metric and the metrics `consumer_{consumption,production}_power` would not be used.

_Originally posted by @cwasicki in https://github.com/frequenz-floss/frequenz-sdk-python/discussions/318#discussioncomment-5589177_

sahas-subramanian-frequenz:
**New power streaming methods** - _public api_

- `BatteryPool.{production_power, consumption_power}`
- `EVChargerPool.{production_power, consumption_power}`
- LogicalMeter:
  - `consumer_power`
  - `grid_production_power`
  - `grid_consumption_power`
  - `chp_power`
  - `chp_production_power`
  - `chp_consumption_power` (probably not necessary, but having just a production method was a bit weird)

**New FormulaEngine steps** - _internal_
- `ConstantValue` - for multiplying formula outputs with `-1` to get production power.
- `Clipper` - for clipping formula outputs to 0.0 in production/consumption power formulas.

**New component graph methods** - _internal_
- `is_pv_inverter`, `is_pv_meter`, `is_pv_chain`
- `is_battery_inverter`, `is_battery_meter`, `is_battery_chain`
- `is_ev_charger`, `is_ev_charger_meter`, `is_ev_charger_chain`,
- `is_chp`, `is_chp_meter`, `is_chp_chain`

These methods were particularly useful in the creation of the `consumer_power` formula.

The `*chain` methods return true if either of the other corresponding methods return true.  Maybe there's a better name, but it is for internal use.


sahas-subramanian-frequenz:
Need to add tests for the `consumer_power` formula,  will be ready tomorrow.

src/frequenz/sdk/timeseries/_formula_engine/_formula_steps.py
```diff
@@ -225,6 +225,41 @@ def apply(self, eval_stack: List[float]) -> None:
         eval_stack.append(self._value)
 
 
+class Clamper(FormulaStep):
```
cwasicki: Isn't this doing clipping instead of clampering? According to https://en.wikipedia.org/wiki/Clamper_(electronics) clampering would shift the whole series, while here the values are clipped inside a range.

src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py
```diff
@@ -239,3 +240,70 @@ def pv_power(self) -> FormulaEngine:
         )
         assert isinstance(engine, FormulaEngine)
         return engine
+
```
cwasicki, line 321: Out of curiosity, why did you choose this kind of design that you need a new function for each combination of component, metric and direction? Couldn't you save a lot of code by providing something like `power(component, direction=None)` where component is `chp`, `pv`, `grid`, etc and direction is either `consumption`, `production` or both?

src/frequenz/sdk/timeseries/_formula_engine/_formula_steps.py
cwasicki: As discussed, in some programming languages the term clamping is used in the way you use it, but in electronics it is used differently. So I would favor `clip` as it's less ambiguous.

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_chp_power_formula.py
```diff
@@ -0,0 +1,107 @@
...
+        component_graph = connection_manager.get().component_graph
+        chps = list(
+            comp
+            for comp in component_graph.components()
+            if comp.category == ComponentCategory.CHP
+        )
+        if not chps:
+            return None
+
+        chp_meters: set[int] = set()
```
cwasicki, line 84: Can the same CHP id occur multiple time in the following loop?

src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool.py
```diff
@@ -141,9 +142,62 @@ def power(self) -> FormulaEngine:
...
+        """Fetch the total power produced by the EV Chargers in the pool.
+
+        If a formula engine to calculate EV Charger power is not already running, it
+        will be started.
+
+        A receiver from the formula engine can be created using the `new_receiver`
+        method.
+
+        Returns:
+            A FormulaEngine that will calculate and stream the total power of all EV
```
cwasicki: `production power`

src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool.py
```diff
@@ -141,9 +142,62 @@ def power(self) -> FormulaEngine:
...
+        """Fetch the total power consumed by the EV Chargers in the pool.
+
+        If a formula engine to calculate EV Charger power is not already running, it
+        will be started.
+
+        A receiver from the formula engine can be created using the `new_receiver`
+        method.
+
+        Returns:
+            A FormulaEngine that will calculate and stream the total power of all EV
```
cwasicki: `consumption power`

src/frequenz/sdk/microgrid/_graph.py
```diff
@@ -352,6 +495,179 @@ def validate(self) -> None:
...
+        Args:
+            component: component to check.
+
+        Returns:
+            Whether the specified component is a CHP meter.
+        """
+        successors = self.successors(component.component_id)
+        return (
+            component.category == ComponentCategory.METER
+            and len(successors) > 0
```
cwasicki, line 675: Can this be `==1`?

src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py
```diff
@@ -239,3 +240,70 @@ def pv_power(self) -> FormulaEngine:
...
+        engine = self._formula_pool.from_generator(
+            "chp_power",
+            CHPPowerFormula,
+            FormulaGeneratorConfig(formula_type=FormulaType.PASSIVE_SIGN_CONVENTION),
+        )
+        assert isinstance(engine, FormulaEngine)
+        return engine
+
+    @property
+    def chp_production_power(self) -> FormulaEngine:
```
cwasicki, line 346: Looks like the methods for PV are missing in this PR?

cwasicki (commented):
Some high-level comments. Most importantly I think the methods for solar are missing.

src/frequenz/sdk/timeseries/_formula_engine/_formula_steps.py
sahas-subramanian-frequenz: I've renamed this,  it is internal to the sdk anyway,  so it would be fine to discuss it further at a later time  if necessary

src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py
sahas-subramanian-frequenz: We have something like that internally,  and the public interface is just one more layer that provides easy to use functions.

Also,  the functions have to be moved to their corresponding *Pools once those are ready,  so a dynamic function will require validation to check if a specific request makes sense.  Instead we just provide functions for the possible combinations,  which won't be too many for each pool.

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_chp_power_formula.py
sahas-subramanian-frequenz: nope. The component graph returns a set,  but I guess you're right, there's some unnecessary complexity here,  I will remove the intermediate list,  and directly iterate over the generator instead.

src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool.py
sahas-subramanian-frequenz: fixed.

src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool.py
sahas-subramanian-frequenz: fixed.

src/frequenz/sdk/microgrid/_graph.py
sahas-subramanian-frequenz: yes,  but the component graph can be generic, doesn't have to care about the tighter requirements of the formula generator (which might go away, especially if we encounter some complicated setups)

src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py
sahas-subramanian-frequenz: added.

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_chp_power_formula.py
sahas-subramanian-frequenz: done

src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py
```diff
@@ -170,13 +215,60 @@ def grid_current(self) -> FormulaEngine3Phase:
...
+        """Fetch the PV power in the microgrid.
+
+        If a formula engine to calculate PV power is not already running, it will be
+        started.
+
+        A receiver from the formula engine can be created using the `new_receiver`
+        method.
+
+        Returns:
+            A FormulaEngine that will calculate and stream PV power production.
```
cwasicki: `PV total power`

sahas-subramanian-frequenz: updated.

sahas-subramanian-frequenz:
Rebased on latest.

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_chp_power_formula.py
```diff
@@ -0,0 +1,105 @@
...
+                    " Expected exactly one."
+                )
+            meter = next(iter(predecessors))
+            if meter.category != ComponentCategory.METER:
+                raise FormulaGenerationError(
+                    f"CHP {chp.component_id} has a predecessor of category "
+                    f"{meter.category}. Expected ComponentCategory.METER."
+                )
+            meter_successors = component_graph.successors(meter.component_id)
+            if len(meter_successors) != 1:
```
matthias-wende-frequenz: Is that check necessary? Seems to me that we would disallow a valid grid topology, where one meter is measuring more then 1 CHPs?

src/frequenz/sdk/microgrid/_graph.py
```diff
@@ -113,6 +113,149 @@ def successors(self, component_id: int) -> Set[Component]:
...
+        Args:
+            component: component to check.
+
+        Returns:
+            Whether the specified component is a battery meter.
+        """
+
+    @abstractmethod
+    def is_battery_chain(self, component: Component) -> bool:
+        """Check if the specified component is part of a battery chain.
```
matthias-wende-frequenz, line 181: I suggest to use the docs to define what a battery chain is.

src/frequenz/sdk/microgrid/_graph.py
```diff
@@ -113,6 +113,149 @@ def successors(self, component_id: int) -> Set[Component]:
...
+        Args:
+            component: component to check.
+
+        Returns:
+            Whether the specified component is an EV charger meter.
+        """
+
+    @abstractmethod
+    def is_ev_charger_chain(self, component: Component) -> bool:
+        """Check if the specified component is part of an EV charger chain.
```
matthias-wende-frequenz, line 220: ditto

src/frequenz/sdk/microgrid/_graph.py
```diff
@@ -113,6 +113,149 @@ def successors(self, component_id: int) -> Set[Component]:
...
+        Args:
+            component: component to check.
+
+        Returns:
+            Whether the specified component is a CHP meter.
+        """
+
+    @abstractmethod
+    def is_chp_chain(self, component: Component) -> bool:
+        """Check if the specified component is part of a CHP chain.
```
matthias-wende-frequenz, line 258: ditto

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_consumer_power_formula.py
```diff
@@ -0,0 +1,121 @@
...
+
+from ....microgrid import connection_manager
+from ....microgrid.component import Component, ComponentCategory, ComponentMetricId
+from .._formula_engine import FormulaEngine
+from .._resampled_formula_builder import ResampledFormulaBuilder
+from ._formula_generator import ComponentNotFound, FormulaGenerator
+
+
+class ConsumerPowerFormula(FormulaGenerator):
+    """Formula generator from component graph for calculating the Consumer Power."""
```
matthias-wende-frequenz: I suggest defining here what is meant by Consumer power.

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_consumer_power_formula.py
```diff
@@ -0,0 +1,121 @@
...
+
+        if len(grid_successors) == 1:
+            grid_meter = next(iter(grid_successors))
+            assert (
+                grid_meter.category == ComponentCategory.METER
+            ), "Only grid successor is not a meter."
+            return self._gen_with_grid_meter(builder, grid_meter)
+        return self._gen_without_grid_meter(builder, grid_successors)
+
+    def _gen_with_grid_meter(
```
matthias-wende-frequenz, line 63: We need to be careful here since this stops working once we add sub panels (Unterverteilungen) into the microgrid api. Maybe it's worth adding a comment here.

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_consumer_power_formula.py
```diff
@@ -0,0 +1,121 @@
...
+
+        grid_successors = component_graph.successors(grid_component.component_id)
+        if not grid_successors:
+            raise ComponentNotFound("No components found in the component graph.")
+
+        if len(grid_successors) == 1:
+            grid_meter = next(iter(grid_successors))
+            assert (
+                grid_meter.category == ComponentCategory.METER
+            ), "Only grid successor is not a meter."
```
matthias-wende-frequenz: Wouldn't it be better to raise an exception here?

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_consumer_power_formula.py
```diff
@@ -0,0 +1,121 @@
...
+            return self._gen_with_grid_meter(builder, grid_meter)
+        return self._gen_without_grid_meter(builder, grid_successors)
+
+    def _gen_with_grid_meter(
+        self,
+        builder: ResampledFormulaBuilder,
+        grid_meter: Component,
+    ) -> FormulaEngine:
+        """Generate formula for calculating consumer power with grid-side meter.
```
matthias-wende-frequenz: We should define what a grid side meter is.

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_consumer_power_formula.py
```diff
@@ -0,0 +1,121 @@
...
+
+        Args:
+            builder: The formula engine builder.
+            grid_meter: The grid-side meter component.
+
+        Returns:
+            A formula engine that will calculate the consumer power.
+        """
+        component_graph = connection_manager.get().component_graph
```
matthias-wende-frequenz, line 77: Why aren't you passing the compenent_graph from the calling function.

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_consumer_power_formula.py
```diff
@@ -0,0 +1,121 @@
...
+        """Generate formula for calculating consumer power without grid-side meter.
+
+        Args:
+            builder: The formula engine builder.
+            grid_successors: The grid successors.
+
+        Returns:
+            A formula engine that will calculate the consumer power.
+        """
+        component_graph = connection_manager.get().component_graph
```
matthias-wende-frequenz, line 115: ditto

matthias-wende-frequenz (commented):
Almost finished. ~I've a couple of open question w.r.t. the consumer power.~ No questions. I missed the continue in the loop over the successors :).

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_consumer_power_formula.py
shsms: Are we going to introduce sub panels into the component graph?  I haven't heard of any such plans.   Those are passive components right?  i.e we can't read from them or control them.

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_consumer_power_formula.py
shsms: maybe just calling it the grid meter makes it more obvious?

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_chp_power_formula.py
shsms: true,  updated.

src/frequenz/sdk/microgrid/_graph.py
shsms: done

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_consumer_power_formula.py
shsms: done

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_consumer_power_formula.py
shsms: true, fixed.

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_consumer_power_formula.py
shsms: renamed to "grid meter" in the docstrings.

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_consumer_power_formula.py
shsms: because then I'd have to import `ComponentGraph` as the argument type,  which I didn't want to do.  This is just a reference,  and is required only once when creating formulas,  so there's no performance cost.

I think it is clearer to a reference from the source every time, rather than passing it around.

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_consumer_power_formula.py
matthias-wende-frequenz: As far as I remember that was the plan. @tiyash-basu-frequenz can you confirm?

matthias-wende-frequenz: I suggest to add a short warning comment to the documentation such that we are aware in the future.

matthias-wende-frequenz (dismissed):
Approved apart from my last comment.

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_consumer_power_formula.py
shsms: I have added a comment: 

If the component graph supports additional types of grid successors in the future, additional checks need to be added here.

tiyash-basu-frequenz: I think the term "sub panels" is misleading. If I have understood it correctly, then it is a section of the microgrid behind a dedicated fuse. I'd call it sub-section.

But I'd suggest ignoring it for now. It is not scheduled to be added to the microgrid API soon. To me even a comment seems unnecessary: because it does not improve the readability and just hints at a future functionality without knowing when we will add it. But up to you.

tiyash-basu-frequenz: If you really want to track this feature, then creating an issue is the proper way to do it.

matthias-wende-frequenz: You are right w.r.t. the issue. We should create one but I'd also leave a comment just to have the behavior documented because it tells that the algorithm isn't going into the depth.

matthias-wende-frequenz (approved):

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_consumer_power_formula.py
tiyash-basu-frequenz: Then I'd suggest to rather create the issue and add a link to the issue in the comment.

I'd suggest doing it in moderation, if never at all, because you do not want to end up adding every proposed feature in the comments.

PR #364, opened 2023-04-26
Validate docstring examples

### LINKED ISSUE #81: Fix broken examples in docstrings
### What happened?

Some examples in the docstrings, like the example for `frequenz.sdk.actor.power_distributing.PowerDistributingActor`, are broken because they are not automatically tested.

The examples should work.

### Affected version(s)

v0.12.0

### Affected part(s)

Documentation (part:docs)

### Extra information

At some point we need to start testing docstrings. There are projects to do that, like:

* https://docs.python.org/3/library/doctest.html

* https://docs.pytest.org/en/7.1.x/how-to/doctest.html

* https://stackoverflow.com/questions/23250068/python-doctest-my-github-markdown-files
  * + https://pypi.org/project/phmdoctest/
  * - https://github.com/ryneeverett/mkcodes

### LINKED ISSUE #253: Remove outdated examples from the `PowerDistributingActor`
### What happened?

The example provided in the comments for the `PowerDistributingActor` is rather outdated and needs updating.

Additionally some general explanation how it works in the first place and what a user can do with it would be helpful.

.

### Affected version(s)

_No response_

### Affected part(s)

Documentation (part:docs)

### Extra information

.

mathias-baumann-frequenz:
- Update stand-alone scripts to be import-friendly
- Add module to validate ``` python wrapped code examples in docstrings
- Fix docstring code example problems
- Patch @actor decorator to retain the origins class' doc, name and module metadata

fixes #81


noxfile.py
```diff
@@ -247,6 +247,9 @@ def docstrings(session: nox.Session, install_deps: bool = True) -> None:
         *darglint_paths,
     )
 
+    # Run the docstring examples validation script on each source file
+    session.run("python", "tests/utils/validate_docstring_examples.py", "src")
```
leandro-lucarella-frequenz: For some reason GitHub commenting on file is not working for me today so I'll comment here. I wouldn't put the script so deep in the `tests` dir, maybe in a top level `scripts/` or in `test/scripts/` to make it more obvious this is a script/program and not something you are supposed to import in tests?

tests/utils/validate_docstring_examples.py
```diff
@@ -0,0 +1,212 @@
...
+        elif inspect.isclass(obj) and obj.__module__ == module.__name__:
+            if obj.__doc__ is not None:
+                docs.append(obj.__doc__)
+            for _, method_obj in inspect.getmembers(obj):
+                if (
+                    inspect.isfunction(method_obj)
+                    and method_obj.__module__ == module.__name__
+                ):
+                    if method_obj.__doc__ is not None:
+                        docs.append(method_obj.__doc__)
```
leandro-lucarella-frequenz: I guess this doesn't catch class attributes or class attributes? Does it catch `properties`? It also doesn't catch module-level variables.

Any reason not to basically look for `__doc__` in any symbol (as long as the symbol is originated in this module)?

tests/utils/validate_docstring_examples.py
```diff
@@ -0,0 +1,212 @@
...
+
+    Args:
+        module: The module to extract code examples from.
+
+    Returns:
+        A list of code examples.
+    """
+    code_blocks = []
+    for doc in get_module_docs(module):
+        code_blocks += re.findall(r"``` ?python(.*?)```", doc, re.DOTALL)
```
leandro-lucarella-frequenz: Probably `r"```\s*python(.*?)```"` (if `\s` doesn't include `\n`, I guess it doesn't, does it?).

Also, I think this regex might be a bit naive, I guess there could be a `\`\`\`` in some code block somehow? Not sure, but at some point I think we should use a markdown parser instead. I'm totally OK to start with the naive approach in this first try though, just a side comment.

tests/utils/validate_docstring_examples.py
```diff
@@ -0,0 +1,212 @@
...
+
+    return code_blocks
+
+
+# pylint: disable=too-many-locals
+def run_code_examples(
```
leandro-lucarella-frequenz: Also in the future I would just write `pytest` files, so we can use the full power of pytest when running tests (no need to implement all the options here, like `fail_fast` or `verbose`) and we could also [mark tests](https://docs.pytest.org/en/7.1.x/how-to/mark.html#registering-marks) like "skip", or "release" to run some tests on releases only.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -67,17 +70,20 @@ class MovingWindow:
...
 
     # ... wait for 5 minutes until the buffer is filled
-    await asyncio.sleep(5)
-
-    # return an numpy array from the window
-    a = window[time_start:time_end]
-    # and use it to for example calculate the mean
-    mean = a.mean()
-    '''
+    await asyncio.sleep(0.1)
```
leandro-lucarella-frequenz: Maybe 0.1 is too little?

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -67,17 +70,20 @@ class MovingWindow:
...
+        # and use it to for example calculate the mean
+        mean = a.mean()
+    except IndexError as e:
+        pass
+    ```
 
     **Example2** (create a polars data frame from a `MovingWindow`):
 
-    ```
+    ``` disabled-no-polars-installed
```
leandro-lucarella-frequenz: Is this something you are interpreting yourself? If so, a few comments:

1. It would be interesting to add a *directive* to `pip install` extra dependencies for tests.
2. We should probably still make the code block a normal python code block, so it is rendered properly when the documentation is rendered
3. We should probably implement the *directives* as markdown comments do they don't show on the rendered docs, or even better, look for some markdown format that would render nicely, as it is probably important info for the people reading the docs that they need to install some extra dependency to try this example. Maybe it can be a special comment in the python example, like:

   ```python
   # example-pip-dependencies: polars
   import polars as pl
   ```

   Then the runner will do a `pip install polars` before running the test.

src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -61,7 +61,7 @@
 
 ``` python
 async some_sink(Sample) -> None:
-    ...
+    pass
```
leandro-lucarella-frequenz, line 64: Did this fail somehow? Even when IMHO `pass` is more semantically correct here, AFAIK `...` should work too, is just an expression (just asking out of curiosity).

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
```diff
@@ -174,11 +174,13 @@ async def apply(self) -> Sample:
     "HigherOrderFormulaBuilder3Phase",
 ]
 
+# pylint: disable=invalid-name
```
leandro-lucarella-frequenz: Why is is an invalid name? Does it say? This looks strange. Please add a comment explaining both in the code and the commit message.

benchmarks/timeseries/benchmark_datasourcing.py
```diff
@@ -31,10 +31,10 @@
     from tests.utils import MockMicrogridClient
 except ImportError:
     print(
-        "Error: Unable to import the `tests` package. "
+        "\nError: Unable to import the `tests` package. "
```
leandro-lucarella-frequenz: Why the extra leading `\n` here? Looks weird.

leandro-lucarella-frequenz (commented):
I have a few comments, but amazing if now the examples are being tested and all really fixed!

If you really did a research and there are no tools available doing this, I'm thinking we might just create a repo for the tool, it looks like something that could be reused by other people and that the ecosystem is missing. But I'm also fine to add it here and split it in the future.

Also, please remove the "fixes #..." because we are still missing the testing of `>>>` blocks, right? Or split the issue in 2.

tests/utils/validate_docstring_examples.py
mathias-baumann-frequenz: From what I learned only a few symbols have the __doc__ property, namely modules, functions, methods and classes.

Well, that's not exactly correct, vars have the property, but it contains type documentation rather than user documentation.

E.g.

```
>>> x = 3
>>> dir(x)
['__abs__', '__add__', '__and__', '__bool__', '__ceil__', '__class__', '__delattr__', '__dir__', '__divmod__', '__doc__', '__eq__', '__float__', '__floor__', '__floordiv__', '__format__', '__ge__', '__getattribute__', '__getnewargs__', '__gt__', '__hash__', '__index__', '__init__', '__init_subclass__', '__int__', '__invert__', '__le__', '__lshift__', '__lt__', '__mod__', '__mul__', '__ne__', '__neg__', '__new__', '__or__', '__pos__', '__pow__', '__radd__', '__rand__', '__rdivmod__', '__reduce__', '__reduce_ex__', '__repr__', '__rfloordiv__', '__rlshift__', '__rmod__', '__rmul__', '__ror__', '__round__', '__rpow__', '__rrshift__', '__rshift__', '__rsub__', '__rtruediv__', '__rxor__', '__setattr__', '__sizeof__', '__str__', '__sub__', '__subclasshook__', '__truediv__', '__trunc__', '__xor__', 'as_integer_ratio', 'bit_count', 'bit_length', 'conjugate', 'denominator', 'from_bytes', 'imag', 'numerator', 'real', 'to_bytes']
>>> x.__doc__
"int([x]) -> integer\nint(x, base=10) -> integer\n\nConvert a number or string to an integer, or return 0 if no arguments\nare given.  If x is a number, return x.__int__().  For floating point\nnumbers, this truncates towards zero.\n\nIf x is not a number or if base is given, then x must be a string,\nbytes, or bytearray instance representing an integer literal in the\ngiven base.  The literal can be preceded by '+' or '-' and be surrounded\nby whitespace.  The base defaults to 10.  Valid bases are 0 and 2-36.\nBase 0 means to interpret the base from the string as an integer literal.\n>>> int('0b100', base=0)\n4"
```

tests/utils/validate_docstring_examples.py
mathias-baumann-frequenz: I did try to use it first but eventually didn't explore that branch fully. I can have a look again

src/frequenz/sdk/timeseries/_resampling.py
mathias-baumann-frequenz: I think I was just overly cautious with my fixes, it probably was fine before

benchmarks/timeseries/benchmark_datasourcing.py
mathias-baumann-frequenz: It just looked a lot nicer in my validate script output :innocent: but eventually figured we don't need to check for docu-wrapped examples in `benchmarks` or `examples` so I can just remove it again, too

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
mathias-baumann-frequenz: I was also really surprised by this but all I can tell you is that pylint complained suddenly. Maybe after I updated to the latest state?

src/frequenz/sdk/timeseries/_moving_window.py
mathias-baumann-frequenz: Nah, it's actually not interpreted at all, it just doesn't match the regex and thus is ignored. I thought about installing deps like that but didn't wanna expand the required installed dependencies for tests at this point.

I did think about directives as well, yes.

src/frequenz/sdk/timeseries/_moving_window.py
mathias-baumann-frequenz: I mean, probably, but it does cause actual wait times when running it. If we want to put realistic values here then I would try to make it use fake-event loops

tests/utils/validate_docstring_examples.py
leandro-lucarella-frequenz: Ah, interesting. Well, the docs extracting tool ([mkdocstrings](https://mkdocstrings.github.io/)) is considering strings following a variable a docstring, I guess it does it by parsing the Python file then. See for example:

* https://frequenz-floss.github.io/frequenz-sdk-python/v0.20/reference/frequenz/sdk/timeseries/#frequenz.sdk.timeseries.UNIX_EPOCH
* https://github.com/frequenz-floss/frequenz-sdk-python/blob/d3528e9670138fe9a292cc07495c2da7c9ea0d00/src/frequenz/sdk/timeseries/_base_types.py#L13-L14

I guess for the first version it is OK to ignore them, but we should support it in the future.

This also makes me think that testing the documentation should probably be part of building the documentation (or an option), so maybe in the future we could plug it to mkdocstring (making it a plugin or something). This will also ensure that the documentation we get to extract examples is the same as the one that will be rendered. The guy doing mkdocstrings seems very nice, he even replied to some issue in our repo when discussing about it.

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
leandro-lucarella-frequenz: Would be interesting to see the `pylint` message to see exactly what it is complaining about.

src/conftest.py
```diff
@@ -0,0 +1,154 @@
...
+    # Remove 'src/' prefix if present
+    if path.startswith("src/"):
+        path = path[4:]
+
+    # Remove the '.py' extension and replace '/' with '.'
+    module_path = path[:-3].replace("/", ".")
+
+    # Create the import statement
+    import_statement = f"from {module_path} import *"
+    return import_statement
```
leandro-lucarella-frequenz, line 62: Not to change now but more FYI, I would recommend using [`pathlib`](https://docs.python.org/3/library/pathlib.html) to manipulate paths, it should be easier and more multi-platform (in case we want to extend the platforms we support).

src/conftest.py
```diff
@@ -0,0 +1,154 @@
...
+    """Code block parser that validates extracted code examples using pylint.
+
+    This parser is a modified version of the default Python code block parser
+    from the Sybil library.
+    It uses pylint to validate the extracted code examples.
+
+    All code examples are preceded by the original file's import statements as
+    well as an wildcard import of the file itself.
+    This allows us to use the code examples as if they were part of the original
+    file.
```
leandro-lucarella-frequenz, line 75: :100:

src/conftest.py
```diff
@@ -0,0 +1,154 @@
...
+
+    Additionally, the code example is padded with empty lines to make sure the
+    line numbers are correct.
```
leandro-lucarella-frequenz, line 78: :boom:

src/conftest.py
```diff
@@ -0,0 +1,154 @@
...
+
+    Pylint warnings which are unimportant for code examples are disabled.
```
leandro-lucarella-frequenz, line 80: It would be good to be able to configure this in the future.

src/conftest.py
```diff
@@ -0,0 +1,154 @@
...
+                "C0114",  # Missing module docstring
+                "C0115",  # Missing class docstring
+                "C0116",  # Missing function or method docstring
+                "W0401",  # Wildcard import
+                "W0404",  # Reimport
+                "W0611",  # Unused import
+                "W0612",  # Unused variable
+                "W0614",  # Unused import from wildcard
+                "E0611",  # No name in module
+                "E1142",  # Await used outside async function
```
leandro-lucarella-frequenz, line 124: Can't you use the string codes instead, that are already much more self descriptive? Like `missing-module-docstring` instead of `C0114`.

pyproject.toml
```diff
@@ -75,6 +75,7 @@ pytest = [
     "pytest-asyncio == 0.21.0",
     "time-machine == 2.9.0",
     "async-solipsism == 0.5",
+    "sybil ~= 5.0.0",
```
leandro-lucarella-frequenz, line 78: Please pin the dependency to a exact version, dependabot will let us know when it's updated. This way we make sure we are always able to "rebuild"/"reproduce" old versions.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -78,7 +81,11 @@ class MovingWindow:
     Example: Create a polars data frame from a `MovingWindow`
 
         ```python
+        # pylint: disable=import-error
```
leandro-lucarella-frequenz, line 84: We need to find a way to remove this from examples, we shouldn't transfer hacks to examples. For me a comment like `# pip install polars` (and then the script will run that line) is fine, because it is also useful documentation for the user. An alternative to not needing to add support to the test script for now would be not support using external libraries (i.e. not use polars in this example, what do you think @matthias-wende-frequenz ?)

leandro-lucarella-frequenz (commented):
So, this is amazing. I love the idea of making examples run in the same *import context* as the file they live in!

What I don't like at all is that this is not really a `pytest` plugin, it is just hijacking the configuration file!

There is zero reason to run this in the context of `pytest`, right? You are not using any `pytest` facilities or features, right?

If so, just create a separate file. You don't even have to make it executable for now, we can just call it using `python -m lint-code-examples` or whatever.

I think this should be definitely a separate tool, so IMHO the next step would be to move it into its own repo.

tests/utils/validate_docstring_examples.py
leandro-lucarella-frequenz: How is it with sybil? I guess it also not considering strings after a variable declaration a docstring for the variable, right?

src/frequenz/sdk/timeseries/_moving_window.py
Marenz: but would we really always install a dependency to run an example in the CI runs?

Marenz:
> There is zero reason to run this in the context of pytest, right?

I liked that it re-uses the existing interface and error reporting

src/frequenz/sdk/timeseries/_moving_window.py
leandro-lucarella-frequenz: To do it proper we should create a separate venv for running each example. I would eventually do it (and run this only on pushes, or nightly, or before releases, depending on how heavy it is, as said in #378), but definitely not for now. I guess the simplest solution if this is the only instance of needing an external library for an example is to just avoid using polars in this example. We can probably demonstrate using `numpy` too, right @matthias-wende-frequenz ?

leandro-lucarella-frequenz:
> I liked that it re-uses the existing interface and error reporting

Oh, really? Can you show an example output? I thought the output was just whatever `pylint` outputs. In any case IMHO it definitely can't be in `conftest.py`, maybe if it is a real `pytest` plugin. But still is hard for me to see an advantage over just producing one file per example and then run `pylint` (and other linting tools) there, that will also use the normal output of the tools. We can even reuse the option and all the noxfile paraphernalia if we just write examples in a new directory like `examples/.generated/` or something like that.

Marenz:
```
❰marenz❙~/frequenz/frequenz-sdk-python(git:validate_docstring_examples)❱✘≻ pytest src
==================================================================== test session starts =====================================================================
platform linux -- Python 3.11.2, pytest-7.2.2, pluggy-1.0.0
rootdir: /home/marenz/frequenz/frequenz-sdk-python, configfile: pyproject.toml
plugins: anyio-3.6.2, mock-3.10.0, asyncio-0.21.0, time-machine-2.9.0
asyncio: mode=Mode.AUTO
collected 17 items

src/frequenz/sdk/actor/_decorator.py ..                                                                                                                [ 11%]
src/frequenz/sdk/actor/power_distributing/power_distributing.py .                                                                                      [ 17%]
src/frequenz/sdk/power/_distribution_algorithm.py ..........                                                                                           [ 76%]
src/frequenz/sdk/timeseries/_moving_window.py ..                                                                                                       [ 88%]
src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py .                                                                                       [ 94%]
src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py .                                                                                          [100%]

```

Marenz:
And here an example with a failure:

```
❰marenz❙~/frequenz/frequenz-sdk-python(git:validate_docstring_examples)❱✘≻ pytest src
==================================================================== test session starts =====================================================================
platform linux -- Python 3.11.2, pytest-7.2.2, pluggy-1.0.0
rootdir: /home/marenz/frequenz/frequenz-sdk-python, configfile: pyproject.toml
plugins: anyio-3.6.2, mock-3.10.0, asyncio-0.21.0, time-machine-2.9.0
asyncio: mode=Mode.AUTO
collected 17 items

src/frequenz/sdk/actor/_decorator.py .F                                                                                                                [ 11%]
src/frequenz/sdk/actor/power_distributing/power_distributing.py .                                                                                      [ 17%]
src/frequenz/sdk/power/_distribution_algorithm.py ..........                                                                                           [ 76%]
src/frequenz/sdk/timeseries/_moving_window.py ..                                                                                                       [ 88%]
src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py .                                                                                       [ 94%]
src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py .                                                                                          [100%]

========================================================================== FAILURES ==========================================================================
______________________________________________________________ _decorator.py line=127 column=1 _______________________________________________________________

Example at /home/marenz/frequenz/frequenz-sdk-python/src/frequenz/sdk/actor/_decorator.py, line 127, column 1 did not evaluate as expected:
Pylint validation failed for code example:
import asyncio
import inspect
import logging
from typing import Any, Generic, Optional, Type, TypeVar
from frequenz.sdk._internal._asyncio import cancel_and_await
from frequenz.sdk.actor._decorator import *

from frequenz.channels import Receiver, Sender, Broadcast
from frequenz.channels.util import Select
@actor
class Actor1:
    def __init__(
        self,
        name: str,
        recv: Receiver[bool],
        output: Sender[bool],
    ) -> None:
        self.name = name
        self._recv = recv
        self._output = output
    async def run(self) -> None:
        async for msg in self._recv:
            await self._output.send(msg)
@actor
class Actor2:
    def __init__(
        self,
        name: str,
        recv: Receiver[bool],
        output: Sender[bool],
    ) -> None:
        self.name = name
        self._recv = recv
        self._output = output
    async def run(self) -> None:
        async for msg in self._recv:
            await self._output.send(msg)
async def main() -> None:
    input_chan: Broadcast[bool] = Broadcast("Input to A1")
    a1_chan: Broadcast[bool] = Broadcast("A1 stream")
    a2_chan: Broadcast[bool] = Broadcast("A2 stream")
    a_1 = Actor1(
        name="ActorOne",
        recv=input_chan.new_receiver(),
        output=a1_chan.new_sender(),
    )
    a_2 = Actor2(
        name="ActorTwo",
        recv=a1_chan.new_receiver(),
        output=a2_chan.new_sender(),
    )
    a2_rx = a2_chan.new_receiver()
    await input_chan.new_seder().send(True)
    msg = await a2_rx.receive()
asyncio.run(main())

Error: Command '['pylint', '--disable', 'missing-module-docstring,missing-class-docstring,missing-function-docstring,wildcard-import,reimported,unused-import,unused-variable,unused-wildcard-import,no-name-in-module,await-outside-async', '--from-stdin', '/home/marenz/frequenz/frequenz-sdk-python/src/frequenz/sdk/actor/_decorator.py']' returned non-zero exit status 2.
Output: ************* Module sdk.actor._decorator
src/frequenz/sdk/actor/_decorator.py:174:10: E1101: Instance of 'Broadcast' has no 'new_seder' member; maybe 'new_sender'? (no-member)

-------------------------------------------------------------------
Your code has been rated at 8.53/10 (previous run: 10.00/10, -1.47)


_decorator.py:127: SybilFailure
================================================================== short test summary info ===================================================================
FAILED src/frequenz/sdk/actor/_decorator.py::line:127,column:1
=============================================================== 1 failed, 16 passed in 21.83s ================================================================
```

Marenz:
I was kind of avoiding to generate actual files and tried to keep it all in memory, most for performance reasons.

Marenz:
Recreated under my new username at https://github.com/frequenz-floss/frequenz-sdk-python/pull/384

tests/utils/validate_docstring_examples.py
leandro-lucarella-frequenz: @Marenz ping, not sure if you could look into this. Is not super high prio, because I think it is much less likely to have examples here, but still it would be good to know.  If it's not supported we can make an extra effort to try to avoid examples in these cases.

Marenz: I believe it is supported, as Sybil doesn't use the `__doc__` property but parses the files itself

leandro-lucarella-frequenz: Can you confirm this?  I think it will be important to decide if we allow writing examples in variable documentation or not.

PR #368, opened 2023-04-28
Fix and improve `MovingWindow` examples in docs

leandro-lucarella-frequenz:
- Fix example formatting
- Add proper python markers to examples
- Improve examples formatting

### DIFF
--- a/src/frequenz/sdk/timeseries/_moving_window.py
+++ b/src/frequenz/sdk/timeseries/_moving_window.py
@@ -54,48 +54,48 @@ class MovingWindow:
     If resampling is not required, the resampler config parameter can be
     set to None in which case the MovingWindow will not perform any resampling.
 
-    **Example1** (calculating the mean of a time interval):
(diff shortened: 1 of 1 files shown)


leandro-lucarella-frequenz:
The fix is for this:

image

And this is how it looks after the improvements in formatting:

image

mathias-baumann-frequenz (approved):

matthias-wende-frequenz (approved):

leandro-lucarella-frequenz:
Ah, one catch, the final output doesn't say "Example: ..." but just "..." because then it renders with a custom title it doesn't add the "Example: " too, just the :test_tube: icon. I tried putting `Example: Example: ...` in the docstring but it looked really weird and like an error, so I guess it is likely that someone will "fix it" at some point, so I was more comfortable with this balance between nice rendered docs and docs in the source that don't look completely weird.

leandro-lucarella-frequenz:
So this is how the output with this PR looks like:

image

PR #373, opened 2023-05-03
Mock Resampler in formula tests

### LINKED ISSUE #249: Some Formula tests are flaky

sahas-subramanian-frequenz:
This allows most of the formula tests to be reliable.

Only the test for the `EVChargerPool.component_data` method still depends on a lower level mock, and it needs it because it combines resampled data with component state values.  If it is flaky, we'll have to add multi stage testing - with a fast test first, and a slower one if that fails, etc.

### DIFF
--- a/src/frequenz/sdk/timeseries/_formula_engine/_formula_engine_pool.py
+++ b/src/frequenz/sdk/timeseries/_formula_engine/_formula_engine_pool.py
@@ -110,3 +110,8 @@ class FormulaEnginePool:
         ).generate()
         self._engines[channel_key] = engine
         return engine
+
+    async def stop(self) -> None:
+        """Stop all formula engines in the pool."""
+        for engine in self._engines.values():
+            await engine._stop()  # pylint: disable=protected-access
--- a/src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_pv_power_formula.py
+++ b/src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_pv_power_formula.py
@@ -3,12 +3,20 @@
 
 """Formula generator for PV Power, from the component graph."""
 
+from __future__ import annotations
+
 import logging
+from collections import abc
 
 from ....microgrid import connection_manager
 from ....microgrid.component import ComponentCategory, ComponentMetricId, InverterType
 from .._formula_engine import FormulaEngine
-from ._formula_generator import NON_EXISTING_COMPONENT_ID, FormulaGenerator, FormulaType
+from ._formula_generator import (
+    NON_EXISTING_COMPONENT_ID,
+    FormulaGenerationError,
+    FormulaGenerator,
+    FormulaType,
+)
 
 _logger = logging.getLogger(__name__)
 
@@ -27,15 +35,8 @@ class PVPowerFormula(FormulaGenerator):
         """
         builder = self._get_builder("pv-power", ComponentMetricId.ACTIVE_POWER)
 
-        component_graph = connection_manager.get().component_graph
-        pv_inverters = list(
-            comp
-            for comp in component_graph.components()
-            if comp.category == ComponentCategory.INVERTER
-            and comp.type == InverterType.SOLAR
-        )
-
-        if not pv_inverters:
+        pv_meters = self._get_pv_meters()
+        if not pv_meters:
             _logger.warning(
                 "Unable to find any PV inverters in the component graph. "
                 "Subscribing to the resampling actor with a non-existing "
@@ -51,11 +52,11 @@ class PVPowerFormula(FormulaGenerator):
 
         builder.push_oper("(")
         builder.push_oper("(")
-        for idx, comp in enumerate(pv_inverters):
+        for idx, comp_id in enumerate(pv_meters):
             if idx > 0:
                 builder.push_oper("+")
 
-            builder.push_component_metric(comp.component_id, nones_are_zeros=True)
+            builder.push_component_metric(comp_id, nones_are_zeros=True)
         builder.push_oper(")")
         if self._config.formula_type == FormulaType.PRODUCTION:
             builder.push_oper("*")
@@ -66,3 +67,40 @@ class PVPowerFormula(FormulaGenerator):
             builder.push_clipper(0.0, None)
 
         return builder.build()
+
+    def _get_pv_meters(self) -> abc.Set[int]:
+        component_graph = connection_manager.get().component_graph
+
+        pv_inverters = list(
+            comp
+            for comp in component_graph.components()
+            if comp.category == ComponentCategory.INVERTER
+            and comp.type == InverterType.SOLAR
+        )
+        pv_meters: set[int] = set()
+
+        if not pv_inverters:
+            return pv_meters
+
+        for pv_inverter in pv_inverters:
+            predecessors = component_graph.predecessors(pv_inverter.component_id)
+            if len(predecessors) != 1:
+                raise FormulaGenerationError(
+                    "Expected exactly one predecessor for PV inverter "
+                    f"{pv_inverter.component_id}, but found {len(predecessors)}."
+                )
+            meter = next(iter(predecessors))
+            if meter.category != ComponentCategory.METER:
+                raise FormulaGenerationError(
+                    f"Expected predecessor of PV inverter {pv_inverter.component_id} "
+                    f"to be a meter, but found {meter.category}."
+                )
+            meter_successors = component_graph.successors(meter.component_id)
+            if len(meter_successors) != 1:
+                raise FormulaGenerationError(
+                    f"Expected exactly one successor for meter {meter.component_id}"
+                    f", connected to PV inverter {pv_inverter.component_id}"
+                    f", but found {len(meter_successors)}."
+                )
+            pv_meters.add(meter.component_id)
+        return pv_meters
--- a/tests/timeseries/_formula_engine/test_formula_composition.py
+++ b/tests/timeseries/_formula_engine/test_formula_composition.py
@@ -4,7 +4,7 @@
 """Tests for formula composition."""
 
 
-from math import isclose
+import math
 
 from pytest_mock import MockerFixture
 
@@ -12,7 +12,7 @@ from frequenz.sdk import microgrid
 from frequenz.sdk.microgrid.component import ComponentMetricId
 
 from ..mock_microgrid import MockMicrogrid
-from .utils import get_resampled_stream, synchronize_receivers
+from .utils import get_resampled_stream
 
 
 class TestFormulaComposition:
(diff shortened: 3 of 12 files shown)


src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_pv_power_formula.py
```diff
@@ -3,12 +3,20 @@
 
 """Formula generator for PV Power, from the component graph."""
```
daniel-zullo-frequenz, line 4: > Fetch PV power from PV meters, rather than PV inverters

Can you please clarify/explain the reason of this change in the commit message?

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_pv_power_formula.py
```diff
@@ -66,3 +67,39 @@ def generate(self) -> FormulaEngine:
             builder.push_clipper(0.0, None)
 
         return builder.build()
+
+    def _get_pv_meters(self) -> abc.Set[int] | None:
```
daniel-zullo-frequenz: Documentation is missing here.

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_pv_power_formula.py
```diff
@@ -66,3 +67,39 @@ def generate(self) -> FormulaEngine:
...
+        pv_inverters = list(
+            comp
+            for comp in component_graph.components()
+            if comp.category == ComponentCategory.INVERTER
+            and comp.type == InverterType.SOLAR
+        )
+        if not pv_inverters:
+            return None
+
+        pv_meters: set[int] = set()
```
daniel-zullo-frequenz: I wonder if it'd be better to return this empty `set` rather than `None` when there is no PV inverters. If this comment applies, changes in the function definition will be needed.

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine_pool.py
```diff
@@ -110,3 +110,8 @@ def from_generator(
         ).generate()
```
daniel-zullo-frequenz, line 110: I found the commit `Add methods for stopping running formula engines` is misleading because the changes in the commit are not only adding methods for stopping running formula engines but also updating existing ones.
So you might want to clarify the motivation or reason to update the existing methods or split those changes into a separate commit

tests/timeseries/mock_datapipeline.py
```diff
@@ -0,0 +1,191 @@
...
+            assert recv is not None
+            self._forward_tasks[request.get_channel_name()] = asyncio.create_task(
+                self._channel_forward_messages(
+                    recv,
+                    self._channel_registry.new_sender(request.get_channel_name()),
+                )
+            )
+
+    async def send_meter_power(self, values: list[float | None]) -> None:
+        """Send meter power."""
```
daniel-zullo-frequenz: Probably is not worth it to add trivial documentation like this so I'm tempted to suggest to remove this kind of method documentation as there are other methods in the same file that don't add any documentation. If we keep it, then we should add documentation for the methods with missing docs

daniel-zullo-frequenz (commented):
First review round done and I have a few suggestions to check for

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_pv_power_formula.py
shsms: It is a private function.  we don't need to document them unless they are not obvious.

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_pv_power_formula.py
shsms: done

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_pv_power_formula.py
shsms: true,  fixed.

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine_pool.py
shsms: The updates to the other stop functions were just to call the formula engine stop functions, if they own formula engine instances.  So that when we call `stop` of a logical meter for example,  that calls the stop of the formula engines it owns as well.

I've updated the commit message.

tests/timeseries/mock_datapipeline.py
shsms: I guess I get lazy when writing tests.  I've improved the docs for these functions.

The docs can't be removed because they're public functions in tests.  Public test functions need to have a description, but not descriptions for args and return values.

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_pv_power_formula.py
daniel-zullo-frequenz: ok, I didn't know we can omit documentation for trivial internal methods

tests/timeseries/_formula_engine/test_formula_composition.py
```diff
@@ -41,33 +42,56 @@ async def test_formula_composition(  # pylint: disable=too-many-locals
...
+        )  # -100 + -200
+        assert (
+            inv_calc_pow is not None
+            and inv_calc_pow.value is not None
+            and math.isclose(inv_calc_pow.value, -264.0)  # -300 + 36
+        )
+        assert (
+            main_pow is not None
+            and main_pow.value is not None
+            and math.isclose(main_pow.value, 100.0)
```
daniel-zullo-frequenz, line 79: Alternatively we can use [pytest.approx](https://docs.pytest.org/en/7.1.x/reference/reference.html#id8) in tests. This is just an optional comment...

tests/timeseries/mock_microgrid.py
```diff
@@ -48,16 +49,28 @@ class MockMicrogrid:  # pylint: disable=too-many-instance-attributes
...
-        self, grid_side_meter: bool, num_values: int = 2000, sample_rate_s: float = 0.01
+    mock_data: MockDataPipeline
+
+    def __init__(  # pylint: disable=too-many-arguments
+        self,
+        grid_side_meter: bool,
+        api_client_streaming: bool = False,
+        num_values: int = 2000,
+        sample_rate_s: float = 0.01,
+        namespaces: int = 1,
```
daniel-zullo-frequenz: optionally I'd rename `namespaces` to `num_spaces` to better describe the variable and be consistent with other variables like `num_values`

tests/timeseries/mock_datapipeline.py
```diff
@@ -0,0 +1,191 @@
...
+    async def send_non_existing_component_value(self) -> None:
+        """Send a value for a non existing component."""
+        sample = Sample(self._next_ts, None)
+        await self._non_existing_component_sender.send(sample)
+
+    async def send_evc_current(self, values: list[list[float | None]]) -> None:
+        """Send the given values as resampler output for ev charger current."""
+        assert len(values) == len(self._ev_current_senders)
+        for chan, evc_values in zip(self._ev_current_senders, values):
+            assert len(evc_values) == 3
```
daniel-zullo-frequenz: Optionally set a variable like `num_phases = 3` and compare it here

tests/timeseries/mock_datapipeline.py
```diff
@@ -0,0 +1,191 @@
...
+
+    async def send_meter_power(self, values: list[float | None]) -> None:
+        """Send the given values as resampler output for meter power."""
+        assert len(values) == len(self._meter_power_senders)
+        for chan, value in zip(self._meter_power_senders, values):
+            sample = Sample(self._next_ts, value)
+            await chan.send(sample)
+
+    async def send_evc_power(self, values: list[float | None]) -> None:
+        """Send the given values as resampler output for ev charger power."""
```
daniel-zullo-frequenz: s/ev/EV

tests/timeseries/mock_datapipeline.py
```diff
@@ -0,0 +1,191 @@
...
+            sample = Sample(self._next_ts, value)
+            await chan.send(sample)
+
+    async def send_non_existing_component_value(self) -> None:
+        """Send a value for a non existing component."""
+        sample = Sample(self._next_ts, None)
+        await self._non_existing_component_sender.send(sample)
+
+    async def send_evc_current(self, values: list[list[float | None]]) -> None:
+        """Send the given values as resampler output for ev charger current."""
```
daniel-zullo-frequenz: s/ev/EV

daniel-zullo-frequenz (dismissed):
I have only optional comments that you might want to check for. LGTM otherwise

tests/timeseries/mock_microgrid.py
shsms: renamed to `num_namespaces`.

tests/timeseries/mock_datapipeline.py
shsms: It feels like this particular constant will lose some meaning if we give it a name.  I've added a comment instead for each occurrence instead.

tests/timeseries/mock_datapipeline.py
shsms: 🤦🏽 done

tests/timeseries/mock_datapipeline.py
shsms: done

daniel-zullo-frequenz (approved):

PR #381, opened 2023-05-08
Make interface for BatteryPool metrics consistent with power methods

sahas-subramanian-frequenz:
The `soc`, `capacity`, `power_bounds` methods of the `BatteryPool`
were directly returning receivers.  They've now been converted to
properties, from which `new_receiver()` has to be called, to get a
receiver.

New:

``` python
soc_recv = battery_pool.soc.new_receiver()
```

Old:

``` python
soc_recv = battery_pool.soc()
```

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -16,6 +16,14 @@ This release drops support for Python versions older than 3.11.
 
 * Now `frequenz.sdk.timeseries.Sample` uses a more sensible comparison.  Before this release `Sample`s were compared only based on the `timestamp`.  This was due to a limitation in Python versions earlier than 3.10.  Now that the minimum supported version is 3.11 this hack is not needed anymore and `Sample`s are compared using both `timestamp` and `value` as most people probably expects.
 
+* `BatteryPool` metric streaming interfaces have changed for `soc`, `capacity` and `power_bounds`:
+
+  ```python
+  soc_rx = battery_pool.soc()    # old
+
+  soc_rx = battery_pool.soc.new_receiver()    # new
+  ```
+
 ## New Features
 
 <!-- Here goes the main new features and examples or instructions on how to use them -->
--- a/examples/battery_pool.py
+++ b/examples/battery_pool.py
@@ -33,9 +33,9 @@ async def main() -> None:
 
     battery_pool = microgrid.battery_pool()
     receivers: Dict[str, Receiver[Any]] = {
-        "soc": await battery_pool.soc(maxsize=1),
-        "capacity": await battery_pool.capacity(maxsize=1),
-        "power_bounds": await battery_pool.power_bounds(maxsize=1),
+        "soc": battery_pool.soc.new_receiver(maxsize=1),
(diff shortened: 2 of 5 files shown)


Marenz (dismissed):

Marenz:
> Merging is blocked
> The base branch requires all commits to be signed.

shsms:
commits are now signed,  also added a release notes entry.

RELEASE_NOTES.md
```diff
@@ -12,6 +12,14 @@
 
 * The `LogicalMeter` no longer takes a `component_graph` parameter.
 
+* `BatteryPool` metric streaming interfaces have changed for `soc`, `capacity` and `power_bounds`:
+
+  ``` python
```
Marenz: syntax should be ```python 

And this makes me realize that we aren't testing the RELEASE_NOTES examples yet

Marenz: @leandro-lucarella-frequenz do we want to test that? If so this example would fail of course

leandro-lucarella-frequenz: Mmmm, not sure about RELEASE_NOTES, sometimes we want to use old syntax/features to show an upgrade path, I would leave it outside.

RELEASE_NOTES.md
```diff
@@ -12,6 +12,14 @@
...
+  soc_rx = battery_pool.soc()    # old
```
leandro-lucarella-frequenz, line 22: Hehe, like here :)

leandro-lucarella-frequenz (dismissed):

RELEASE_NOTES.md
Marenz: Well, that could be solved by either commenting the wold syntax/way or by having an extra ``` wrapped block that is not marked as python

shsms: removed the extra space.

shsms:
Rebased, resolved conflicts.

RELEASE_NOTES.md
leandro-lucarella-frequenz: I think that adds more noise than it helps. Examples in the RELEASE NOTES are a one-off thing, if you checked the time you write them, then they don't need to be kept up to date ever again. I really think that would make writing release notes more annoying and harder to read with very little gain.

examples/battery_pool.py
```diff
@@ -33,9 +33,9 @@ async def main() -> None:
 
     battery_pool = microgrid.battery_pool()
     receivers: Dict[str, Receiver[Any]] = {
-        "soc": await battery_pool.soc(maxsize=1),
-        "capacity": await battery_pool.capacity(maxsize=1),
-        "power_bounds": await battery_pool.power_bounds(maxsize=1),
+        "soc": battery_pool.soc.new_receiver(maxsize=1),
+        "capacity": battery_pool.capacity.new_receiver(maxsize=1),
+        "power_bounds": battery_pool.power_bounds.new_receiver(maxsize=1),
```
leandro-lucarella-frequenz, line 38: Is so nice not having an `await` here :heart:

leandro-lucarella-frequenz (approved):

PR #383, opened 2023-05-09
Remove r-prefix of documentation string

Marenz:
???


leandro-lucarella-frequenz:
This is just a *raw* string I think we can leave it as it is, as it's a very long a complex docstring.

https://docs.python.org/3/reference/lexical_analysis.html
> Both string and bytes literals may optionally be prefixed with a letter 'r' or 'R'; such strings are called raw strings and treat backslashes as literal characters. As a result, in string literals, '\U' and '\u' escapes in raw strings are not treated specially. Given that Python 2.x’s raw unicode literals behave differently than Python 3.x’s the 'ur' syntax is not supported.

PR #384, opened 2023-05-10
Validate docstring examples

### LINKED ISSUE #81: Fix broken examples in docstrings

Marenz:
- Add module to validate ``` python wrapped code examples in docstrings
- Make all examples validate correctly

fixes #81


Marenz:
@leandro-lucarella-frequenz I tried having the script in another file/dir than `src/conftest.py` but it seems to be more involved, so I left it for now.

src/conftest.py
```diff
@@ -0,0 +1,182 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Pytest plugin to validate and run docstring code examples.
+
+Code examples are often wrapped in triple backticks (```) within our docstrings.
```
leandro-lucarella-frequenz, line 6: I think this probably doesn't render correctly as markdown (i.e. with `mkdocs`). I know for now we are not rendering it, but if it's moved to its own repo, it probably will, so I rather have the comment being valid markdown from the start.

src/conftest.py
```diff
@@ -0,0 +1,182 @@
...
+        path: The path to convert.
+
+    Returns:
+        The import statement.
+
+    Raises:
+        ValueError: If the path does not point to a Python file.
+    """
+    # Create Path objects for the path and present working directory
+    path_obj = Path(path)
```
leandro-lucarella-frequenz: This complements the suggestion below of just using `pathlib.Path` as early as possible:

```suggestion
def path_to_import_statement(path_obj: pathlib.Path) -> str:
    """Convert a path to a Python file to an import statement.

    Args:
        path: The path to convert.

    Returns:
        The import statement.

    Raises:
        ValueError: If the path does not point to a Python file.
    """
```

(I would probably just use `path` instead of `path_obj` but wanted to minimize the diff :D )

src/conftest.py
```diff
@@ -0,0 +1,182 @@
...
+    pwd_obj = Path(os.getcwd())
```
leandro-lucarella-frequenz:
```suggestion
    pwd_obj = Path.cwd()
```

src/conftest.py
```diff
@@ -0,0 +1,182 @@
...
+    if relative_path.suffix != ".py":
+        raise ValueError("Path must point to a Python file (.py)")
+
+    # Remove 'src' prefix if present
+    parts = relative_path.parts
+    if parts[0] == "src":
+        parts = parts[1:]
+
+    # Remove the '.py' extension and join parts with '.'
+    module_path = ".".join(parts)[:-3]
```
leandro-lucarella-frequenz: Funny I had to implement almost the same for converting files to python modules/packages for repo-config:

https://github.com/frequenz-floss/frequenz-repo-config-python/blob/c3239e426e503822bd7d454c9d6621b639014c33/src/frequenz/repo/config/nox/util.py#L81-L110

src/frequenz/sdk/actor/_decorator.py
```diff
@@ -101,25 +103,30 @@ async def run(self) -> None:
...
+            output=echo_chan.new_sender(),
+        )
+        echo_rx = echo_chan.new_receiver()
+
+        await input_chan_2.new_sender().send(True)
+        msg = await echo_rx.receive()
 
-    await input_chan_2.new_sender().send(True)
-    msg = await echo_rx.receive()
+    asyncio.run(main())
```
leandro-lucarella-frequenz: I think in the previous version you did it. Was it too difficult to automatically wrap all examples in a `async def main()`? Or was more like a design choice that is better to have more complete examples?

src/frequenz/sdk/actor/power_distributing/power_distributing.py
```diff
@@ -89,11 +89,13 @@ class PowerDistributingActor:
     printed.
 
     Example:
-        ``` python
+        ```python
         import grpc.aio as grpcaio
 
-        from frequenz.sdk.microgrid.graph import _MicrogridComponentGraph
+        from frequenz.sdk.microgrid._graph import _MicrogridComponentGraph
```
leandro-lucarella-frequenz:
```suggestion
```

Now it seems to be taken from `graph = microgrid.connection_manager.get().component_graph` so the import shouldn't be necessary. Doesn't `pylint` complains about unused imports?

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -78,7 +81,11 @@ class MovingWindow:
     Example: Create a polars data frame from a `MovingWindow`
 
         ```python
+        # pylint: disable=import-error
         import polars as pl
```
leandro-lucarella-frequenz, line 85: There is still the question about what to do about this. @matthias-wende-frequenz do you think we can avoid using `polars` in the example? Or is it very important for `polars` to be used here? I don't think it is terrible to have a `pylint: disable=xxx` in an example, but if we can avoid it, much better.

src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py
```diff
@@ -30,7 +30,13 @@ class LogicalMeter:
     normal `Receiver`s, but can also be composed to form higher-order formula streams.
 
     Example:
-        ``` python
+        ```python
+        from frequenz.channels import Receiver, Sender, Broadcast
+        from frequenz.sdk.actor import DataSourcingActor, ComponentMetricsResamplingActor
+        from frequenz.sdk.timeseries._resampling import ResamplerConfig
```
leandro-lucarella-frequenz:
```suggestion
        from frequenz.sdk.timeseries import ResamplerConfig
```

We should avoid doing any internal `_xxx` imports in examples.

leandro-lucarella-frequenz (commented):
Just a few minor things.

src/frequenz/sdk/actor/_decorator.py
Marenz: The latter. I didn't want to add too much noise to the examples and as we no longer run/exec the examples but just lint them the requirements where more relaxed.

src/frequenz/sdk/actor/power_distributing/power_distributing.py
Marenz: >  Doesn't pylint complains about unused imports?

It would, but I disabled that as I wildcard import everything in the file as well as the original imports for the examples

Marenz: Though, thinking about this again, I suppose I could disable all those auito-imports with comments and keep it enabled in the "global" options

src/conftest.py
Marenz: Nope, seems fine: 
image

src/frequenz/sdk/actor/power_distributing/power_distributing.py
Marenz: I played around a bit and managed to enable a few more import warnings and thus fixed a few more examples.

src/conftest.py
leandro-lucarella-frequenz: Cool! Thanks for checking!

src/frequenz/sdk/actor/power_distributing/power_distributing.py
leandro-lucarella-frequenz: Great!

src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py
```diff
@@ -62,16 +74,15 @@ class LogicalMeter:
         )
 
         # Get a receiver for a builtin formula
-        grid_power_recv = logical_meter.grid_power()
+        grid_power_recv = logical_meter.grid_power
```
leandro-lucarella-frequenz: This is the type of error that `pylint` can't detect :(

`mypy` should though.

```suggestion
        grid_power_recv = logical_meter.grid_power.new_receiver()
```

leandro-lucarella-frequenz (commented):
Found one error in the examples undetected by `pylint`.

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: If we remove polars we can remove the whole example. So the discussion is more about do we need the example and my feeling is, that removing an example to make a tool work is not the best practice.

leandro-lucarella-frequenz: OK, I see, although the discussion is not if we need the example or not. I think we all agree we want the example, I just thought maybe it was possible to do some example/demonstration just using `numpy` that was equivalent (since we already have it as a dependency). This is just exposing my ignorance about pandas/numpy if I'm not making any sense, but I didn't wanted to remove examples to adapt to the tool :)

In any case I think we need to strike a balance, because also adding obscure `# pylint: disable=import-error` (that could be carried out blinding by users) is also not great. 

Adding support to specify dependencies in code examples is also quite a bit of work and will make running this examples much slower. An alternative could be to *offload* examples needing external libraries to `examples/`, where it should be easier to specify dependencies for example, but I agree in this case that defeats the whole purpose of having a quick small example embedded.

In any case, I agree that for now if there is no consensus about this, the best approach is to leave the `# pylint: disable=import-error`, but I think the goal should be to somehow remove it in the future. I will create a discussion about this so it's not forgotten.

leandro-lucarella-frequenz: OK, I just moved the topic here for now:

* https://github.com/frequenz-floss/frequenz-repo-config-python/issues/27#issuecomment-1547692254

src/frequenz/sdk/actor/_decorator.py
```diff
@@ -101,25 +103,29 @@ async def run(self) -> None:
                     await self._output.send(msg.inner)
 
 
-    input_chan_1: Broadcast[bool] = Broadcast("input_chan_1")
-    input_chan_2: Broadcast[bool] = Broadcast("input_chan_2")
+    async def main() -> None:
```
leandro-lucarella-frequenz: Revert adding the `async def main()`

src/frequenz/sdk/actor/_decorator.py
leandro-lucarella-frequenz: OK, so we can just revert this change then.

src/frequenz/sdk/actor/_decorator.py
```diff
@@ -154,24 +160,27 @@ async def run(self) -> None:
...
-        name="ActorTwo",
-        recv=a1_chan.new_receiver(),
-        output=a2_chan.new_sender(),
-    )
-
-    a2_rx = a2_chan.new_receiver()
-
-    await input_chan.new_sender().send(True)
-    msg = await a2_rx.receive()
+    async def main() -> None:
```
leandro-lucarella-frequenz: Revert adding the `async def main()`

leandro-lucarella-frequenz (dismissed):
:tada:

leandro-lucarella-frequenz (approved):

PR #387, opened 2023-05-15
Add method to MovingWindow that waits for a number of samples

### LINKED ISSUE #967: [MovingWindow] Add a trigger that fires after having received a fixed number of samples
### What's needed?

Currently there is no clean way to wait until the MovingWindow got updated with a certain number of samples.
This feature has been requested in the past and there has been an initial implemented attempt https://github.com/frequenz-floss/frequenz-sdk-python/pull/387.

This feature has been requested again now by @cwasicki.

### Proposed solution

Introduce an asynchronous `wait_for_samples` method that returns once the requested number of new samples has been stored in the `MovingWindow`.

### Use cases

* replace a while loop in the `Forecast actor` (https://github.com/frequenz-floss/frequenz-sdk-python/pull/387#issuecomment-2152273322)

_No response_

### Additional context

_No response_

matthias-wende-frequenz:
Up to now there was no clean way to wait until the MovingWindow got updated with a certain number of samples.
In this PR we introduce a `wait_for_samples` method that finishes once the number of samples arrived.

### DIFF
--- a/src/frequenz/sdk/timeseries/_moving_window.py
+++ b/src/frequenz/sdk/timeseries/_moving_window.py
@@ -138,6 +138,13 @@ class MovingWindow:
         self._resampler_sender: Sender[Sample] | None = None
         self._resampler_task: asyncio.Task[None] | None = None
 
+        self._wait_for_num_samples: int = 0
+        """The number of samples to wait for before the wait_for_num_samples channels
+        sends out an event."""
+        self._wait_for_samples_channel = Broadcast[None](
+            "Wait for number of samples channel."
+        )
+
         if resampler_config:
             assert (
                 resampler_config.resampling_period <= size
@@ -169,6 +176,9 @@ class MovingWindow:
         Raises:
             asyncio.CancelledError: if the MovingWindow task is cancelled.
         """
+        received_samples_count = 0
+        wait_for_samples_sender = self._wait_for_samples_channel.new_sender()
+
         try:
             async for sample in self._resampled_data_recv:
                 _logger.debug("Received new sample: %s", sample)
@@ -177,11 +187,47 @@ class MovingWindow:
                 else:
                     self._buffer.update(sample)
 
+                # count the number of samples and send out a trigger when it matches
+                # the number of samples to wait for.
+                received_samples_count += 1
+                if self._wait_for_num_samples != 0:
+                    if received_samples_count == self._wait_for_num_samples:
+                        received_samples_count = 0
+                        await wait_for_samples_sender.send(None)
+
         except asyncio.CancelledError:
             _logger.info("MovingWindow task has been cancelled.")
             raise
 
         _logger.error("Channel has been closed")
+
+    def set_sample_counter(self, num_samples: int) -> None:
+        """Set the number of samples to wait for until the sample counter triggers.
+
+        Args:
+            num_samples: The number of samples to wait for.
+
+        Raises:
+            ValueError: if the number of samples is less than or equal to zero.
+        """
+        if num_samples <= 0:
+            raise ValueError(
+                "The number of samples to wait for should be greater than zero."
+            )
+        self._wait_for_num_samples = num_samples
+
+    def new_sample_count_receiver(self) -> Receiver[None]:
+        """Wait until a given number of samples has been received.
+
+        The sample counter is updated irrespective of whether this
+        method is called or not. Thus this might trigger when
+        a smaller number of samples than the given number has been
+        updated.
+
+        Returns:
+            A receiver that triggers after a number of samples arrived.
+        """
+        return self._wait_for_samples_channel.new_receiver()
 
     async def stop(self) -> None:
         """Cancel the running tasks and stop the MovingWindow."""
--- a/tests/timeseries/test_moving_window.py
+++ b/tests/timeseries/test_moving_window.py
@@ -104,6 +104,27 @@ async def test_window_size() -> None:
     assert len(window) == 5
 
 
+@pytest.mark.parametrize("samples_to_wait_for", [-1, 0, 1, 10])
+async def test_wait_for_samples(samples_to_wait_for: int) -> None:
+    """Test waiting for samples."""
+    window, sender = init_moving_window(timedelta(seconds=1))
+
+    if samples_to_wait_for <= 0:
+        with pytest.raises(ValueError):
+            window.set_sample_counter(samples_to_wait_for)
+        return
+    sample_count_recv = window.new_sample_count_receiver()
+
+    window.set_sample_counter(samples_to_wait_for)
+
+    # asyncio.create_task(push_data_delayed())
+    for i in range(0, samples_to_wait_for):
+        await sender.send(
+            Sample(datetime.now(tz=timezone.utc) + timedelta(seconds=i), 1.0)
+        )
+    await sample_count_recv.receive()
+
+
 # pylint: disable=redefined-outer-name
 async def test_resampling_window(fake_time: time_machine.Coordinates) -> None:
     """Test resampling in MovingWindow."""


matthias-wende-frequenz:
FYI @idlir-shkurti-frequenz

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -138,6 +138,16 @@ def __init__(  # pylint: disable=too-many-arguments
         self._resampler_sender: Sender[Sample] | None = None
         self._resampler_task: asyncio.Task[None] | None = None
 
+        """The number of samples that have been received."""
+        self.count_samples = 0
+        """The number of samples to wait for before
+        the wait_for_samples method triggers."""
+        self.wait_for_num_samples = 0
```
shsms: docstrings need to be below the item they are docing.
```suggestion
        self.count_samples = 0
        """The number of samples that have been received."""
        self.wait_for_num_samples = 0
        """The number of samples to wait for before
        the wait_for_samples method triggers."""
```

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -138,6 +138,16 @@ def __init__(  # pylint: disable=too-many-arguments
...
+
+        self.wait_for_samples_channel = Anycast[None]()
```
shsms: Anycast is slower than Broadcast, so unless there's a reason we can't use Broadcast, we should use Broadcast in the sdk.

But in this case,  because this channel will always remain inside a process,  and it is just a signal,  maybe an `asyncio.Event` is better.

matthias-wende-frequenz: `asyncio.Event` sounds lovely.
I was hoping for a nice recommendation in the review :). Thanks Sahas!

matthias-wende-frequenz: updated!

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: done

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -138,6 +138,13 @@ def __init__(  # pylint: disable=too-many-arguments
         self._resampler_sender: Sender[Sample] | None = None
         self._resampler_task: asyncio.Task[None] | None = None
 
+        self.count_samples = 0
+        """The number of samples that have been received."""
```
leandro-lucarella-frequenz: I think this one should probably be a read-only public attribute. Also for attributes it's better to use nouns instead of verbs (and it's missing the type), for example:

```suggestion
        self._received_samples_count: int = 0
        """The number of samples that have been received."""
```

Plus

```py        
    @property
    def received_samples_count(self) -> int
        return self._received_samples_count
```

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -138,6 +138,13 @@ def __init__(  # pylint: disable=too-many-arguments
...
+        self.wait_for_num_samples = 0
+        """The number of samples to wait for before
+        the wait_for_samples method triggers."""
+        self.wait_for_samples_event = asyncio.Event()
```
leandro-lucarella-frequenz: These two **really** look like they should be *private*. Same about names (and missing types):

```suggestion
        self._expected_samples_count: int = 0
        """The number of samples to wait for before `wait_for_samples()` triggers."""
        self._wait_for_samples_event: asyncio.Event = asyncio.Event()
        """The event to signal `wait_for_samples()` that the wait is over."""
```

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -177,12 +184,43 @@ async def _run_impl(self) -> None:
...
 
         _logger.error("Channel has been closed")
 
+    async def wait_for_samples(self, num_samples: int) -> None:
+        """Wait until a given number of samples has been received.
+
+        The sample counter is updated irrespective of whether this
+        method is called or not. Thus this might trigger when
+        a smaller number of samples than the given number has been
+        updated.
```
leandro-lucarella-frequenz: I guess you decided to go with modulo for a reason, but it is not clear at all for me why. Was this designed with some use case in mind? Maybe you could an example? If feels very strange that if I call `mw.wait_for_samples(10)` it could trigger after receiving only 5 samples or even immediately!

Also, not sure if it will be useful, but given there is this inconsistency between requested number of samples and actual amount of received samples since this method was called, maybe you could return the actual number of samples that were received (à la `read()`)?

leandro-lucarella-frequenz (commented):
I also feel something smells here, I don't like having some class variable to only track a counter that is used by one method.  My instinct says that there should be a way to split this functionality to a different class that uses, can be plugged to or proxies to the `MovingWindow`, but I don't have anything in particular to propose right. At least nothing that's not a lot of extra work (and probably worsen the usability of this too), so I'm fine to leave it as is, but just mentioning it in case anyone suddenly gets a brilliant idea :bulb:

leandro-lucarella-frequenz:
Oh, I also noticed in your last PRs that it seems you have your editor configured to a maximum line length of maybe 50 chars? Ideally it should be 88 to match `black`'s config.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -177,12 +184,43 @@ async def _run_impl(self) -> None:
...
+
+        Raises:
+            ValueError: if the number of samples is less than or equal to zero.
+        """
+        if num_samples <= 0:
+            raise ValueError(
+                "The number of samples to wait for should be greater than zero."
+            )
+
+        if self.count_samples % num_samples == 0:
```
Marenz: This will return for `self.count_samples == 0` immediately...

tests/timeseries/test_moving_window.py
```diff
@@ -104,6 +104,23 @@ async def test_window_size() -> None:
     assert len(window) == 5
 
 
+async def test_wait_for_samples() -> None:
```
Marenz: I think we def. need more tests with different wait parameters

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -177,12 +184,43 @@ async def _run_impl(self) -> None:
...
+            return
+
+        self.wait_for_num_samples = num_samples
+        await self.wait_for_samples_event.wait()
```
Marenz: If you never call `self.wait_for_samples_event.clear()` it will always fire immediately in `.wait()`

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -177,12 +184,43 @@ async def _run_impl(self) -> None:
                 else:
                     self._buffer.update(sample)
 
+                self.count_samples += 1
+                if self.wait_for_num_samples != 0:
+                    if self.count_samples % self.wait_for_num_samples == 0:
```
Marenz: You're using the same check in line 217, maybe put it in a common private function that way if you change one part it's also correct in the other part

src/frequenz/sdk/timeseries/_moving_window.py
cwasicki: Also it could fail if no new samples were received since last call of this method.

tests/timeseries/test_moving_window.py
leandro-lucarella-frequenz: Agreed, if there were bugs that weren't caught by at least the author and 2 reviewers, then we definitely need more tests, specially for corner cases.

[hypothesis](https://hypothesis.readthedocs.io/en/latest/quickstart.html) could be a good match for this, but if it's too much effort to add, then something like `pytest.mark.parametrize` could help:

```py
pytest.mark.parametrize('samples_to_wait_for', [-1, 0, 1, 4, 100])
def test_xxx(samples_to_wait_for: int) -> None:
    ...
```

matthias-wende-frequenz: done

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: modulo is gone

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: changed

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: this check is gone

matthias-wende-frequenz:
I've reworked the implementation. Now the user can get a channel that sends a `None` when the a certain number of samples arrived. That has the advantage that we can plug it into our channels select mechanism.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -138,6 +138,13 @@ def __init__(  # pylint: disable=too-many-arguments
         self._resampler_sender: Sender[Sample] | None = None
         self._resampler_task: asyncio.Task[None] | None = None
 
+        self._wait_for_num_samples: int = 0
+        """The number of samples to wait for before the wait_for_num_samples channels
+        sends out an event."""
```
leandro-lucarella-frequenz, line 143: I recommend using nouns for non-boolean attributes:

```suggestion
        self._num_samples_to_wait_for: int = 0
        """The number of samples to wait for before triggering an event through the channel."""
```

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -138,6 +138,13 @@ def __init__(  # pylint: disable=too-many-arguments
...
+        self._wait_for_samples_channel = Broadcast[None](
+            "Wait for number of samples channel."
+        )
```
leandro-lucarella-frequenz, line 146: So the text here is for debugging purposes only, right @shsms?

I'd suggest using a shorter string for that and using the long string as documentation for the variable:

```suggestion
        self._wait_for_samples_channel = Broadcast[None]("wait-for-samples")
        """Channel to send events to when wait for number of samples is triggered."""

```

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -169,6 +176,9 @@ async def _run_impl(self) -> None:
         Raises:
             asyncio.CancelledError: if the MovingWindow task is cancelled.
         """
+        received_samples_count = 0
+        wait_for_samples_sender = self._wait_for_samples_channel.new_sender()
```
leandro-lucarella-frequenz, line 180: I guess once a MW was stopped there is no way to start it again, right? Otherwise the sender should probably be created in the constructor instead to avoid *leaking* sender objects.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -177,12 +187,48 @@ async def _run_impl(self) -> None:
                 else:
                     self._buffer.update(sample)
 
+                # count the number of samples and send out a trigger when it matches
+                # the number of samples to wait for.
+                received_samples_count += 1
+                if self._wait_for_num_samples != 0:
+                    if received_samples_count == self._wait_for_num_samples:
+                        received_samples_count = 0
+                        await wait_for_samples_sender.send(None)
```
leandro-lucarella-frequenz, line 196: Since it's free and even when the user should know it beforehand anyway, we could also send the number of samples received instead of `None`, maybe it could become a handy shortcut.
```suggestion
                        await wait_for_samples_sender.send(received_samples_count)
                        received_samples_count = 0
```

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -177,12 +187,48 @@ async def _run_impl(self) -> None:
...
+            num_samples: The number of samples to wait for.
+
+        Raises:
+            ValueError: if the number of samples is less than or equal to zero.
+        """
+        if num_samples <= 0:
+            raise ValueError(
+                "The number of samples to wait for should be greater than zero."
+            )
+        self._wait_for_num_samples = num_samples
```
leandro-lucarella-frequenz, line 217: If I understand correctly, this indirectly enables sending the events if `num_samples > 0`, right? If so I would rename this method to something that makes it easier to realize that happens. What about:

```py

@property
def is_wait_for_samples_event_enabled(self) -> bool:
    # Returns `self._wait_for_num_samples != 0`

def enable_wait_for_samples_event(self, num_samples: int) -> bool:
    # Same as above but check for `num_samples > 0`, return whether it was enabled before

def disable_wait_for_samples_event(self) -> bool:
    # Set `num_samples = 0`, return whether it was enabled before
```

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -177,12 +187,48 @@ async def _run_impl(self) -> None:
...
+
+    def new_sample_count_receiver(self) -> Receiver[None]:
```
leandro-lucarella-frequenz, line 219: If the above suggestion is applied, I would rename this for consistency:

```suggestion
    def new_wait_for_samples_event_receiver(self) -> Receiver[None]:
```

We could go with something different than `wait_for_samples_event` as it is a bit long, but the 4 methods should use the same term IMHO.

tests/timeseries/test_moving_window.py
```diff
@@ -104,6 +104,27 @@ async def test_window_size() -> None:
...
+
+    if samples_to_wait_for <= 0:
+        with pytest.raises(ValueError):
+            window.set_sample_counter(samples_to_wait_for)
+        return
+    sample_count_recv = window.new_sample_count_receiver()
+
+    window.set_sample_counter(samples_to_wait_for)
+
+    # asyncio.create_task(push_data_delayed())
```
leandro-lucarella-frequenz, line 120:
```suggestion
```

?

tests/timeseries/test_moving_window.py
```diff
@@ -104,6 +104,27 @@ async def test_window_size() -> None:
...
+    for i in range(0, samples_to_wait_for):
+        await sender.send(
+            Sample(datetime.now(tz=timezone.utc) + timedelta(seconds=i), 1.0)
+        )
+    await sample_count_recv.receive()
```
leandro-lucarella-frequenz, line 125: There are some failures that this will not detect or detect wrongly. For example if you trigger before the `samples_to_wait_for` this will succeed too. If the event doesn't trigger, this test will hang forever, which might be quite annoying.

Maybe could check for the internal counter to verify it is incremented? If you also add my suggestion to return the number of samples received you could add the check here too.

Maybe you can also move the `await sample_count_recv.receive()` to a different task so you can check it didn't fire before you sent enough samples and that it was fired after you sent all the expected samples but without blocking the *main thread* if it fails?

Maybe you could also push different sample values and then check that the state of the moving window is what it would be expected after receiving all that samples, so if the even triggered before that check should also fail.

leandro-lucarella-frequenz (commented):
Besides a few comments on improving consistency, the API and the tests, the commit message needs to be updated because it mentions a `wait_for_samples` method.

cwasicki:
Will this not be part of v1 milestone?

matthias-wende-frequenz:
> Will this not be part of v1 milestone?

Let's see. This change is not breaking so it can be v1.1.x too. But let's try to get it in.

matthias-wende-frequenz:
I think we should close it and come back to it, once @cwasicki has finished all the changes.

cwasicki:
The code changes here look unrelated to what I am working on.

matthias-wende-frequenz:
> The code changes here look unrelated to what I am working on.

Right, ignore my comment ;)

matthias-wende-frequenz:
This is quite old. It we need it we can reopen the PR or start from scratch.

cwasicki:
Is there another way to know if the moving window was updated?

matthias-wende-frequenz:
> Is there another way to know if the moving window was updated?

No there isn't.
This pr was closed, not because we don't want to support it, but rather since nobody is working nor anyone was asking for this feature. So as it appears there is no need for this feature.

shsms:
It turns out this is required in the Forecast actor.  Only yesterday evening, Christoph showed me the place where he's using a while loop that could be replaced with this, just a little bit after we were talking about closing this issue.  But I think not top priority, so we can pick this up again soon.

matthias-wende-frequenz:
> It turns out this is required in the Forecast actor. Only yesterday evening, Christoph showed me the place where he's using a while loop that could be replaced with this, just a little bit after we were talking about closing this issue. But I think not top priority, so we can pick this up again soon.

Okay, thanks for the notice. I'll put it back on my todo list.

cwasicki:
Cool, is there an issue to track?

matthias-wende-frequenz:
> Cool, is there an issue to track?

https://github.com/frequenz-floss/frequenz-sdk-python/issues/967

PR #400, opened 2023-05-23
Log host:port when connecting to a microgrid

### LINKED ISSUE #344: Add a log message with the host/port of the microgrid API we are connecting to
### What's needed?

To add a log message with the host/port of the microgrid API we are connecting to.

### Proposed solution

Just log a message saying where are we connecting to when we do so, probably with level `INFO`.

### Use cases

Every app out there.

Log it in every app out there.

### Additional context

@christianparpart mentined having some issues connecting to the wrong microgrid API, and that a log message mentioning where the connection is being made could have been useful.

Marenz:


leandro-lucarella-frequenz (approved):
LGTM

PR #401, opened 2023-05-23
Drop support for Python versions older than 3.11

### LINKED ISSUE #148: Move minimal supported Python version to 3.11
### What's needed?

It would be very nice to be able to use new improvements in Python 3.10/3.11, which introduced some big features and some major improvements to typing:

* `dataclass_transform`
* `Self`
* Variadic generics
* Pattern matching (`match` statement)
* Exception groups

Currently we need to use a few hacks to be able to do simple things, like forward declarations (using `from __future__ import annoations`) or include extra backported modules like `typing_extensions` or `exceptiongroups`.

### Proposed solution

Move to python 3.11 as the minimum required version.

### Blockers

In README there is sentence:
`For arm64 only Python 3.8 is supported (due to some dependencies that only support 3.8).`

These dependencies was an xgboost package that I've build locally, since we wanted to have gpu support for our nvidia jetson xaviar nx edge controllers
Actually these dependencies were never used anywhere.
It is not build and needed now.

- [x] `xgboost` compiled for controllers GPU: not being used right now.
- [x] Ubuntu shipping 3.11: We can use the official Python docker images (based on Debian/alpine)
- [x] `mypy` doesn't support all 3.11 features yet (not a hard blocker though): https://github.com/python/mypy/issues/12840 (is not all done but only very minor and obscure things are missing, so it should not be a blocker)
- [x] Tensorflow only supports 3.10 for now, but 3.11 is coming: https://github.com/tensorflow/tensorflow/issues/58032
- [x] Ask @frequenz-floss/datasci-team if there are other required ML libraries needed
- [ ] Disable tests for older python versions in branch protection

### Check if a dependency is Python 3.11-compatible

* https://observablehq.com/d/eb336fc983fcdf52

### LINKED ISSUE #399: Make `Sample`s comparable by value again
### What happened?

Equality comparison between two `Sample` objects is passing even when they have different values.

Two `Sample` objects should be considered equal only if they have the same timestamp and value.

### Affected version(s)

_No response_

### Affected part(s)

### Extra information

_No response_

leandro-lucarella-frequenz:
- Reformat `pyproject.toml` using `taplo`
- Drop support for Python versions older than 3.11
- Improve typing annotations
- Remove Sample comparison hack

Fixes #148, fixes #399.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -33,6 +33,7 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python }}
+          cache: 'pip'
 
       - uses: actions/cache@v3
         with:
@@ -60,6 +61,7 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: '3.11'
+          cache: 'pip'
 
       - name: Install build dependencies
         run: |
@@ -90,6 +92,7 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ env.DEFAULT_PYTHON_VERSION }}
+          cache: 'pip'
 
       - name: Install build dependencies
         run: |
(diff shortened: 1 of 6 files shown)


Marenz (dismissed):

leandro-lucarella-frequenz:
Test were failing, now they should be fixed, so this needs one more approval. Enabled auto-merge.

Marenz (approved):

PR #404, opened 2023-05-24
Add microgrid as default import to conftest.py

matthias-wende-frequenz:
Simplifies examples when using high level microgrid sdk tools like `microgrid.logical_meter()`.


Marenz (approved):

Marenz:
We probably need to put this in its own repo soon, it's already diverging from the channels repo (this is the second diversion)

Issue #276, opened 2023-03-29
Create separate charge/discharge interfaces for data, bounds and control commands


shsms:
### What's needed?

- [x] #398
- [x] `grid_power, battery_pool.power, etc` should have separate consumption/supply streams all positive.

More details to be added soon.

shsms:
We don't need this issue to track #413 anymore,  it is being tracked in #542

PR #408, opened 2023-05-26
Control interface for BatteryPool

### LINKED ISSUE #376: Remove multi-user capability from `PowerDistributor`
### What's needed?

The `PowerDistributor` has the capacity to accept requests from multiple actors,  but it doesn't have an algorithm to coordinate requests from multiple actors,  and just sends out the latest request from an actor, overriding requests from others.  So it can't be used in its current form in multi-user scenarios.

This coordination is going to be implemented in the `PowerArbitrator` instead,  and that will be the only component that communicates with the `PowerDistributor`.

So the PowerDistributor's features for supporting multiple users can be removed.

### Proposed solution

_No response_

### Use cases

_No response_

_No response_

### Additional context

_No response_

### LINKED ISSUE #398: `battery_pool.{charge, discharge}` methods,  both taking positive values

shsms:
This PR updates the PowerDistributingActor to remove its multi-user interface, make its accessible only through the BatteryPool,  with these methods:

1. set_power :: send a PSC power request to the power distributor.
2. charge :: send a charge power (+ve) to the power distributor.
3. discharge :: send a discharge power (+ve) to the power distributor.
4. power_distribution_results :: get a receiver to power distributor responses.


leandro-lucarella-frequenz:
I linked a few issues that should be closed by this, I guess this doesn't cover the whole #276, right?  If this closes any other issues, please remember to link them!

shsms:
> I linked a few issues that should be closed by this, I guess this doesn't cover the whole https://github.com/frequenz-floss/frequenz-sdk-python/issues/276, right?

Thanks Luca, for #276,  only the bounds part would be pending after this.  I've made an issue for that.

daniel-zullo-frequenz:
@sahas-subramanian-frequenz and @leandro-lucarella-frequenz 
based on https://github.com/frequenz-floss/frequenz-sdk-python/pull/365#discussion_r1205082778, do you want to discuss or address the alternative names for `include_broken` for the [request attribute](https://github.com/frequenz-floss/frequenz-sdk-python/blob/v0.x.x/src/frequenz/sdk/actor/power_distributing/request.py#L33) here in this PR? Otherwise I'll create a new issue if this still needs to be discussed/addressed

shsms:
I was thinking of `include_broken_batteries`, because that would be slightly more obvious with the `charge/discharge` methods.  Does that make sense?  

But I have no opinion about renaming further,  it turns out it can't be fully explained with just the name,  and we'll just have to use what make sense most of the time (which is include_broken),  and describe it better in the docstrings.

I'll add a further line saying that using this option is still not a guarantee that all/any batteries will get used.

daniel-zullo-frequenz:
`include_broken_batteries` is just too long for a flag IMO but as it is an important flag it makes sense to make it explicitly obvious for the user.

> I'll add a further line saying that using this option is still not a guarantee that all/any batteries will get used.

As you gonna add a line, can you also update the documentation for `include_broken_batteries` to something around this:
```
    If set to `True`, the power distribution algorithm will consider all batteries, including the broken ones, when distributing power. 
    In such cases, any remaining power after distributing among the available batteries will be distributed equally among the unavailable (broken) batteries.
    If all batteries in the set are unavailable, the power will be equally distributed among all the unavailable batteries in the request.

    If set to `False`, the power distribution will only take into account the available batteries, excluding any broken ones.
```

leandro-lucarella-frequenz:
> `include_broken_batteries` is just too long for a flag IMO but as it is an important flag it makes sense to make it explicitly obvious for the user.

For me it is OK because it is an obscure flag, is not something that we expect users be using all the time. I'm fine with leaving it at `include_broken`, as Sahas said we won't be able to convey the full meaning there, so it should be enough with something that is not clearly confusing or misleading.  I'd rather be as clear as possible though, even if the flag name is on the long side because, again, it should be a rare flag to use anyway.

src/frequenz/sdk/actor/power_distributing/power_distributing.py
```diff
@@ -188,21 +178,26 @@ class PowerDistributingActor:
 
     def __init__(
         self,
-        users_channels: Dict[str, Bidirectional.Handle[Result, Request]],
+        requests_rx: Receiver[Request],
```
leandro-lucarella-frequenz: I would name this `request_receiver` even if it is a bit longer. I think the `rx`/`tx` idiom might be a bit cryptic for the SDK, at least for the public facing interface.

src/frequenz/sdk/actor/power_distributing/power_distributing.py
```diff
@@ -188,21 +178,26 @@ class PowerDistributingActor:
...
+                for each request namespace.
             battery_status_sender: Channel for sending information which batteries are
                 working.
             wait_for_data_sec: How long actor should wait before processing first
                 request. It is a time needed to collect first components data.
         """
+        self._requests_rx = requests_rx
+        self._channel_registry = channel_registry
         self._wait_for_data_sec = wait_for_data_sec
+        self._result_senders: Dict[str, Sender[Result]] = {}
```
leandro-lucarella-frequenz, line 200: So these are matched to the request namespace, right? Or how do you map which response should be sent to which channel?

Maybe it would be good to clarify adding a docstring to the attribute even if it's private.

src/frequenz/sdk/timeseries/battery_pool/battery_pool.py
```diff
@@ -84,15 +90,129 @@ def __init__(  # pylint: disable=too-many-arguments
...
+        power: float,
+        *,
+        adjust_power: bool = True,
+        request_timeout_sec: float = 5.0,
+        include_broken: bool = False,
+    ) -> None:
+        """Set the given power for the batteries in the pool.
+
+        Power values need to follow the Passive Sign Convention (PSC). That is, positive
+        values indicate charge power and negative values indicate discharge power.
```
leandro-lucarella-frequenz, line 117: What about mentioning the existence of `charge()` and `discharge()` here and recommending their use instead (unless you are already working with a power value following PSC)?

src/frequenz/sdk/timeseries/battery_pool/battery_pool.py
```diff
@@ -84,15 +90,129 @@ def __init__(  # pylint: disable=too-many-arguments
...
             channel_registry,
             resampler_subscription_sender,
         )
 
+    async def set_power(
+        self,
+        power: float,
+        *,
+        adjust_power: bool = True,
+        request_timeout_sec: float = 5.0,
```
leandro-lucarella-frequenz: Shouldn't this be a `timedelta` now?

examples/power_distribution.py
leandro-lucarella-frequenz, line 1: I wonder if by now we should just remove this example, I guess users should be encouraged to use the battery pool instead of communicating directly to the power distirbuting actor. Maybe not in this PR, but in a near future?

We can leave a small example in the docstring, but the bulk of this example shows how to build the data pipeline more than how to distribute power among a set of batteries.

benchmarks/power_distribution/power_distributor.py
leandro-lucarella-frequenz, line 1: Did you run the benchmarks before and after? Where there any differences?

leandro-lucarella-frequenz (commented):
I keep forgetting why we want to remove the multi-user support for `PowerDistributor`, it might be a good idea to explain the why in the commit message removing it.

Same about the commit adding the `namespace` to the request (I saw it is used in following commits, but when looking at the commit in order it is a bit mysterious).

Besides the small comments, LGTM!

src/frequenz/sdk/actor/power_distributing/power_distributing.py
shsms: done

src/frequenz/sdk/actor/power_distributing/power_distributing.py
shsms: done

src/frequenz/sdk/timeseries/battery_pool/battery_pool.py
shsms: done, also in the charge/discharge methods.

src/frequenz/sdk/timeseries/battery_pool/battery_pool.py
shsms: updated.

examples/power_distribution.py
shsms: True,  we have a number of examples that became more confusing when we updated them to pass pylint as well.  I guess we have a big example cleanup project pending.

benchmarks/power_distribution/power_distributor.py
shsms: It was the same time for the single user tests.

shsms:
Renamed to `include_broken_batteries` and updated docstrings.

Updated commit messages explaining removal of multi-user support and introduction of namespaces.

leandro-lucarella-frequenz (approved):

src/frequenz/sdk/timeseries/battery_pool/battery_pool.py
```diff
@@ -84,15 +90,144 @@ def __init__(  # pylint: disable=too-many-arguments
...
+        values indicate charge power and negative values indicate discharge power.
+
+        When not using the Passive Sign Convention, the `charge` and `discharge` methods
+        might be more convenient.
+
+        Args:
+            power: The power to set for the batteries in the pool.
+            adjust_power: If True, the power will be adjusted to fit the power bounds,
+                if necessary. If False, then power requests outside the bounds will be
+                rejected.
```
christianparpart, line 126: maybe we should have a better name for `adjust_power`, for example `apply_power_bounds`, because the former does not tell me what this parameter is doing, in fact, I was scratching my face and wondering the whole purpose of this function just by looking at that boolean in that context. :-)

With a better name, hopefully (like proposed), it should be more clear what impact this boolean parameter will have.

Same applies to the other `adjust_power` occurrences below.

leandro-lucarella-frequenz: Another alternative could be `adjust_to_bounds`. at least for me it is not super clear what applying a bound means.

shsms: I think it is impossible to encode everything the flag does into its name,  we will have to read the docs,  because it is not trivial and it needs info about
1. when/why is it required - when requests ~exceed the bounds~ are out of bounds
2. what does it do when true - adjust power to fit within the bounds and forward request
3. what does it do when false - reject request,  not forward without adjusting.

And `apply_power_bounds` always happens.  What to do after that is the question - adjust the power value to fit the bounds, or reject the request.

shsms: But I agree, `adjust_to_bounds` is nicer.  If there's code that says:

```python
await battery_pool.charge(2000, adjust_to_bounds=True)
```
maybe it is obvious to first time users without reading the docs.  Hopefully,  in most scenarios they want our default value (True) anyway and won't have to deal with it.

llucax: > I think it is impossible to encode everything the flag does into its name, we will have to read the docs, because it is not trivial and it needs info about

Yeah, I was about to say the same thing, but in this case I also felt that there was room for improvement :)

We need to strike a balance. And like `sop`, if something is super hard to make clear in a *indetifier*, I think it is better to have it obscure to force people to look at the docs instead of having something more meaningful but misleading.

PR #422, opened 2023-06-05
Add Quantity types and update formulas to produce typed Samples

shsms:


src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -0,0 +1,310 @@
...
+                or not fspec_parts[1].isdigit()
+            ):
+                raise ValueError(
+                    "Invalid format specifier. Must be empty or `[0].{precision}`"
+                )
+            if fspec_parts[0] == "0":
+                keep_trailing_zeros = True
+            precision = int(fspec_parts[1])
+        else:
+            precision = 3
```
shsms, line 160: Add exception for types without `_exponent_unit_map`

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -0,0 +1,310 @@
...
+class Quantity:
+    """A quantity with a unit."""
+
+    _base_value: float
+    """The value of this quantity in the base unit."""
+
+    _exponent_unit_map: dict[int, str]
+    """A mapping from the exponent of the base unit to the unit symbol."""
+
+    def __init__(self, value: float, *, exponent: int = 0) -> None:
```
shsms: Based on the words used in https://en.wikipedia.org/wiki/Metric_prefix, we could change
- `exponent` to `base10_multiple` or similar
- `_exponent_unit_map` to `metric_prefix_map`

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -0,0 +1,310 @@
...
+        """Initialize a new quantity.
+
+        Args:
+            value: The value of this quantity in a given exponent of the base unit.
+            exponent: The exponent of the base unit the given value is in.
+        """
+        self._base_value = value * 10**exponent
+
+    def __init_subclass__(cls, exponent_unit_map: dict[int, str]) -> None:
```
leandro-lucarella-frequenz, line 37: I didn't know about this `__init__subclass__` thing :scream:

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -0,0 +1,310 @@
...
+
+
+class Quantity:
+    """A quantity with a unit."""
+
+    _base_value: float
+    """The value of this quantity in the base unit."""
+
+    _exponent_unit_map: dict[int, str]
+    """A mapping from the exponent of the base unit to the unit symbol."""
```
leandro-lucarella-frequenz: I think this might actually be public (read-only). It might be useful if you need to so some custom string conversion or something.

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -0,0 +1,310 @@
...
+        Raises:
+            TypeError: If the given exponent_unit_map is not a dict.
+            ValueError: If the given exponent_unit_map does not contain a base unit
+                (exponent 0).
+        """
+        if not isinstance(exponent_unit_map, dict):
+            raise TypeError(
+                "Expected exponent_unit_map to be a dict, "
+                f"got {type(exponent_unit_map)}"
+            )
```
leandro-lucarella-frequenz: This is only useful if you don't trust `mypy`, right?

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -0,0 +1,310 @@
...
+        """
+        return math.isinf(self._base_value)
+
+    def __repr__(self) -> str:
+        """Return a representation of this quantity.
+
+        Returns:
+            A representation of this quantity.
+        """
+        return f"{type(self).__name__}(value={self._base_value}, exponent=0)"
```
leandro-lucarella-frequenz, line 106: You could use `{self._base_value!r}` to get the `repr` of `self._base_value` instead of the `str`, but I guess for floats is exactly the same.

src/frequenz/sdk/timeseries/_quantities.py
leandro-lucarella-frequenz: Can this happen? So `mypy` doesn't check that a `class` declaration has the arguments that are required for `__init_subclass__`?

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -0,0 +1,310 @@
...
+        super().__init_subclass__()
+
+    @property
+    def base_value(self) -> float:
+        """Return the value of this quantity in the base unit.
+
+        Returns:
+            The value of this quantity in the base unit.
+        """
+        return self._base_value
```
leandro-lucarella-frequenz, line 61: What about adding a `base_unit` that returns `_exponent_unit_map[0]` for convenience/completeness?

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -0,0 +1,310 @@
...
+
+        Args:
+            other: The other quantity.
+
+        Returns:
+            Whether this quantity is equal to another.
+        """
+        if not isinstance(other, self.__class__):
+            return NotImplemented
+        return self._base_value == other._base_value
```
leandro-lucarella-frequenz, line 275: Shouldn't we add `__hash__` too so we can used them as `dict` keys and in `set`s?

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -0,0 +1,310 @@
...
+        """Return whether this quantity is equal to another.
+
+        Args:
+            other: The other quantity.
+
+        Returns:
+            Whether this quantity is equal to another.
+        """
+        if not isinstance(other, self.__class__):
+            return NotImplemented
```
leandro-lucarella-frequenz, line 270: This should return `False` IMHO, otherwise if will break *generic* code, for example if type info is lost and types are considered `Any`, you should still be able to use `a == b`. In Python you should be able to compare any 2 values and only get `True` or `False`.

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -0,0 +1,310 @@
...
+    def __gt__(self, other: Self) -> bool:
+        """Return whether this quantity is greater than another.
+
+        Args:
+            other: The other quantity.
+
+        Returns:
+            Whether this quantity is greater than another.
+        """
+        return self._base_value > other._base_value
```
leandro-lucarella-frequenz, line 219: Does `mypy` check that `other` is of type `Self`? So `mypy` will forbid `some_power > some_energy` or `some_power > 1.0`? I wonder if type info is lost and values are considered `Any`, the `mypy` can't do the check, so we probably should also add a runtime check here raising a `TypeError` if the types are not compatible (pretty much like tz aware and naive `datetime`).

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -0,0 +1,310 @@
...
+
+
+class Power(
+    Quantity,
+    exponent_unit_map={
+        -3: "mW",
+        0: "W",
+        3: "kW",
+        6: "MW",
+    },
```
leandro-lucarella-frequenz, line 285: I know for us it doesn't make a lot of sense, but thinking that this should probably be turned into its own library in the future, any reasons not to include at least also `uW` and `GW`?

Even more, I think usually all exponents should be the same `p`, `n`, `u`, `m`, `k`, `M`, `G`, `T`, etc., so maybe you want to also add an option to just specify the base unit (`W` here) and build the map automatically. Then the standard map can include all exponents:

[Exponents](https://convertermaniacs.com/metric-prefixes/mega-to-nano-converter.html)

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -0,0 +1,310 @@
...
+
+        Args:
+            megawatts: The power in megawatts.
+
+        Returns:
+            A new power quantity.
+        """
+        return cls(value=megawatts, exponent=6)
+
+    def as_watts(self) -> float:
```
leandro-lucarella-frequenz, line 337: I would make all of these `as_` `@property`s...

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -0,0 +1,310 @@
...
+    def __add__(self, other: Self) -> Self:
+        """Return the sum of this power and another.
+
+        Args:
+            other: The other power.
+
+        Returns:
+            The sum of this power and another.
+        """
+        return self.from_watts(self.as_watts() + other.as_watts())
```
leandro-lucarella-frequenz: Why this? Can't you just move this to the `Quantity` class and do `return type(self)(self._base_value + other._base_value, exponent=0)`?

Similar to the other methods taking `other: Self`, we should probably add a check that `other` is actually of type `Self` in case this might be called with values typed as `Any`.

tests/timeseries/test_quantities.py
```diff
@@ -0,0 +1,47 @@
...
+
+def test_power() -> None:
+    """Test the power class."""
+    power = Power.from_kilowatts(1.2)
+    assert f"{power}" == "1.2 kW"
+    assert power.as_watts() == 1200.0
+    assert power.as_megawatts() == 0.0012
+    assert power == Power.from_milliwatts(1200000.0)
+    assert power == Power.from_megawatts(0.0012)
+    assert power != Power.from_watts(1000.0)
```
leandro-lucarella-frequenz, line 171: To check what I mention above about checking for `other: Self` maybe you can add a test:

```py
def sum(a: Any, b: Any) -> Any:
    return a + b
```

I think `mypy` will not catch this.

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -309,6 +310,55 @@ def __sub__(self, other: Self) -> Self:
         """
         return self.from_watts(self.as_watts() - other.as_watts())
 
+    def __mul__(self, duration: timedelta) -> "Energy":
```
leandro-lucarella-frequenz: `from __future__ import annotations`?

leandro-lucarella-frequenz: Also I think in general we should also provide `__mul__` and `__truediv__` with `other: float`: `Power.from_watts(10) / 2 == Power.from_watts(5)`.

benchmarks/timeseries/periodic_feature_extractor.py
```diff
@@ -34,7 +35,7 @@ async def init_feature_extractor(period: int) -> PeriodicFeatureExtractor:
         timedelta(seconds=1), lm_chan.new_receiver(), timedelta(seconds=1)
     )
 
-    await lm_chan.new_sender().send(Sample(datetime.now(tz=timezone.utc), 0))
+    await lm_chan.new_sender().send(Sample(datetime.now(tz=timezone.utc), Quantity(0)))
```
leandro-lucarella-frequenz, line 38: I feel like `Quantity` should be an ABC, so an abstract `Quantity` should be used.

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
```diff
@@ -62,6 +63,7 @@ def __init__(
         name: str,
         steps: List[FormulaStep],
         metric_fetchers: Dict[str, MetricFetcher],
+        output_type: Type[Quantity],
```
leandro-lucarella-frequenz: What about making this `Generic[T]` with `T = TypeVar("T", bound=Quantity)`? Maybe then you don't need to store the `output_type` and can make type checking easier?

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
```diff
@@ -222,7 +226,7 @@ def __add__(
             A formula builder that can take further expressions, or can be built
                 into a formula engine.
         """
-        return self._higher_order_builder(self) + other  # type: ignore
+        return self._higher_order_builder(self, self._output_type) + other  # type: ignore
```
leandro-lucarella-frequenz, line 245: Is this (and the following) `type: ignore` still needed?

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
```diff
@@ -189,20 +191,24 @@ async def apply(self) -> Sample[Quantity]:
     "HigherOrderFormulaBuilder3Phase",
 ]
 
-_GenericEngine = TypeVar("_GenericEngine", "FormulaEngine", "FormulaEngine3Phase")
+_GenericEngine = TypeVar(
+    "_GenericEngine",
+    "FormulaEngine",  # type: ignore
+    "FormulaEngine3Phase",  # type: ignore
```
leandro-lucarella-frequenz, line 204: Why are these `type: ignore` needed?

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
```diff
@@ -74,12 +74,14 @@ def __init__(
         """
         self._name = name
         self._steps = steps
-        self._metric_fetchers = metric_fetchers
+        self._metric_fetchers: Dict[
+            str, MetricFetcher[GenericQuantity]
+        ] = metric_fetchers
         self._first_run = True
-        self._output_type = output_type
+        self._output_type: type[GenericQuantity] = output_type
```
leandro-lucarella-frequenz: Oh, interesting that you can use `type` as a *generic type*, maybe you should replace the other `Type` uses?

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
```diff
@@ -55,15 +55,15 @@
...
     """A post-fix formula evaluator that operates on `Sample` receivers."""
 
     def __init__(
         self,
         name: str,
         steps: List[FormulaStep],
-        metric_fetchers: Dict[str, MetricFetcher],
-        output_type: Type[Quantity],
+        metric_fetchers: Dict[str, MetricFetcher[GenericQuantity]],
+        output_type: Type[GenericQuantity],
```
leandro-lucarella-frequenz: Can't you just use `GenericQuantity` when you need `output_type`? Do you really need to store it?

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine_pool.py
```diff
@@ -47,14 +47,14 @@ def __init__(
         self._namespace = namespace
         self._channel_registry = channel_registry
         self._resampler_subscription_sender = resampler_subscription_sender
-        self._engines: dict[str, "FormulaEngine|FormulaEngine3Phase"] = {}
+        self._engines: dict[str, "FormulaEngine[Any]|FormulaEngine3Phase[Any]"] = {}
```
leandro-lucarella-frequenz: How badly does this fucks the type system? Can't we have one `FormulaEnginePool` per `FormulaEngine[T]`? I'm a bit concerned that we are making a lot of effort to have typing everywhere but then if we use the pool we lose all that unless we are explicitly casting.

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine_pool.py
```diff
@@ -47,14 +47,14 @@ def __init__(
...
 
     def from_string(
         self,
         formula: str,
         component_metric_id: ComponentMetricId,
         nones_are_zeros: bool = False,
-    ) -> FormulaEngine:
+    ) -> FormulaEngine[Quantity]:
```
leandro-lucarella-frequenz, line 59: Also I don't understand why sometimes is `Quantity` and sometimes `Any`.

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine_pool.py
```diff
@@ -99,19 +103,62 @@ def from_generator(
             A FormulaReceiver or a FormulaReceiver3Phase instance based on what the
                 FormulaGenerator returns.
         """
-        if channel_key in self._engines:
-            return self._engines[channel_key]
+        from ._formula_engine import (  # pylint: disable=import-outside-toplevel
```
leandro-lucarella-frequenz, line 106: Can you maybe clarify why this is needed? I guess to break a cyclic import?

leandro-lucarella-frequenz (commented):
Lots of comments but this is **amazing**! Great work! :tada:

src/frequenz/sdk/timeseries/_quantities.py
leandro-lucarella-frequenz: After reviewing the whole PR, I'm pretty sure we need to add runtime checks for this, because also then the type is just a general `Quantity` `mypy` won't be able to do checks.

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
leandro-lucarella-frequenz: This was done in another commit :+1:

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine_pool.py
leandro-lucarella-frequenz: This was also done in another commit. :+1:

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -0,0 +1,656 @@
...
+
+"""Types for holding quantities with units."""
+
+import math
+from datetime import timedelta
+from typing import Self, TypeVar, overload
+
+GenericQuantity = TypeVar(
+    "GenericQuantity", "Quantity", "Power", "Current", "Voltage", "Energy"
+)
```
leandro-lucarella-frequenz: I would rename this as `QuantityT` or `TQ` or something like that, `GenericQuantity` sounds more like an alias for `Generic[Quantity]`.

```suggestion
QuantityT = TypeVar(
    "QuantityT", "Quantity", "Power", "Current", "Voltage", "Energy"
)
```

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -0,0 +1,656 @@
...
+        """
+        return math.isnan(self._base_value)
+
+    def isinf(self) -> bool:
+        """Return whether this quantity is infinite.
+
+        Returns:
+            Whether this quantity is infinite.
+        """
+        return math.isinf(self._base_value)
```
leandro-lucarella-frequenz, line 90: I wonder if we should add other convenience aliases from `math`, like `fabs()`, `isfinite()`. I didn't know about `fsum()` but it also looks interesting, although not too easy to integrate here.

src/frequenz/sdk/timeseries/_quantities.py
shsms: ~this was required based on an earlier version of the code.  You reviewed too quickly, before I could clean up the PR and mark it as ready for review. :-)~

shsms: mypy does check this.

src/frequenz/sdk/timeseries/_quantities.py
shsms: 😱 I didn't either, until I went looking for it for this PR. 😆

src/frequenz/sdk/timeseries/_quantities.py
shsms: we can change it if required.  no need to solve imaginary problems already.

src/frequenz/sdk/timeseries/_quantities.py
shsms: yup, left over from older version, removed.

src/frequenz/sdk/timeseries/_quantities.py
shsms: no, we actually need an exception,  because the `Quantity` base type will be used for dimensionless quantities.

shsms: handled differently, with a default None value.

src/frequenz/sdk/timeseries/_quantities.py
shsms: it won't always be there, for dimensionless quantities.

src/frequenz/sdk/timeseries/_quantities.py
shsms: No,  they recommend this: https://docs.python.org/3/library/constants.html#NotImplemented

src/frequenz/sdk/timeseries/_quantities.py
shsms: no,  `as_` is a function.  just `quantity.watts` is a property.  we can discuss later when I make the PR ready for review.

src/frequenz/sdk/timeseries/_quantities.py
shsms: I've added a `base_unit` method that returns an a string, or None for dimensionless quantities.

src/frequenz/sdk/timeseries/_quantities.py
shsms: true, done.

src/frequenz/sdk/timeseries/_quantities.py
shsms: Mypy can actually match `Self`, except when checking against the base class,  but we are treating the base class as a distinct dimensionless type as well.

We now have runtime checks.  Using `return NotImplemented`.  Those are able to provide nice meaningful error messages.

src/frequenz/sdk/timeseries/_quantities.py
shsms: I tried to keep only the commonly used units,  so that it doesn't get confusing.  For current, for example, `kA` feels unfamiliar.

But maybe you're right.  Strict adherence to the standards doesn't sound so bad.  I'll make the change.

src/frequenz/sdk/timeseries/_quantities.py
shsms: Should have used `_base_value` in the subclasses too,  there was a 10% perf difference, using the superclass constructor with _base_value.

But these can't be implemented in the superclass because not all quantities can get all operations.  For example voltages don't have addition.  And there might be other qtys with other requirements.

tests/timeseries/test_quantities.py
shsms: Tests have been improved.

src/frequenz/sdk/timeseries/_quantities.py
shsms: True, at least for some types,  I will get to it.

benchmarks/timeseries/periodic_feature_extractor.py
shsms: `Quantity` is being used as a dimensionless type,  so the resampler for example would return objects of this type.

src/frequenz/sdk/timeseries/_quantities.py
shsms: done

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
shsms: It is because of a python limitation.  Those types are generic,  but when putting them in a TypeVar,  we are unable to keep them generic.  For example, we can't say

```python
_GenericEngine = TypeVar(
    "_GenericEngine",
    "FormulaEngine[QuantityT]",
    "FormulaEngine3Phase[QuantityT]",
)
```
This also makes it slightly harder to use the `_GenericEngine` type.  But it is something.

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
shsms: Yup,  it was originally required because the type system is unable to establish a relationship between the 3 generic types in:

```python
class _ComposableFormulaEngine(
    ABC, Generic[_GenericEngine, _GenericHigherOrderBuilder, QuantityT]
):
```

And complains that only one combination is covered in the return statement, not all possible combinations.  I ran into trouble when trying overload methods as well,  don't remember the exact issues.

But anyway, the whole thing has become almost impossible to represent with the super limited Python type system, with the generic quantity types, for reasons mentioned here: https://github.com/frequenz-floss/frequenz-sdk-python/pull/422#discussion_r1226478466

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
shsms: We need to store it in places where objects of that type need to be created.  For example,  a FormulaEngine might take dimensionless values directly from the resampler and return Power values.  Because the interpreter doesn't understand typehints,  we have to give it an actual class to use.

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
shsms: yup,  found this out near the end.  There are a lot of uses of `typing.*` in these files,  will replace them all in a separate PR.  Tracked in #433

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine_pool.py
shsms: yup, documented at the top under the `if TYPE_CHECKING` section.

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine_pool.py
shsms: I think `Any` is used only in an intermediate commit,  only in the FormulaEnginePool's storage,  which was very weakly typed,  and using `Quantity` was causing errors that I didn't want to deal with in that commit.  I could have gone ahead and made them Quantity anyway, because we're not expecting intermediate commits to pass tests,  I guess I just didn't do it like that.

src/frequenz/sdk/timeseries/_quantities.py
shsms: Although might not be so easy to make general, because of types like `°C` for temperature.

src/frequenz/sdk/timeseries/_quantities.py
shsms: I've added `from __future__ import annotations`.

Regarding the mul and truediv methods,  I'm not sure it would always be possible for all types,  no a general method is not a good idea.   

Also from a performance point of view,  this might be bad,  especially if we also want to support division by multiple types,  and we have to call `isinstance` 3 times to get to the right type.  And then `isinstance(v, float)` will reject ints, but the typing system will accept ints, so have to have an explicit runtime type check for ints also.

Actually I'm having second thoughts about this whole PR because of the performance costs.  Or at least about all arithmetic operations on these types.  Because we are encouraging code like 

```python
c = current1 + current2
```

but that line takes around 15 times as long as if it were just integers.

src/frequenz/sdk/timeseries/_quantities.py
shsms: I think boolean methods are fine.  But I'd rather not have further methods that implicitly create new objects.

I'm actually thinking we should remove all arithmetic operations because they all need to create new Quantity objects.  The comparison methods are fine.  If people want to do math,  they'll have to convert to fundamental types first.

src/frequenz/sdk/timeseries/_quantities.py
shsms: any opinions on this? @matthias-wende-frequenz @llucax

llucax: I don't an opinion in particular, `exponent` sounds more natural and `base10_multiple` sounds more specific. I don't have a strong preference.

If we are going with the metric system explicitly, I would make `Quantity` provide the map instead of allowing the subclasses to use any arbitrary "exponent" (as I said in some other comment). It should be enough to specify the unit and then all the sub-units can be generated from that and the metric specification.

src/frequenz/sdk/timeseries/_quantities.py
llucax: I don't mean to change anything, just exposing through a `@property`, but OK. This also depends a bit on providing the map ourselves or letting the user provide it when creating new sub-classes.

src/frequenz/sdk/timeseries/_quantities.py
llucax: Oh, interesting, I didn't know about `NotImplemented` :+1:

llucax: I'm pretty sure there are a few places where we are using this wrongly (returning `False`) :grimacing:

src/frequenz/sdk/timeseries/_quantities.py
llucax: > I'll make the change.

:+1: Just a few more comments for completeness...

> I tried to keep only the commonly used units, so that it doesn't get confusing. For current, for example, kA feels unfamiliar.

Yeah, but the thing is that should never happen unless something is really wrong. We can still not provide `from_xxx()` and `as_xxx()` methods that are not natural, this will only be used for printing, and if for some weird reason you are working with 2389484A, I think it is still more legible to have 2.38 MA than 2389484 A, even if the MA looks weird, because it is pretty damn weird to have 2.38 MA. So IMHO we gain in safety (people can't add a map with some error), DRY (people don't need to create a new map over and over again for each unit) and predictability (all units use the same system).

> Although might not be so easy to make general, because of types like °C for temperature.

What about `°C`? You mean we should use Kelvin because that's the base SI unit or just that `K°C` also looks weird? If it is the later, IMHO is the same as I say above, that should never happen because otherwise we have burning batteries, which is a bigger problem that weirdness while reading logs :laughing: . If you mean the later, then :grimacing: . I guess we can make an exception and not use the SI unit for that.

shsms: Also this doesn't work for things like area, volume, etc,  because 1 sq.km is 1000000 sq.m.

src/frequenz/sdk/timeseries/_quantities.py
llucax: OK, not sure what do you mean about `quantity.watts` as we said we didn't want to call it `watts` because it can be more ambiguous, specially considering what `timedelta` does, where `.days` for example will only return the `days` part, not convert the whole delta to `days` as a `float`.

Anyway, it is not important, I think for methods that are a simple math operation, we can make them `@property` to cut a bit the syntactic noise, but it is just a matter of taste.  Maybe it would be good to have a convention, so we don't have different styles in the SDK.

src/frequenz/sdk/timeseries/_quantities.py
llucax: Why can't voltage have addition?

src/frequenz/sdk/timeseries/_quantities.py
shsms: And we're not using SI units everywhere,  for example energy is J,  but we use kWh,  which simplifies to a weird J/s*s. :D

tests/timeseries/test_quantities.py
llucax: Unless I'm missing it hidden somewhere, I still don't see the `Any` case being tested.

shsms: Looks like I added them only for comparisons: https://github.com/frequenz-floss/frequenz-sdk-python/blob/5c159c85c084742f54fc8fd90bac9c6b8c1a8344/tests/timeseries/test_quantities.py#L118-L123

I will add similar for all types.

src/frequenz/sdk/timeseries/_quantities.py
llucax: > Regarding the mul and truediv methods, I'm not sure it would always be possible for all types, no a general method is not a good idea.

Why it shouldn't be possible for all types?

>  And then isinstance(v, float) will reject ints, but the typing system will accept ints, so have to have an explicit runtime type check for ints also.

Why wouldn't you allow `int`s?

> Also from a performance point of view

Yeah, this is a hard question besides this case in particular. Still, if we are already supporting several types for these operations, I guess one more or one less is not a huge deal. Just for the records, we discussed the performance and we concluded that for live data this should be fine, for historical data we might need to rethink the design in general.

How do you feel about supporting `float` or `int` ([`Real`](https://peps.python.org/pep-3141/) I would say) after that conversation?

benchmarks/timeseries/periodic_feature_extractor.py
llucax: I think I mentioned this before somewhere, but I can't find it. Can't we make `Quantity` named something like `BaseQuantity` and make it an ABC and then have `Quantity = Power | Current | Voltage | Energy | ...`. Of course the downside of this is that if we add new units, we need to update the `Quantity` type alias, but for me it sounds weird to have a unit-less quantity.

Is it only needed for type hinting or do you see any other use cases for it?

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
llucax: OK, maybe it would be good to add a comment somewhere, even if it is vague about why these ignores are needed.

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
llucax: Same here.  Is the same file, so maybe just a global comment at the start of the file mentioning that some `type: ignore` were necessary because of X and Y issues is enough.

benchmarks/power_distribution/power_distributor.py
```diff
@@ -49,7 +50,7 @@ async def send_requests(batteries: Set[int], request_num: int) -> List[Result]:
     results_rx = battery_pool.power_distribution_results()
     result: List[Result] = []
     for _ in range(request_num):
-        await battery_pool.set_power(float(random.randrange(100000, 1000000)))
+        await battery_pool.set_power(Power(float(random.randrange(100000, 1000000))))
```
llucax, line 53: What do you think about disabling the *default* constructor for quantities with units (sub-classes) to force users to use the more explicit `from_xxx()` constructors?

tests/timeseries/test_quantities.py
```diff
@@ -0,0 +1,220 @@
...
+
+    assert f"{Fz1(1.024445, exponent=-6)}" == "0 Hz"
+    assert f"{Fz1(1.024445, exponent=-6):.6}" == "0.000001 Hz"
+    assert f"{Fz2(1.024445, exponent=-6)}" == "1.024 uHz"
+
+    assert f"{Fz1(1.024445, exponent=-12)}" == "0 Hz"
+    assert f"{Fz2(1.024445, exponent=-12)}" == "0 Hz"
+
+
+def test_comparison() -> None:
```
llucax, line 96: Suggestion: this might be a good candidate to do some property testing using `hypothesis`.

benchmarks/timeseries/periodic_feature_extractor.py
llucax: BTW, this could be explored in a followup PR (if we believe it might work and be better, I don't think it is worth to address it now).

examples/power_distribution.py
```diff
@@ -126,7 +127,9 @@ async def run(self) -> None:
                 if (datetime.now(timezone.utc) - time_data).total_seconds() > 30:
                     _logger.error("Active power data are stale")
                     continue
-                queue.put_nowait(active_power.value)
+                queue.put_nowait(
+                    None if not active_power.value else active_power.value.base_value
```
llucax, line 131: Looking at this, I think it might be better to leave `base_value` as private too, in code like this it is not clear at all what `base_value` is, is it Watt? Kilowatt? `active_power.value.as_watts()` is much more clear.

pyproject.toml
```diff
@@ -113,6 +113,9 @@ disable = [
   # disabled because it conflicts with isort
   "wrong-import-order",
   "ungrouped-imports",
+  # pylint's unsubscriptable check is buggy and is not needed because
+  # it is a type-check, for which we already have mypy.
+  "unsubscriptable-object",
```
llucax, line 118: FYI:

* https://github.com/frequenz-floss/frequenz-repo-config-python/issues/54

src/frequenz/sdk/timeseries/__init__.py
```diff
@@ -47,4 +48,13 @@
     "Sample",
     "Sample3Phase",
     "UNIX_EPOCH",
+    #
+    # Quantities
+    #
+    "Quantity",
+    "QuantityT",
```
llucax: I wouldn't export `QuantityT`, I think it is better to define `TypeVar`s explicitly in every other module. Internally in the SDK you can always import it from `_quantities` if you prefer. Maybe, as with properties, it would be good to have a guideline so we don't have to discuss this over and over again and we ensure a consistent style through the SDK.

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
```diff
@@ -411,10 +449,12 @@ class FormulaBuilder:
         following calls need to be made:
 
         ```python
-        channel = Broadcast[Sample]("channel")
+        from frequenz.sdk.timeseries._quantities import Quantity
```
llucax: Better not to expose internal modules in examples.

Also, besides allowing instantiating a unit-less `Quantity` or not, I would use a concrete quantity with a unit in examples to avoid encouraging users to use `Quantity` directly, which should be pretty rare.

Same for all the other examples in this file.

```suggestion
        from frequenz.sdk.timeseries import Power
```

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
```diff
@@ -411,10 +449,12 @@ class FormulaBuilder:
...
+
+        channel = Broadcast[Sample[Quantity]]("channel")
```
llucax:
```suggestion
        channel = Broadcast[Sample[Power]]("channel")
```

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
```diff
@@ -411,10 +449,12 @@ class FormulaBuilder:
...
         receiver_1 = channel.new_receiver("receiver_1")
         receiver_2 = channel.new_receiver("receiver_2")
-        builder = FormulaBuilder("addition")
+        builder = FormulaBuilder("addition", Quantity)
```
llucax:
```suggestion
        builder = FormulaBuilder("addition", Power)
```

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
```diff
@@ -506,8 +550,10 @@ def push_clipper(self, min_value: float | None, max_value: float | None) -> None
         For example, this clips the output of the entire expression:
 
         ```python
-        builder = FormulaBuilder("example")
-        channel = Broadcast[Sample]("channel")
+        from frequenz.sdk.timeseries._quantities import Quantity
+
+        builder = FormulaBuilder("example", Quantity)
+        channel = Broadcast[Sample[Quantity]]("channel")
```
llucax:
```suggestion
        from frequenz.sdk.timeseries import Power

        builder = FormulaBuilder("example", Power)
        channel = Broadcast[Sample[Power]]("channel")
```

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
```diff
@@ -522,8 +568,10 @@ def push_clipper(self, min_value: float | None, max_value: float | None) -> None
         And this clips the output of metric_2 only, and not the final result:
 
         ```python
-        builder = FormulaBuilder("example")
-        channel = Broadcast[Sample]("channel")
+        from frequenz.sdk.timeseries._quantities import Quantity
+
+        builder = FormulaBuilder("example", Quantity)
+        channel = Broadcast[Sample[Quantity]]("channel")
```
llucax:
```suggestion
        from frequenz.sdk.timeseries import Power

        builder = FormulaBuilder("example", Power)
        channel = Broadcast[Sample[Power]]("channel")
```

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -132,7 +133,7 @@ async def run() -> None:
     def __init__(  # pylint: disable=too-many-arguments
         self,
         size: timedelta,
-        resampled_data_recv: Receiver[Sample],
+        resampled_data_recv: Receiver[Sample[Quantity]],
```
llucax, line 136: Since the `MovingWindow` will probably instantiated for a very specific unit, shouldn't it be `Generic[QuantityT]` instead of using a unit-less `Quantity`? I see it more like the formulas and less like the resampler, they user should know by the time they create a moving window which unit they are using.

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -0,0 +1,686 @@
...
+        super().__init_subclass__()
+
+    @property
+    def base_value(self) -> float:
+        """Return the value of this quantity in the base unit.
+
+        Returns:
+            The value of this quantity in the base unit.
+        """
+        return self._base_value
```
llucax, line 61: As said in the benchmark comment I wouldn't provide direct access to the `base_value` as it is not clear what the unit is.

I guess this is only present because is the only way to get the value for unit-less quantities. Again, if we can't make `Quantity` an ABC, maybe we can call this `as_unitless` to math the other "getters" and make it more explicit that you are *losing* the unit information for other concrete types? I think people will feel much more discouraged to use `as_unitless` than `base_value` in concrete types with units.

Or maybe we can just override the property `base_value` with something that raises for types with a unit? Or maybe still have a ACB `BaseQuantity` and have a `UnitlessQuantity` type that also derives from it and doesn't have a unit associated and provides the `base_value` (and `BaseQuanity` doesn't). This might also help with the edge cases where we don't really want a `Power` for example to be mixed with a `Quantity` (the `type(self) is X and type(other) is X` vs `isinstance(x, X)` checks.

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -0,0 +1,686 @@
...
+        Returns:
+            A string representation of this quantity.
+        """
+        return self.__format__("")
+
+    def __format__(self, __format_spec: str) -> str:
+        """Return a formatted string representation of this quantity.
+
+        If specified, must be of this form: `[0].{precision}`.  If a 0 is not given, the
+        trailing zeros will be omitted.  If no precision is given, the default is 3.
```
llucax, line 120: I think it would be good to specify a bit better how units are chosen (when is the base unit used, when a "KU" or "mU" could be used, etc.).  In the example it is not clear if the used unit is related to how the object was created (`from_milliamperes()` -> `mA`) or if it is just coincidence and `mA` is used because is the unit that will produce the smallest number bigger than 1 for example.

src/frequenz/sdk/timeseries/_quantities.py
llucax: > Also this doesn't work for things like area, volume, etc, because 1 sq.km is 1000000 sq.m.

Well, this is only if you want to make a sq.km a *first-class citizen* with its own type, not if you just represent a sq.km simply as `sq_km = Distance.from_kilometer(1) * Distance.from_kilometer(1)`. If we would want to have an `Area`/`Volume` units, then we can allow providing a custom map or using the default map considering that these are powers.

tests/timeseries/test_periodic_feature_extractor.py
llucax, line 1: As the moving window, shouldn't this be parametrized on a `QuantityT` instead of using a unit-less `Quantity`?

tests/timeseries/test_quantities.py
llucax: I was referring more about the case where the type system sees both types as `Any`, but I'm not sure if it's relevant anymore, because the type system is not rejecting `Fz1(1.024445, exponent=0) < Fz2(1.024445, exponent=3)` anyways.

But still testing that operations fails between different unit it is still probably worth testing.

src/frequenz/sdk/timeseries/_quantities.py
llucax: I looked at https://google.github.io/styleguide/pyguide.html#213-properties and even when it is vague, it always talks about attributes, and I agree `as_xxx` is not really an attribute, so I agree to leave it as a function.

llucax (commented):
Again quite a few comments, but many are things to consider in the future, not something to address in this PR as it is already huge and there is much more value in having it merged soon than in discussing improvements for ages :)

I'm going to make a summary with the main points I see pending:

* `as_xxx()` as properties: I looked at https://google.github.io/styleguide/pyguide.html#213-properties and even when it is vague, it always talks about attributes, and I agree `as_xxx` is not really an attribute, so I agree to leave it as a function.
* default `exponent_unit_map` provided by `Quantity`: I still think this is the best option, but if it is controversial, we can leave it for a future PR, specially because it doesn't affect users API.
* About moving operations (like `__add__`) to `Quantity`: Same as above.
* About providing also `__mul__` and `__truediv__` with `other: float`: Same as above.
* About making  `Quantity` an ABC (and the different alternatives to do this): Also too big of a change for this PR and it doesn't affect the API, so we can do it separately if it makes sense.
* Add comments about `type: ignore`: in this PR.
* Disallow the default constructor for types with concrete units (`Power(10)`): I would address in this PR, as it shouldn't be a big change and affects the API. But could be left for another PR if it turns out not to be that trivial (or controversial)
* Using hypothesis: optional but in my experience it helped catching some corner cases. It took me a non-trivial amount of effort to do it right, this is why I wouldn't block the PR on this and has the *optional status*.
* Forbidding `.base_value` for types with concrete units: ideally in this PR but it might depend on what we decide for `Quantity` as ABC, so it is also fine to leave to address in the future. Maybe in this PR we can add a note to the doc saying that this should not be used for concrete types.
* Exporting/importing `TypeVar`s: we can create a discussion, but in this case in particular I think it shouldn't be exported in `timeseries` because is not something that users will need often, and if they need to write a generic function or class they can declare the `TypeVar` themselves.
* Internal imports and use concrete unit types in examples: fix in this PR.
* Moving window and feature extractor as generic types: fix in this PR (if appropriate).
* Clarify `__format__` behavior in the docstring: this PR.

For the stuff that you agree we should leave for another PR, can you either create new issues/discussions gathering the info in the relevant comments and then resolve them with a link to the new issue/discussion, or just add a comment saying you agree to leave for another PR and I'll do the issue/discussion creation and resolving.
*

src/frequenz/sdk/timeseries/_quantities.py
shsms: True, we can make it optional,  but will keep that for a subsequent PR.

src/frequenz/sdk/timeseries/_quantities.py
shsms: True,  it can have addition in serial connections,  and I guess they all can.  I was thinking only of parallel connections, for voltage. Will make it a common method.

shsms: done

tests/timeseries/test_quantities.py
shsms: The type system *is* rejecting that comparison, hence the `# type: ignore`.

I've added similar for addition and subtraction as well.

src/frequenz/sdk/timeseries/_quantities.py
shsms: >> And then isinstance(v, float) will reject ints, but the typing system will accept ints, so have to have an explicit runtime type check for ints also.
>
> Why wouldn't you allow ints?

We would allow ints,  my comment above is about how we would have to have two `isinstance` checks, one for float and one for int.

So it would look like (for Power):

```python
    def __truediv__(self, other: Current | Voltage | float) -> Voltage | Current | Power:
        if isinstance(other, Current):
            return Voltage(self._base_value / other._base_value, exponent=0)
        if isinstance(other, Voltage):
            return Current(self._base_value / other._base_value, exponent=0)
        if isinstance(other, float) or isinstance(other, int):
            return type(self)(self._base_value / other, exponent=0)
        raise TypeError(
            f"unsupported operand type(s) for /: '{type(self)}' and '{type(other)}'"
        )

Comment (shsms) [reply src/frequenz/sdk/timeseries/_quantities.py]: > How do you feel about supporting float or int ([Real](https://peps.python.org/pep-3141/) I would say) after that conversation?

Is this about the solution you proposed during the call?  Can you explain it again with a code example?

File: benchmarks/timeseries/periodic_feature_extractor.py (hunk 16 above)
Comment (shsms) [reply benchmarks/timeseries/periodic_feature_extractor.py:38]: > Is it only needed for type hinting or do you see any other use cases for it?

The reason it it like this is because in cases like the FeatureExtractor,  we need a fallback type that everything would decompose into.  So we pass any receiver to the FeatureExtractor,  and it would be treated as a `Quantity`.  Because from that point we'll only deal with numpy arrays, for example,  we can't propagate the types downstream.

File: src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py (hunk 18 above)
Comment (shsms) [reply src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py:245]: done

File: src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py (hunk 19 above)
Comment (shsms) [reply src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py:204]: done

File: benchmarks/power_distribution/power_distributor.py (hunk 27 above)
Comment (shsms) [reply benchmarks/power_distribution/power_distributor.py:53]: I like the idea of course,  but it makes it harder to have generic methods.  Unless it can be done with just using `__new__` and setting the params?

File: examples/power_distribution.py (hunk 29 above)
Comment (shsms) [reply examples/power_distribution.py:131]: The `base_value` is really just for the generic cases described here:  https://github.com/frequenz-floss/frequenz-sdk-python/pull/422#discussion_r1232042848

And for the dimensionless cases like reading from the resampler directly.

I guess the `power_distribution` example is no longer relevant anyway, because people should use the battery pool instead.  I have an in-progress branch for that,  so won't forget.  But will do it after this one is merged.

File: src/frequenz/sdk/timeseries/__init__.py (hunk 31 above)
Comment (shsms) [reply src/frequenz/sdk/timeseries/__init__.py]: true, removed.

File: src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py (hunk 32 above)
Comment (shsms) [reply src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py]: done

File: src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py (hunk 33 above)
Comment (shsms) [reply src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py]: done

File: src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py (hunk 34 above)
Comment (shsms) [reply src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py]: done

File: src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py (hunk 35 above)
Comment (shsms) [reply src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py]: done

File: src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py (hunk 36 above)
Comment (shsms) [reply src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py]: done

File: src/frequenz/sdk/timeseries/_quantities.py (hunk 11 above)
Comment (shsms) [reply src/frequenz/sdk/timeseries/_quantities.py:285]: https://github.com/frequenz-floss/frequenz-sdk-python/issues/449

File: src/frequenz/sdk/timeseries/_quantities.py (hunk 2 above)
Comment (shsms) [reply src/frequenz/sdk/timeseries/_quantities.py]: > I don't an opinion in particular, exponent sounds more natural and base10_multiple sounds more specific. I don't have a strong preference.

ok,  will keep the names as they are for now.  They are internal for the time being anyway.

> If we are going with the metric system explicitly, I would make Quantity provide the map instead of allowing the subclasses to use any arbitrary "exponent" (as I said in some other comment). It should be enough to specify the unit and then all the sub-units can be generated from that and the metric specification.

lets discuss here:  https://github.com/frequenz-floss/frequenz-sdk-python/pull/422#discussion_r1232042848

File: src/frequenz/sdk/timeseries/_quantities.py (hunk 38 above)
Comment (shsms) [reply src/frequenz/sdk/timeseries/_quantities.py:61]: > II think people will feel much more discouraged to use `as_unitless` than base_value in concrete types with units.

how about `as_base_unit` instead?  Because `unitless` is not clear that there's no transformations going on.

> Or maybe we can just override the property base_value with something that raises for types with a unit?

this is not possible, because then we won't be able to have all quantities decompose into the base values.  Then formula engines won't be able to get the base value from Power values, for example.  Or in the feature extractor.

> Or maybe still have a ACB BaseQuantity and have a UnitlessQuantity type 

No, we want any kind of objects to be de-composable to the unitless quantity.  If Power and UnitlessQuantity are in the same level in the hierarchy, then I can't have a type that accepts UnitlessQuantity and accept Power values.  More specific case here:  https://github.com/frequenz-floss/frequenz-sdk-python/pull/422#discussion_r1232042848

File: src/frequenz/sdk/timeseries/_quantities.py (hunk 39 above)
Comment (shsms) [reply src/frequenz/sdk/timeseries/_quantities.py:120]: Added a docu line, and added an additional example that makes it clear.

File: tests/timeseries/test_periodic_feature_extractor.py
Comment (shsms) [reply tests/timeseries/test_periodic_feature_extractor.py:1]: It turns out MovingWindow etc can't propagate types because the most likely output types for these classes is a numpy array.  So it doesn't have to be generic,  just needs to accept any thing that gets put in here.  And using the base Quantity type achieves that.

File: src/frequenz/sdk/timeseries/_moving_window.py (hunk 37 above)
Comment (shsms) [reply src/frequenz/sdk/timeseries/_moving_window.py:136]: answered here: https://github.com/frequenz-floss/frequenz-sdk-python/pull/422/commits/e4f118f4b29b82064880237978e322b014d617f5#r1234053164

File: benchmarks/power_distribution/power_distributor.py (hunk 27 above)
Comment (llucax) [reply benchmarks/power_distribution/power_distributor.py:53]: Yup, it should be possible to use `__new__` to construct the object.

It can also be done in a follow-up PR also, I wouldn't block this PR on this.

I think we did that with some other class. Ah, yeah, the `AsyncConstructible`:

* https://github.com/frequenz-floss/frequenz-sdk-python/blob/650d0586c73d8cda4f1f496540ace4484abebe79/src/frequenz/sdk/_internal/_asyncio.py#L36-L48
* https://github.com/frequenz-floss/frequenz-sdk-python/blob/650d0586c73d8cda4f1f496540ace4484abebe79/src/frequenz/sdk/timeseries/battery_pool/_component_metric_fetcher.py#L38-L63

File: tests/timeseries/test_quantities.py (hunk 14 above)
Comment (llucax) [reply tests/timeseries/test_quantities.py:171]: Oh, OK. Can't you just use a function that takes `Any` as arguments instead to test this and circumvent the type system to check the exception is raised? I think it makes it more clear what are you testing for, because using adding random `# type: ignore` shouldn't be something we really care about. But having the type relaxed to be `Any` is a real world case. Maybe you can even leave your example as is and just type-hint variables with any:

```py
 fz1: Any = Fz1(1.024445, exponent=0)
 fz2: Any = Fz2(1.024445, exponent=3)
 with pytest.raises(TypeError) as excinfo: 
     assert fz1 < fz2
 assert ( 
     excinfo.value.args[0] 
     == "'<' not supported between instances of 'Fz1' and 'Fz2'" 
 ) 
 ```

File: src/frequenz/sdk/timeseries/_quantities.py (hunk 15 above)
Comment (llucax) [reply src/frequenz/sdk/timeseries/_quantities.py]: Sure. This is all I mean:

```py
    def __truediv__(self, other: Current | Voltage | float) -> Voltage | Current | Power:
        if isinstance(other, Current):
            return Voltage(self._base_value / other._base_value, exponent=0)
        if isinstance(other, Voltage):
            return Current(self._base_value / other._base_value, exponent=0)
        if isinstance(other, numbers.Real):  # Only 1 extra `isinstance()` for all reals (`int` and `float` and whatever user-defined type that inherits from `numbers.Real`)
            return type(self)(self._base_value / other, exponent=0)
        raise TypeError(
            f"unsupported operand type(s) for /: '{type(self)}' and '{type(other)}'"
        )
```

```py
>>> import numbers
>>> isinstance(1, numbers.Real)
True
>>> isinstance(1.0, numbers.Real)
True
```

benchmarks/timeseries/periodic_feature_extractor.py
llucax: I still don't understand why you need the unit-less `Quantity` type, if you need it as a step to pass it to `numpy` or something like that, then you can *decompose into* a `float` already, right? But maybe we can discuss in a follow up discussion:

* https://github.com/frequenz-floss/frequenz-sdk-python/discussions/450

examples/power_distribution.py
llucax: The issue I see about it is it can be used in sub-classes, completely by-passing the unit checks. This is why it would be nice (if we don't go with https://github.com/frequenz-floss/frequenz-sdk-python/discussions/450) to find some way to make it available only in the base unit-less `Quantity`.

src/frequenz/sdk/timeseries/_moving_window.py
llucax: The link doesn't seem to be pointing to any file:line in particular, maybe you can mention the file:line in the comment so I can look for it?

src/frequenz/sdk/timeseries/_quantities.py
llucax: > how about `as_base_unit` instead? Because unitless is not clear that there's no transformations going on.

Maybe, I'm not sure if it is as discouraging. I think the root problem for me is you can have a `Quantity` without a unit in general and this is just a small consequence of that, so maybe we don't focus on this detail until we have the major design topic sorted out. By now I'm pretty sure I'm missing some fundamental use case that needs this. Maybe you can clarify it in a call and continue the conversation in https://github.com/frequenz-floss/frequenz-sdk-python/discussions/450).

tests/timeseries/test_periodic_feature_extractor.py
llucax: https://github.com/frequenz-floss/frequenz-sdk-python/discussions/450#discussioncomment-6221135

src/frequenz/sdk/timeseries/_quantities.py
llucax: * https://github.com/frequenz-floss/frequenz-sdk-python/discussions/451

llucax (approved):
OK, I created a couple of discussions:

* https://github.com/frequenz-floss/frequenz-sdk-python/discussions/450
* https://github.com/frequenz-floss/frequenz-sdk-python/discussions/451

Feel free to close all the discussions where I point to either of those 2 discussions.

The remaining points are all minor and non-blockers to me, so I will approve. If you decide not to go with the disabled `__init__`, please create a follow-up issue so we don't forget.

benchmarks/power_distribution/power_distributor.py
shsms: https://github.com/frequenz-floss/frequenz-sdk-python/issues/452

PR #426, opened 2023-06-06
ci: Fix building of tags

leandro-lucarella-frequenz:
Tags were not built because `branches-ignore` was added so only branches were built.

We also specify that only tags starting with `v` should produce a GitHub release (and pypi upload), just in case some tag is added that is not really a new version.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,6 +4,8 @@ on:
   merge_group:
   pull_request:
   push:
+    tags:
+      - '*'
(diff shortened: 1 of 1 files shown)


leandro-lucarella-frequenz:
Seems to be working (the build triggered on tags) in my fork: https://github.com/leandro-lucarella-frequenz/frequenz-sdk-python/actions/runs/5190701685

matthias-wende-frequenz (approved):

PR #437, opened 2023-06-14
Fix seconds in MovingWindow doc example

cwasicki:
Sleep uses seconds and the example does not work for a 5 minute window


llucax (approved):

llucax:
Queuing to merge, thanks!

PR #445, opened 2023-06-16
Fix ring buffer len

### LINKED ISSUE #442: OrderedRingBuffer.__len__ returning wrong values
### What happened?

The following function will trigger the assert

```
def test_len_with_gaps() -> None:
    """Test the length when there are gaps in the buffer."""
    buffer = OrderedRingBuffer(
        np.empty(shape=10, dtype=float),
        sampling_period=timedelta(seconds=1),
        align_to=datetime(1, 1, 1, tzinfo=timezone.utc),
    )

    for i in range(10):
        buffer.update(Sample(datetime(2, 2, 2, 0, 0, i, tzinfo=timezone.utc), i))
        assert len(buffer) == i + 1
```

It probably happens because gaps are not properly considered.

Returning the right values

### Affected version(s)

_No response_

### Affected part(s)

### Extra information

_No response_

Marenz:
- RingBuffer: Fix off-by-one error in gap calculation
- RingBuffer: Rename variables and add comments for more clarity
- RingBuffer: Decrease test samples to speedup a long test
- RingBuffer: Fix gaps being ignored in __len__()

fixes #442

### DIFF
--- a/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
+++ b/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
@@ -281,19 +281,18 @@ class OrderedRingBuffer(Generic[FloatArray]):
         return any(map(lambda gap: gap.contains(timestamp), self._gaps))
 
     def _update_gaps(
-        self, timestamp: datetime, newest: datetime, new_missing: bool
+        self, timestamp: datetime, newest: datetime, record_as_missing: bool
     ) -> None:
         """Update gap list with new timestamp.
 
         Args:
             timestamp: Timestamp of the new value.
             newest: Timestamp of the newest value before the current update.
-            new_missing: if true, the given timestamp will be recorded as missing.
-
+            record_as_missing: if `True`, the given timestamp will be recorded as missing.
         """
-        currently_missing = self.is_missing(timestamp)
+        found_in_gaps = self.is_missing(timestamp)
 
-        if not new_missing:
+        if not record_as_missing:
             # Replace all gaps with one if we went far into then future
             if self._datetime_newest - newest >= self._full_time_range:
                 self._gaps = [
@@ -301,19 +300,20 @@ class OrderedRingBuffer(Generic[FloatArray]):
                 ]
                 return
 
-            if not currently_missing and timestamp > newest + self._sampling_period:
+            # Check if we created a gap with the addition of the new value
+            if not found_in_gaps and timestamp > newest + self._sampling_period:
                 self._gaps.append(
                     Gap(start=newest + self._sampling_period, end=timestamp)
                 )
 
         # New missing entry that is not already in a gap?
-        if new_missing:
-            if not currently_missing:
+        if record_as_missing:
+            if not found_in_gaps:
                 self._gaps.append(
                     Gap(start=timestamp, end=timestamp + self._sampling_period)
                 )
         elif len(self._gaps) > 0:
-            if currently_missing:
+            if found_in_gaps:
                 self._remove_gap(timestamp)
 
         self._cleanup_gaps()
--- a/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
+++ b/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
@@ -517,10 +517,24 @@ class OrderedRingBuffer(Generic[FloatArray]):
         if self._datetime_newest == self._DATETIME_MIN:
             return 0
 
+        # Sum of all elements in the gap ranges
+        sum_missing_entries = max(
+            0,
+            sum(
+                (
+                    gap.end
+                    # Don't look further back than oldest timestamp
+                    - max(gap.start, self._datetime_oldest)
+                )
+                // self._sampling_period
+                for gap in self._gaps
+            ),
+        )
+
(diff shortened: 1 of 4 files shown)


src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -278,39 +279,41 @@ def is_missing(self, timestamp: datetime) -> bool:
...
+        self, timestamp: datetime, newest: datetime, should_be_in_gaps: bool
     ) -> None:
         """Update gap list with new timestamp.
 
         Args:
             timestamp: Timestamp of the new value.
             newest: Timestamp of the newest value before the current update.
-            new_missing: if true, the given timestamp will be recorded as missing.
-
+            should_be_in_gaps: if true, the given timestamp will be recorded as missing.
```
llucax: Silly but just for consistency:

```suggestion
            should_be_in_gaps: If `True`, the given timestamp will be recorded as missing.
```

Also from the docs it looks like maybe `record_as_missing` might be even more clear, I find `should_be_in_gaps` not super clear TBH.

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -278,39 +279,41 @@ def is_missing(self, timestamp: datetime) -> bool:
...
-        currently_missing = self.is_missing(timestamp)
+        found_in_gaps = self.is_missing(timestamp)
 
-        if not new_missing:
+        if not should_be_in_gaps:
             # Replace all gaps with one if we went far into then future
-            if self._datetime_newest - newest >= self._time_range:
+            if self._datetime_newest - newest >= self._sampling_period * len(
+                self._buffer
+            ):
```
llucax: So the bug fix in the first commit is here, right? Two comments:

1. In terms of commit, maybe it would have made much more obvious where the bug fix was if you split the commit in removing `self._time_range` first (which still has the bug) and then a second commit just fixing the bug.
2. Would it be too complicated to add a test that fails before the bug fix and succeeds after to make sure this doesn't regress in the future?

llucax:
This probably needs a release notes entry, but probably it should be mentioned as a bug fix in the `MovingWindow`, as the ring buffer is only exposed via it, right? @matthias-wende-frequenz

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
Marenz: I mean, we have a test like that, the added test will fail if this logic fails

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -145,7 +144,9 @@ def update(self, sample: Sample[QuantityT]) -> None:
         # Update timestamps
         prev_newest = self._datetime_newest
         self._datetime_newest = max(self._datetime_newest, timestamp)
-        self._datetime_oldest = self._datetime_newest - self._time_range
+        self._datetime_oldest = (
+            self._datetime_newest - (len(self._buffer) - 1) * self._sampling_period
+        )
```
shsms, line 150: but now the calculation has moved from the `init` method to the hot path.  maybe we can find a better name instead?  like `buffer_time_range`, `buffer_duration`, etc?

shsms: or maybe `max_buffer_duration`,  or `max_time_span`?

Marenz: I mean, it's a rather simple calculation.. it shouldn't impact the total runtime noticeable

shsms: Maybe,  but they add up.  We are of course adding additional features that make things slower, so that we can have a nicer external interface, but here it looks like an easy performance win, however tiny,  especially because there's no async here, and can be reused for the historical pipeline, if necessary, etc.

But I'll leave it up to you.

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
llucax: And why didn't tests failed before? I'm lost.

Marenz: Basically this PR fixes two bugs:
* In __len__() we forgot to consider gaps, that was added and a test for it as well.
* For specific situations, the gaps were wrong which the bug in __len__() exposed and thus is also tested in the same tests as for __len__() 

But after your comments I added an extra test for the second (gap) bug as well now

llucax: OK, thanks for the clarification and the extra test :+1:

llucax (commented):
LGTM, leaving the approval to @shsms

shsms (dismissed):

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
Marenz: updated

shsms (approved):

Marenz:
Someone please merge this as I don't have permission.
image

PR #459, opened 2023-06-21
Expose only usable SoC and Capacity through the BatteryPool

### LINKED ISSUE #417: Expose only the Usable SOC and Capacity values from the BatteryPool
### What's needed?

The BatteryPool currently exposes aggregated for all batteries in the pool:

1. total capacity
2. average SoC and SoC bounds, weighted by capacity.

Both of these can be scaled to the usable SoC bounds, so that the calculations are easier for users.  We could also have methods with `*_unscaled` in their names, to get the unscaled values.

So if total capacity is 100kWh, and soc bounds are 10-90%, and soc is 10%,  then BatteryPool should report a capacity of 80kWh and 0% soc.  And it doesn't have to report scaled soc bounds, because that is always 0-100%.

### Proposed solution

_No response_

### Use cases

_No response_

_No response_

### Additional context

_No response_

shsms:
With this PR,
 - soc values from the battery pool are always normalized to the 0-100% range,  and they are now streamed as `Sample[Quantity]` objects.
 - only the usable capacity is streamed, eliminating the need for capacity bounds,  so they are now streamed as `Sample[Energy]` objects.

Closes #417

### DIFF
--- a/src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
+++ b/src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
@@ -206,9 +206,7 @@ class CapacityCalculator(MetricCalculator[CapacityMetrics]):
             High level metric calculated from the given metrics.
             Return None if there are no component metrics.
         """
-        result = CapacityMetrics(
-            timestamp=_MIN_TIMESTAMP, total_capacity=0, bound=Bound(lower=0, upper=0)
-        )
+        result = CapacityMetrics(timestamp=_MIN_TIMESTAMP, total_capacity=0)
 
         for battery_id in working_batteries:
             if battery_id not in metrics_data:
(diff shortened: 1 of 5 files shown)


src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
```diff
@@ -323,11 +323,14 @@ def calculate(
             ):
                 continue
 
+            usable_capacity = capacity * (soc_upper_bound - soc_lower_bound)
```
Marenz: It's a bit confusing that this is named `usable_capacity` when it is in fact just capacity multiplied with with a number `<100`..

shsms: But what else can we call it?

If rated capacity is 100kWh,  and soc bounds are 10-90%,  then usable capacity is 80kWh.  Because if we start discharging when the battery is full (i.e 90%) and stop when it is empty (i.e 10%), we'll get 80kWh out of it.

Maybe a comment with this example helps?

shsms: oh you mean because it is multiplied by numbers > 1.  Yes,  that's just to do fewer calculations,  and it gets cancelled out in a later stage.  I can add a comment clarifying that.

Marenz: You used the same variable name in a different place but also did `/ 100` there. I just feel that's a confusing inconsistency.
Maybe call it `capacity_multiplied`?

Marenz: or `usable_capacity_multiplied`

Marenz: `capacity_calculation_step`

shsms: true, I've added a `_x100` prefix to all the related variables, and added a comment explaining what's going on.

Marenz: Very nice!

Marenz (approved):

PR #465, opened 2023-06-26
Disable default constructor in specialized `Quantity` types

### LINKED ISSUE #452: Disable default constructor for derived `Quantity` types
Allow only the `from_` methods for constructing these objects.

From @llucax:

Yup, it should be possible to use `__new__` to construct the object.

It can also be done in a follow-up PR also, I wouldn't block this PR on this.

I think we did that with some other class. Ah, yeah, the `AsyncConstructible`:

* https://github.com/frequenz-floss/frequenz-sdk-python/blob/650d0586c73d8cda4f1f496540ace4484abebe79/src/frequenz/sdk/_internal/_asyncio.py#L36-L48
* https://github.com/frequenz-floss/frequenz-sdk-python/blob/650d0586c73d8cda4f1f496540ace4484abebe79/src/frequenz/sdk/timeseries/battery_pool/_component_metric_fetcher.py#L38-L63

_Originally posted by @llucax in https://github.com/frequenz-floss/frequenz-sdk-python/pull/422#discussion_r1234195497_

shsms:


src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -190,7 +190,9 @@ def __add__(self, other: Self) -> Self:
         """
         if not type(other) is type(self):
             return NotImplemented
-        return type(self)(self._base_value + other._base_value)
+        summe = type(self).__new__(type(self))
```
llucax, line 193: Did you asked @matthias-wende-frequenz -GTP to write this code? :de:

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -275,6 +279,38 @@ def __eq__(self, other: object) -> bool:
...
+        raise TypeError(f"No default constructor allowed for {cls.__name__}")
+
+    def __init__(self: Any, *args: Any, **kwargs: Any) -> None:
+        raise TypeError(
+            f"Use of default constructor NOT allowed for {cls.__name__}. "
+            f"Use one of the `{cls.__name__}.from_*()` methods instead."
+        )
+
+    cls.__init__ = __init__  # type: ignore[method-assign]
+    return cls
```
llucax: I'm a bit scared about patching classes via a decorator now, after the headaches this brought with the `@actor` decorator. I guess for this case where you are only adding a `__init__` that probably won't confuse `mypy` it should be OK, but I fear in the future someone might think it is a good idea to add some other common method here and things go south again.

Other approaches that might be more `mypy` friendly are just defining a base type `_NoDefaultConstructible` and inherit from it (but that won't catch if the error if someone adds an `__init__` to it, although it would be weird to inherit from `_NoDefaultConstructible` and add an `__init__`. But another alternative would be metaclasses, I'm not sure how `mypy` feels about those, but I think is the most widely method used to manipulate classes, so my guess is that probably better than decorators.

For example: https://stackoverflow.com/questions/8212053/private-constructor-in-python/64682734#64682734

This also adds the `_create()` method that could actually be useful to abstract sub-classes from dealing with `__new__()` directly.

llucax (commented):
Nice! Only 1 (serious) comment.

src/frequenz/sdk/timeseries/_quantities.py
shsms: python apparently has a `sum` keyword,  so pylint doesn't like it.  And of course,  this was inspired by @matthias-wende-frequenz GPT's original creation 🇩🇪

llucax: Is a built-in, not a keyword, and it is probably `pylint` being anal, it doesn't like people redefining built-ins :+1:

src/frequenz/sdk/timeseries/_quantities.py
shsms: I've replaced it with a meta class.  Didn't add a private constructor,  because it has to be typed with `Any`,  whereas with `__new__`, we get proper typing.

llucax (dismissed):
:tada: LGTM.

I noticed there is no `RELEASE_NOTES.md` entry for this (I mean the whole quantities new feature), so maybe you can add some in this PR? We should have some notes before the release :)

shsms:
> I noticed there is no RELEASE_NOTES.md entry for this (I mean the whole quantities new feature), so maybe you can add some in this PR? We should have some notes before the release :)

added for all my recently merged PRs.

llucax (approved):

PR #470, opened 2023-06-29
A `Percentage` quantity

shsms:
Percentage is is a unitless quantity,  but with a symbol,  and with multiple representations (as pct, or as a fractional value),  so it somehow fits as a `Quantity`.  The biggest benefit would allow for better user experience for `BatteryPool.soc`,  for example:

```python
soc_sample = await battery_pool.soc.new_receiver().receive()
soc_pct = soc_sample.value.base_value   ## old
soc_pct = soc_sample.value.as_percent()  ## new

### DIFF
--- a/src/frequenz/sdk/timeseries/_quantities.py
+++ b/src/frequenz/sdk/timeseries/_quantities.py
@@ -9,7 +9,9 @@ import math
 from datetime import timedelta
 from typing import Any, NoReturn, Self, TypeVar, overload
 
-QuantityT = TypeVar("QuantityT", "Quantity", "Power", "Current", "Voltage", "Energy")
+QuantityT = TypeVar(
+    "QuantityT", "Quantity", "Power", "Current", "Voltage", "Energy", "Percentage"
+)
 
 
 class Quantity:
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/frequenz/sdk/timeseries/_quantities.py [hunk 1]
```diff
@@ -704,3 +706,55 @@ def __truediv__(self, other: timedelta | Power) -> Power | timedelta:
         raise TypeError(
             f"unsupported operand type(s) for /: '{type(self)}' and '{type(other)}'"
         )
+
+
+class Percentage(
+    Quantity,
+    metaclass=_NoDefaultConstructible,
+    exponent_unit_map={0: "%"},
```
llucax, line 714: Super picky, but maybe you could add ‰ (per mille) and ‱ (per ten thousand) :grimacing: :sunglasses:

llucax (dismissed):

src/frequenz/sdk/timeseries/_quantities.py
shsms: I have to read up on what those are,  never seen them before.  Maybe later,  or you can make a PR at some point as well,  now that the basic requirements for `soc` are satisfied.

shsms:
oops, I just pushed some docstring improvements. @llucax could you approve again?

src/frequenz/sdk/timeseries/_quantities.py
llucax: Yeah, I wasn't being serious, I think they will cause more confusion than being useful :)

llucax (approved):

PR #475, opened 2023-07-03
Limit pydantic to v1

shsms:
It appears v2 has breaking changes which we need to adjust for later.

This PR adds the fix to the v0.16.x branch, and also pins the version of the grpc-stubs package.


llucax (approved):

llucax (changes requested):
Wait, no. What do you mean 0.16?

llucax:
Oh, you really mean a very old version of the SDK? It seems like many other things are breaking. I don't think we pinned deps in older versions, so it might be a huge challenge to get them working at all.

shsms:
> Oh, you really mean a very old version of the SDK? It seems like many other things are breaking. I don't think we pinned deps in older versions, so it might be a huge challenge to get them working at all.

we did.  I think only the grpc typing is failing.  I will try and fix it.

llucax (approved):

PR #478, opened 2023-07-03
Store a reference to the engine in receiver objects

shsms:
This PR fixes a bug which occurs when creating and using a higher order formula engine, without holding a reference to it.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -37,3 +37,5 @@ New `Quantity` types! These types can have units (power, current, voltage, etc.)
 - Two bugs in the ring buffer which is used by the `MovingWindow` class were fixed:
   - `len(buffer)` was not considering potentially existing gaps (areas without elements) in the buffer.
   - A off-by-one error in the gap calculation logic was fixed that recorded a gap when there was none if an element with a future timestamp was added that would create a gap of exactly 1.
+
+- A formula engine lifetime issue, when creating higher order formula receivers without holding on to a reference to the engine, was fixed.
--- a/src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
+++ b/src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
@@ -360,7 +360,19 @@ class FormulaEngine(
         if self._task is None:
             self._task = asyncio.create_task(self._run())
 
-        return self._channel.new_receiver(name, max_size)
+        recv = self._channel.new_receiver(name, max_size)
+
+        # This is a hack to ensure that the lifetime of the engine is tied to the
+        # lifetime of the receiver.  This is necessary because the engine is a task that
+        # runs forever, and in cases where higher order built for example with the below
+        # idiom, the user would hold no references to the engine and it could get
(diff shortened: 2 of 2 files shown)


matthias-wende-frequenz (approved):

matthias-wende-frequenz (dismissed):

llucax (changes requested):
I believe this is not necessary. Let's discuss it first.

llucax (commented):
OK, we agreed this is necessary because tasks as saved as weak references by the loop (https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task). I would just add the link to the docs saying this for extra clarity and for me it is good to go.

If you create a new specific formula engine receiver that has an explicit formula engine attribute, then this wouldn't be a hack, right? You say this is a hack only because we are sticking an attribute to a receiver that doesn't have one, right?

shsms:
> we agreed this is necessary because tasks as saved as weak references by the loop (https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task).

right, so without this, the engine holds a strong reference to the task that runs the formula evaluator.  But there might be cases where users don't have a reference to the engine,  like in the example above.

Then the user would have a reference to the receiver,  and it will have a reference to the channel,  but the channel doesn't hold on to anything,  so the engine would be dangling, ready for clean up.

> I would just add the link to the docs saying this for extra clarity and for me it is good to go.

done.

> If you create a new specific formula engine receiver that has an explicit formula engine attribute, then this wouldn't be a hack, right? You say this is a hack only because we are sticking an attribute to a receiver that doesn't have one, right?

I think if we do that, we'd be transferring the hack to the receiver,  because it would have to have generics to get the typing right, not that we're short on typing hacks in this package.  Also,  it won't be enough to just create a receiver,  we'd have to create a specialized channel also,  because the receivers are actually created by the channels,  and the `Broadcast` channel would only create its default receiver, unless it is patched or overridden.  Right now,  it is a hack, but it is at least easy and readable.

matthias-wende-frequenz:
> Right now, it is a hack, but it is at least easy and readable.

given that this actually gives a dangling reference a place, I wouldn't call it a hack but rather a bugfix.

shsms:
> I wouldn't call it a hack but rather a bugfix.

It is a hack, because we are injecting an attribute with an undocumented name that can't be found anywhere in the channels repo.  But as @llucax accurately put it,

> the hack is not saving the reference but **how** are you saving the reference,

But yes,  it is also a bug fix.  So I've updated the PR description to say just the what, and not the how.

llucax:
> I think if we do that, we'd be transferring the hack to the receiver, because it would have to have generics to get the typing right, not that we're short on typing hacks in this package. Also, it won't be enough to just create a receiver, we'd have to create a specialized channel also, because the receivers are actually created by the channels, and the Broadcast channel would only create its default receiver, unless it is patched or overridden. Right now, it is a hack, but it is at least easy and readable.

Yeah, true, it is not that easy to make it proper into the receiver. I didn't think it was worth writing a new receiver to make it less hacky, I think the hack is super contained and hidden, but more thinking about it there was a la hacky way to fix the bug.

llucax (approved):

PR #479, opened 2023-07-03
replaced parse_raw_as with parse_obj_as, to work with pydantic v2

jack-herrmann:
this pull request is a work in progress, please bear with me


llucax:
I created #500, so if we don't want to proceed with this we can close it now.

shsms:
I agree,  we can leave this in maintenance mode until there's cloud-sync support, or until we start implementing schemas+validations.

PR #483, opened 2023-07-05
Add methods for sending individual mock component data messages in tests

shsms:
With this PR, there will be no remaining test dependencies on the flaky automatic raw-data streaming functionality of `MockMicrogrid`.  The benchmarks for the data sourcing actor still depends on this, but that can be dealt with later.

This PR also allows for the tests for the power distributing actor to use `MockMicrogrid`, because that would make it easy for the tests to be run on many different component graph configs,  rather than the single config it uses now.

### DIFF
--- a/tests/timeseries/mock_microgrid.py
+++ b/tests/timeseries/mock_microgrid.py
@@ -105,8 +105,8 @@ class MockMicrogrid:  # pylint: disable=too-many-instance-attributes
         self._streaming_tasks: list[asyncio.Task[None]] = []
         self._start_meter_streaming(4)
 
-    async def start_mock_datapipeline(self, mocker: MockerFixture) -> None:
-        """Start the MockDataPipeline."""
+    async def start(self, mocker: MockerFixture) -> None:
+        """Init the mock microgrid client and start the mock resampler."""
         self.init_mock_client(lambda mock_client: mock_client.initialize(mocker))
         self.mock_resampler = MockResampler(
             mocker,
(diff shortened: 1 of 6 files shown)


tests/timeseries/mock_microgrid.py
```diff
@@ -363,12 +345,104 @@ def add_ev_chargers(self, count: int) -> None:
...
+        Args:
+            values: list of active power values for each ev charger.
+        """
+        assert len(values) == len(self.evc_ids)
+        timestamp = datetime.now(tz=timezone.utc)
+        for comp_id, value in zip(self.evc_ids, values):
+            await self._microgrid.send(
+                EvChargerDataWrapper(
+                    component_id=comp_id,
+                    timestamp=timestamp,
```
daniel-zullo-frequenz, line 425: The timestamp will be always the same for all the samples in all these new methods, I'm just checking if this the behavior we want  here

daniel-zullo-frequenz (commented):
I only have a question and LGTM

tests/timeseries/mock_microgrid.py
shsms: I think that's fine for now, because we're not using this for testing anything that relies on the timestamp.

But there are some cases where we have to test the behaviour of the battery status tracker with stale data.  So the timestamp param will become configurable in the next PR.

daniel-zullo-frequenz: cool, thanks for the clarification!

daniel-zullo-frequenz (approved):

PR #485, opened 2023-07-05
Add BatteryPool.temperature and quantity Temperature

### LINKED ISSUE #466: `BatteryPool` should expose a temperature receiver
### What's needed?

By migrating away from `PowerDistributor` and use of `BatteryData` / `InverterData`, the only thing missing is access to the `temperature` of a battery(-pool).

### Proposed solution

`BatteryPool.temperature` should stream objects that contain the min/max/avg block temperatures of all batteries in the pool.  The returned objects could implement `__iter__`, so they can be directly used in a loop.  But it is nice to use an class, rather than a list, because later on we can add more methods for finding the hottest block, avg of all batteries, etc.

The fields could probably be called {temp_lowest/temp_highest/temp_avg} to indicate they are not bounds, but actual measurements.

This can be implemented with the same approach as `soc` and `capacity`.

### Use cases

Analogous to `soc`, `power`, `power_bounds`, `capacity`, add a `temperature` receiver.

still depend on `BatteryData` and legacy APIs :)

### Additional context

This is my second attempt to migrate our PQ actor to the BatteryPool :)

christianparpart:
Closes #466.

### DIFF
--- a/src/frequenz/sdk/timeseries/_quantities.py
+++ b/src/frequenz/sdk/timeseries/_quantities.py
@@ -10,7 +10,14 @@ from datetime import timedelta
 from typing import Any, NoReturn, Self, TypeVar, overload
 
 QuantityT = TypeVar(
-    "QuantityT", "Quantity", "Power", "Current", "Voltage", "Energy", "Percentage"
+    "QuantityT",
+    "Current",
+    "Energy",
+    "Percentage",
+    "Power",
+    "Quantity",
+    "Temperature",
+    "Voltage",
 )
 
 
@@ -299,6 +306,38 @@ class _NoDefaultConstructible(type):
             f"{cls.__module__}.{cls.__qualname__}, "
             f"use one of the `{cls.__name__}.from_*()` methods instead."
         )
+
+
+class Temperature(
+    Quantity,
+    metaclass=_NoDefaultConstructible,
+    exponent_unit_map={
+        0: "°C",
+    },
+):
+    """A temperature quantity (in degrees Celsius)."""
+
+    @classmethod
+    def from_celsius(cls, value: float) -> Self:
+        """Initialize a new temperature quantity.
+
+        Args:
+            value: The temperature in degrees Celsius.
+
+        Returns:
+            A new temperature quantity.
+        """
+        power = cls.__new__(cls)
+        power._base_value = value
+        return power
+
+    def as_celsius(self) -> float:
+        """Return the temperature in degrees Celsius.
+
+        Returns:
+            The temperature in degrees Celsius.
+        """
+        return self._base_value
 
 
 class Power(
--- a/src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
+++ b/src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
@@ -14,7 +14,7 @@ from ...microgrid import connection_manager
 from ...microgrid.component import ComponentCategory, ComponentMetricId, InverterType
 from ...timeseries import Energy, Percentage, Sample, Temperature
 from ._component_metrics import ComponentMetricsData
-from ._result_types import Bound, PowerMetrics
+from ._result_types import Bound, PowerMetrics, TemperatureMetrics
 
 _logger = logging.getLogger(__name__)
 _MIN_TIMESTAMP = datetime.min.replace(tzinfo=timezone.utc)
@@ -59,7 +59,7 @@ def battery_inverter_mapping(batteries: Iterable[int]) -> dict[int, int]:
 
 # Formula output types class have no common interface
 # Print all possible types here.
-T = TypeVar("T", Sample[Percentage], Sample[Energy], PowerMetrics)
+T = TypeVar("T", Sample[Percentage], Sample[Energy], PowerMetrics, TemperatureMetrics)
 
 
 class MetricCalculator(ABC, Generic[T]):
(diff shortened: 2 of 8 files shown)


src/frequenz/sdk/timeseries/battery_pool/battery_pool.py
```diff
@@ -373,6 +378,24 @@ def soc(self) -> MetricAggregator[Sample[Percentage]]:
 
         return self._active_methods[method_name]
 
+    @property
+    def temperature_max(self) -> MetricAggregator[Sample[Temperature]]:
```
llucax: I think we need some sort of convention for metrics that have a max/min/avg, etc. Adding `temperature_min`, `temperature_avg`, etc. doesn't look super nice. A few alternatives that come to mind:

* `pool.temperature.max`: A bit annoying as we need to add an aggregation (some extra class)
* `pool.temperature(MAX)`: A bit annoying that it is not a property like the other metrics

This might also apply to apparent/real/reactive/active power for example, which we have discussed in the last SDK meeting.

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -301,6 +308,16 @@ def __call__(cls, *_args: Any, **_kwargs: Any) -> NoReturn:
         )
 
 
+class Temperature(
```
llucax, line 311: Could you split adding the `Temperature` quantity into a separate commit? This way it is easier to focus on a particular change during the review.

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -301,6 +308,16 @@ def __call__(cls, *_args: Any, **_kwargs: Any) -> NoReturn:
...
+    Quantity,
+    metaclass=_NoDefaultConstructible,
+    exponent_unit_map={
+        0: "°C",
+    },
+):
+    """A temperature quantity (in degrees Celsius)."""
```
llucax, line 318: You should add a `from_celsius()` constructor and `as_celsius()` as in the other quantities, so users will always have to specify the used unit to create and to use a value.

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
```diff
@@ -234,6 +234,70 @@ def calculate(
...
+            metrics = metrics_data[battery_id]
+            temperature = metrics.get(ComponentMetricId.TEMPERATURE_MAX)
+            if temperature is None:
+                continue
+            timestamp = max(timestamp, metrics.timestamp)
+            temperature_max = max(temperature_max, temperature)
+        if timestamp == _MIN_TIMESTAMP:
+            return None
+
+        return Sample(timestamp=timestamp, value=Temperature(value=temperature_max))
```
llucax: This constructor shouldn't work, as the default constructor is disabled.

```suggestion
        return Sample(timestamp=timestamp, value=Temperature.from_celsius(temperature_max))
```

llucax (commented):
This is also missing tests.

src/frequenz/sdk/timeseries/battery_pool/battery_pool.py
christianparpart: I've just pushed over it, i went out of draft one push to early

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
```diff
@@ -57,9 +58,19 @@ def battery_inverter_mapping(batteries: Iterable[int]) -> dict[int, int]:
...
 
 
+@dataclass
+class TemperatureMetrics:
+    """Container for temperature metrics."""
+
+    timestamp: datetime
+    max: Temperature
+    min: Temperature
+    avg: Temperature
```
llucax: I think this is too specific and breaking `Sample` as the main type used by formula engines. If we go this route I would much rather see a generic `SampleRange` or something like that, that provides the `max`,  `min` and `avg` similar to `Sample3Phase`. But them I'm not sure if then this can be easily composed using formulas, as probably formulas expect only one value per sample. The more I think about this approach of packed samples the more I think it is not a good idea. When we designed the pipeline we specifically agreed to make each metric a single value, otherwise we could have packed all metrics of a single component as one single sample, but there were issues with that (which I can't remember right now, although formula composition seems to be a good one).

But I guess @shsms should clarify this as he is the author of both `Sample3Phase` and the formula engine.

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
```diff
@@ -234,6 +245,99 @@ def calculate(
...
+        timestamp = _MIN_TIMESTAMP
+        temperature_sum: float = 0
+        temperature_min: float = float("inf")
+        temperature_max: float = 0
+        temperature_count: int = 0
+        for battery_id in working_batteries:
+            if battery_id not in metrics_data:
+                continue
+            metrics = metrics_data[battery_id]
+            temperature = metrics.get(ComponentMetricId.TEMPERATURE_MAX)
```
llucax, line 309: I don't get this, so you are actually only getting the `TEMPERATURE_MAX` metric and then calculating the min, max and avg for `TEMPERATURE_MAX`? If so I find this very misleading that the `battery_pool` metric is then called simply `temperature`. And I don't think the formula engine is the place to do calculations over a single metric, this should be probably done outside of the battery pool, as again is something we might want to do for many other stuff than just the `TEMPERATURE_MAX`.

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
```diff
@@ -234,6 +245,99 @@ def calculate(
...
+                constructor.
+
+        Returns:
+            High level metric calculated from the given metrics.
+            Return None if there are no component metrics.
+        """
+        timestamp = _MIN_TIMESTAMP
+        temperature_sum: float = 0
+        temperature_min: float = float("inf")
+        temperature_max: float = 0
```
llucax: Can't a temperature be negative? I think the lack of a value is much better represented by `None`.

src/frequenz/sdk/timeseries/battery_pool/battery_pool.py
llucax: OK, I gave this another look but GitHub said the diff was updated again while I was writing comments, so I will back off now, please let us know when the PR has settled and it is ready for a review :)

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
christianparpart: sure, default-initializaed to negative infinity now. thx.

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
christianparpart: The `temperature_max` i'm getting out of that component is the max temperature for that particular battery (which seems to be usually a *group* of battery cells, and the max of THAT is the value I get). that is the temperature I am getting per battery component.

Out of that value, given the fact that we're talking about BatteryPool here, that may contain more than one battery in it, i'm calculating the min/avg/max metrics for the set of all batteries in this pool.

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
christianparpart: So what i could do instead is unfold the idea and expose a `BatteryPool.temperature_max` along with `...temperature_min` and `...temperature_avg` receivers. This way we do not feed the concern you're raising. I'm fine with that.

christianparpart: p.s.: that means i can refactor back to `Sample[Tenmperature]` that way then, too, if you agree.

shsms: This won't work,  because of the throttling provided by `SendOnUpdate`.  Temperatures for some components would get dropped.  It might be better to use something like this:

``` python
class BatteryTemperatures:
    @dataclass
    class Metrics:
        component_id: int
        lowest: Temperature
        """Temperature of block with the lowest temperature."""
        highest: Temperature
        """Temperature of block with the highest temperature."""
        average: Temperature
        """Average temperature of all blocks."""

    def __init__(self, values: list[BatteryTemperatures.Metrics]):
        self._values = values

    def __iter__(self) -> Iterator[BatteryTemperatures.Metrics]:
        return iter(self._values)
```

also it was suggested that calling them `max` and `min` was confusing, because they sound like bounds.  `highest` and `lowest` might be better.

shsms: @llucax formulas can only be applied on resampled values.  For a generic SampleRange with highest/lowest, etc.  I don't mind,  if you think there will be additional use cases for these in the future.

shsms: > p.s.: that means i can refactor back to Sample[Tenmperature] that way then, too, if you agree.

that won't work,  because we have to stream temperatures of individual batteries.  Not an aggregate temperature.

llucax: > @llucax formulas can only be applied on resampled values. For a generic SampleRange with highest/lowest, etc. I don't mind, if you think there will be additional use cases for these in the future.

Moving to https://github.com/frequenz-floss/frequenz-sdk-python/discussions/494

I'm fine with finishing this PR with an intermediary solution, but I think the discussion is worth having, specially with v1.0 approaching.

PR #488, opened 2023-07-06
Migrate to repo-config

llucax:
- Migrate to use `repo-config`
- Fix pylint false positive for `pydantic.parse_raw_as`
- Add more cross-linking to generated docs
- Drop dependency on the `toml` package
- Fix missing `import` in a code example
- Fix some typing issues in `src/conftest.py`
- Add explicit dependency to `types-protobuf`


src/conftest.py
```diff
@@ -122,8 +125,6 @@ def evaluate(self, example: Example) -> None | str:
         import_header.append(
             path_to_import_statement(Path(os.path.relpath(example.path)))
         )
-        # Add microgrid as default import
-        import_header.append("from frequenz.sdk import microgrid")
```
daniel-zullo-frequenz, line 126: It is not clear to me why this import has been removed in `Fix some typing issues in src/conftest.py`

.github/workflows/ci.yaml
```diff
@@ -37,23 +44,19 @@ jobs:
...
-
       - name: Install required Python packages
         run: |
           python -m pip install --upgrade pip
-          python -m pip install nox toml
+          python -m pip install -e .[dev-noxfile]
 
-      - name: run nox
+      - name: Run nox
+        # To speed things up a bit we use the speciall ci_checks_max session
```
daniel-zullo-frequenz, line 53: typo in `speciall`

mkdocs.yml
```diff
@@ -2,13 +2,14 @@
...
 
 # Project information
-site_name: Frequenz Python SDK
-site_description: Frequenz Python SDK.
-site_author: Frequenz Energy-as-a-Service GmbH
-copyright: Frequenz Energy-as-a-Service GmbH
+site_name: "Frequenz Python SDK"
+site_description: "A development kit to interact with the Frequenz development platform"
+site_author: "Frequenz Energy-as-a-Service GmbH"
+copyright: "Copyright © 2022 Frequenz Energy-as-a-Service GmbH"
```
daniel-zullo-frequenz, line 8: Maybe 2023 as the year it's been just added?

pyproject.toml
```diff
@@ -1,14 +1,17 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
```
daniel-zullo-frequenz, line 2: Also here 2023?

daniel-zullo-frequenz (commented):
I've just have a few cosmetics to check for, LGTM

src/conftest.py
llucax: Mmm, that's weird, I guess this was some problem with rebasing the PR because it wasn't intended. Good catch!

.github/workflows/ci.yaml
llucax: Woops, but this needs to be fixed in `repo-config`. I think it is fine to introduce the typo here as it is an auto-generated file, but I will fix in `repo-config`.

llucax: https://github.com/frequenz-floss/frequenz-repo-config-python/pull/81

mkdocs.yml
llucax: No, the file was there before, it wasn't added by this PR.

pyproject.toml
llucax: No, it is a file that existed before.

src/conftest.py
llucax: Oh, I know. This is because I copied the "fixed" version from the channels repo, and the channels repo doesn't have the auto-import. I will remove the auto-import anyways (but in a different commit) to have the `conftest.py` unified between repos, as we should move it to `repo-config` and be general enough for all repos.

llucax:
Updated, it also fixes the CI (isort) issues

daniel-zullo-frequenz (approved):

PR #490, opened 2023-07-06
Update to channels 0.16.x

### LINKED ISSUE #407: Bump the channels dependency to use 0.16.x
### What's needed?

There was a channels release and even when the SDK is not desperately needing to use the new version, users might need to.

### Proposed solution

Bump the dependency and adapt the SDK code to work with the new channels release.

### Use cases

_No response_

_No response_

### Additional context

_No response_

jack-herrmann:
Fixes #407

### DIFF
--- a/src/frequenz/sdk/actor/_config_managing.py
+++ b/src/frequenz/sdk/actor/_config_managing.py
@@ -6,6 +6,7 @@
 import logging
 import os
 import tomllib
+from collections import abc
 from typing import Any, Dict, Optional, Set
 
 from frequenz.channels import Sender
@@ -31,7 +32,7 @@ class ConfigManagingActor:
         self,
         conf_file: str,
         output: Sender[Config],
-        event_types: Optional[Set[FileWatcher.EventType]] = None,
+        event_types: abc.Set[FileWatcher.EventType] = frozenset(FileWatcher.EventType),
     ) -> None:
         """Read config variables from the file.
 
@@ -79,8 +80,8 @@ class ConfigManagingActor:
         """
         await self.send_config()
 
-        async for path in self._file_watcher:
-            if str(path) == self._conf_file:
+        async for event in self._file_watcher:
+            if str(event.path) == self._conf_file:
                 _logger.info(
                     "Update configs, because file %s was modified.",
                     self._conf_file,
--- a/src/frequenz/sdk/actor/_config_managing.py
+++ b/src/frequenz/sdk/actor/_config_managing.py
@@ -7,7 +7,7 @@ import logging
 import os
 import tomllib
 from collections import abc
-from typing import Any, Dict, Optional, Set
+from typing import Any, Dict
 
 from frequenz.channels import Sender
 from frequenz.channels.util import FileWatcher
@@ -81,11 +81,12 @@ class ConfigManagingActor:
         await self.send_config()
 
         async for event in self._file_watcher:
-            if str(event.path) == self._conf_file:
-                _logger.info(
-                    "Update configs, because file %s was modified.",
-                    self._conf_file,
-                )
-                await self.send_config()
+            if event.type != FileWatcher.EventType.DELETE:
+                if str(event.path) == self._conf_file:
+                    _logger.info(
+                        "Update configs, because file %s was modified.",
+                        self._conf_file,
+                    )
+                    await self.send_config()
 
         _logger.debug("ConfigManager stopped.")
--- a/src/frequenz/sdk/actor/_decorator.py
+++ b/src/frequenz/sdk/actor/_decorator.py
@@ -82,7 +82,7 @@ def actor(cls: Callable[_P, _R]) -> Type[_R]:
     Example (one actor receiving from two receivers):
     ```python
     from frequenz.channels import Broadcast, Receiver, Sender
-    from frequenz.channels.util import Select
+    from frequenz.channels.util import select, selected_from
     @actor
     class EchoActor:
         def __init__(
@@ -99,13 +99,9 @@ def actor(cls: Callable[_P, _R]) -> Type[_R]:
             self._output = output
 
         async def run(self) -> None:
-            select = Select(channel_1=self._recv1, channel_2=self._recv2)
-            while await select.ready():
-                if msg := select.channel_1:
-                    await self._output.send(msg.inner)
-                elif msg := select.channel_2:
-                    await self._output.send(msg.inner)
-
+            async for selected in select(self._recv1, self._recv2):
+                if selected_from(selected, self._recv1):
+                    await self._output.send(selected.value)
 
     input_chan_1: Broadcast[bool] = Broadcast("input_chan_1")
     input_chan_2: Broadcast[bool] = Broadcast("input_chan_2")
--- a/src/frequenz/sdk/actor/power_distributing/_battery_status.py
+++ b/src/frequenz/sdk/actor/power_distributing/_battery_status.py
@@ -17,7 +17,7 @@ from frequenz.api.microgrid.battery_pb2 import RelayState as BatteryRelayState
 from frequenz.api.microgrid.common_pb2 import ErrorLevel
 from frequenz.api.microgrid.inverter_pb2 import ComponentState as InverterComponentState
 from frequenz.channels import Receiver, Sender
-from frequenz.channels.util import Select, Timer
+from frequenz.channels.util import Timer, select, selected_from
 
 from frequenz.sdk._internal._asyncio import cancel_and_await
 from frequenz.sdk.microgrid import connection_manager
@@ -197,14 +197,15 @@ class BatteryStatusTracker:
             raise RuntimeError(f"Can't find inverter adjacent to battery: {battery_id}")
 
         self._battery: _ComponentStreamStatus = _ComponentStreamStatus(
-            battery_id, data_recv_timer=Timer(max_data_age_sec)
+            battery_id,
+            data_recv_timer=Timer.timeout(timedelta(max_data_age_sec)),
         )
         self._inverter: _ComponentStreamStatus = _ComponentStreamStatus(
-            inverter_id, data_recv_timer=Timer(max_data_age_sec)
+            inverter_id,
+            data_recv_timer=Timer.timeout(timedelta(max_data_age_sec)),
         )
 
         # Select needs receivers that can be get in async way only.
-        self._select: Select | None = None
 
         self._task: asyncio.Task[None] = asyncio.create_task(
             self._run(status_sender, set_power_result_receiver)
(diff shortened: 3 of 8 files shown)


src/frequenz/sdk/actor/power_distributing/_battery_status.py
```diff
@@ -223,8 +224,80 @@ async def stop(self) -> None:
...
+        self, bat_data: BatteryData
+    ) -> None:  # Any is probably not the best
+        self._battery.last_msg_correct = (
+            self._is_message_reliable(bat_data)
+            and self._is_battery_state_correct(bat_data)
+            and self._no_critical_error(bat_data)
+            and self._is_capacity_present(bat_data)
+        )
+        self._battery.last_msg_timestamp = bat_data.timestamp
+        self._battery.data_recv_timer.reset()
```
shsms: These functions were necessary, in order to get the tests working with the new select, with the least amount of work.  The tests are written in this way to limit the amount of async calls made in tests,  but that's making it harder to test and some parts of the code outside the select are not getting tested.

I'm already working on improving the test coverage of the battery status tracker,  so I will rewrite the tests in such a way that they don't get so much in the way of code structure.  That will come after this PR is merged.

src/frequenz/sdk/actor/_config_managing.py
```diff
@@ -78,8 +79,8 @@ async def run(self) -> None:
         """
         await self.send_config()
 
-        async for path in self._file_watcher:
-            if str(path) == self._conf_file:
+        async for event in self._file_watcher:
+            if str(event.path) == self._conf_file:
```
llucax: Now that we also have the [event type](https://frequenz-floss.github.io/frequenz-channels-python/v0.16/reference/frequenz/channels/util/#frequenz.channels.util._file_watcher.FileWatcher.Event.type) available, we should check what the type is, and should not try to read the file if it is a [`DELETE` event](https://github.com/frequenz-floss/frequenz-channels-python/blob/a4b873ba8c3b38df78ce8f6883dc82d143894859/src/frequenz/channels/util/_file_watcher.py#L29), as it will fail.

This is a pre-existing issue, so we can address it in a separate PR too, please create a new issue if you prefer to do that. If you prefer to fix it in this PR, please do the fix in a separate commit.

src/frequenz/sdk/actor/_decorator.py
```diff
@@ -79,7 +79,7 @@ def actor(cls: Callable[_P, _R]) -> Type[_R]:
     Raises:
         TypeError: when the class doesn't have a `run` method as per spec.
 
-    Example (one actor receiving from two receivers):
+    Example (one actor receiving from two receivers): !this example is not updated to channels 0.16!
```
llucax: To update it, you should probably only need to change the select loop to:

```py
        async def run(self) -> None:
            async for selected in select(self._recv1, self._recv2)
                if selected_from(selected, self._recv1):
                    await self._output.send(selected.value)
```

It is a pretty bad example still, but we'll address this when we revamp the actor decorator, so don't worry about it, just make sure the current example is minimally fixed.

src/frequenz/sdk/actor/power_distributing/_battery_status.py
```diff
@@ -17,7 +18,7 @@
 from frequenz.api.microgrid.common_pb2 import ErrorLevel
 from frequenz.api.microgrid.inverter_pb2 import ComponentState as InverterComponentState
 from frequenz.channels import Receiver, Sender
-from frequenz.channels.util import Select, Timer
+from frequenz.channels.util import Timer, select, selected_from  # Select
```
llucax:
```suggestion
from frequenz.channels.util import Timer, select, selected_from
```

src/frequenz/sdk/actor/power_distributing/_battery_status.py
```diff
@@ -5,6 +5,7 @@
 from __future__ import annotations
 
 import asyncio
+import datetime as dt
```
llucax: Please don't use the `dt` alias to keep a consistent style in the SDK (we usually don't use it in the SDK).
```suggestion
import datetime
```

src/frequenz/sdk/actor/power_distributing/_battery_status.py
```diff
@@ -197,14 +198,14 @@ def __init__(  # pylint: disable=too-many-arguments
             raise RuntimeError(f"Can't find inverter adjacent to battery: {battery_id}")
 
         self._battery: _ComponentStreamStatus = _ComponentStreamStatus(
-            battery_id, data_recv_timer=Timer(max_data_age_sec)
+            battery_id, data_recv_timer=Timer.periodic(dt.timedelta(max_data_age_sec))
```
llucax:
```suggestion
            battery_id, data_recv_timer=Timer.periodic(datetime.timedelta(max_data_age_sec))
```

src/frequenz/sdk/actor/power_distributing/_battery_status.py
```diff
@@ -197,14 +198,14 @@ def __init__(  # pylint: disable=too-many-arguments
...
         )
         self._inverter: _ComponentStreamStatus = _ComponentStreamStatus(
-            inverter_id, data_recv_timer=Timer(max_data_age_sec)
+            inverter_id, data_recv_timer=Timer.periodic(dt.timedelta(max_data_age_sec))
```
llucax:
```suggestion
            inverter_id, data_recv_timer=Timer.periodic(datetime.timedelta(max_data_age_sec))
```

src/frequenz/sdk/actor/power_distributing/_battery_status.py
```diff
@@ -197,14 +198,14 @@ def __init__(  # pylint: disable=too-many-arguments
...
         )
 
         # Select needs receivers that can be get in async way only.
-        self._select: Select | None = None
+        # self._select: Select | None = None
```
llucax:
```suggestion
```

src/frequenz/sdk/actor/power_distributing/_battery_status.py
```diff
@@ -223,8 +224,80 @@ async def stop(self) -> None:
         """Stop tracking battery status."""
         await cancel_and_await(self._task)
 
-        if self._select is not None:
-            await self._select.stop()
+    def _handle_status_battery(
+        self, bat_data: BatteryData
+    ) -> None:  # Any is probably not the best
```
llucax: I guess the comment about `Any` is outdated? Same for the rest of the new methods.

```suggestion
    ) -> None
```

src/frequenz/sdk/actor/power_distributing/_battery_status.py
```diff
@@ -223,8 +224,80 @@ async def stop(self) -> None:
...
+
+            if duration > 0:
+                _logger.warning(
+                    "battery %d failed last response. block it for %f sec",
+                    self.battery_id,
+                    duration,
+                )
+
+    def _handle_status_battery_timer(
+        self, _timedelta: timedelta
```
llucax: If you don't need the drift, then maybe don't pass it at all? But if you do pass it, it would be better to call the argument `timer_drift` instead of `_timedelta`.

Same for the inverter timer method.

src/frequenz/sdk/actor/power_distributing/_battery_status.py
```diff
@@ -223,8 +224,80 @@ async def stop(self) -> None:
         """Stop tracking battery status."""
         await cancel_and_await(self._task)
 
-        if self._select is not None:
-            await self._select.stop()
+    def _handle_status_battery(
```
llucax: Very nice refactoring here with one method to handle each receiver. The diff ended up a bit hard to read though, so in the future I would recommend to split the refactoring to a separate commit and in this commit try to make the code structure as similar as it originally was.

In this case keeping the `_update_status` method and pass the `selected` object instead.

src/frequenz/sdk/actor/power_distributing/_battery_status.py
```diff
@@ -223,8 +224,80 @@ async def stop(self) -> None:
...
+    ) -> None:  # Any is probably not the best
+        if self._inverter.last_msg_correct:
+            self._inverter.last_msg_correct = False
+            _logger.warning(
+                "Inverter %d stopped sending data, last timestamp: %s",
+                self._inverter.component_id,
+                self._inverter.last_msg_timestamp,
+            )
+
+    def _check_status(self) -> Optional[Status]:
```
llucax: I think it is more clear if this remains close to the loop, because we are looking if the status was updated or not, and send it only if it did. This function is not really *checking* the status, it is only checking if it changed or not. It wasn't clear to me what was going on when reading the main loop.

If you prefer to keep it in a separate function, maybe rename it to something more clear, even if it is longer, for example:

```suggestion
    def _get_new_status_if_changed(self) -> Optional[Status]:
```

src/frequenz/sdk/timeseries/ev_charger_pool/_set_current_bounds.py
```diff
@@ -10,7 +10,7 @@
 from typing import Dict
 
 from frequenz.channels import Broadcast, Sender
-from frequenz.channels.util import Select, Timer
+from frequenz.channels.util import Timer, select, selected_from  # Select
```
llucax:
```suggestion
from frequenz.channels.util import Timer, select, selected_from
```

src/frequenz/sdk/timeseries/ev_charger_pool/_set_current_bounds.py
```diff
@@ -95,17 +95,16 @@ async def _run(self) -> None:
         ).into_peekable()
         latest_bound: Dict[int, ComponentCurrentLimit] = {}
 
-        select = Select(
-            bound_chan=self._bounds_rx,
-            timer=Timer(self._repeat_interval.total_seconds()),
-        )
-        while await select.ready():
+        bound_chan = self._bounds_rx
+        timer = Timer.periodic(timedelta(self._repeat_interval.total_seconds()))
```
llucax: I think this should be `Timer.timeout()` or use `skip_missed_ticks=True` for `periodic()`, because I guess we don't want to send the bounds again as a burst if the loop is delayed for too long, right @shsms ?

tests/actor/test_battery_status.py
```diff
@@ -235,52 +235,52 @@ async def test_sync_update_status_with_messages(
         assert tracker.battery_id == BATTERY_ID
         assert tracker._last_status == Status.NOT_WORKING
 
-        select = FakeSelect(inverter=inverter_data(component_id=INVERTER_ID))
-        assert tracker._update_status(select) is None  # type: ignore[arg-type]
+        tracker._handle_status_inverter(inverter_data(component_id=INVERTER_ID))
+        assert tracker._check_status() is None
```
llucax: Ah, ok, for tests it might be good to keep this status check separated then...

llucax (commented):
Looks good in general, but there are a few minor things to address.

Also, can you add a note in the `RELEASE_NOTES.md` file, in the Upgrading section that the channels version as upgraded so SDK users should also upgrade? Please also add a link to the channels repo release, so users can follow upgrade the instructions there (https://github.com/frequenz-floss/frequenz-channels-python/releases/tag/v0.16.0).

src/frequenz/sdk/actor/power_distributing/_battery_status.py
llucax: Disclaimer: I did the review before reading this comment :)

src/frequenz/sdk/timeseries/ev_charger_pool/_set_current_bounds.py
shsms: oh true,  I guess `timeout` definitely makes more sense here.  Same for the two timers created in the `__init__` method of the `BatteryStatusTracker` class.

src/frequenz/sdk/actor/power_distributing/_battery_status.py
```diff
@@ -197,14 +197,15 @@ def __init__(  # pylint: disable=too-many-arguments
...
 
         self._battery: _ComponentStreamStatus = _ComponentStreamStatus(
-            battery_id, data_recv_timer=Timer(max_data_age_sec)
+            battery_id,
+            data_recv_timer=Timer.timeout(timedelta(max_data_age_sec)),
         )
         self._inverter: _ComponentStreamStatus = _ComponentStreamStatus(
-            inverter_id, data_recv_timer=Timer(max_data_age_sec)
+            inverter_id,
+            data_recv_timer=Timer.timeout(timedelta(max_data_age_sec)),
```
shsms: Took a while to identify why the timers weren't triggering.  These have to say `timedelta(seconds=max_data_age_sec)`, else it assumes days.  

@llucax next pylint project?

llucax: I think it should be part of `pylint-datetime`.

jack-herrmann: Yes, it will be

shsms: @jh2007github could you please update how the timedelta objects are constructed here,  and let us know when the PR is ready for review again?

jack-herrmann: I have done that, sorry it took so long, I forgot all about this needed change

jack-herrmann:
By the way I am aware of the misspelled branch name, I just don't know how to rename it without breaking anything

src/frequenz/sdk/actor/power_distributing/_battery_status.py
shsms: have you pushed?  I don't see it yet.

shsms: oh it is done in a different commit.  I guess that's fine.

shsms:
> By the way I am aware of the misspelled branch name, I just don't know how to rename it without breaking anything

That's fine, it won't be mentioned in the merge commit.

shsms (approved):
lgtm, nice work! 🎉

PR #495, opened 2023-07-10
Rename power to active_power

jack-herrmann:
still two formatting issues, that black wasn't able to fix
(I am not sure how that channels commit got in here, it is definitely not supposed to)

### DIFF
--- a/benchmarks/power_distribution/power_distributor.py
+++ b/benchmarks/power_distribution/power_distributor.py
@@ -1,7 +1,7 @@
 # License: MIT
 # Copyright © 2022 Frequenz Energy-as-a-Service GmbH
 
-"""Check how long it takes to distribute power."""
+"""Check how long it takes to distribute active_power."""
 
 import asyncio
 import csv
@@ -37,7 +37,7 @@ async def send_requests(batteries: Set[int], request_num: int) -> List[Result]:
 
     Args:
         user: user that should send request
-        batteries: set of batteries where the power should be set
+        batteries: set of batteries where the active_power should be set
         request_num: number of requests that should be send
 
     Raises:
@@ -47,10 +47,12 @@ async def send_requests(batteries: Set[int], request_num: int) -> List[Result]:
         List of the results from the PowerDistributingActor.
     """
     battery_pool = microgrid.battery_pool(batteries)
-    results_rx = battery_pool.power_distribution_results()
+    results_rx = battery_pool.active_power_distribution_results()
     result: List[Result] = []
     for _ in range(request_num):
-        await battery_pool.set_power(Power(float(random.randrange(100000, 1000000))))
+        await battery_pool.set_active_power(
+            Power(float(random.randrange(100000, 1000000)))
+        )
         try:
             output = await asyncio.wait_for(results_rx.receive(), timeout=3)
             if output is None:
@@ -105,12 +107,12 @@ async def run_test(  # pylint: disable=too-many-locals
     """
     start = timeit.default_timer()
 
-    power_request_channel = Broadcast[Request]("power-request")
+    active_power_request_channel = Broadcast[Request]("active_power-request")
     battery_status_channel = Broadcast[BatteryStatus]("battery-status")
-    channel_registry = ChannelRegistry(name="power_distributor")
+    channel_registry = ChannelRegistry(name="active_power_distributor")
     distributor = PowerDistributingActor(
         channel_registry=channel_registry,
-        requests_receiver=power_request_channel.new_receiver(),
+        requests_receiver=active_power_request_channel.new_receiver(),
         battery_status_sender=battery_status_channel.new_sender(),
     )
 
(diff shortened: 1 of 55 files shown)


llucax:
> (I am not sure how that channels commit got in here, it is definitely not supposed to)

This is because you based your changes on the branch that was updating the channels repo. You could do a `git rebase -i upstream/v0.x.x` (if your remote pointing to this repository is called `upstream`) and then just remove the commit updating the channels dependency. Then you can do a `git push --force-with-lease` to push the changes.

llucax:
You also have quite a few conflicts, you probably need to rebase your changes on the current `v0.x.x` branch (this will be done by the `git rebase` command I mentioned above).

benchmarks/power_distribution/power_distributor.py
```diff
@@ -1,7 +1,7 @@
 # License: MIT
 # Copyright © 2022 Frequenz Energy-as-a-Service GmbH
 
-"""Check how long it takes to distribute power."""
+"""Check how long it takes to distribute active_power."""
```
llucax, line 4: In the documentation I suggest to either use `active power` as normal text, or use `` `active_power` `` to highlight it is talking about code (a property) if that is the case.

benchmarks/power_distribution/power_distributor.py
```diff
@@ -47,10 +47,12 @@ async def send_requests(batteries: Set[int], request_num: int) -> List[Result]:
         List of the results from the PowerDistributingActor.
     """
     battery_pool = microgrid.battery_pool(batteries)
-    results_rx = battery_pool.power_distribution_results()
+    results_rx = battery_pool.active_power_distribution_results()
```
llucax, line 50: I guess this should stay simply as `power_distribution_results`? @shsms ?

benchmarks/power_distribution/power_distributor.py
```diff
@@ -105,12 +107,12 @@ async def run_test(  # pylint: disable=too-many-locals
     """
     start = timeit.default_timer()
 
-    power_request_channel = Broadcast[Request]("power-request")
+    active_power_request_channel = Broadcast[Request]("active_power-request")
```
llucax, line 110:
```suggestion
    active_power_request_channel = Broadcast[Request]("active-power-request")
```

benchmarks/power_distribution/power_distributor.py
```diff
@@ -105,12 +107,12 @@ async def run_test(  # pylint: disable=too-many-locals
...
     battery_status_channel = Broadcast[BatteryStatus]("battery-status")
-    channel_registry = ChannelRegistry(name="power_distributor")
+    channel_registry = ChannelRegistry(name="active_power_distributor")
```
llucax, line 112: Also this probably needs to stay as `power_distributor`, as this is talking about an entity that has that name, but not sure what kind of power the power distributor distributes :D Also maybe @shsms or @mathias-baumann-frequenz know better.

examples/battery_pool.py
```diff
@@ -35,7 +35,7 @@ async def main() -> None:
     receivers: Dict[str, Receiver[Any]] = {
         "soc": battery_pool.soc.new_receiver(maxsize=1),
         "capacity": battery_pool.capacity.new_receiver(maxsize=1),
-        "power_bounds": battery_pool.power_bounds.new_receiver(maxsize=1),
+        "active_power_bounds": battery_pool.active_power_bounds.new_receiver(maxsize=1),
```
llucax, line 38: I guess this applies to all power, not just active?

llucax (commented):
I made a few comments, but I guess it makes more sense that someone more familiar with how and where we use active power or we refer more generally to power reviews this instead.

benchmarks/power_distribution/power_distributor.py
shsms: We are currently setting only the active power, so these are active power results.  We might start supporting setting reactive power,  and the new api specs already support it.

I'm actually doubtful of this whole exercise, because I imagine the maintenance of power factor to be a lower level function.  I'd rather just leave everything as it is,  and if we decide to introduce support for reactive power, call those methods `reactive`.

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -311,125 +311,125 @@ class Power(
         6: "MW",
     },
 ):
-    """A power quantity."""
+    """A active_power quantity."""
```
shsms, line 314: We'd have to rename the type to `ActivePower` as well,  because it has a different unit from reactive (VAR) and apparent (VA) powers.

shsms:
We decided not to go ahead with this, because the word power already implies active power,  and most users of the sdk won't have to deal with any other kind of power.  Maintenance of power factor and other user of reactive power are all low level functionalities,  that can be exposed through secondary/lower level interfaces, if required.

PR #512, opened 2023-07-13
Add capability to multiply quantities with percentage

Marenz:


Marenz:
I was hoping [this error](https://github.com/frequenz-floss/frequenz-sdk-python/actions/runs/5544167247/jobs/10121236213?pr=512#step:5:21) was just a local thing, but it seems not :/

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -218,6 +218,19 @@ def __sub__(self, other: Self) -> Self:
...
+        """Return the product of this quantity and a percentage.
+
+        Args:
+            percent: The percentage.
+
+        Returns:
+            The product of this quantity and a percentage.
+        """
+        product = type(self).__new__(type(self))
+        product._base_value = self._base_value * percent._base_value / 100.0
```
llucax:
```suggestion
        product._base_value = self._base_value * percent.as_fraction()
```

?

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -400,17 +413,41 @@ def as_megawatts(self) -> float:
         """
         return self._base_value / 1e6
 
+    @overload
+    def __mul__(self, percent: Percentage) -> Power:
+        """Return a power from multiplying this power by the given percentage.
+
+        Args:
+            percent: The percentage to multiply by.
```
llucax: Missing `Return:`, strange that darglint doesn't complain...

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -400,17 +413,41 @@ def as_megawatts(self) -> float:
         """
         return self._base_value / 1e6
 
+    @overload
+    def __mul__(self, percent: Percentage) -> Power:
```
llucax:
```suggestion
    def __mul__(self, percent: Percentage) -> Self:
```

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -400,17 +413,41 @@ def as_megawatts(self) -> float:
...
     def __mul__(self, duration: timedelta) -> Energy:
         """Return an energy from multiplying this power by the given duration.
 
         Args:
             duration: The duration to multiply by.
+        """
 
-        Returns:
-            An energy from multiplying this power by the given duration.
+    def __mul__(self, other: Percentage | timedelta) -> Power | Energy:
```
llucax:
```suggestion
    def __mul__(self, other: Percentage | timedelta) -> Self | Energy:
```

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -400,17 +413,41 @@ def as_megawatts(self) -> float:
...
+        if isinstance(other, Percentage):
+            return super().__mul__(other)
+        if isinstance(other, timedelta):
+            return Energy.from_watt_hours(
+                self._base_value * other.total_seconds() / 3600.0
+            )
+
+        raise TypeError(
+            f"unsupported operand type(s) for *: '{self.__class__.__name__}' and "
+            f"'{other.__class__.__name__}'"
```
llucax: I think this should raise a `NotImplementedError` to signal Python to either try other operators or print a nicer error message, right @shsms ?

Marenz:
I am giving up on this for now, I can't solve the mypy issue. 
Inside mypy the following check fails:

```
      elif isinstance(override, Overloaded) and self.is_forward_op_method(name):
            # Operator method overrides cannot extend the domain, as
            # this could be unsafe with reverse operator methods.
            original_domain = self.get_op_other_domain(original)
            override_domain = self.get_op_other_domain(override)
            if (
                original_domain
                and override_domain
                and not is_subtype(override_domain, original_domain)
            ):
                fail = True
                op_method_wider_note = True
```

The comment reads as if what I want to do is unsafe according to them. I couldn't find other sources that claim whether it's good or not...

shsms:
This seems to shut up both mypy and pyright:

``` diff
diff --git a/src/frequenz/sdk/timeseries/_quantities.py b/src/frequenz/sdk/timeseries/_quantities.py
index 4db55c99..e808b28e 100644
--- a/src/frequenz/sdk/timeseries/_quantities.py
+++ b/src/frequenz/sdk/timeseries/_quantities.py
@@ -413,8 +413,8 @@ class Power(
         """
         return self._base_value / 1e6
 
-    @overload
-    def __mul__(self, percent: Percentage) -> Power:
+    @overload  # type: ignore
+    def __mul__(self, other: Percentage) -> Power:
         """Return a power from multiplying this power by the given percentage.
 
         Args:
@@ -422,7 +422,7 @@ class Power(
         """
 
     @overload
-    def __mul__(self, duration: timedelta) -> Energy:
+    def __mul__(self, other: timedelta) -> Energy:
         """Return an energy from multiplying this power by the given duration.
 
         Args:
@@ -541,8 +541,8 @@ class Current(
         """
         return self._base_value * 1e3
 
-    @overload
-    def __mul__(self, percent: Percentage) -> Self:
+    @overload  # type: ignore
+    def __mul__(self, other: Percentage) -> Self:
         """Return a power from multiplying this power by the given percentage.
 
         Args:
@@ -550,7 +550,7 @@ class Current(
         """
 
     @overload
-    def __mul__(self, voltage: Voltage) -> Power:
+    def __mul__(self, other: Voltage) -> Power:
         """Multiply the current by a voltage to get a power.
 
         Args:
@@ -653,8 +653,8 @@ class Voltage(
         """
         return self._base_value / 1e3
 
-    @overload
-    def __mul__(self, percent: Percentage) -> Self:
+    @overload  # type: ignore
+    def __mul__(self, other: Percentage) -> Self:
         """Return a power from multiplying this power by the given percentage.
 
         Args:
@@ -662,7 +662,7 @@ class Voltage(
         """
 
     @overload
-    def __mul__(self, current: Current) -> Power:
+    def __mul__(self, other: Current) -> Power:
         """Multiply the voltage by the current to get the power.
 
         Args:

File: src/frequenz/sdk/timeseries/_quantities.py (hunk 5 above)
Comment (shsms) [reply src/frequenz/sdk/timeseries/_quantities.py]: yup,  `return NotImplemented`,  and it generates a standard error message automatically.

File: src/frequenz/sdk/timeseries/_quantities.py (hunk 4 above)
Comment (Marenz) [reply src/frequenz/sdk/timeseries/_quantities.py]: May I ask why? :)

File: src/frequenz/sdk/timeseries/_quantities.py [hunk 6]
```diff
@@ -221,6 +221,19 @@ def __sub__(self, other: Self) -> Self:
...
+    def __mul__(self, percent: Percentage) -> Self:
+        """Return the product of this quantity and a percentage.
+
+        Args:
+            percent: The percentage.
+
+        Returns:
+            The product of this quantity and a percentage.
+        """
+        product = type(self).__new__(type(self))
```
shsms, line 236: I think here it has to say

```python
        if type(percent) is not Percentage:
            return NotImplemented

File: tests/timeseries/test_quantities.py [hunk 7]
```diff
@@ -377,3 +377,19 @@ def test_abs() -> None:
...
+    voltage = Voltage.from_volts(230.0)
+    current = Current.from_amperes(2)
+    energy = Energy.from_kilowatt_hours(12)
+    percentage_ = Percentage.from_percent(50)
+
+    assert power * percentage == Power.from_watts(500.0)
+    assert voltage * percentage == Voltage.from_volts(115.0)
+    assert current * percentage == Current.from_amperes(1)
+    assert energy * percentage == Energy.from_kilowatt_hours(6)
+    assert percentage_ * percentage == Percentage.from_percent(25)
```
shsms, line 395: Could you add a couple of tests for the unhappy cases as well, making sure that comparisons fail where they are not implemented?

tests/timeseries/test_quantities.py
```diff
@@ -377,3 +377,23 @@ def test_abs() -> None:
...
+
+    for quantity in [power, voltage, current, energy, percentage_]:
+        with pytest.raises(TypeError):
+            _ = quantity * 200.0
```
shsms: I was thinking more like `Energy * Power` and `Energy * Energy`, etc.  Things that can have a `base_value` field.  I was thinking of this just because such a check was not there in the `__mul__` method and having such tests would have caught that. (although you've now added it).

Marenz:
updated: Added the requested tests

shsms (dismissed):

shsms (approved):

PR #514, opened 2023-07-14
Support float.inf and float.nan in quantities

Marenz:
This might be controversial, but my use case is the following:

```
        boundaries: list[tuple[Percentage, SoC]] = [
            (config.battery_bounds.lower, SoC.DISCHARGED),
            (reserve_bounds.lower, SoC.RESERVE),
            (config.target_soc_bounds_pct.lower, SoC.NORMAL),
            (config.target_soc_bounds_pct.upper, SoC.TARGET),
            (reserve_bounds.upper, SoC.NORMAL),
            (config.battery_bounds.upper, SoC.RESERVE),
            (Percentage.from_percent(float("inf")), SoC.CHARGED),
        ]

        for boundary, state in boundaries:
            if bat_soc_pct < boundary:
                return state

```

I am using inf here as a stop-marker that will always be larger than the last boundary. This way I can avoid having to add a special case for the last boundary.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,16 +6,17 @@
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with --> 
+<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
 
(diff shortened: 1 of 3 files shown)


shsms:
This looks good to me,  but could you change the release notes to say it is a bug fix to get the formatting of quantities to not raise when encountering inf values?  And maybe you could add fixes/tests for `nan` as well.

Marenz:
done

shsms (approved):

PR #520, opened 2023-07-18
Use `Timer.periodic` in the `Resampler`

### LINKED ISSUE #264: Replace the custom timer in the resampler with a `Timer.periodic()` receiver
### What's needed?

Make sure that the resampler emits samples as regularly as possible.

### Proposed solution

Use the new `PeriodicTimer` in the channels package, which uses a monotonic clock and makes sure it doesn't drift.

### Use cases

_No response_

The resampler is currently implementing its own timer as it was needed before the `PeriodicTimer` was available.

### Additional context

* https://github.com/frequenz-floss/frequenz-sdk-python/issues/170
* https://github.com/frequenz-floss/frequenz-sdk-python/pull/255
* https://github.com/frequenz-floss/frequenz-sdk-python/pull/236
* https://github.com/frequenz-floss/frequenz-channels-python/pull/79

jack-herrmann:
Fixes #264

### DIFF
--- a/src/frequenz/sdk/timeseries/_resampling.py
+++ b/src/frequenz/sdk/timeseries/_resampling.py
@@ -16,6 +16,7 @@ from datetime import datetime, timedelta, timezone
 from typing import AsyncIterator, Callable, Coroutine, Optional, Sequence
 
 from frequenz.channels.util import Timer
+from frequenz.channels.util._timer import _to_microseconds
 
 from .._internal._asyncio import cancel_and_await
 from ._base_types import UNIX_EPOCH, Sample
@@ -369,7 +370,8 @@ class Resampler:
         self._resamplers: dict[Source, _StreamingHelper] = {}
         """A mapping between sources and the streaming helper handling that source."""
 
-        self._window_end: datetime = self._calculate_window_end()
+        window_end, start_delay_time = self._calculate_window_end()
+        self._window_end: datetime = window_end
         """The time in which the current window ends.
 
         This is used to make sure every resampling window is generated at
@@ -384,6 +386,15 @@ class Resampler:
 
         self._timer: Timer = Timer.periodic(config.resampling_period)
         """The timer used to trigger the resampling windows."""
+
+        # Hack to align the timer, this should be implemented in the Timer class
+        # If there is no delay, then we need to add one period, because the timer
+        # shouldn't fire right away
+        if not start_delay_time:
+            start_delay_time += config.resampling_period
+        self._timer._next_tick_time = _to_microseconds(
+            timedelta(seconds=asyncio.get_running_loop().time()) + start_delay_time
+        )  # pylint: disable=protected-access
 
     @property
     def config(self) -> ResamplerConfig:
@@ -504,7 +515,7 @@ class Resampler:
             if one_shot:
                 break
 
-    def _calculate_window_end(self) -> datetime:
+    def _calculate_window_end(self) -> tuple[datetime, timedelta]:
         """Calculate the end of the current resampling window.
 
         The calculated resampling window end is a multiple of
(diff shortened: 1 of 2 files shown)


src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -461,15 +463,33 @@ async def resample(self, *, one_shot: bool = False) -> None:
...
+        async for _ in Timer.periodic(
+            timedelta(seconds=self._config.resampling_period.total_seconds())
+        ):
+            now = datetime.now(tz=timezone.utc)
+            timer_error = now - self._window_end
+            # We use a tolerance of 10% of the resampling period
+            tolerance = timedelta(
+                seconds=self._config.resampling_period.total_seconds() / 10.0
+            )
+            if timer_error > tolerance:
```
shsms: I think all this code can go away now and we can just use the drift that the timer returns.

```suggestion
        # We use a tolerance of 10% of the resampling period
        tolerance = timedelta(
            seconds=self._config.resampling_period.total_seconds() / 10.0
        )
        async for drift in Timer.periodic(
            timedelta(seconds=self._config.resampling_period.total_seconds())
        ):
            if drift > tolerance:

Comment (jack-herrmann) [reply src/frequenz/sdk/timeseries/_resampling.py]: now and time_error also need to be updated in the logger message, unfortunately I saw this after accepting the change

File: src/frequenz/sdk/timeseries/_resampling.py [hunk 2]
```diff
@@ -461,15 +463,33 @@ async def resample(self, *, one_shot: bool = False) -> None:
...
         """
-        while True:
-            await self._wait_for_next_resampling_period()
+        # We use a tolerance of 10% of the resampling period
+        tolerance = timedelta(
+            seconds=self._config.resampling_period.total_seconds() / 10.0
+        )
+        async for drift in Timer.periodic(
+            timedelta(seconds=self._config.resampling_period.total_seconds())
+        ):
```
llucax: This timer needs to be a instance member, otherwise the `drift` will always be zero, as you creating the timer on each resample call. For the Resampling actor this is not an issue because `resample()` is called once and then we loop on it, but in the tests we use `one_shot` and in that case we call `resample()` multiple times, thus triggering the permanent zero drift.

Also the tests need to be adjusted to also do a `asyncio.sleep()` with the same time as `fake_time.shift()` to make sure the loop monotonic clock (which is the one the new channels timer uses) is also advanced to produce a drift, otherwise the timer will always trigger at the interval time.

llucax (changes requested):
I will send a PR to your branch with a fix, as it is not trivial to explain in comments.

llucax:
There is another issue with this change, the `align_to` is also broken, and fixing it is not all that trivial. Adding alignment to the channels timer is hard because it doesn't know anything about `datetime` and we want to align to a `datetime`.

So to align the timer we need to actually have an initial `Timer.timeout()` to just wait until the next aligned window and then start the `Timer.periodic()` after that. Having channels (having https://github.com/frequenz-floss/frequenz-channels-python/issues/116 would make this **much** easier), but for now I think we need to refactor things a bit to make it work. I'll still send a proposal fix but it is taking me a bit more time than expected.

llucax:
I'm starting to think that a periodic timer is not a good fit for this :grimacing: 

If I add a timer to align, but **that** timer drifts too much, then what? :grimacing: :gun: 

We are really getting into a mess here, this issue was created before we realized we needed to align the resampling window, and the alignment only makes sense using a `datetime`.

We either need to live with the fact that the periodic time will be a bit shifted (we can still emit the timestamps for samples and do all the calculations with the desired precise window we want, it is only that the sample will be sent a bit later than that timestamp) and don't care about **alignment of the timer**, or we should probably leave things as they are.

And to be honest, I'm leaning more towards leave what we have, it is less effort and probably more accurate (less chances of having a considerable shift between the timer triggering and the timestamp being emitted).

llucax:
Proposed hacky fix for this: https://github.com/jh2007github/frequenz-sdk-python/pull/1

I would leave this open and implement the hack properly in `Timer` and only merge it when we upgraded the SDK to use the version with a `start_delay: timedelta` option, since there is no real urgency to merge it.

llucax (dismissed):
We decided to merge this as is and update when the `Timer` gets the new `start_delay` option.

llucax:
BTW, this needs someone else's approval, as I made the latest commits in this PR :innocent:

llucax:
Setting auto-merge.

llucax:
* https://github.com/frequenz-floss/frequenz-sdk-python/issues/556

shsms (dismissed):

shsms (dismissed):

matthias-wende-frequenz (dismissed):

llucax:
@jh2007github not sure what's going on here with:

image

llucax:
It would be good to merge this before releasing v0.23.0 as it is ready already, but the approvals keep getting dismissed by the base change.

llucax (approved):

Issue #214, opened 2023-02-21
MovingWindow: Create Interface for requesting data and handling the data copy behaviour


matthias-wende-frequenz:
When asking the `MovingWindow` for a subwindow we might return a slice of the buffer, that means a reference to the underlying numpy based ringbuffer.
There are cases, e.g. when dealing with small window sizes and high sample rates, where the data might get updated from the data stream before the requested subwindow has been processed.

As a solution we want to tell the `MovingWindow` to return a copy of the data instead of a reference. In order to do that we need to add an interface, such that users can control copy behavior.

We might implement a method for the `MovingWindow` as follows.

```python
window(
	self,
	start: datetime,
	end: datetime,
	fill_gaps: Optional[Quantity] = None,
	force_copy: Bool = False
)
```

This method should return a sorted numpy array that holds the resampled data from start to end and fills the gaps with the Value that is provided by `fill_gaps`.

------------

Summary of the (ongoing) work related to this and what I see essential for v1.0.  

Essential for this issue:
* For v1.0:
  * https://github.com/frequenz-floss/frequenz-sdk-python/pull/641
  * https://github.com/frequenz-floss/frequenz-sdk-python/pull/598
  * https://github.com/frequenz-floss/frequenz-sdk-python/pull/638
  * https://github.com/frequenz-floss/frequenz-sdk-python/pull/668
  * https://github.com/frequenz-floss/frequenz-sdk-python/pull/672
  * https://github.com/frequenz-floss/frequenz-sdk-python/issues/676 (If not ready for v1.0 getitem magic should be removed before the release)
  
* Post v1.0?
  * https://github.com/frequenz-floss/frequenz-sdk-python/pull/669.
  * https://github.com/frequenz-floss/frequenz-sdk-python/issues/646

Related to moving window:
* For v1.0
  * https://github.com/frequenz-floss/frequenz-sdk-python/pull/664
  * https://github.com/frequenz-floss/frequenz-sdk-python/pull/670
* Post v1.0?
  * https://github.com/frequenz-floss/frequenz-sdk-python/issues/558

matthias-wende-frequenz:
FYI @cwasicki @thomas-nicolai-frequenz

thomas-nicolai-frequenz:
I don't understand why we could not design it in a way where the user passes his numpy array by reference the moving window is writing to. This way we could avoid the need for the fill_gaps parameter. In that case the numpy array could be initialised with whatever values the someone wants.

[...]

PR #523, opened 2023-07-19
MovingWindow returns sorted np array, when wrapping around and new test

### LINKED ISSUE #212: Add Stream Weighted Average
### What's needed?

A weighted moving average implementation that updates the avg whenever a new data sample arrives from the stream.

### Proposed solution

We can use the [`MovingWindow`](https://github.com/frequenz-floss/frequenz-sdk-python/blob/v0.x.x/src/frequenz/sdk/timeseries/_moving_window.py) class. Whenever the `MovingWindow` gets updated it triggers the update of the mean.

Right now, the `MovingWindow` doesn't support sending events on update.

We propose two different implementation approaches.

* Add a output trigger channel that informs downstream actors about an update of the window.
* The moving average is implemented as part of the `MovingWindow` and downstream users can request an output channel.

The weights can be user defined by an iterator of weights. When no weight is provided the average is supposed to return the arithmetic average. We might also want to provide a couple of defaults, e.g. exponential.

**Example**

```python
window = MovingWindow(100, resampled_data_recv)
# use the latest three values for the weighted average
ma_recv = window.moving_avg(weights=[0.2, 0.3, 0.5])
...
```

**Example** (exponential)

```python
window = MovingWindow(100, resampled_data_recv)
# use the latest three values for the weighted average
ma_recv = window.moving_avg(exponential_weights(num=10))
...
```

### Use cases

The main use case is feature extraction for ML applications.

_No response_

### Additional context

_No response_

jack-herrmann:
I am not quite sure about the if-block when new_stop is < 0, but I thought it would be a convenient addition.
Fixes #212

### DIFF
--- a/src/frequenz/sdk/timeseries/_moving_window.py
+++ b/src/frequenz/sdk/timeseries/_moving_window.py
@@ -322,6 +322,17 @@ class MovingWindow:
             if isinstance(key.start, datetime) and isinstance(key.stop, datetime):
                 return self._buffer.window(key.start, key.stop)
             if isinstance(key.start, int) and isinstance(key.stop, int):
+                new_start = self._buffer.wrap(key.start)
+                new_stop = self._buffer.wrap(key.stop)
+                if new_start >= new_stop:
+                    if (
+                        new_stop < 0
+                        and (self._buffer.maxlen - 1) + new_stop >= new_start
+                    ):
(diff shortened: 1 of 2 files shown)


cwasicki:
Will be addressed in https://github.com/frequenz-floss/frequenz-sdk-python/issues/214

PR #527, opened 2023-07-21
Implement dunder iadd, isub, imul methods for Quantity

shsms:
imul is implemented only for `Percentage` because, all other
multiplications that we have produce outputs of a new type, which is
not possible with imul.

With this usage like this becomes possible:

```python
power = Power.from_watts(1000.0)
power += Power.from_watts(3.2)
```


Marenz (approved):

llucax:
This should probably be reverted because of this: https://github.com/frequenz-floss/frequenz-sdk-python/pull/535#discussion_r1274774909

PR #533, opened 2023-07-24
Quantity improvements

### LINKED ISSUE #531: Clarify documentation for quantities
### What's needed?

Currently, the documentation for quantities lacks clear information about the underlying value implementation of a quantity.
It is important to explicitly state that the underlying value of a quantity is a float, and users need to be aware of the
considerations and caveats associated with using floats when working with quantities.

### Proposed solution

We should update the documentation to include a dedicated section that clearly states the underlying value of a quantity is a float.  Users/developers should take into account regular cares regarding using floats apply for quantities too.

### LINKED ISSUE #532: Remove hashing capabilities from Quantity
### What's needed?

Using `Quantity` hashing capabilities could potentially cause unexpected behavior on the users side given that the underlying implementation of quantities uses floating-point values. So if users do not know that they are exactly doing or dealing with, this could just create problems and bugs on their side.

### Proposed solution

Remove hashing capabilities from quantities (function `__hash__()`)

shsms:
Closes #531, Closes #532

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -9,13 +9,15 @@
 <!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
 
 - `Channels` has been upgraded to version 0.16.0, for information on how to upgrade visit https://github.com/frequenz-floss/frequenz-channels-python/releases/tag/v0.16.0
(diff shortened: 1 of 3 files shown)


llucax (commented):
LGTM but release notes should be added.

shsms:
> release notes should be added.

done

llucax (approved):

PR #535, opened 2023-07-26
Add zero() constructor for Quantities

llucax:
It is a very common case which becomes very verbose, and it includes unit information that is completely irrelevant (zero is zero in any unit).

### DIFF
--- a/src/frequenz/sdk/timeseries/_quantities.py
+++ b/src/frequenz/sdk/timeseries/_quantities.py
@@ -24,7 +24,10 @@ QuantityT = TypeVar(
 
 
 class Quantity:
-    """A quantity with a unit."""
+    """A quantity with a unit.
+
+    Quantities try to behave like float and are also immutable.
+    """
 
     _base_value: float
     """The value of this quantity in the base unit."""
@@ -249,48 +252,6 @@ class Quantity:
         product._base_value = self._base_value * percent.as_fraction()
         return product
 
-    def __iadd__(self, other: Self) -> Self:
-        """Add another quantity to this one.
-
-        Args:
-            other: The other quantity.
-
-        Returns:
-            This quantity.
-        """
-        if not type(other) is type(self):
-            return NotImplemented
-        self._base_value += other._base_value
-        return self
-
-    def __isub__(self, other: Self) -> Self:
-        """Subtract another quantity from this one.
-
-        Args:
-            other: The other quantity.
-
-        Returns:
-            This quantity.
-        """
-        if not type(other) is type(self):
-            return NotImplemented
-        self._base_value -= other._base_value
-        return self
-
-    def __imul__(self, percent: Percentage) -> Self:
-        """Multiply this quantity by a percentage.
-
-        Args:
-            percent: The percentage.
-
-        Returns:
-            This quantity.
-        """
-        if not isinstance(percent, Percentage):
-            return NotImplemented
-        self._base_value *= percent.as_fraction()
-        return self
-
     def __gt__(self, other: Self) -> bool:
         """Return whether this quantity is greater than another.
 
@@ -413,7 +374,7 @@ class Power(
 ):
     """A power quantity.
 
-    Objects of this type are wrappers around `float` values.
+    Objects of this type are wrappers around `float` values and are immutable.
 
     The constructors accept a single `float` value, the `as_*()` methods return a
     `float` value, and each of the arithmetic operators supported by this type are
@@ -586,7 +547,7 @@ class Current(
 ):
     """A current quantity.
 
-    Objects of this type are wrappers around `float` values.
+    Objects of this type are wrappers around `float` values and are immutable.
 
     The constructors accept a single `float` value, the `as_*()` methods return a
     `float` value, and each of the arithmetic operators supported by this type are
@@ -682,7 +643,7 @@ class Voltage(
 ):
     """A voltage quantity.
 
-    Objects of this type are wrappers around `float` values.
+    Objects of this type are wrappers around `float` values and are immutable.
 
     The constructors accept a single `float` value, the `as_*()` methods return a
     `float` value, and each of the arithmetic operators supported by this type are
@@ -804,7 +765,7 @@ class Energy(
 ):
     """An energy quantity.
 
-    Objects of this type are wrappers around `float` values.
+    Objects of this type are wrappers around `float` values and are immutable.
 
     The constructors accept a single `float` value, the `as_*()` methods return a
     `float` value, and each of the arithmetic operators supported by this type are
@@ -923,7 +884,7 @@ class Frequency(
 ):
     """A frequency quantity.
 
-    Objects of this type are wrappers around `float` values.
+    Objects of this type are wrappers around `float` values and are immutable.
 
     The constructors accept a single `float` value, the `as_*()` methods return a
     `float` value, and each of the arithmetic operators supported by this type are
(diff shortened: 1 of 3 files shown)


tests/timeseries/test_quantities.py
```diff
@@ -42,6 +42,56 @@ class Fz2(
...
+    assert Frequency.zero().as_hertz() == 0.0
+    assert Frequency.zero().as_megahertz() == 0.0
+    assert Frequency.zero() is Frequency.zero()  # It is a "singleton"
+
+    assert Percentage.from_percent(0.0) == Percentage.zero()
+    assert Percentage.from_fraction(0.0) == Percentage.zero()
+    assert isinstance(Percentage.zero(), Percentage)
+    assert Percentage.zero().as_percent() == 0.0
+    assert Percentage.zero().as_fraction() == 0.0
+    assert Percentage.zero() is Percentage.zero()  # It is a "singleton"
```
Marenz, line 98:
```py
x = Percentage.zero()
x = x + Percentage.from_percent(1)
assert Percentage.zero() == Percentage.from_percent(0)
```

would this assert fail?

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -62,6 +62,30 @@ def __init_subclass__(cls, exponent_unit_map: dict[int, str]) -> None:
...
+    """Cache for zero singletons.
+
+    This is a workaround for mypy getting confused when using @functools.cache and
+    @classmethod combined with returning Self. It believes the resulting type of this
+    method is Self and complains that members of the actual class don't exist in Self,
+    so we need to implement the cache ourselves.
+    """
+
+    @classmethod
+    def zero(cls) -> Self:
```
shsms, line 78: I guess class methods can't be properties?

tests/timeseries/test_quantities.py
shsms: I think it wouldn't,  because `x + ...` generates a new object,  and that gets bound to `x`,  and the zero object's reference count would just get decremented.  It is not possible to get the deref behaviour of other languages in Python I think (at least easily).

Marenz:
```py
        x = Percentage.zero()
        x += Percentage.from_percent(1.0)
>       assert Percentage.zero() == Percentage.from_percent(0.0)
E       AssertionError: assert Percentage(value=1.0, exponent=0) == Percentage(value=0.0, exponent=0)
```

Marenz: but true, with `x = x + ..` it doesn't fail

src/frequenz/sdk/timeseries/_quantities.py
Marenz: No, I think it works

tests/timeseries/test_quantities.py
shsms: :o I forgot we implemented +=, etc.  So now we have to decide whether to keep += or not cache zeros?  My feeling is that zero values are rarely created as part of the hot-path,  so we can just create a new object and not cache them.

Marenz: You could do something terrible and check in `+=` if you are a singleton and reset the according cache

llucax: You can't change the object returned by `Percentage.zero()` unless you do this: `Percentage._zero_cache[Percentage] = x`. Variables in Python are bound to objects, you are just rebinding `x` when doing `x = ...`, nothing more.

src/frequenz/sdk/timeseries/_quantities.py
llucax: Interesting, it occurred to me but I didn't check, I can try.

llucax: Python (tests) works, but `pylint` doesn't :roll_eyes: 
```
tests/timeseries/test_quantities.py:47:11: W0143: Comparing against a callable, did you omit the parenthesis? (comparison-with-callable)
```

And `mypy` explodes (again) :boom: (lately I found `mypy` is very unstable, it happens very often that I make a silly change and `mypy` with start throwing an `INTERNAL ERROR`. Sometimes if I run it outside `nox` it will give an error, but sometimes it doesn't. I guess we are taking it to the limit. :unamused: 

Running with `--show-traceback` at least I get which line is making it crash and then looking inside vim, I see the mypy error, so I guess what's killing it is processing all the files at once, and the `mypy` error is:

```
Only instance methods can be decorated with @property  [misc] (mypy error)
```

So weird, because again, Python does support it. OK, researching some more, it seems like it used to work but it was deprecated in 3.11:

> Chaining classmethod descriptors (introduced in bpo-19072) is now deprecated. It can no longer be used to wrap other descriptors such as property. The core design of this feature was flawed and caused a number of downstream problems.

https://github.com/python/mypy/issues/13746#issuecomment-1258864382

Soooooo.... No, we can't do that :P

It would have been nice though, because then it is obvious that it is a singleton.

tests/timeseries/test_quantities.py
shsms: @llucax we have `+=` now, which mutate the objects.  that's what @Marenz is showing in his example above.

llucax: OK, I didn't see the `+=`, that is a problem indeed because it is mutating the object, so yeah, if objects are not immutable it is an issue. I guess we then need to drop the singleton idea, good catch. I guess it is not **that** bad.

Or, drop mutable method. I'm not sure really, there are very nice properties of objects being immutable (like being able to use them as defaults in function arguments). Now I wonder how built-in types work, because they are immutable but `+=` also work on them.

llucax: I guess you can just return a new object in `__iadd__`, etc. Should we do that?

llucax: It seems like `__i*__` are the only mutating methods. I think for number we really want them to be immutable.

llucax: I just tried removing all `__i*__` methods and then `x += ...` with `x = Quantity.zero()` works. I really think we should do that. I think users would expect to be able to use defaults in function arguments, I think `x += ...` should be rare enough to trade a bit of worse performance for safety.

shsms: you mean `x += ...` works without `__iadd__`,  and acts as alias for `x = x + ...`?  If that's the case, we can remove that.  Maybe in this PR?

src/frequenz/sdk/timeseries/_quantities.py
Marenz: Yeah, I noticed that the mypy people seem to have their own ideas of what is and isn't possible, same with the percentage __mul__ stuff

tests/timeseries/test_quantities.py
llucax: > you mean x += ... works without __iadd__, and acts as alias for x = x + ...?

Exactly.

>  If that's the case, we can remove that. Maybe in this PR?

I can do that.

src/frequenz/sdk/timeseries/_quantities.py
llucax: Well, in this case it is deprecated by Python itself, so it might actually stop working at some point.

tests/timeseries/test_quantities.py
llucax: Done.

llucax:
Updated to revert adding the mutable methods and added a test to make sure zero is immutable.

shsms (approved):

PR #416, opened 2023-05-31
Update `frequenz-api-microgrid` to v0.15.1

shsms:
Closes #412 

To be merged after v0.21 is released.

PR #537, opened 2023-07-26
Stream exclusion bounds from the battery pool

### LINKED ISSUE #524: Define power bound attributes as Power type
### What's needed?

Currently the lower and upper power `Bound` attributes are defined as plain float values. They are used for sending PowerMetrics.

### Proposed solution

Update the`Bound` attributes to type `Power` instead

### Additional context

https://github.com/frequenz-floss/frequenz-sdk-python/blob/v0.x.x/src/frequenz/sdk/timeseries/battery_pool/_result_types.py#L10-L18

shsms:
Based on #416, to be merged after upgrading the Microgrid API dependency to v0.15.1.

Also closes #524

### DIFF
--- a/src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
+++ b/src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
@@ -542,8 +542,8 @@ class PowerBoundsCalculator(MetricCalculator[PowerMetrics]):
         )
 
         for battery_id in working_batteries:
-            supply_upper_bounds: list[float] = []
-            consume_upper_bounds: list[float] = []
+            inclusion_lower_bounds: list[float] = []
+            inclusion_upper_bounds: list[float] = []
 
             if battery_id in metrics_data:
                 data = metrics_data[battery_id]
@@ -553,12 +553,12 @@ class PowerBoundsCalculator(MetricCalculator[PowerMetrics]):
                 value = data.get(ComponentMetricId.POWER_INCLUSION_UPPER_BOUND)
                 if value is not None:
                     result.timestamp = max(result.timestamp, data.timestamp)
-                    consume_upper_bounds.append(value)
+                    inclusion_upper_bounds.append(value)
 
                 value = data.get(ComponentMetricId.POWER_INCLUSION_LOWER_BOUND)
                 if value is not None:
                     result.timestamp = max(result.timestamp, data.timestamp)
-                    supply_upper_bounds.append(value)
+                    inclusion_lower_bounds.append(value)
 
             inverter_id = self._bat_inv_map[battery_id]
             if inverter_id in metrics_data:
@@ -567,17 +567,17 @@ class PowerBoundsCalculator(MetricCalculator[PowerMetrics]):
                 value = data.get(ComponentMetricId.ACTIVE_POWER_INCLUSION_UPPER_BOUND)
                 if value is not None:
                     result.timestamp = max(data.timestamp, result.timestamp)
-                    consume_upper_bounds.append(value)
+                    inclusion_upper_bounds.append(value)
 
                 value = data.get(ComponentMetricId.ACTIVE_POWER_INCLUSION_LOWER_BOUND)
                 if value is not None:
                     result.timestamp = max(data.timestamp, result.timestamp)
-                    supply_upper_bounds.append(value)
+                    inclusion_lower_bounds.append(value)
 
-            if len(consume_upper_bounds) > 0:
-                result.consume_bound.upper += min(consume_upper_bounds)
-            if len(supply_upper_bounds) > 0:
-                result.supply_bound.lower += max(supply_upper_bounds)
+            if len(inclusion_upper_bounds) > 0:
+                result.consume_bound.upper += min(inclusion_upper_bounds)
+            if len(inclusion_lower_bounds) > 0:
+                result.supply_bound.lower += max(inclusion_lower_bounds)
 
         if result.timestamp == _MIN_TIMESTAMP:
             return None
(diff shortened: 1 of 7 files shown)


shsms:
Rebased on latest

src/frequenz/sdk/timeseries/battery_pool/_result_types.py
```diff
@@ -26,38 +28,24 @@ class PowerMetrics:
...
+    # pylint: disable=line-too-long
+    inclusion_bounds: Bounds
+    """Inclusion power bounds for all batteries in the battery pool instance.
+
+    This is the range within which power requests are allowed by the battery pool.
+
+    When exclusion bounds are present, they will exclude a subset of the inclusion
+    bounds.
+
+    More details [here](https://github.com/frequenz-floss/frequenz-api-common/blob/v0.3.0/proto/frequenz/api/common/metrics.proto#L37-L91).
```
llucax, line 40: This could now point to the generated docs: https://frequenz-floss.github.io/frequenz-api-common/next/python-reference/frequenz/api/common/metrics_pb2/#frequenz.api.common.metrics_pb2.Metric.system_exclusion_bounds

We could even just use python symbols for cross referencing if we add these object indexes:

```diff
diff --git a/mkdocs.yml b/mkdocs.yml
index e8ddb46..13fe0f1 100644
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -105,7 +105,9 @@ plugins:
           import:
             # See https://mkdocstrings.github.io/python/usage/#import for details
             - https://docs.python.org/3/objects.inv
+            - https://frequenz-floss.github.io/frequenz-api-common/v0.3/objects.inv
             - https://frequenz-floss.github.io/frequenz-channels-python/v0.14/objects.inv
+            - https://frequenz-floss.github.io/frequenz-api-microgrid/v0.15/objects.inv
             - https://grpc.github.io/grpc/python/objects.inv
             - https://networkx.org/documentation/stable/objects.inv
             - https://numpy.org/doc/stable/objects.inv
```

```suggestion
    See [`frequenz.api.common.metrics_pb2.Metric.system_exclusion_bounds`][] for more details.
```

But for this to work we need to initialize the microgrid API generated docs and we need to tag `v0.3.1` in the common API so we can have a stable link to the docs (for now only the `next` version is provided), so we can update this in a followup PR.

(as a nice side-effect we can remove the `# pylint: disable=line-too-long` too, probably :D)

src/frequenz/sdk/timeseries/battery_pool/_result_types.py
```diff
@@ -26,38 +28,24 @@ class PowerMetrics:
...
-            )
-        )
-    ```
+    exclusion_bounds: Bounds
+    """Exclusion power bounds for all batteries in the battery pool instance.
+
+    This is the range within which power requests are NOT allowed by the battery pool.
+    If present, they will be a subset of the inclusion bounds.
+
+    More details [here](https://github.com/frequenz-floss/frequenz-api-common/blob/v0.3.0/proto/frequenz/api/common/metrics.proto#L37-L91).
```
llucax, line 49: Same.

tests/timeseries/_battery_pool/test_battery_pool.py
```diff
@@ -891,37 +925,68 @@ async def run_power_bounds_test(  # pylint: disable=too-many-locals
...
             {
                 "active_power_inclusion_lower_bound": math.nan,
                 "active_power_inclusion_upper_bound": math.nan,
+                "active_power_exclusion_lower_bound": math.nan,
+                "active_power_exclusion_upper_bound": math.nan,
             },
-            PowerMetrics(now, Bound(-100, 0), Bound(0, 0)),
+            PowerMetrics(
+                now,
+                Bounds(Power.from_watts(-100), Power.from_watts(0)),
```
llucax: Not super important but this could be:
```suggestion
                Bounds(Power.from_watts(-100), Power.zero()),
```

llucax (dismissed):
Only a few very minor comments that can be addressed in a separate PR.

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
```diff
@@ -566,20 +565,27 @@ def calculate(
...
             return None
 
-        return result
+        return PowerMetrics(
+            timestamp=timestamp,
+            inclusion_bounds=Bounds(
+                Power.from_watts(inclusion_bounds_lower),
+                Power.from_watts(inclusion_bounds_upper),
+            ),
+            exclusion_bounds=Bounds(Power.from_watts(0.0), Power.from_watts(0.0)),
```
daniel-zullo-frequenz: ~Here also `Power.zero()`~

daniel-zullo-frequenz (approved):
Luca's suggestions about documentation sound good to me.
I can only add that the release notes should include a reference or link to the documentation of inclusion/exclusion bounds, otherwise the user needs to go through the code to get the info/links. But that can be also done in a separate PR

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
daniel-zullo-frequenz: It does not apply

llucax:
> I can only add that the release notes should include a reference or link to the documentation of inclusion/exclusion bounds, otherwise the user needs to go through the code to get the info/links. But that can be also done is a separate PR

Lets do that in a separate commit, when also improving the links in the code.

src/frequenz/sdk/timeseries/battery_pool/_result_types.py
llucax: * https://github.com/frequenz-floss/frequenz-sdk-python/pull/571

PR #541, opened 2023-07-27
New access functions to the moving window ringbuffer

### LINKED ISSUE #214: MovingWindow: Create Interface for requesting data and handling the data copy behaviour

jack-herrmann:
Fixes #214.

### DIFF
--- a/src/frequenz/sdk/timeseries/_moving_window.py
+++ b/src/frequenz/sdk/timeseries/_moving_window.py
@@ -272,11 +272,13 @@ class MovingWindow:
         Args:
             start: The start index or datetime of the sub window.
             end: The end index or datetime of the sub window.
-            fill_gaps: The value to fill gaps with.
+            fill_gaps: The value to fill gaps with, fill_gaps implies force_copy = True.
             force_copy: If True, a copy of the underlying buffer is returned.
 
         Raises:
             IndexError: if the end index is smaller than the start index.
+            TypeError: if the start and end parameters are not of type int or datetime.
+            ValueError: if fill_gaps is not None and force_copy is False.
 
         Returns:
             A sub window of the `MovingWindow`.
@@ -291,9 +293,16 @@ class MovingWindow:
 
             start_index = self._buffer.datetime_to_index(start)
             end_index = self._buffer.datetime_to_index(end)
+        elif isinstance(start, int) and isinstance(end, int):
+            start_index = start
+            end_index = end
         else:
-            start_index = start  # type: ignore
-            end_index = end  # type: ignore
+            raise TypeError(
+                f"start and end parameters have to be of type int or datetime, but are {type(start)} and {type(end)}"  # pylint: disable=line-too-long
+            )
+
+        if fill_gaps is not None and not force_copy:
+            raise ValueError("fill_gaps can only be used with force_copy=True")
 
         # Make copy of buffer
         buffer_to_use = deepcopy(self._buffer) if (force_copy) else self._buffer
@@ -327,14 +336,7 @@ class MovingWindow:
         # noqa: DAR101 key
         """
 
-    @overload
-    def __getitem__(self, key: slice) -> ArrayLike:
-        """See the main __getitem__ method.
-
-        # noqa: DAR101 key
-        """
-
-    def __getitem__(self, key: SupportsIndex | datetime | slice) -> float:
+    def __getitem__(self, key: SupportsIndex | datetime) -> float:
         """
         Return a sub window of the `MovingWindow`.
 
--- a/src/frequenz/sdk/timeseries/_moving_window.py
+++ b/src/frequenz/sdk/timeseries/_moving_window.py
@@ -9,8 +9,9 @@ import asyncio
 import logging
 import math
 from collections.abc import Sequence
+from copy import deepcopy
 from datetime import datetime, timedelta
-from typing import SupportsIndex, overload
+from typing import Optional, SupportsIndex, overload
 
 import numpy as np
 from frequenz.channels import Broadcast, Receiver, Sender
@@ -258,6 +259,60 @@ class MovingWindow:
         """
         return len(self._buffer)
 
+    def window(
+        self,
+        start: int | datetime,
+        end: int | datetime,
+        fill_gaps: Optional[float] = None,
+        force_copy: bool = False,
+    ) -> ArrayLike:
+        """
+        Access a sub window of the `MovingWindow`.
+
+        Args:
+            start: The start index or datetime of the sub window.
+            end: The end index or datetime of the sub window.
+            fill_gaps: The value to fill gaps with.
+            force_copy: If True, a copy of the underlying buffer is returned.
+
+        Raises:
+            IndexError: if the end index is smaller than the start index.
+
+        Returns:
+            A sub window of the `MovingWindow`.
+        """
+        if len(self._buffer) == 0:
+            raise IndexError("The buffer is empty.")
+        if isinstance(start, datetime) and isinstance(end, datetime):
+            if start > end:
+                raise IndexError(
+                    f"end parameter {end} has to predate start parameter {start}"
+                )
+
+            start_index = self._buffer.datetime_to_index(start)
+            end_index = self._buffer.datetime_to_index(end)
+        else:
+            start_index = start  # type: ignore
+            end_index = end  # type: ignore
+
+        # Make copy of buffer
+        buffer_to_use = deepcopy(self._buffer) if (force_copy) else self._buffer
+
+        # Fill gaps with the given value
+        # Assumption that if fill_gaps is used then also force_copy is used
+        # To make this even more efficient use interval tree
+        if fill_gaps is not None and force_copy:
+            for gap_range in buffer_to_use.gaps:
+                buffer_to_use[gap_range.start : gap_range.end] = [fill_gaps] * len(range(gap_range.start, gap_range.end))  # type: ignore  # pylint: disable=line-too-long
+
+        len_to_use = len(buffer_to_use) if (fill_gaps is None) else buffer_to_use.maxlen
+
+        # Requested window wraps around the ends
+        if start_index >= end_index:
+            if end_index >= 0 or not (len_to_use - 1) + end_index >= start_index:
+                return np.concatenate((buffer_to_use[start_index:], buffer_to_use[0:end_index]))  # type: ignore  # pylint: disable=line-too-long
+        return buffer_to_use[start_index:end_index]
+
     @overload
     def __getitem__(self, key: SupportsIndex) -> float:
         """See the main __getitem__ method.
@@ -279,7 +334,7 @@ class MovingWindow:
         # noqa: DAR101 key
         """
 
-    def __getitem__(self, key: SupportsIndex | datetime | slice) -> float | ArrayLike:
+    def __getitem__(self, key: SupportsIndex | datetime | slice) -> float:
         """
         Return a sub window of the `MovingWindow`.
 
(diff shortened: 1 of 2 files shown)


src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -279,7 +334,7 @@ def __getitem__(self, key: slice) -> ArrayLike:
         # noqa: DAR101 key
         """
 
-    def __getitem__(self, key: SupportsIndex | datetime | slice) -> float | ArrayLike:
+    def __getitem__(self, key: SupportsIndex | datetime | slice) -> float:
```
llucax: Shouldn't `datetime` and `slice` be removed from here?

llucax (commented):
I'll leave the code itself to @Marenz and @matthias-wende-frequenz as they are more familiar, but regarding commits:

1. The merge commit shouldn't be there, you should be able to get rid of it by rebasing.
2. The `New access functions to the moving window ringbuffer` should follow the commit message guideline (which we need to make public as it is only internal right now :-/).

Basically:

---

A commit message consists of two components:

* a single-line title of 72 characters or less written in the
  [imperative mood](https://en.wikipedia.org/wiki/Imperative_mood),
  without any terminating punctuation (no period at the end!)

* one or more detailed paragraphs explaining the changes made in the
  commit, and their motivation and context (i.e. why as well as what),
  with maximum line width of 72 characters

  - commits that implement something related to an issue should
    include a line `Part of {link-to-issue}`

  - commits that fix an issue, or finish a series of changes that
    fix an issue, should include a line `Fixes {link-to-issue}`

  - the full link should be used to cross-reference issues, not
    just an issue number

  - if a commit relates to more than one issue, links to all of
    them should be included

---

So maybe something like this for the title: `Move slicing using datetime to a new `window()` method`. Then the body of the message should be a more in-depth description why this is needed, probably can be extracted from https://github.com/frequenz-floss/frequenz-sdk-python/issues/214.

This also needs an update to the release notes.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -258,6 +259,60 @@ def __len__(self) -> int:
...
+        if isinstance(start, datetime) and isinstance(end, datetime):
+            if start > end:
+                raise IndexError(
+                    f"end parameter {end} has to predate start parameter {start}"
+                )
+
+            start_index = self._buffer.datetime_to_index(start)
+            end_index = self._buffer.datetime_to_index(end)
+        else:
+            start_index = start  # type: ignore
```
Marenz: Please use another `isinstance(..)` check here for int, not `type: ignore`. Ignore should only be used as the very last resort.

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: only slice should be removed, datetime is fine

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -258,6 +259,60 @@ def __len__(self) -> int:
...
+            end_index = end  # type: ignore
+
+        # Make copy of buffer
+        buffer_to_use = deepcopy(self._buffer) if (force_copy) else self._buffer
```
matthias-wende-frequenz, line 312: We should deepcopy also if fill_gaps is not None, i.e. we imply a force copy. If fill_gaps is not None and force_copy is false we should raise an exception.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -258,6 +259,60 @@ def __len__(self) -> int:
...
+        fill_gaps: Optional[float] = None,
+        force_copy: bool = False,
+    ) -> ArrayLike:
+        """
+        Access a sub window of the `MovingWindow`.
+
+        Args:
+            start: The start index or datetime of the sub window.
+            end: The end index or datetime of the sub window.
+            fill_gaps: The value to fill gaps with.
```
matthias-wende-frequenz: Can you please add a comment, that this implies force_copy.

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: agreed. We should only avoid runtime type checks in hot paths. Also you want to assure that the object to proceed with is an integer.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -258,6 +259,60 @@ def __len__(self) -> int:
...
+
+        # Make copy of buffer
+        buffer_to_use = deepcopy(self._buffer) if (force_copy) else self._buffer
+
+        # Fill gaps with the given value
+        # Assumption that if fill_gaps is used then also force_copy is used
+        # To make this even more efficient use interval tree
+        if fill_gaps is not None and force_copy:
+            for gap_range in buffer_to_use.gaps:
+                buffer_to_use[gap_range.start : gap_range.end] = [fill_gaps] * len(range(gap_range.start, gap_range.end))  # type: ignore  # pylint: disable=line-too-long
```
matthias-wende-frequenz, line 319: I'm not sure if we should disable the line-too-long rule. @llucax what do you think?

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: Don't we need to wrap here too? I think we do have to. Might be a bit tricky w.r.t. supporting negative end indices.

src/frequenz/sdk/timeseries/_moving_window.py
llucax: I don't see a reason to remove it, `black` should fix most cases of lines being too long, and it has a soft limit of 88 but I think it allows going up to 110 chars, which IMHO is reasonable. What might happen is that `pylint` is not properly configured and complaining for lines longer than 88.

src/frequenz/sdk/timeseries/_moving_window.py
llucax: I don't see a reason to remove it, `black` should fix most cases of lines being too long, and it has a soft limit of 88 but I think it allows going up to 110 chars, which IMHO is reasonable. What might happen is that `pylint` is not properly configured and complaining for lines longer than 88.

src/frequenz/sdk/timeseries/_moving_window.py
jack-herrmann: Matthias if I understand correctly what you mean, there are checks for the wrapping later on

src/frequenz/sdk/timeseries/_moving_window.py
llucax: I thought mypy complained about using non-ints as indexes, but maybe I'm mixing up things...

matthias-wende-frequenz: are damn! of course you are right, that was one of the motivations for this change.

src/frequenz/sdk/timeseries/_moving_window.py
matthias-wende-frequenz: Sorry, my comment might have be a bit misleading.

The point is, that we want to pass an index as if that `MovingWindow` would be a regular array. But it isn't, since in fact it's a ringbuffer. Therefore, what we have to do is to map the passed index to the index of the actual ringbuffer, which doesn't happen here.

cwasicki:
Ideally follow conventions of moving window in pandas: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.rolling.html

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -258,28 +259,88 @@ def __len__(self) -> int:
...
-    @overload
-    def __getitem__(self, key: SupportsIndex) -> float:
-        """See the main __getitem__ method.
+    def window(
+        self,
+        start: int | datetime,
+        end: int | datetime,
+        fill_gaps: Optional[float] = None,
+        force_copy: bool = False,
+    ) -> ArrayLike:
```
llucax, line 268: I strongly suggest to make some arguments keyword-only, in particular I think `bool`s should always be keyword-only (unless they are the only argument or extremely obvious), as having `window(start, end, 1.0, True)` at the call site makes the call very obscure to read. In particular in this case I think all optional arguments should be keyword only:

```suggestion
    def window(
        self,
        start: int | datetime,
        end: int | datetime,
        *,
        fill_gaps: Optional[float] = None,
        force_copy: bool = False,
    ) -> ArrayLike:
```

So users will be forced to write `window(start, end, fill_gaps=1.0, force_copy=True)` which is much clearer.

Since `start` and `end` have the same type, we could even consider making all keyword-only arguments, as a call `window(end, start)` might not be obviously wrong, while `window(start=end, end=start)` looks more obviously wrong and `window(end=end, start=start)` is still correct. But I guess this error it is less likely to happen so I'm not fully convinced.

At some point it would be nice to write some guideline/convention to recommend common cases where using keyword-only and positional-only arguments work best. I would like to see them used more often.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -258,28 +259,88 @@ def __len__(self) -> int:
...
+        """
+        Access a sub window of the `MovingWindow`.
 
-        # noqa: DAR101 key
+        Args:
+            start: The start index or datetime of the sub window.
+            end: The end index or datetime of the sub window.
+            fill_gaps: The value to fill gaps with, fill_gaps implies force_copy = True.
+            force_copy: If True, a copy of the underlying buffer is returned.
```
cwasicki, line 276: I would add a note here that usually the data is copied even if `False` if the requested window size is of similar size than the buffer.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -258,28 +259,88 @@ def __len__(self) -> int:
...
         return len(self._buffer)
 
-    @overload
-    def __getitem__(self, key: SupportsIndex) -> float:
-        """See the main __getitem__ method.
+    def window(
+        self,
+        start: int | datetime,
+        end: int | datetime,
+        fill_gaps: Optional[float] = None,
```
cwasicki, line 266: Are there really gaps or only the single, initially uninitialized part of the memory?

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -258,28 +259,88 @@ def __len__(self) -> int:
...
+        force_copy: bool = False,
+    ) -> ArrayLike:
+        """
+        Access a sub window of the `MovingWindow`.
```
cwasicki, line 270: This needs more documentation about what gaps are and how they are treated.

cwasicki (changes requested):

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -258,28 +259,88 @@ def __len__(self) -> int:
...
+
+        # Make copy of buffer
+        buffer_to_use = deepcopy(self._buffer) if (force_copy) else self._buffer
+
+        # Fill gaps with the given value
+        # Assumption that if fill_gaps is used then also force_copy is used
+        # To make this even more efficient use interval tree
+        if fill_gaps is not None and force_copy:
+            for gap_range in buffer_to_use.gaps:
+                buffer_to_use[gap_range.start : gap_range.end] = [fill_gaps] * len(range(gap_range.start, gap_range.end))  # type: ignore  # pylint: disable=line-too-long
```
cwasicki, line 319: This line fails with 
```
TypeError: 'datetime.datetime' object cannot be interpreted as an integer
```

tests/timeseries/test_moving_window.py
```diff
@@ -94,12 +94,12 @@ async def test_access_window_by_int_slice() -> None:
...
     window, sender = init_moving_window(timedelta(seconds=14))
-    await push_logical_meter_data(sender, range(0, 5))
-    assert np.array_equal(window[3:5], np.array([3.0, 4.0]))
+    await push_logical_meter_data(sender, range(0, 14))
+    assert np.array_equal(window.window(3, 5), np.array([3.0, 4.0]))
 
     data = [1, 2, 2.5, 1, 1, 1, 2, 2, 1, 1, 1, 1, 1, 1]
     await push_logical_meter_data(sender, data)
-    assert np.array_equal(window[5:14], np.array(data[5:14]))
+    assert np.array_equal(window.window(5, 14), np.array(data[5:14]))
```
cwasicki, line 102: The new window method needs more tests.

cwasicki (changes requested):

src/frequenz/sdk/timeseries/_moving_window.py
cwasicki: @llucax Agree to make most of the arguments keyword only.

With respect to `start` and `end`, both are fine. It depends which behavior we want if start > end. We could
* return an empty array,
* raise and exception,
* or return the array in reverse order.

To me the last option would only make sense if we support more array-like functionality, like real slices including step sizes. However, I would leave this for a later version since it complicates this quite a bit. We would have all functionality on the array that the method returns. Instead I would go with the first option. In fact I would stick as close to other array-like types as possible, but just support start and end. Therefore I don't think we need keywords for start and end as it feels natural to use it this way. As long as we clearly define this in the doc.

llucax:
Closing as we are taking another approach.

PR #544, opened 2023-07-27
Add CI check that RELEASE_NOTES.md was updated

Marenz:
This PR adds a CI check to ensure that all PRs have a release note.
Except when a commit message contains `[skip relnotes]`

This is just a suggestion, if you don't like this solution or have a better idea, please let me know.

### DIFF
--- a/.github/workflows/release_notes_check.yml
+++ b/.github/workflows/release_notes_check.yml
@@ -1,0 +1,25 @@
+name: Release Notes Check
+
+on:
+  pull_request:
+    types:
+      # On by default if you specify no types.
+      - "opened"
+      - "reopened"
+      - "synchronize"
+      # For `skip-label` only.
+      - "labeled"
+      - "unlabeled"
+
+jobs:
+  check-release-notes:
+    name: Check release notes are updated
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check for a release notes update
+        uses: brettcannon/check-for-changed-files@294a99714e0d350b5083472a293d41bc91804e68 # v1.1.1
+        with:
+          file-pattern: "RELEASE_NOTES.md"
+          prereq-pattern: "src/**"
+          skip-label: "cmd:skip-release-notes"
+          failure-message: "Missing a release notes update. Please add one or apply the ${skip-label} label to the pull request"


Marenz:
Someone with the appropriate rights needs to white-list `dorny/paths-filter@v2` in case we decide to use this.

```
[Error: .github#L1](https://github.com/frequenz-floss/frequenz-sdk-python/commit/9f4b625e4a2069ed0c8107dec09ef7ee66f02398#annotation_12952395949)
dorny/paths-filter@v2 is not allowed to be used in frequenz-floss/frequenz-sdk-python. Actions in this workflow must be: within a repository owned by frequenz-floss, created by GitHub, verified in the GitHub Marketplace, or matching the following: yoheimuta/action-protolint@*.
```

.github/workflows/release_notes_check.yml
```diff
@@ -0,0 +1,28 @@
...
+      # On by default if you specify no types.
+      - "opened"
+      - "reopened"
+      - "synchronize"
+      # For `skip-label` only.
+      - "labeled"
+      - "unlabeled"
+
+jobs:
+  check_release_notes:
```
llucax:
```suggestion
  check-release-notes:
```

Just to be consistent with the style in other workflows.

.github/workflows/release_notes_check.yml
```diff
@@ -0,0 +1,28 @@
...
+    runs-on: ubuntu-latest
```
llucax, line 17:
```suggestion
    name: Check release notes are updated
    runs-on: ubuntu-latest
```

.github/workflows/release_notes_check.yml
```diff
@@ -0,0 +1,28 @@
...
+      - name: Check for a release notes update
+        # Should be used once the PR is merged:
+        # https://github.com/brettcannon/check-for-changed-files/pull/83
+        #uses: brettcannon/check-for-changed-files@v1
+        uses: marenz/check-for-changed-files@token-support-packaged
+        with:
+          token: "${{ secrets.GITHUB_TOKEN }}"
+          file-pattern: "RELEASE_NOTES.md"
+          prereq-pattern: "src/**"
+          skip-label: "skip release notes"
```
llucax:
```suggestion
          skip-label: "cmd:skip-release-notes"
```

Also to make it more consistent with our label style. I updated the label already.

llucax (commented):
Small nitpicking but LGTM! I say we wait a bit to see if the upstream change is merged. This should be copied to repo-config too, and we need to add instructions about the labels that need to be created for this (to repo-config I mean, not here).

.github/workflows/release_notes_check.yml
```diff
@@ -0,0 +1,26 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check for a release notes update
+        uses: frequenz-floss/check-for-changed-files@token-support
+        with:
+          token: "${{ secrets.GITHUB_TOKEN }}"
+          file-pattern: "RELEASE_NOTES.md"
+          prereq-pattern: "src/**"
+          skip-label: "cmd:skip-release-notes"
+          failure-message: "Missing a release notes update. Please add one or apply the ${skip-label} label to the pull request"
```
llucax, line 25: Side note: does this `${skip-label}` really work? Do you have a sample repo where this is running (and failing) to see how it looks like?

Marenz: Yes, I used it and it worked. Example: https://github.com/frequenz-io/frequenz-actor-fcr-balancing/actions/runs/5715881521

.github/workflows/release_notes_check.yml
```diff
@@ -0,0 +1,26 @@
...
+      - "labeled"
+      - "unlabeled"
+
+jobs:
+  check-release-notes:
+    name: Check release notes are updated
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check for a release notes update
+        uses: frequenz-floss/check-for-changed-files@token-support
```
llucax: Since this is a public repo I think it is safe to use the upstream repo, and we don't need the token support either, so if you update to use upstream we can merge already (maybe use a hash instead of the version for extra security).

Marenz:
Then you have to white-list it now

```
[Error: .github#L1](https://github.com/frequenz-floss/frequenz-sdk-python/commit/7018f2833a3e5c25ab4add9bd82d6f502e5d1885#annotation_13076322087)
brettcannon/check-for-changed-files@294a99714e0d350b5083472a293d41bc91804e68 is not allowed to be used in frequenz-floss/frequenz-sdk-python. Actions in this workflow must be: within a repository owned by frequenz-floss, created by GitHub, verified in the GitHub Marketplace, or matching the following: yoheimuta/action-protolint@*.
```

.github/workflows/release_notes_check.yml
```diff
@@ -0,0 +1,25 @@
...
+      - "labeled"
+      - "unlabeled"
+
+jobs:
+  check-release-notes:
+    name: Check release notes are updated
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check for a release notes update
+        uses: brettcannon/check-for-changed-files@294a99714e0d350b5083472a293d41bc91804e68
```
llucax: Can you add a comment with the tag or branch being used? I think it's needed for dependabot to take action. See https://github.com/frequenz-floss/frequenz-sdk-python/blob/v0.x.x/.github/workflows/labeler.yml#L21 for example.

Marenz:
Updated. Still missing the white-list entry

llucax:
Done!

llucax (approved):

llucax:
Can you create a PR with this in repo-config?

PR #552, opened 2023-07-28
Temperature streaming from the BatteryPool

### LINKED ISSUE #466: `BatteryPool` should expose a temperature receiver

shsms:
Closes #466

### DIFF
--- a/src/frequenz/sdk/microgrid/component/_component_data.py
+++ b/src/frequenz/sdk/microgrid/component/_component_data.py
@@ -142,7 +142,7 @@ class BatteryData(ComponentData):
     """
 
     temperature: float
-    """The (average) temperature reported by the battery a battery, in Celcius (°C)."""
+    """The (average) temperature reported by the battery, in Celcius (°C)."""
 
     _relay_state: battery_pb.RelayState.ValueType
     """State of the battery relay."""
--- a/src/frequenz/sdk/microgrid/component/_component_data.py
+++ b/src/frequenz/sdk/microgrid/component/_component_data.py
@@ -141,8 +141,8 @@ class BatteryData(ComponentData):
         This will be a positive number, or zero if no charging is possible.
     """
 
-    temperature_max: float
-    """The maximum temperature of all the blocks in a battery, in Celcius (°C)."""
+    temperature: float
+    """The (average) temperature reported by the battery a battery, in Celcius (°C)."""
 
     _relay_state: battery_pb.RelayState.ValueType
     """State of the battery relay."""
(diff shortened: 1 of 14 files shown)


llucax (commented):
LGTM, needs release notes.

shsms:
> LGTM, needs release notes.

done

src/frequenz/sdk/microgrid/component/_component_data.py
```diff
@@ -141,8 +141,8 @@ class BatteryData(ComponentData):
         This will be a positive number, or zero if no charging is possible.
     """
 
-    temperature_max: float
-    """The maximum temperature of all the blocks in a battery, in Celcius (°C)."""
+    temperature: float
+    """The (average) temperature reported by the battery a battery, in Celcius (°C)."""
```
llucax:
```suggestion
    """The (average) temperature reported by the battery, in Celcius (°C)."""
```

llucax (commented):
Sorry, just discovered some typo/left-over.

As a side note I'm also still wondering if we don't want to explicitly say that this is average temperature, but I guess this was already discussed and was agreed to leave it without?

shsms:
> As a side note I'm also still wondering if we don't want to explicitly say that this is average temperature, but I guess this was already discussed and was agreed to leave it without?

do you mean in the field name?  because it says average in the docs: ```The (average) temperature reported by the battery, in Celcius (°C)```

llucax:
Yeah, field name.

llucax (dismissed):
Approving as I think it is very likely that we don't want to change it, and if we do it is easier to do in a followup PR.

shsms:
Had to merge 0.x.x because of conflict.  Couldn't rebase, because the PR already has a merge commit.

llucax (approved):

PR #557, opened 2023-08-01
Update battery pool SoC calculation

matthias-wende-frequenz:
Fix the SoC calculation such that it never exceeds 100%.

### DIFF
--- a/src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
+++ b/src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
@@ -424,7 +424,8 @@ class SoCCalculator(MetricCalculator[Sample[Percentage]]):
             soc_scaled = (
                 (soc - soc_lower_bound) / (soc_upper_bound - soc_lower_bound) * 100
             )
-            soc_scaled = max(soc_scaled, 0)
+            # we are clamping here because the SoC might be out of bounds
+            soc_scaled = min(max(soc_scaled, 0), 100)
             timestamp = max(timestamp, metrics.timestamp)
             used_capacity_x100 += usable_capacity_x100 * soc_scaled
             total_capacity_x100 += usable_capacity_x100
--- a/src/frequenz/sdk/timeseries/battery_pool/battery_pool.py
+++ b/src/frequenz/sdk/timeseries/battery_pool/battery_pool.py
@@ -347,16 +347,17 @@ class BatteryPool:
     def soc(self) -> MetricAggregator[Sample[Percentage]]:
         """Fetch the normalized average weighted-by-capacity SoC values for the pool.
 
-        The values are normalized to the 0-100% range.
+        The values are normalized to the 0-100% range and clamped if the SoC is out of
+        bounds.
 
         Average soc is calculated with the formula:
         ```
         working_batteries: Set[BatteryData] # working batteries from the battery pool
 
-        soc_scaled = max(
+        soc_scaled = min(max(
             0,
             (soc - soc_lower_bound) / (soc_upper_bound - soc_lower_bound) * 100,
-        )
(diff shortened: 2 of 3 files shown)


src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
```diff
@@ -424,7 +424,7 @@ def calculate(
             soc_scaled = (
                 (soc - soc_lower_bound) / (soc_upper_bound - soc_lower_bound) * 100
             )
-            soc_scaled = max(soc_scaled, 0)
+            soc_scaled = min(max(soc_scaled, 0), 100)
```
daniel-zullo-frequenz, line 428: This is clamping `soc_scaled` from 0 to 100 and I understand it fixes the error. However I have the feeling the previous code and the current patch are hiding the fact that `soc` is not within the bounds to get a value lower than 0 or greater than 100. If that's the case I'd suggest to explicitly clamp the `soc` to the bounds instead.
Something around:
```python
soc = max(min(soc, soc_upper_bound), soc_lower_bound)
soc_scaled = (soc - soc_lower_bound) / (soc_upper_bound - soc_lower_bound) * 100
```
Though my suggestion does not protect us for cases where the bounds were not set correctly. So in that sense clamping from 0 to 100 is safer.

src/frequenz/sdk/timeseries/battery_pool/battery_pool.py
```diff
@@ -353,10 +353,11 @@ def soc(self) -> MetricAggregator[Sample[Percentage]]:
         ```
         working_batteries: Set[BatteryData] # working batteries from the battery pool
 
-        soc_scaled = max(
+        # TODO what bound is used?
```
daniel-zullo-frequenz: do you want to keep this TODO message? just checking...

daniel-zullo-frequenz (commented):
Other than a suggestion to consider and a question, LGTM
Also this might need an entry to the release notes as it is fixing the SoC metric.

llucax:
Also release notes.

src/frequenz/sdk/timeseries/battery_pool/battery_pool.py
matthias-wende-frequenz: no, classic oversight from my side :).

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
daniel-zullo-frequenz: We agreed IRL to keep the current changes and add an inline comment to clarify `soc` might be out of bounds

src/frequenz/sdk/timeseries/battery_pool/battery_pool.py
```diff
@@ -347,16 +347,17 @@ def consumption_power(self) -> FormulaEngine[Power]:
     def soc(self) -> MetricAggregator[Sample[Percentage]]:
         """Fetch the normalized average weighted-by-capacity SoC values for the pool.
 
-        The values are normalized to the 0-100% range.
+        The values are normalized to the 0-100% range and if clamped if the SoC is
```
daniel-zullo-frequenz:
```suggestion
        The values are normalized to the 0-100% range and clamped if the SoC is
```

daniel-zullo-frequenz (commented):
A last cosmetic suggestion and LGTM

src/frequenz/sdk/timeseries/battery_pool/battery_pool.py
matthias-wende-frequenz: fixed

daniel-zullo-frequenz (approved):

PR #561, opened 2023-08-03
Add min and max operations to formula engine

matthias-wende-frequenz:


matthias-wende-frequenz:
needs update of release notes

tests/timeseries/_formula_engine/test_formula_composition.py
```diff
@@ -171,6 +171,61 @@ async def test_formula_composition_missing_bat(self, mocker: MockerFixture) -> N
...
+        await mockgrid.start(mocker)
+
+        logical_meter = microgrid.logical_meter()
+        engine_min = logical_meter.grid_power.min(Power.from_watts(0)).build(
+            "grid_power_min"
+        )
+        engine_min_rx = engine_min.new_receiver()
+        engine_max = logical_meter.grid_power.max(Power.from_watts(0)).build(
+            "grid_power_max"
+        )
```
shsms: I think because this is modifying the formula engine,  and has interactions with other operations that are part of the formula engine,  we have to add a lot more test cases,  like between formulas,  and with `+`, `-`, etc.

And I think the best place to do that is under this class: https://github.com/frequenz-floss/frequenz-sdk-python/blob/cfede0a9fa8d4b7ff133941b8a84b8ef1178f664/tests/timeseries/test_formula_engine.py#L300, because that already has a framework for adding many cases easily.

src/frequenz/sdk/timeseries/_formula_engine/_formula_steps.py
```diff
@@ -132,6 +132,52 @@ def apply(self, eval_stack: List[float]) -> None:
...
+class Maximizer(FormulaStep):
+    """A formula step that represents the max function."""
+
+    def __repr__(self) -> str:
+        """Return a string representation of the step.
+
+        Returns:
+            A string representation of the step.
+        """
+        return "max"
```
daniel-zullo-frequenz, line 144: Not related to this PR, but do you know why all `FormulaStep` are using `__repr__` rather than `__str__`? It seems to me `__str__` is more suitable for these use-cases but I might be missing something.

tests/timeseries/_formula_engine/test_formula_composition.py
```diff
@@ -171,6 +171,61 @@ async def test_formula_composition_missing_bat(self, mocker: MockerFixture) -> N
...
+
+        await mockgrid.mock_resampler.send_meter_power([100.0])
+
+        # Test min
+        min_pow = await engine_min_rx.receive()
+        assert (
+            min_pow is not None
+            and min_pow.value is not None
+            and math.isclose(min_pow.value.base_value, 0.0)
+        )
```
daniel-zullo-frequenz: Optionally you might want to define this assertion and similar ones in a more "pythonic" way
```suggestion
        assert (
            min_pow
            and min_pow.value
            and math.isclose(min_pow.value.base_value, 0.0)
        )
```

tests/timeseries/_formula_engine/test_formula_composition.py
```diff
@@ -171,6 +171,61 @@ async def test_formula_composition_missing_bat(self, mocker: MockerFixture) -> N
 
         assert count == 10
 
+    async def test_formula_composition_min_max(self, mocker: MockerFixture) -> None:
+        """Test the composition of formulas with min/max values."""
+        mockgrid = MockMicrogrid(grid_side_meter=True)
+        await mockgrid.start(mocker)
+
+        logical_meter = microgrid.logical_meter()
+        engine_min = logical_meter.grid_power.min(Power.from_watts(0)).build(
```
daniel-zullo-frequenz: You might want to use the quantity zero constructor from SDK v0.23.0, `Power.zero()` instead

tests/timeseries/_formula_engine/test_formula_composition.py
```diff
@@ -171,6 +171,61 @@ async def test_formula_composition_missing_bat(self, mocker: MockerFixture) -> N
...
+            and min_pow.value is not None
+            and math.isclose(min_pow.value.base_value, 0.0)
+        )
+
+        # Test max
+        max_pow = await engine_max_rx.receive()
+        assert (
+            max_pow is not None
+            and max_pow.value is not None
+            and math.isclose(max_pow.value.base_value, 100.0)
```
daniel-zullo-frequenz: `isclose()` is avalilable for quantities since SDK v0.23.0

src/frequenz/sdk/timeseries/_formula_engine/_formula_steps.py
christianparpart: `__repr__` is also supposed to yield a Python compatible syntax that can be used for reconstruction.

Mind, I **might** be touching `__str__` in my own PR right now / this week, because I'm implementing an actual pretty printer for formula.

tests/timeseries/_formula_engine/test_formula_composition.py
christianparpart: @daniel-zullo-frequenz but the `isclose()` inside SDK is a private implementation that's not exposed. Not sure how you want to handle such cases for SDK's own test cases though.

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
```diff
@@ -45,12 +47,14 @@
...
 
 _operator_precedence = {
-    "(": 0,
-    "/": 1,
-    "*": 2,
-    "-": 3,
-    "+": 4,
-    ")": 5,
+    "max": 0,
+    "min": 1,
```
christianparpart, line 51: why is `min` and `max` having two different operator precedences?

And it's a global non-constant map, why is it all lower-case and not upper-case? (just asking, trying to understand) :)

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
```diff
@@ -650,15 +688,15 @@ def _push(
...
         self._steps.append((TokenType.OPER, oper))
 
-        # pylint: disable=protected-access
         if isinstance(other, (FormulaEngine, FormulaEngine3Phase)):
             self._steps.append((TokenType.COMPONENT_METRIC, other))
-        elif isinstance(other, (Quantity, float)):
+        elif isinstance(other, (Quantity, float, int)):
             match oper:
-                case "+" | "-":
+                case "+" | "-" | "max" | "min":
```
christianparpart, line 695: is min and max really supposed to be a binary expression, such as `A min B` rather than `min(A, B)`? If so, why? :)

tests/timeseries/_formula_engine/test_formula_composition.py
daniel-zullo-frequenz: `isclose()` is a public functionality for `quantities`, see https://github.com/frequenz-floss/frequenz-sdk-python/blob/v0.x.x/src/frequenz/sdk/timeseries/_quantities.py#L131C9-L131C16

So my suggestion was to write that line as `max_pow.value.isclose(Power.from_watts(100.0))`  or something around that

daniel-zullo-frequenz: >  the isclose() inside SDK is a private implementation that's not exposed.

Where did you get that from? Maybe you got confused by `is_close_to_zero()` which it is indeed a SDK internal function?

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
shsms: I guess we're starting to see cracks in the postfix evaluation approach.

src/frequenz/sdk/timeseries/_formula_engine/_formula_steps.py
matthias-wende-frequenz: You are right. The way we use it, it seems that `__str__`is mor suitable. I will not touch it in this PR though and rather change it later.

matthias-wende-frequenz: Actually that is used when pushing operators to the stack and there is a dict with corresponding names. So it makes sense after all.

matthias-wende-frequenz: we can still rename them to longer names representing the classes and add str representation. I've created an issue for later :).
https://github.com/frequenz-floss/frequenz-sdk-python/issues/614

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
matthias-wende-frequenz: ~actually I think it doesn't make much of a difference. But you are right technically they can have the same precedence.~ That's was wrong :)

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
matthias-wende-frequenz: can you please give an example for the cracks?

matthias-wende-frequenz: btw. it's implemented as A.max(B) ;)

matthias-wende-frequenz:
Updated the PR and addressed all comments except that it needs more tests ~and CI is still failing due to formatting issues~.

tests/timeseries/_formula_engine/test_formula_composition.py
```diff
@@ -170,6 +170,45 @@ async def test_formula_composition_missing_bat(self, mocker: MockerFixture) -> N
...
+        assert (
+            max_pow and max_pow.value and max_pow.value.isclose(Power.from_watts(100.0))
+        )
+
+        await mockgrid.mock_resampler.send_meter_power([-100.0])
+
+        # Test min
+        min_pow = await engine_min_rx.receive()
+        assert (
+            min_pow and min_pow.value and max_pow.value.isclose(Power.from_watts(100.0))
```
daniel-zullo-frequenz: this is checking `max_pow` and I understand it should be checking `min_pow` instead

daniel-zullo-frequenz (commented):
I have only a comment to check/amend a test, LGTM otherwise

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
shsms: I guess this PR is the example.  As @christianparpart points out, `min` and `max` are not binary operators, but functions.  We do seem to be able to hack it into our postfix expression evaluator, but it is not ideal, and we still have to add tests and prove that it works in all cases, because this kind of thing can get very buggy very quickly.

I was thinking about an ast walker when doing the original implementation, but the requirements were so simple that it didn't make sense.  If we want to extend the possible operations further after this, we should replace the implementation first.

tests/timeseries/_formula_engine/test_formula_composition.py
matthias-wende-frequenz: ups nice catch. Also it should have been close to -100, the value that has been send. Two errors accumulating to a working test.

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
matthias-wende-frequenz: There is no difference between operators and functions here. The + and * operators are simply functions mapping from float x float → float same as min and max.

daniel-zullo-frequenz (approved):

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
christianparpart: Whenever that topic might be prioritized, I'd love to implement the AST and related work

matthias-wende-frequenz: > Whenever that topic might be prioritized, I'd love to implement the AST and related work

What do you mean? There is no problem here so far.

PR #562, opened 2023-08-03
Support exclusion bounds in power distributor

### LINKED ISSUE #152: PowerDistributingActor should use min SoP (inclusion/exclusion bounds)
### What's needed?

Microgrid will expose min bounds for charge/discharge power soon.
After that we should update PowerDistributing algorithm to look at it.

### Proposed solution

Implementation should be in old MOS repository, because in past we had this min bounds.
 To be checked

### Use cases

_No response_

_No response_

### Additional context

_No response_

shsms:
To be merged after #537 

Closes #152

### DIFF
--- a/src/frequenz/sdk/actor/power_distributing/power_distributing.py
+++ b/src/frequenz/sdk/actor/power_distributing/power_distributing.py
@@ -226,20 +226,18 @@ class PowerDistributingActor:
             bat_id: None for bat_id, _ in self._bat_inv_map.items()
         }
 
-    def _get_bounds(self, batteries: abc.Set[int], include_broken: bool) -> PowerBounds:
+    def _get_bounds(
+        self,
+        pairs_data: list[InvBatPair],
+    ) -> PowerBounds:
         """Get power bounds for given batteries.
 
         Args:
-            batteries: List of batteries
-            include_broken: whether all batteries in the batteries set in the
-                request must be used regardless the status.
+            pairs_data: list of battery and adjacent inverter data pairs.
 
         Returns:
             Power bounds for given batteries.
         """
-        pairs_data: List[InvBatPair] = self._get_components_data(
-            batteries, include_broken
-        )
         return PowerBounds(
             inclusion_lower=sum(
                 max(
@@ -352,11 +350,6 @@ class PowerDistributingActor:
         await asyncio.sleep(self._wait_for_data_sec)
 
         async for request in self._requests_receiver:
-            error = self._check_request(request)
-            if error:
-                await self._send_result(request.namespace, error)
-                continue
-
             try:
                 pairs_data: List[InvBatPair] = self._get_components_data(
                     request.batteries, request.include_broken_batteries
@@ -372,6 +365,11 @@ class PowerDistributingActor:
                 await self._send_result(
                     request.namespace, Error(request=request, msg=str(error_msg))
                 )
+                continue
+
+            error = self._check_request(request, pairs_data)
+            if error:
+                await self._send_result(request.namespace, error)
                 continue
 
             try:
@@ -497,11 +495,16 @@ class PowerDistributingActor:
 
         return result
 
-    def _check_request(self, request: Request) -> Optional[Result]:
+    def _check_request(
+        self,
+        request: Request,
+        pairs_data: List[InvBatPair],
+    ) -> Optional[Result]:
         """Check whether the given request if correct.
 
         Args:
             request: request to check
+            pairs_data: list of battery and adjacent inverter data pairs.
 
         Returns:
             Result for the user if the request is wrong, None otherwise.
@@ -517,7 +520,7 @@ class PowerDistributingActor:
                 )
                 return Error(request=request, msg=msg)
 
-        bounds = self._get_bounds(request.batteries, request.include_broken_batteries)
+        bounds = self._get_bounds(pairs_data)
         if request.adjust_power:
             # Automatic power adjustments can only bring down the requested power down
             # to the inclusion bounds.
--- a/src/frequenz/sdk/actor/power_distributing/power_distributing.py
+++ b/src/frequenz/sdk/actor/power_distributing/power_distributing.py
@@ -517,19 +517,27 @@ class PowerDistributingActor:
                 )
                 return Error(request=request, msg=msg)
 
-        if not request.adjust_power:
-            if request.power < 0:
-                bound = self._get_lower_bound(
-                    request.batteries, request.include_broken_batteries
-                )
-                if request.power < bound:
-                    return OutOfBound(request=request, bound=bound)
-            else:
-                bound = self._get_upper_bound(
-                    request.batteries, request.include_broken_batteries
-                )
-                if request.power > bound:
-                    return OutOfBound(request=request, bound=bound)
+        bounds = self._get_bounds(request.batteries, request.include_broken_batteries)
+        if request.adjust_power:
+            # Automatic power adjustments can only bring down the requested power down
+            # to the inclusion bounds.
+            #
+            # If the requested power is in the exclusion bounds, it is NOT possible to
+            # increase it so that it is outside the exclusion bounds.
+            if (
+                request.power > bounds.exclusion_lower
+                and request.power < bounds.exclusion_upper
+            ):
+                return OutOfBound(request=request, bound=bounds)
+        else:
+            in_lower_range = (
+                bounds.inclusion_lower <= request.power <= bounds.exclusion_lower
+            )
+            in_upper_range = (
+                bounds.exclusion_upper <= request.power <= bounds.inclusion_upper
+            )
+            if not (in_lower_range or in_upper_range):
+                return OutOfBound(request=request, bound=bounds)
 
         return None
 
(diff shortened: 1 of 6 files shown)


shsms:
Still need to add tests.

src/frequenz/sdk/actor/power_distributing/power_distributing.py
```diff
@@ -472,19 +520,27 @@ def _check_request(self, request: Request) -> Optional[Result]:
...
-                )
-                if request.power > bound:
-                    return OutOfBound(request=request, bound=bound)
+        bounds = self._get_bounds(pairs_data)
+        if request.adjust_power:
+            # Automatic power adjustments can only bring down the requested power down
+            # to the inclusion bounds.
+            #
+            # If the requested power is in the exclusion bounds, it is NOT possible to
+            # increase it so that it is outside the exclusion bounds.
```
llucax, line 529: Why is it not OK to increase power but it's OK to decrease power when adjusting was requested?

tests/actor/test_power_distributing.py
```diff
@@ -396,7 +397,7 @@ async def test_power_distributor_invalid_battery_id(
         result: Result = done.pop().result()
         assert isinstance(result, Error)
         assert result.request == request
-        err_msg = re.search(r"^No battery 100, available batteries:", result.msg)
+        err_msg = re.search(r"No battery 100, available batteries:", result.msg)
```
llucax, line 400: Why is the `^` removed? It doesn't look like the message changed.

```suggestion
        err_msg = re.search(r"^No battery 100, available batteries:", result.msg)
```

src/frequenz/sdk/power/_distribution_algorithm.py
```diff
@@ -336,57 +346,98 @@ def _distribute_power(
...
             battery_availability_ratio,
             sum_ratio,
-        ) = self._compute_battery_availability_ratio(components, available_soc)
+        ) = self._compute_battery_availability_ratio(
+            components, available_soc, excl_bounds
+        )
 
         distribution: Dict[int, float] = {}
-
+        print(f"{power_w=}")
```
llucax: Leftover `print`

```suggestion
```

src/frequenz/sdk/power/_distribution_algorithm.py
```diff
@@ -336,57 +346,98 @@ def _distribute_power(
...
+                deficits[inverter.component_id] = calculated_power - excl_bound
             else:
-                distributed_power += distribution[inverter.component_id]
+                excess_reserved[inverter.component_id] = calculated_power - excl_bound
+
+            distributed_power += excl_bound
+            distribution[inverter.component_id] = excl_bound
+
+        for inverter_id, deficit in deficits.items():
+            while not math.isclose(deficit, 0.0, abs_tol=1e-6) and deficit < 0.0:
```
llucax: You could use `._internal._math.is_close_to_zero()` here.
```suggestion
            while not is_close_to_zero(deficit) and deficit < 0.0:
```

src/frequenz/sdk/power/_distribution_algorithm.py
```diff
@@ -336,57 +346,98 @@ def _distribute_power(
...
+                take_from = max(excess_reserved.items(), key=lambda item: item[1])
+                if math.isclose(take_from[1], 0.0, abs_tol=1e-6) or take_from[1] < 0.0:
```
llucax:
```suggestion
                if is_close_to_zero(take_from[1]) or take_from[1] < 0.0:
```

src/frequenz/sdk/power/_distribution_algorithm.py
```diff
@@ -336,57 +346,98 @@ def _distribute_power(
...
+                    break
+                if take_from[1] >= -deficit or math.isclose(
+                    take_from[1], -deficit, abs_tol=1e-6
```
llucax, line 408: I `deficit` supposed to be close to zero? `abs_tol` seems to be [mainly useful for comparing to near-zero values](https://docs.python.org/3/library/math.html#math.isclose) AFAIK, not sure if it is generally OK to use it for other values or not. Maybe @matthias-wende-frequenz can speak his mathematician mind...

llucax (commented):
Except for the `abs_tol` for `deficit` (which I'm not sure it is an issue or not), nothing major from my side. I didn't dig very deep into the distribution algorithm itself. The rest can be easily applied by accepting the suggestions.

src/frequenz/sdk/actor/power_distributing/power_distributing.py
shsms: This has been our practice I think.  Going towards zero can be automatic/not away from.  Increasing power has to be a considered decision, not automatic, because for example the minimum might be too high for sustained maintenance, and could lead to peaks or unexpected soc drops, etc.

tests/actor/test_power_distributing.py
shsms: If you look at the corresponding commit,  https://github.com/frequenz-floss/frequenz-sdk-python/pull/562/commits/de1331d6bbc6adbca8dafd1198534456e087c91d, `check_request` call is moved downwards,  and so this error is caught and raised by a different `send_result` which wraps an error,  so the string is wrapped by quotes.

src/frequenz/sdk/power/_distribution_algorithm.py
shsms: This is a `>=`.  But the `=` part is not exact, value is sometimes slightly on the other side.  hence the `isclose`.

src/frequenz/sdk/actor/power_distributing/power_distributing.py
```diff
@@ -472,19 +520,27 @@ def _check_request(self, request: Request) -> Optional[Result]:
...
+            if (
+                request.power > bounds.exclusion_lower
+                and request.power < bounds.exclusion_upper
+            ):
```
daniel-zullo-frequenz: python:tm: 
```suggestion
            if bounds.exclusion_lower < request.power < bounds.exclusion_upper:
```

src/frequenz/sdk/power/_distribution_algorithm.py
```diff
@@ -336,57 +346,98 @@ def _distribute_power(
...
+            distribution[inverter_id] += excess
+            distributed_power += excess
+
+        for inverter_id, deficit in deficits.items():
+            if deficit < -0.1:
+                left_over = power_w - distributed_power
+                if left_over > -deficit:
+                    distributed_power += deficit
+                    deficit = 0.0
+                    deficits[inverter_id] = 0.0
```
daniel-zullo-frequenz, line 428: I think `deficits` does not need to be updated in this for-loop

src/frequenz/sdk/power/_distribution_algorithm.py
```diff
@@ -336,57 +346,98 @@ def _distribute_power(
...
+                elif left_over > 0.0:
+                    deficit += left_over
+                    distributed_power += left_over
+                    deficits[inverter_id] = deficit
```
daniel-zullo-frequenz, line 432: Updating `deficits` might be redundant here as it will no longer be used after the for-loop

src/frequenz/sdk/power/_distribution_algorithm.py
```diff
@@ -336,57 +346,98 @@ def _distribute_power(
...
+                else:
+                    deficit += excess_reserved[take_from[0]]
+                    deficits[inverter_id] = deficit
+                    excess_reserved[take_from[0]] = 0.0
+
+        for inverter_id, excess in excess_reserved.items():
+            distribution[inverter_id] += excess
+            distributed_power += excess
+
+        for inverter_id, deficit in deficits.items():
```
daniel-zullo-frequenz, line 422: This duplicated for-loop is not needed as its content could be added in the first loop for deficits, right after the end of the `while`-loop

tests/actor/test_power_distributing.py
llucax: Ah, ok, It was hidden in the diff, expanding it now I can see it, thanks for the clarification!

src/frequenz/sdk/power/_distribution_algorithm.py
llucax: Yeah, what I'm asking if it is OK is specifically the `abs_tol` argument.

daniel-zullo-frequenz (approved):

src/frequenz/sdk/power/_distribution_algorithm.py
matthias-wende-frequenz: I don't understand the full algorithm but the check is fine. It's just about having a higher certainty the the `deficit` is close the excess_reserved under consideration (the `take_from[1]`).

daniel-zullo-frequenz: I think this should use the relative tolerance instead as `take_from[1]` and `deficit` are already compared to zero above where the absolute tolerance makes more sense

src/frequenz/sdk/power/_distribution_algorithm.py
shsms: true,  only the `distributed_power` needs to be updated.  I guess I added the other two just to be able to properly reason about what was going on as I was developing,  but they don't have to be there.

llucax: Maybe add a comment about it in the refactor PR if it would make reasoning about it easier.

src/frequenz/sdk/power/_distribution_algorithm.py
llucax: The relative tolerance is defined by default (`rel_tol=1e-09`), this is why I asked only about removing `tol_abs`.

llucax:
* https://github.com/frequenz-floss/frequenz-sdk-python/issues/572

PR #564, opened 2023-08-04
Implement BackgroundService and new Actor class

### LINKED ISSUE #45: Fix typing for Actor's `join()` method
### What happened?

When calling the `join()` method of an actor (or `_stop()` or probably any method defined by the `@actor` decorator) we get an error from `mypy` complaining about the method not existing in the actor class, for example:

```
test.py:10: error: "SampleActor" has no attribute "join"
```

`pylint` seems to be complaining too.

No errors when calling the method.

### Affected version(s)

All

### Affected part(s)

Actors

### Extra information

The `typing_extensions` package (#19), in particular the `@dataclass_transform()` decorator, might help with this, it needs further investigation.

As a workaround, `# noqa` can be used when the method is called.

Here is an example where we need to use `# noqa`:

https://github.com/frequenz-floss/frequenz-sdk-python/blob/20e1dffadcb728443190e1dcfae45d68545ba962/tests/power_distribution/test_power_distributor.py#L193

### LINKED ISSUE #196: Don't start actor automatically in the __init__()
### What's needed?

Probably all applications will need to start some actors and then wait for them to finish indefinitely. So far this is on the SDK users, but it could be handled by SDK. Also this allows us to find better methods to wait for all actors (or tasks) to finish.

In this case the SDK can also provide sanity checks, like the `microgrid_api` is properly initialized, before starting all the actors, or even do the initialization itself if we can provide the run function some config.

This issue is complementary to:
* #36

### Proposed solution

Stop auto-starting actors in the init, and provide a `start()` method instead. Then we can have an even more simplified run function:

Taking this more to the extreme, if we get all the initialization parameters from
This makes me think we should probably make an `initialize()` function at the `sdk` level too).

```py
async def run(config: Config, *actors: Actor) -> None:
    await microgrid_api.initialize(config.microgrid.host, config.microgrid.port)
    await asyncio.gather(actor.start() for actor in actors)
    await asyncio.gather(actor.join() for actor in actors)
```

Then clients will be like:

```python
from frequenz import sdk

config = sdk.Config(...)

async def main() -> None:
    actor1 = Actor1()
    actor2 = Actor2()
    actor3 = Actor3()

    await sdk.run(config, actor1, actor2, actor3)

if __name__ == '__main__':
    asyncio.run(main())
```

We should probably allow the user to initialize the `microgrid_api` itself, in case they need to do some extra config using the `micrgrid_api` before calling the `run` method, so at the end we can also provide smaller, more specific functions, that the user can use, and a very simple `run()` that should cover 90% of the use cases.

### Use cases

We have a few scenarios where not making the actors auto-start could lead to better patterns, in particular actors could own their own channels without the risk of circular dependencies, or initialization order issues.

For the auto-start dependencies issue, we are just creating all channels outside of the actors first, and then passing the channels to the actors.

### Additional context

_No response_

### LINKED ISSUE #240: Create a BackgroundService class
### What's needed?

Right now we have no common approach on how to manage classes/objects that spawn tasks. The most core example are actors, but we also have a lot of other classes that spawn tasks (like the resampler, moving window, etc.).

It should be possible to clearly control the lifespan of these objects, and be able to cleanly and clearly start and stop the tasks they spawn.

For now every class is doing their own thing, sometimes there is no way to stop things, sometimes there is but it is private, etc.

We should have a common way to do this.

### Proposed solution

Add a `BackgroundService` class to the SDK that provides a simple framework to handle this.

Here is a quick sketch of a possible implementation:

```py

class BackgroundService(ABC):

    def __init__(self):
        self.tasks: set[asyncio.Task] = set

    # async might not be really needed here, but allows sub-classes
    # to do async work as part of starting a task, which can be a
    # good thing since async initializaiton work can't be done in
    # __init__()
    @abstract
    async def start(self) -> None:
        pass

    @property
    def is_stopped(self) -> bool:
        for it in self.tasks:
            if t.done():
                return False
        return True
    def cancel(self) -> None:
        for t in self.tasks:
            t.cancel()

    async def join(self) -> None:
        await asyncio.gather(*[t.join() for t in self.tasks])

    async def stop(self) -> None:
        for t in self.tasks:
            await cancel_and_await(t)

    async __aenter__(self) -> None:
        await self.start()

    async def __aexit__(self, exc_type, exc, tb) -> None:
        await self.stop()

    def __del__(self) -> None:
        self.cancel()
        # Maybe see if there is a loop running and try to run await self.stop() in the loop?
```

(`join()` and `stop()` should probably be implemented using `asyncio.wait()` in a more resilient way)

Then all other classes can inherit from `BackgroundService`:

```py

class MovingWindow(BackgroundService):
    def __init__(self, other_params: ..., ...):
        super().__init__()

    async def start(self) -> None:
        self.tasks.add(asyncio.create_task(...))

mw = MovingWindow(...)
await mw.start()
...
await mw.stop()
```

### Use cases

* `Actor`
* `MovingWindow`
* `Resampler`
* etc.

_No response_

### Additional context

# Advantages

* Having a common, well-known interface (consistency, ease of use, less cognitive burden for users to know what's going on)
* It allows full control over the lifespan of a worker
* It separates initialization from running the task, which could help avoid dependency problems if 2 *workers* need each other to run.
* It is explicit in terms of when something is running in the background or not.

# Disadvantages

* If people forget to call `start()` it might take a while to figure out what's going on.
* 

# Class name

I first thought of `Worker`, but it's not a great name, as it is usually used for something that runs in a thread and can run a task, but I could not think of a better name.

I thought of `Task` but it might be confusing given `asyncio.Task`, I thought of `BackgroundTask` as it is something that it is supposed to run in the background, opposed to a general `Task` that can also be some calculation for which you want some result, but want to run several calculations in parallel.

I thought of `Daemon` as the functionality is similar to a unix daemon. Or maybe `Service` in the same sense.  To make it more explicit I ended up with `BackgroundService`, as it will only be used to sub-class it shouldn't be too annoying to have a long name.

# References

* Coming originally from this PR discussion: https://github.com/frequenz-floss/frequenz-sdk-python/pull/190#discussion_r1100080476
* Discussed in: https://github.com/frequenz-floss/frequenz-sdk-python/discussions/197

llucax:
`BackgroundService` is a new abstract base class can be used to write other classes that runs one or more tasks in the background. It provides a consistent API to start and stop these services and also takes care of the handling of the background tasks. It can also work as an `async` context manager, giving the service a deterministic lifetime and guaranteed cleanup.

The new `Actor` class brings quite a few new improvements over the old `@actor` decorator. These are the main differences:

* It doesn't start automatically, `start()` needs to be called to start an actor.
* The method to implement the main logic was renamed from `run()` to `_run()`, as it is not intended to be run externally.
* Actors can have an optional `name` (useful for debugging/logging purposes) and `loop` (if the actor should run in a loop different from the currently running loop).
* The actor will only be restarted if an unhandled `Exception` is raised by `_run()`. It will not be restarted if the `_run()` method finishes normally. If an unhandled `BaseException` is raised instead, it will be re-raised. For normal cancellation the `_run()` method should handle `asyncio.CancelledError` if the cancellation shouldn't be propagated (this is the same as with the decorator).
* The `_stop()` method is public (`stop()`) and will `cancel()` and `await` for the task to finish, catching the `asyncio.CancelledError`.
* The `join()` method is renamed to `wait()`, but they can also be awaited directly ( `await actor`).
* For deterministic cleanup, actors can now be used as `async` context managers.

The base actors (`ConfigManagingActor`, `ComponentMetricsResamplingActor`, `DataSourcingActor`, `PowerDistributingActor`) now inherit from the new `Actor` class, as well as the `MovingWindow`.

Fixes #240, fixes #45, fixes #196.


llucax:
This is a WIP because it is lacking tests and release notes (and updating the rest of the actors). It would be good to have some early feedback while I update the rest of the actors.

src/frequenz/sdk/actor/_background_service.py
```diff
@@ -0,0 +1,214 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
```
Marenz: 2023

src/frequenz/sdk/actor/_background_service.py
```diff
@@ -0,0 +1,214 @@
...
+
+        Args:
+            exc_type: The type of the exception raised, if any.
+            exc_val: The exception raised, if any.
+            exc_tb: The traceback of the exception raised, if any.
+        """
+        await self.stop()
+
+    def __await__(self) -> collections.abc.Generator[None, None, None]:
+        """Await this background service.
```
Marenz, line 223: as in, all tasks that still running? Maybe add this info to the docu

Marenz:
In your commit message, `insted` -> `instead`, `raise` -> `raised` (printed bold in the quoted text)

> The main differences with the `@actor` decorator are:
> [...]
>* The actor will only be restarted if an unhandled `Exception` is **raise**
  by `_run()`. It will not be restarted if the `_run()` method finishes
  normally. If an unhandled `BaseException` is raised **insted**, it will be

src/frequenz/sdk/actor/_actor.py
```diff
@@ -0,0 +1,212 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
```
Marenz: 2023

src/frequenz/sdk/actor/_actor.py
```diff
@@ -0,0 +1,212 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+class Actor(BackgroundService, abc.ABC):
+    """A primitive unit of computation that runs autonomously.
+
+    From [Wikipedia](https://en.wikipedia.org/wiki/Actor_model):
+
+    > An actor as the basic building block of concurrent computation. In response to
```
Marenz: `as` -> `is`? You quoted it a bit in the middle, here `as` no longer makes sense ;)

llucax: Yeah, good catch. To keep the quote as a quote I changed it to `From Wikipedia an actor is: > [...] the basic building block....`.

src/frequenz/sdk/actor/_background_service.py
llucax: Not sure what's unclear, this implements `await backgroud_service`. I just added a clarification that will wait until all the tasks are finished, hope that makes it more clear.

llucax:
Updated with the suggestions by @Marenz

Marenz:
Other than those small things, design wise I didn't notice any problems with the code. I would leave a more detailed review once the draft part is over.

llucax:
OK, now this has all actor classes migrated. I'm pretty happy about the results.

Next step are:

- [x] Remove the actor decorator
- [x] Write tests
- [x] Write release notes
- [x] Update PR description

And in a follow-up PR:

- Migrate other classes to use `BackgroundService` (like `MovingWindow`, `BatteryPoolStatus`, and whatever uses some background tasks and has `run`/`stop`/`join` methods.

src/frequenz/sdk/actor/_background_service.py
```diff
@@ -0,0 +1,307 @@
...
+                name: str | None = None,
+                loop: asyncio.AbstractEventLoop | None = None,
+            ) -> None:
+                super().__init__(name=name, loop=loop)
+                self._resolution_s = resolution_s
+
+            async def start(self) -> None:
+                # Note that you should always use the self._loop attribute access loop
+                # functions instead of asyncio directly.
+                self._tasks.add(self.loop.create_task(self._tick()))
```
shsms: the comment says "use self._loop", but the code uses self.loop

src/frequenz/sdk/actor/_background_service.py
```diff
@@ -0,0 +1,307 @@
...
+        Args:
+            name: The name of this background service. If `None`, `str(id(self))` will
+                be used. This is used mostly for debugging purposes.
+            loop: The event loop this background service will run on. If `None`, the
+                current running event loop will be used.
+        """
+        self._name: str = str(id(self)) if name is None else name
+        self._loop: asyncio.AbstractEventLoop = (
+            loop if loop else asyncio.get_running_loop()
+        )
```
shsms: `loop or asyncio.get_running_loop()` would have been so much cooler.  But I guess the loop parameter is going away for now, until we are clear on why it is needed, as discussed.

src/frequenz/sdk/actor/_background_service.py
llucax: Damn `pylint` didn't catch this... Thanks!

Marenz:
coonflicts

llucax:
> coonflicts

I guess nothing too radical to resolve. I won't update the PR until the review is done, as it is a big PR and it could be very annoying to have updates while reviewing.

tests/actor/test_resampling.py
```diff
@@ -122,39 +122,37 @@ async def test_single_request(
...
+        )
+
+        await _assert_resampling_works(
+            channel_registry,
+            fake_time,
+            resampling_chan_name=subs_req.get_channel_name(),
+            data_source_chan_name=data_source_req.get_channel_name(),
+        )
+
+        await resampling_actor._resampler.stop()  # pylint: disable=protected-access
```
shsms, line 155: probably not relevant to the PR, but just to understand, is it not possible to make the resampler stop, when the actor stops as well?

Maybe we need a cleanup method in the `BackgroundService` or `Actor` base class that gets called before the `stop` of the background service,  that actors can override if they have additional cleanups?  Or maybe the actor is able to get the background service to manage the task of the resampler as well.

src/frequenz/sdk/actor/_background_service.py
```diff
@@ -0,0 +1,307 @@
...
+        self._tasks: set[asyncio.Task[Any]] = set()
```
shsms: I wonder if `self._name` and `self._tasks` can be **also** initialized in the `start` method,  so that people don't have to call `super.__init__()` from actor init methods,  even if they don't want to set a name.

(oops, wanted to add to review, but clicked wrong button, because reviewing in multiple tabs)

src/frequenz/sdk/actor/_background_service.py
```diff
@@ -0,0 +1,307 @@
...
+    ) -> None:
+        """Initialize this BackgroundService.
+
+        Args:
+            name: The name of this background service. If `None`, `str(id(self))` will
+                be used. This is used mostly for debugging purposes.
+            loop: The event loop this background service will run on. If `None`, the
+                current running event loop will be used.
+        """
+        self._name: str = str(id(self)) if name is None else name
```
shsms, line 79: can we get it to say have the name of the class of `self`, followed by the id,  to make it a bit more readable?

src/frequenz/sdk/actor/_background_service.py
llucax: I don't understand this. You mean initialize it if it wasn't set in the constructor?

In any case I think it will make things more complicated, for example the `run()` function emits some logs that actors will be started before they are started, so in that case they won't have the names set. Also we'll need to make `_name` and `_tasks` optional (`type | None`, which will force us to litter the implementation with `assert _ not None`). I think having to write the `super().__init__()` is not that bad, also some of the linters (I don't remember if it is `mypy` or `pylint`) catches the issue it if you forget.

tests/actor/test_resampling.py
llucax: Yes, we can override the `stop()` method itself, I'm not sure why I didn't do it for the resampler, I think I did it for some other actor or background service (probably the moving window).

I will update!

benchmarks/timeseries/periodic_feature_extractor.py
```diff
@@ -27,19 +29,23 @@
...
+        await lm_chan.new_sender().send(
+            Sample(datetime.now(tz=timezone.utc), Quantity(0))
+        )
 
-    # Initialize the PeriodicFeatureExtractor class with a period of period seconds.
-    # This works since the sampling period is set to 1 second.
-    return PeriodicFeatureExtractor(moving_window, timedelta(seconds=period))
+        # Initialize the PeriodicFeatureExtractor class with a period of period seconds.
+        # This works since the sampling period is set to 1 second.
+        yield PeriodicFeatureExtractor(moving_window, timedelta(seconds=period))
```
shsms, line 48: looks like this usage of `asynccontextmanager` will mean this function can't be used for creating two simultaneous feature extractor instances, for example,  because when a second call is made,  the first one and its corresponding `MovingWindow` will get cleaned up?

So it has to used only with a `async with` right?

Because I can easily imagine people wanting to create multiple PeriodicFeatureExtractor instances for different data streams, etc. because they might want multiple features.

tests/actor/test_resampling.py
llucax: Not so simple: This is related to the comment in the resampling actor's `_run()` cleanup (`finally`):

```py
            # XXX: Here we should probably do a:  pylint: disable=fixme
            # await self._resampler.stop()
            # But since the actor will be restarted, the internal state would
            # be broken if we stop the resampler.
            #
            # We have an even bigger problem with this naive restarting
            # approach, as restarting this actor without really resetting its
            # state would be mostly the same as not really leaving the run()
            # method and just swallow any exception, which doesn't look super
            # smart.
```

This is why I'm planning to move the restart logic from the `Actor` to the `run()` method. Since the actor is monitoring himself, there is one task that is special, as we can't kill it because is the task monitoring the `_run()` method and restarting it. The self-auto-restarting makes everything more complicated and it is not very resilient,as the actor's internal state is not reset, so if it was left in an inconsistent state because of some bug, re-starting the `_run()` won't fix it. Actually restarting it in `run()` fixes part of the problem but not this, maybe we need some explicit `restart()` method that cleans up the internal state.

So I would leave this for future PRs, I also didn't converted more classes to background services (the resampler and its helper classes should be also background services) because they are not breaking changes, and I wanted to prioritize breaking changes for 1.0.

benchmarks/timeseries/periodic_feature_extractor.py
shsms: I wonder if it is possible to just make PeriodicFeatureExtractor tie its lifetime with that of its corresponding moving window.  But I guess for that it would need full ownership of the MovingWindow,  which we are not giving.  

People might want to keep using the same MovingWindow instance after or independent of the life time of the PeriodicFeatureExtractor.

shsms (commented):
big PR, but still it was very easy to review because of the well organized commits!

src/frequenz/sdk/actor/_background_service.py
llucax: This is what `str` (and `repr`) do. I think having the class name always in `str` and `repr` is a good thing, so it will be duplicate if I add it to the name too.

benchmarks/timeseries/periodic_feature_extractor.py
llucax: Not sure if I'm not understanding the issue here but please note that this is just the benchmark, I used `asynccontextmanager` because it was convenient in this context, but users don't have to do the same. The periodic feature extractor itself is not a context manager.

llucax:
> big PR, but still it was very easy to review because of the well organized commits!

Yeah, I think it was a **huge** win in terms of reviewing dropping the requirement of every commit being a working piece and focusing more on splitting commits to make changes very isolated. Very glad we tried it out!

src/frequenz/sdk/actor/_background_service.py
shsms: > You mean initialize it if it wasn't set in the constructor?

yup,  but if pylint, etc is able to catch a missing `super().__init__()` that is just as good.  Is there a test for this case already?

llucax: You mean a test that `pylint` catches it?

benchmarks/timeseries/periodic_feature_extractor.py
shsms: it was just about what impact the use of `asynccontextmanager` would have if the decorated function has an `async with` inside it, when the decorated function is called multiple times.

Anyway, just wanted to make sure we don't give people the wrong idea about how this can be used,  if people try to read the benchmark script and try to take some inspiration from it.

But it can be examined later and resolved if it is a problem.

src/frequenz/sdk/actor/_background_service.py
shsms: yup

shsms: Oh,  but pylint doesn't complain if there are unused `disable=` in the code.  Unlike mypy which complains if there are unused `type: ignore`.  So if the feature is provided by pylint, then maybe it is hard to test.

benchmarks/timeseries/periodic_feature_extractor.py
llucax: Oh, ok, if the context manager is used more than once, then it is executed twice and will create a new periodic feature extractor. I think everything should work as expected

src/frequenz/sdk/actor/_background_service.py
llucax: Yeah, I'm not sure how I would be able to test if pylint keeps catching this 😕

llucax: BTW, it is pylint that catches this: https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/super-init-not-called.html

llucax:
Updated:

* Rebased on the current head (I force pushed the rebase separately, so it is easier to see the updates separately from the [rebase diff](https://github.com/frequenz-floss/frequenz-sdk-python/compare/f0bc7ba9a311778c5a985fa67cf490e119183514..59eb30c838bafae769acb093ba8be90c37adaaa2)).
* Removed the `loop` argument from both `BackgroundService` and `Actor`
* Removed the *complete* example from `BackgroundService` and the simple example now forwards the `name` argument
* Same with the README, now only one example is provided for the migration.
* Merged the commit [Add names to MovingWindow tasks](https://github.com/frequenz-floss/frequenz-sdk-python/pull/564/commits/ebebdccc92b092a6f183119a4f6380314b301c32) into [Make MovingWindow a BackgroundService](https://github.com/frequenz-floss/frequenz-sdk-python/pull/564/commits/98eecef28ca8b436553e38201a3d0492d31f93b9)
* Squashed `fixup!` commits
* Other minor fixes
* [Changes diff](https://github.com/frequenz-floss/frequenz-sdk-python/compare/59eb30c838bafae769acb093ba8be90c37adaaa2..ace41e1f9dd4d904e17c80e0843039740d8c36c4)

llucax:
Enabling auto-merge

shsms (approved):
Awesome, everybody get ready to refactor everything now!

shsms:
> Enabling auto-merge

I think you also have to click a button,  looks like you just typed the message. :D

llucax:
It seems like GitHub copilot is not there yet...

PR #565, opened 2023-08-04
Improvements to the `ConfigManagingActor`

llucax:
While working on the migration to the new `Actor` class I noticed several small issues with the `ConfigManagingActor`:

- Use built-in types for typing
- Improve documentation for `ConfigManagingActor`
- Improve logging for `ConfigManagingActor`
- Add missing type hints to attributes
- Rename `conf_file` to `config_path` and accept `pathlib.Path`
- Add clarification about watching the parent directory
- Use match to process the file watcher events

### DIFF
--- a/src/frequenz/sdk/actor/_config_managing.py
+++ b/src/frequenz/sdk/actor/_config_managing.py
@@ -7,7 +7,7 @@ import logging
 import os
 import tomllib
 from collections import abc
-from typing import Any, Dict
+from typing import Any
 
 from frequenz.channels import Sender
 from frequenz.channels.util import FileWatcher
@@ -49,7 +49,7 @@ class ConfigManagingActor:
         )
         self._output = output
 
-    def _read_config(self) -> Dict[str, Any]:
+    def _read_config(self) -> dict[str, Any]:
         """Read the contents of the config file.
 
         Raises:
--- a/src/frequenz/sdk/actor/_config_managing.py
+++ b/src/frequenz/sdk/actor/_config_managing.py
@@ -64,7 +64,7 @@ class ConfigManagingActor:
             with open(self._conf_file, "rb") as toml_file:
                 return tomllib.load(toml_file)
         except ValueError as err:
-            logging.error("Can't read config file, err: %s", err)
+            logging.error("%s: Can't read config file, err: %s", self, err)
             raise
 
     async def send_config(self) -> None:
@@ -81,9 +81,8 @@ class ConfigManagingActor:
             if event.type != FileWatcher.EventType.DELETE:
                 if str(event.path) == self._conf_file:
                     _logger.info(
-                        "Update configs, because file %s was modified.",
+                        "%s: Update configs, because file %s was modified.",
+                        self,
                         self._conf_file,
                     )
                     await self.send_config()
-
-        _logger.debug("ConfigManager stopped.")
--- a/src/frequenz/sdk/actor/_config_managing.py
+++ b/src/frequenz/sdk/actor/_config_managing.py
@@ -49,6 +49,9 @@ class ConfigManagingActor:
             if isinstance(config_path, pathlib.Path)
             else pathlib.Path(config_path)
         )
+        # FileWatcher can't watch for non-existing files, so we need to watch for the
+        # parent directory instead just in case a configuration file doesn't exist yet
+        # or it is deleted and recreated again.
         self._file_watcher: FileWatcher = FileWatcher(
             paths=[self._config_path.parent], event_types=event_types
         )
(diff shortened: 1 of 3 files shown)


src/frequenz/sdk/actor/_config_managing.py
```diff
@@ -50,13 +52,13 @@ def __init__(
         self._output = output
 
     def _read_config(self) -> dict[str, Any]:
-        """Read the contents of the config file.
-
-        Raises:
-            ValueError: if config file cannot be read.
+        """Read the contents of the configuration file.
```
daniel-zullo-frequenz, line 61: Maybe `Read the contents of the TOML configuration file.` instead?

src/frequenz/sdk/actor/_config_managing.py
```diff
@@ -64,7 +64,7 @@ def _read_config(self) -> dict[str, Any]:
             with open(self._conf_file, "rb") as toml_file:
                 return tomllib.load(toml_file)
         except ValueError as err:
-            logging.error("Can't read config file, err: %s", err)
+            logging.error("%s: Can't read config file, err: %s", self, err)
```
daniel-zullo-frequenz, line 73: would it make sense to log the exception instead? I mean `logging.exception(...)`

tests/actor/test_config_manager.py
```diff
@@ -80,9 +80,7 @@ async def test_update(self, config_file: pathlib.Path) -> None:
         config_channel: Broadcast[Config] = Broadcast(
             "Config Channel", resend_latest=True
         )
-        _config_manager = ConfigManagingActor(
-            conf_file=str(config_file), output=config_channel.new_sender()
-        )
+        _config_manager = ConfigManagingActor(config_file, config_channel.new_sender())
```
daniel-zullo-frequenz, line 83: the parameter names were deleted, `config_file`  and `output`. Was this deliberately done? Just checking...

src/frequenz/sdk/actor/_config_managing.py
```diff
@@ -84,13 +84,31 @@ async def run(self) -> None:
...
                         self,
                         self._config_path,
                     )
                     await self.send_config()
+                case FileWatcher.EventType.DELETE:
+                    _logger.info(
+                        "%s: The configuration file %s was deleted, ignoring...",
+                        self,
+                        self._config_path,
+                    )
```
daniel-zullo-frequenz, line 112: The deletion case is still under discussion, right?

daniel-zullo-frequenz (commented):
A few minor comments/questions to check for, LGTM otherwise

src/frequenz/sdk/actor/_config_managing.py
llucax: I don't think so, we just want to know that a toml field value is invalid, having the whole stack trace doesn't help much and adds quite a bit of noise, right?

tests/actor/test_config_manager.py
llucax: Yes, I was hoping it passed unnoticed, damn you Dani! 😁

In a following commit I'm changing one of keyword arguments and since the original order in which I did the changes don't match the current commits it was more work to revise and do this in steps, so I thought it was fine to just remove the keywords from the call site to avoid breakage in the following commit. I even thought about making the arguments positional-only (as it is pretty clear what each argument is from the type alone) to avoid future breakages in general if we want to rename keyword args.

src/frequenz/sdk/actor/_config_managing.py
llucax: Yes, this is keeping the original behaviour.

src/frequenz/sdk/actor/_config_managing.py
llucax: I could but is it really important? This is an internal method anyway, I don't think it adds much clarity for a developer reading the code.

tests/actor/test_config_manager.py
llucax: Oh, I need to add a release note because this is a breaking change.

llucax: Done!

llucax:
Added release notes, enabled auto-merge.

src/frequenz/sdk/actor/_config_managing.py
daniel-zullo-frequenz: No, it isn't important. I was reading the doc in the public method and noticed the difference with the doc here and just mentioned it. I'll resolve this conversation

src/frequenz/sdk/actor/_config_managing.py
daniel-zullo-frequenz: Ok. Actually the exception will be logged as this method raises an error that only the actor will capture and log the exception before getting restarted. So we don't need to do it here

tests/actor/test_config_manager.py
daniel-zullo-frequenz: ok, then we only need an entry into the release notes here

daniel-zullo-frequenz: It was already added :)

daniel-zullo-frequenz (approved):

PR #566, opened 2023-08-04
Allow running release notes check in merge queues

llucax:
The check will do nothing for merge queues, as we need the PR information to be able to check anything. But in any case, checking again is not necessary because PRs should be queued for merge only when they already pass the release notes check.


llucax:
Force merging so it hopefully fixes the merge queue and we can merge again :grimacing:

PR #571, opened 2023-08-09
Add docs cross-linking to APIs

llucax:
For APIs now we can use cross-referencing, which is more user-friendly but also easier to maintain as we only need to update the version we want to point to in one place (the `mkdocs.yml` file).

- mkdocs: Add cross-linking object inventory for APIs
- Change links to the source with links to the docs


llucax:
Enabling auto-merge.

Marenz (approved):

PR #574, opened 2023-08-09
Clear release notes

llucax:

PR #578, opened 2023-08-10
Fix ring buffer gap cleanup code

matthias-wende-frequenz:
When cleaning up gaps the case that a gaps start date can become older then the oldest stored value in the buffer wasn't covered. Furthermore the cases for cleaning up a single gap didn't run for the last gaps, since the loop was stopping to early.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -16,3 +16,7 @@ Now the microgrid API v0.15.x is being used. The SDK can only connect to microgr
 - The `ConfigManagingActor` constructor now can accept a `pathlib.Path` as `config_path` too (before it accepted only a `str`).
 
 - The `PowerDistributingActor` now considers exclusion bounds, when finding an optimal distribution for power between batteries.
+
+## Bug Fixes
+
+- Fixes a bug in the ring buffer updating the end timestamp of gaps when they are outdated.
--- a/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
+++ b/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
@@ -328,17 +328,24 @@ class OrderedRingBuffer(Generic[FloatArray]):
         self._gaps = sorted(self._gaps, key=lambda x: x.start.timestamp())
 
         i = 0
-        while i < len(self._gaps) - 1:
+        while i < len(self._gaps):
             w_1 = self._gaps[i]
-            w_2 = self._gaps[i + 1]
+            if i < len(self._gaps) - 1:
+                w_2 = self._gaps[i + 1]
+            else:
+                w_2 = None
+
             # Delete out-of-date gaps
             if w_1.end <= self._datetime_oldest:
                 del self._gaps[i]
+            # Update start of gap if it's rolled out of the buffer
+            elif w_1.start < self._datetime_oldest:
+                self._gaps[i].start = self._datetime_oldest
             # If w2 is a subset of w1 we can delete it
-            elif w_1.start <= w_2.start and w_1.end >= w_2.end:
+            elif w_2 and w_1.start <= w_2.start and w_1.end >= w_2.end:
                 del self._gaps[i + 1]
             # If the gaps are direct neighbors, merge them
-            elif w_1.end >= w_2.start:
+            elif w_2 and w_1.end >= w_2.start:
                 w_1.end = w_2.end
                 del self._gaps[i + 1]
             else:
--- a/tests/timeseries/test_ringbuffer.py
+++ b/tests/timeseries/test_ringbuffer.py
@@ -369,3 +369,24 @@ def test_off_by_one_gap_logic_bug() -> None:
 
     assert buffer.is_missing(times[0]) is False
     assert buffer.is_missing(times[1]) is False
+
+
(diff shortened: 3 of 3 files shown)


src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -328,17 +328,24 @@ def _cleanup_gaps(self) -> None:
         self._gaps = sorted(self._gaps, key=lambda x: x.start.timestamp())
 
         i = 0
-        while i < len(self._gaps) - 1:
+        while i < len(self._gaps):
```
Marenz, line 331: Def should have a test for this and the other new case

Marenz:
Your commit message claims

> When cleaning up gaps the case that a gaps end date can become older
then the oldest stored value in the buffer wasn't covered. 

But I don't think it's true. It's covered right [here](https://github.com/frequenz-floss/frequenz-sdk-python/pull/578/commits/2c396b2a95f3866344b98628f90cc6bee24c684d#diff-c33ddb1d8130e18c9629f82d1f7baef27997e49c9248e1d4c770be068e938c7fR339):

```py
            # Delete out-of-date gaps
            if w_1.end <= self._datetime_oldest:
                del self._gaps[i]
```

Marenz:
About having gaps whos start time is older than the oldest.. I am not sure that's really important, but I guess it doesn't hurt to clamp it

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
matthias-wende-frequenz: There is a case for the new case at least

matthias-wende-frequenz: Added another test case for the "old" bug.

Marenz:
Also, in your commit message you're wrongly using "then" when it should be "than" :)

matthias-wende-frequenz:
> Your commit message claims
> 
> > When cleaning up gaps the case that a gaps end date can become older
> > then the oldest stored value in the buffer wasn't covered.
> 
> But I don't think it's true. It's covered right [here](https://github.com/frequenz-floss/frequenz-sdk-python/pull/578/commits/2c396b2a95f3866344b98628f90cc6bee24c684d#diff-c33ddb1d8130e18c9629f82d1f7baef27997e49c9248e1d4c770be068e938c7fR339):
> 
> ```python
>             # Delete out-of-date gaps
>             if w_1.end <= self._datetime_oldest:
>                 del self._gaps[i]
> ```

No, that case deletes the whole gap once it's fully outside of the buffer, i.e. it looks only at the end of the gap. The new case looks at and updates the beginning of the gap.

matthias-wende-frequenz:
> About having gaps whos start time is older than the oldest.. I am not sure that's really important, but I guess it doesn't hurt to clamp it

We need that in the moving window. But maybe it's not a bug, but more like we covering an unforeseen use case. We want to know the number of missing items in the window and now we can simply sum up the length of all gaps.

Marenz:
I am sorry, but I am confused. The commit message talks literally about the **end** date being older than oldest timestamp

> When cleaning up gaps the case that **_a gaps end date_** can become older
then the oldest stored value in the buffer wasn't covered.

And this matches also what you say my case is..

> No, that case deletes the whole gap once it's fully outside of the buffer, **_i.e. it looks only at the end of the gap._** 

This though

> The new case looks at and updates the beginning of the gap.

is true, however the message should then say "Clean up gap starts if they are before the oldest timestamp" or so

Marenz:
PR LGTM, just the first commit message greatly confuses

matthias-wende-frequenz:
> I am sorry, but I am confused. The commit message talks literally about the **end** date being older than oldest timestamp
> 
> > When cleaning up gaps the case that **_a gaps end date_** can become older
> > then the oldest stored value in the buffer wasn't covered.
> 
> And this matches also what you say my case is..
> 
> > No, that case deletes the whole gap once it's fully outside of the buffer, **_i.e. it looks only at the end of the gap._**
> 
> This though
> 
> > The new case looks at and updates the beginning of the gap.
> 
> is true, however the message should then say "Clean up gap starts if they are before the oldest timestamp" or so

Did I ever mention that I'm having troubles with signs :D. Indeed the the start time of the gap is meant. I'll update the message accordingly.
To be fair. The code tells the truth :).

Marenz (approved):

RELEASE_NOTES.md
```diff
@@ -16,3 +16,7 @@ Now the microgrid API v0.15.x is being used. The SDK can only connect to microgr
 - The `ConfigManagingActor` constructor now can accept a `pathlib.Path` as `config_path` too (before it accepted only a `str`).
 
 - The `PowerDistributingActor` now considers exclusion bounds, when finding an optimal distribution for power between batteries.
+
+## Bug Fixes
+
+- Fixes a bug in the ring buffer updating the end timestamp of gaps when they are outdated.
```
llucax, line 22: You are a bad bad boy. You should have first approved and merged #574.

Marenz: Wops

PR #579, opened 2023-08-10
Add microgrid.frequency()

### LINKED ISSUE #341: Interface for fetching grid metrics: `frequency`
### What's needed?

A high level interface for fetching grid frequency.

### Proposed solution

This can be done in two steps:

1. A method `microgrid.grid.frequency()` that takes an optional `component_id` of a meter,  from which it would stream frequency values. This value does not need to be resampled (so no need to get it from the resampler actor).  If no `component_id` is specified, use the meter closest to the grid, in the component graph.  If there are multiple meters close to the grid,  pick any one that has no successors.
2. Once the microgrid API starts supporting metadata for components,  a query interface around the component graph,  to identify meters with the best resolution, and get their component IDs,  which can be used in the methods from step 1, instead of having to specify component IDs in config files.

### Use cases

_No response_

_No response_

Marenz:
fixes https://github.com/frequenz-floss/frequenz-sdk-python/issues/341

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -16,6 +16,8 @@
 
   Note that a microgrid is allowed to have zero or one grid connection point. Microgrids configured as islands will have zero grid connection points, and microgrids configured as grid-connected will have one grid connection point.
 
+- A new method `microgrid.frequeny()` was added to allow easy access to the current frequency of the grid.
+
 - A new class `Fuse` has been added to represent fuses. This class has a member variable `max_current` which represents the maximum current that can course through the fuse. If the current flowing through a fuse is greater than this limit, then the fuse will break the circuit.
 
 
--- a/src/frequenz/sdk/timeseries/_grid_frequency.py
+++ b/src/frequenz/sdk/timeseries/_grid_frequency.py
@@ -1,0 +1,166 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Fetches the Grid Frequency."""
+
+from __future__ import annotations
+
+import asyncio
+import logging
+from typing import TYPE_CHECKING
+
+from frequenz.channels import Receiver, Sender
+
+from ..actor import ChannelRegistry
+from ..microgrid import connection_manager
+from ..microgrid.component import Component, ComponentCategory, ComponentMetricId
+from ..timeseries._base_types import Sample
+from ..timeseries._quantities import Frequency
+
+if TYPE_CHECKING:
+    # Imported here to avoid a circular import.
+    from ..actor import ComponentMetricRequest
+
+
+def create_request(component_id: int) -> ComponentMetricRequest:
+    """Create a request for grid frequency.
+
+    Args:
+        component_id: The component id to use for the request.
+
+    Returns:
+        A component metric request for grid frequency.
+    """
+    # Imported here to avoid a circular import.
+    # pylint: disable=import-outside-toplevel
+    from ..actor import ComponentMetricRequest
+
+    return ComponentMetricRequest(
+        "grid-frequency", component_id, ComponentMetricId.FREQUENCY, None
+    )
+
+
+class GridFrequency:
+    """Grid Frequency."""
+
+    def __init__(
+        self,
+        data_sourcing_request_sender: Sender[ComponentMetricRequest],
+        channel_registry: ChannelRegistry,
+        component: Component,
+    ):
+        """Initialize the grid frequency formula generator.
+
+        Args:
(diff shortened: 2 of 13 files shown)


src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_grid_frequency_formula.py
```diff
@@ -0,0 +1,59 @@
...
+            )
+
+        grid_successors = component_graph.successors(grid_component.component_id)
+
+        for comp in grid_successors:
+            if comp.category != ComponentCategory.METER:
+                continue
+
+            builder.push_component_metric(comp.component_id, nones_are_zeros=False)
+            break
```
daniel-zullo-frequenz: It's likely this `break` is the reason why the test get stuck while waiting for values

Marenz:
fixes #341

RELEASE_NOTES.md
```diff
@@ -10,7 +10,8 @@
 
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+- A new method `microgrid.frequeny()` was added to allow easy access to the
+  current frequency of the grid.
```
llucax: This goes into GH Release as the description, and for some reason the markdown rendered there is silly about new lines, it will render them as `</br>`.

```suggestion
- A new method `microgrid.frequeny()` was added to allow easy access to the current frequency of the grid.
```

src/frequenz/sdk/timeseries/_grid_frequency.py
```diff
@@ -0,0 +1,157 @@
...
+        receiver = self._channel_registry.new_receiver(
+            self._component_metric_request.get_channel_name()
+        )
+
+        if not self._task:
+            self._task = asyncio.create_task(self._send_request())
+        else:
+            logging.info("Grid frequency request already sent: %s", self._component)
+
+        return receiver
```
llucax, line 92: Just to put more accessible something that I said/asked somewhere else, this is hiding the `async` nature of creating a new receiver. Well, the receiver is created immediately, but it won't be receiving anything until the component metric request is fulfilled, so even if you have a receiver, some samples could be missed at first.

It was mentioned that for testing this is problematic because we need a sync point, because we want to verify we receive what we sent in tests. This is related to https://github.com/frequenz-floss/frequenz-channels-python/pull/154.

One solution would be to create a special receiver here that can be awaited to be ready, but I think that's a bit overkill for testing, same as adding the feature to channels. The other solution would be to add more mocking/instrumentation to tests, or to make tests resilient to the fact that a few first messages could be missed. To be honest, I think the later is the best option for tests, because this is the actual behavior of this method and we should test that.

src/frequenz/sdk/timeseries/_grid_frequency.py
```diff
@@ -0,0 +1,157 @@
...
+
+    async def _send_request(self) -> None:
+        """Send the request for grid frequency."""
+        logging.info("Sending request for grid frequency: %s", self._component)
+        await self._request_sender.send(self._component_metric_request)
+        logging.info("Sent request for grid frequency: %s", self._component)
+
+    @staticmethod
+    def find_frequency_component() -> Component:
+        """Find the component that will be used for grid frequency.
```
llucax, line 102: Maybe it would be good to explain the algorithm here? What does it try to find first? What happens if that doesn't exist,etc.?

tests/timeseries/test_logical_meter.py
```diff
@@ -111,6 +116,147 @@ async def test_grid_power_2(
...
+        mockgrid.add_batteries(2)
+        mockgrid.add_solar_inverters(1)
+        await mockgrid.start(mocker)
+
+        grid_freq = microgrid.frequency()
+        grid_freq_recv = grid_freq.new_receiver()
+
+        assert grid_freq._task is not None
+        # We have to wait for the metric request to be sent
+        await grid_freq._task
```
llucax, line 131: But I think this hack for syncing is also perfectly fine, you are only accessing the implementation of what you are testing.

Marenz:
rebased on newest content, explained the search better, fixed release notes

llucax (commented):
Great thanks. I forgot to mention the eternal "full stops in commit titles" :grimacing:

llucax:
Still full stops in commits, and now the CI is failing too :(

Marenz:
Depends on #603

Marenz:
> Still full stops in commits, and now the CI is failing too :(

First you break it then you complain about it :P

Marenz:
This is now ready

llucax (dismissed):
Needs rebasing, but LGTM.

llucax (approved):

PR #584, opened 2023-08-15
Support PV configurations with no or only some meters

Marenz:
- Put frequently used code in base function.
- Properly handle PV power requests when no or only some meters exist.
  So far we only had configurations like this: Meter -> Inverter -> PV. However
the scenario with Inverter -> PV is also possible and now handled correctly.

### DIFF
--- a/src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_formula_generator.py
+++ b/src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_formula_generator.py
@@ -15,6 +15,7 @@ from typing import Callable, Generic
 from frequenz.channels import Sender
 
 from ....actor import ChannelRegistry, ComponentMetricRequest
+from ....microgrid import component, connection_manager
 from ....microgrid.component import ComponentMetricId
 from ..._quantities import QuantityT
 from .._formula_engine import FormulaEngine, FormulaEngine3Phase
@@ -99,6 +100,36 @@ class FormulaGenerator(ABC, Generic[QuantityT]):
             create_method,
         )
         return builder
+
+    def _get_grid_component_successors(self) -> set[component.Component]:
+        """Get the set of grid component successors in the component graph.
+
+        Returns:
+            A set of grid component successors.
+
+        Raises:
+            ComponentNotFound: If the grid component is not found in the component graph.
+            ComponentNotFound: If no successor components are found in the component graph.
+        """
+        component_graph = connection_manager.get().component_graph
+        grid_component = next(
+            iter(
+                component_graph.components(
+                    component_category={component.ComponentCategory.GRID}
+                )
+            ),
+            None,
+        )
+
+        if grid_component is None:
+            raise ComponentNotFound("Grid component not found in the component graph.")
+
+        grid_successors = component_graph.successors(grid_component.component_id)
+
+        if not grid_successors:
+            raise ComponentNotFound("No components found in the component graph.")
+
+        return grid_successors
 
     @abstractmethod
     def generate(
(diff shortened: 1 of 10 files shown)


src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_formula_generator.py
```diff
@@ -100,6 +101,38 @@ def _get_builder(
         )
         return builder
 
+    def _get_graph_and_grid_component_successors(
+        self,
+    ) -> tuple[ComponentGraph, set[component.Component]]:
```
llucax: Getting the component graph is only one line of code, wouldn't it make more sense to have this function instead?

```suggestion
    def _get_grid_component_successors(
        self,
        component_graph: ComponentGraph,
    ) -> set[component.Component]:
```

I find the name and return type too convoluted otherwise.

llucax (commented):
It might be good to give this PR and the commit changing the logic a bit more context, it is hard to figure out what's going on and why the change is needed otherwise, at least from someone not very involved in PV stuff.

Also small nitpick, the commit style in the SDK is **not** to end commit title with a `.`.

I reviewed the code but not the business logic, as I don't fully understand what the requirements were, so I'll leave that to @matthias-wende-frequenz

llucax:
Also please remember to set either an issue that this is fixing, or select the project and milestone.

Marenz:
> set either an issue that this is fixing, 

It's a virtual issue :sweat_smile: it exists in the Mat(t)hias' heads

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_pv_power_formula.py
```diff
@@ -71,39 +65,35 @@ def generate(self) -> FormulaEngine[Power]:
...
+        (
+            component_graph,
+            grid_successors,
+        ) = self._get_graph_and_grid_component_successors()
+
+        if len(grid_successors) == 1:
+            successor = next(iter(grid_successors))
+            if successor.category not in (
+                ComponentCategory.METER,
+                ComponentCategory.INVERTER,
```
matthias-wende-frequenz: If there is only one component behind the grid component, that has to be a meter! Otherwise the configuration is screwed. An inverter doesn't has an successor. Thus in that case we are having an inverter directly connected to a grid which implies the whole microgrid is a single inverter. I think we don't have to support such an edge cases in formulas.

matthias-wende-frequenz: So as consequence, please remove the check for `ComponentCategory.INVERTER`

Marenz:
Updated based on comments

llucax:
> It's a virtual issue 😅 it exists in the Mat(t)hias' heads

That's fine, but then we need a better description for the PR :)

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_formula_generator.py
llucax: Not sure why didn't you take the component graph as an argument, you could also make it optional and get it inside the function if `None` was provided. But I guess it is OK like this too, at the end it shouldn't be expensive to get the component graph anyway.

Marenz:
> That's fine, but then we need a better description for the PR :)

Really? I always considered the code, release notes and commit message more important than the PR description.. 
But I updated the PR text as well now

matthias-wende-frequenz (approved):

llucax:
> Really? I always considered the code, release notes and commit message more important than the PR description..

Yeah, a better description for the commit is also good, I make it in general, but in this case that there is no issue for me it is more difficult to review if I have to dig deep in some commit to know what's going on.

IMHO in general the PR description should have a higher level explanation, and the commit something more technical about the implementation.

But again, if there is an issue with the details then there is no need to repeat everything in the PR.

Also now PR descriptions become a merge commit, and part of the repo history, si we should have that into consideration.

> But I updated the PR text as well now

Thanks!

shsms:
I agree with Luca about the descriptions.  It is probably not that big of a deal when everyone already has enough context,  but when people are explicitly asking for more context, while reviewing, it should be provided.  they shouldn't have to ask twice or have to justify the request.

PR #587, opened 2023-08-17
Upgrade to repo-config v0.5.2

llucax:
All changes here are derived from regenerating files using the templates in repo-config v0.5.2.


llucax:
Not sure what's going on with the tests, locally it passes and if I remove the `# type: ignore` locally, it fails :angry:

Marenz:
test is failing

llucax:
OK, after a long investigation it seems like the problem with `mypy` is caching.

This failed in the CI but worked locally because it seems like the first uncached run is the one failing, but if you run `mypy` again it works and sometimes, if you run it again it will fail again and alternate. I found a similar issue that was fixes, so I guess this is a persistent problem (not sure if it is worth reporting unless we spend some time to reduce it to a small test case): https://github.com/python/mypy/issues/2960

Using the `--no-incremental` option (which avoids caching) the issue is fixed, so I added that as a temporary fix in a new commit (the new repo-config will probably use this option by default).

Marenz (approved):

matthias-wende-frequenz (approved):

PR #589, opened 2023-08-18
Fix `consumer_power()` not working certain configurations.

Marenz:
- Fix MockMicrogrid bug not properly disabling main meter
- Fix `consumer_power()` not working certain configurations.
    In microgrids without consumers and no main meter, the formula
    would never return any values.
    This was undiscovered as we didn't test that scenario in the MockMicrogrid class used
    for testing.

### DIFF
--- a/benchmarks/timeseries/benchmark_datasourcing.py
+++ b/benchmarks/timeseries/benchmark_datasourcing.py
@@ -70,7 +70,7 @@ async def benchmark_data_sourcing(
         num_ev_chargers * len(COMPONENT_METRIC_IDS) * num_msgs_per_battery
     )
     mock_grid = MockMicrogrid(
-        grid_side_meter=False, num_values=num_msgs_per_battery, sample_rate_s=0.0
+        grid_meter=False, num_values=num_msgs_per_battery, sample_rate_s=0.0
     )
 
     mock_grid.add_ev_chargers(num_ev_chargers)
--- a/tests/timeseries/mock_microgrid.py
+++ b/tests/timeseries/mock_microgrid.py
@@ -40,7 +40,7 @@ class MockMicrogrid:  # pylint: disable=too-many-instance-attributes
     """Setup a MockApi instance with multiple component layouts for tests."""
 
     grid_id = 1
-    main_meter_id = 4
+    _grid_meter_id = 4
 
     chp_id_suffix = 5
     evc_id_suffix = 6
@@ -53,7 +53,7 @@ class MockMicrogrid:  # pylint: disable=too-many-instance-attributes
 
     def __init__(  # pylint: disable=too-many-arguments
         self,
-        grid_side_meter: bool,
+        grid_meter: bool,
         api_client_streaming: bool = False,
         num_values: int = 2000,
         sample_rate_s: float = 0.01,
@@ -62,7 +62,7 @@ class MockMicrogrid:  # pylint: disable=too-many-instance-attributes
         """Create a new instance.
 
         Args:
-            grid_side_meter: whether the main meter should be on the grid side or not.
+            grid_meter: whether there is a meter successor of the GRID component.
             api_client_streaming: whether the mock client should be configured to stream
                 raw data from the API client.
             num_values: number of values to generate for each component.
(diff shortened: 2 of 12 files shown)


benchmarks/timeseries/benchmark_datasourcing.py
```diff
@@ -70,7 +70,7 @@ async def benchmark_data_sourcing(
         num_ev_chargers * len(COMPONENT_METRIC_IDS) * num_msgs_per_battery
     )
     mock_grid = MockMicrogrid(
-        grid_side_meter=False, num_values=num_msgs_per_battery, sample_rate_s=0.0
+        main_meter=False, num_values=num_msgs_per_battery, sample_rate_s=0.0
```
matthias-wende-frequenz: I suggest to call it `grid_meter` as it's the successor to the `GRID` component. The term `main_meter` isn't helpful w.r.t. the meter placement.

tests/timeseries/mock_microgrid.py
```diff
@@ -62,7 +62,7 @@ def __init__(  # pylint: disable=too-many-arguments
         """Create a new instance.
 
         Args:
-            grid_side_meter: whether the main meter should be on the grid side or not.
+            main_meter: whether there is a main meter that is between grid and microgrid.
```
matthias-wende-frequenz: whether there is a meter successor of the `GRID` component.

tests/timeseries/test_logical_meter.py
```diff
@@ -88,7 +88,7 @@ async def test_grid_power_2(
         results: list[Quantity] = []
         meter_sums: list[Quantity] = []
         for count in range(10):
-            await mockgrid.mock_resampler.send_meter_power([20.0 + count, 12.0, -13.0])
+            await mockgrid.mock_resampler.send_meter_power([20.0 + count, 12.0])
```
matthias-wende-frequenz: shouldn't the first value in the list send to the non existing meter?

tests/timeseries/test_logical_meter.py
```diff
@@ -62,7 +62,7 @@ async def test_grid_power_2(
         mocker: MockerFixture,
     ) -> None:
         """Test the grid power formula without a grid side meter."""
-        mockgrid = MockMicrogrid(grid_side_meter=False)
+        mockgrid = MockMicrogrid(main_meter=False)
```
matthias-wende-frequenz: Although one meter is removed, this test still works, since we compare the `logical_meter`s power against the sum of all the other meters.

tests/timeseries/test_logical_meter.py
```diff
@@ -136,6 +136,31 @@ async def test_grid_production_consumption_power_consumer_meter(
...
+        mockgrid.add_batteries(2)
+        mockgrid.add_solar_inverters(1)
+        await mockgrid.start(mocker)
+
+        logical_meter = microgrid.logical_meter()
+        grid_recv = logical_meter.grid_power.new_receiver()
+        grid_production_recv = logical_meter.grid_production_power.new_receiver()
+        grid_consumption_recv = logical_meter.grid_consumption_power.new_receiver()
+
+        await mockgrid.mock_resampler.send_meter_power([2.5, 3.5, 4.0])
```
matthias-wende-frequenz, line 155: A Solar inverter that charges the PV panels? Interesting case :joy:.

Marenz: It's the same power sum as the test above...

tests/timeseries/test_logical_meter.py
Marenz: it doesn't really matter which values are used here as we simply sum them up and compare

benchmarks/timeseries/benchmark_datasourcing.py
shsms: if you're adding a `add_consumer_meters`, then you might as well add a `add_grid_meters` method and drop the flag.  There has already been some talk about locations with multiple grid meters anyway.

Marenz: that' s the next step actually

Marenz:
updated

matthias-wende-frequenz (dismissed):

benchmarks/timeseries/benchmark_datasourcing.py
shsms: I feel like this is not a good idea,  because this makes our tests rely on the implementation rather than on just the expected behaviour.  I think instead we should always add a main meter,  and have special cases for graphs without main meters.  Maybe we can discuss tomorrow during the meeting.

shsms: I have added it to the agenda.

shsms (changes requested):
I have some comments: https://github.com/frequenz-floss/frequenz-sdk-python/pull/589#discussion_r1300214661

blocking until we discuss tomorrow.

tests/timeseries/test_logical_meter.py
matthias-wende-frequenz: After discussing with Sahas, we should add a consumption meter here, because we are not testing the same case any longer.

Marenz: fixed

shsms (approved):
It turns out my big objections were incorrect, I guess we just have to write more granular test cases in the future, so unblocking.

matthias-wende-frequenz (approved):

PR #591, opened 2023-08-21
Forward zero power requests always to the microgrid API

shsms:
Only non-zero requests should be rejected if they fall within the
exclusion bounds.

This PR also:
- moves distribution algorithm into the power distributing actor's dir.
- remove unused methods
- rename `OutOfBound` -> `OutOfBounds`
- Improve power distribution algorithm docstrings.


tests/actor/power_distributing/test_power_distributing.py
```diff
@@ -155,6 +155,103 @@ async def test_power_distributor_one_user(self, mocker: MockerFixture) -> None:
         assert result.excess_power == approx(200.0)
         assert result.request == request
 
+    async def test_power_distributor_exclusion_bounds(
+        self, mocker: MockerFixture
+    ) -> None:
+        """Test if power distributing actor rejects non-zero requests  in exclusion bounds."""
```
Marenz: double whitespace `requests  in`

shsms: fixed.

llucax:
I guess this is not addressing the unclear documentation issues in the power distributor that we talked about today?

llucax (commented):
LGTM, but it looks weird to have a special case for 0 and send it even if it is out of bounds. I wonder if we shouldn't have another special value (like `None`) that means "I don't care about these batteries anymore" instead of zero. I guess you already discussed it and this makes sense and setting power to zero means that already, but mentioning it just in case.

On a different topic, I was thinking maybe we should go in the opposite direction, having `frequenz.sdk.power.DistributingActor`, as currently the `frequenz.sdk.actor` package is a bag of cats. But I think that's for post 1.0, so in the current context I think the moving of the distribution algorithm makes sense.

shsms:
> but it looks weird to have a special case for 0 and send it even if it is out of bounds.

most of the time there won't be any exclusion bounds, and people send zero when they want to stop power.  When there are exclusion bounds, when they send 0, it does get set on the device, but then other algorithms would kick in, if necessary, on the microgrid service side, that could change the power.  So maybe I was not being accurate when I said earlier that the meaning of 0 has changed.

llucax:
> most of the time there won't be any exclusion bounds, and people send zero when they want to stop power. When there are exclusion bounds, when they send 0, it does get set on the device, but then other algorithms would kick in, if necessary, on the microgrid service side, that could change the power. So maybe I was not being accurate when I said earlier that the meaning of 0 has changed.

I still think it is confusing as an interface. When there are exclusion bounds (let's not analyze the case where there aren't because then we agree all is fine :D), as an user it is confusing, if I want to stop using a battery but the bounds say I can't set power to 0, I still have to send a `set_power(0)`. I understand that for now doing that was the way to *release* a battery, but with exclusion bounds IMHO is not clear enough, maybe someone will just send the lowest power accepted by the bounds.

Wouldn't it be better to have an explicit way to say "release this battery(pool)"? Either `set_power(None)` or maybe even a different method, like `battery_pool.release()` or something. Then the implementation can send `0` even if it is out of bounds, that's fine.

@matthias-wende-frequenz @thomas-nicolai-frequenz opinions?

thomas-nicolai-frequenz:
@llucax I agree. Just one note:

> if I want to stop using a battery but the bounds say I can't set power to 0, I still have to send a set_power(0)

If the bounds are set to 0 they would still allow 0 to be send right? Maybe I'm misunderstanding. If min is 0 and max is 0, 0 would still feel like a valid value. 

> battery_pool.release()

Agreed. I would just vote for a different method name. 

@tiyash-basu-frequenz any additional thoughts?

tiyash-basu-frequenz:
> any additional thoughts?

None at the moment. I agree that having a specific method for releasing components makes the interface clearer. That would also simplify things when multiple actors run in an app. Actor A, instead of asking for 0W power, could just release it, therefore, at an interface level, not interfering with an Actor B's request of, say 1000W charge.

shsms:
I feel like we should take this in and deal with the whole topic in the power manager, because that would be the top level interface.

If it looks like we won't be able to get the power manager in time (I think we will make it),  we can come back here and add a function later.

llucax:
OK, I created:

* https://github.com/frequenz-floss/frequenz-sdk-python/issues/600

Assigned it to v1.0 because if we change the behavior it will be technically a breaking change...

llucax (dismissed):

shsms:
rebased and resolved conflicts.

llucax (approved):

PR #595, opened 2023-08-22
Add depth first search for components 

### LINKED ISSUE #593: DFS component graph search
### What's needed?

We need a better approach for finding components in nested graphs. Currently, when creating certain formulas, namely pv and producer power, we only look at direct successors of either the grid component or the grid meter.
Furthermore we're only supporting graphs with a single grid meter.

### Proposed solution

A search algorithm that goes into the depth of the graph and stops when it finds the desired component.

### Use cases

_No response_

Nested component graphs and graphs with more the one main meter will be supported.

### Additional context

_No response_

matthias-wende-frequenz:
In this PR we add a dfs component search to the component graph. The search continues to go into depth until the condition is fulfilled. The condition has to be provided as a lambda.

Furthermore we add function that returns the grid component such that it can be used in the dfs search, whenever the search should begin at the `GRID` component.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -10,7 +10,7 @@
 
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+- DFS for compentent graph
 
 ## Bug Fixes
 
--- a/src/frequenz/sdk/microgrid/_graph.py
+++ b/src/frequenz/sdk/microgrid/_graph.py
@@ -266,6 +266,29 @@ class ComponentGraph(ABC):
             Whether the specified component is part of a CHP chain.
         """
 
+    @abstractmethod
+    def dfs(
+        self,
+        current_node: Component,
+        visited: Set[Component],
+        condition: Callable[[Component], bool],
+    ) -> Set[Component]:
+        """
+        Search for components that fulfill the condition in the Graph.
+
+        DFS is used for searching the graph. The graph travarsal is stopped
+        once a component fulfills the condition.
+
+        Args:
+            current_node: The current node to search from.
+            visited: The set of visited nodes.
+            condition: The condition function to check for.
+
+        Returns:
+            A set of component ids where the coresponding components fulfill
+            the condition function.
+        """
+
 
 class _MicrogridComponentGraph(ComponentGraph):
     """ComponentGraph implementation designed to work with the microgrid API.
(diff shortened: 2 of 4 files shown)


src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_formula_generator.py
```diff
@@ -120,10 +120,22 @@ def _get_grid_component_successors(self) -> set[component.Component]:
             ),
             None,
         )
-
         if grid_component is None:
             raise ComponentNotFound("Grid component not found in the component graph.")
 
+        return grid_component
+
+    def _get_grid_component_successors(self) -> set[component.Component]:
```
shsms, line 128: maybe just `_get_grid_successors`?

src/frequenz/sdk/microgrid/_graph.py
```diff
@@ -266,6 +266,29 @@ def is_chp_chain(self, component: Component) -> bool:
             Whether the specified component is part of a CHP chain.
         """
 
+    @abstractmethod
+    def dfs_components_with_condition(
```
shsms: maybe `filter` or `filter_by` is a better name?  and in the docstring you can say it is done with a dfs.

shsms: But I guess that doesn't show that we are looking for components and not selections.  And maybe `filter` isn't really graph terminology.  So `dfs_components_by`?

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_formula_generator.py
matthias-wende-frequenz: I don't mind. But that change is not part of this PR. It's just a weird diff.

src/frequenz/sdk/microgrid/_graph.py
matthias-wende-frequenz: Changed it to simply dfs. There is nothing special about this dfs implementation that makes it deserve a special name.

src/frequenz/sdk/timeseries/_formula_engine/_formula_generators/_formula_generator.py
matthias-wende-frequenz: And I wouldn't like to touch it in this PR either, because I'd assume that it involves changing more code. If we want that, let's do it in a separate PR.

RELEASE_NOTES.md
```diff
@@ -10,7 +10,7 @@
 
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+- DFS search for compentent graph
```
shsms: redundant `search`?

shsms (dismissed):
just an optional comment, lgtm otherwise.

RELEASE_NOTES.md
matthias-wende-frequenz: YES

shsms (approved):

src/frequenz/sdk/microgrid/_graph.py
```diff
@@ -688,6 +711,42 @@ def is_chp_chain(self, component: Component) -> bool:
         """
         return self.is_chp(component) or self.is_chp_meter(component)
 
+    def dfs(
```
llucax, line 714: Sorry for coming late to the party, but wouldn't it make more sense to call this just `search()`? Or `depth_first_search()` if it is very important to note that uses DFS. The name is just too cryptic IMHO if you are not familiar with graphs, I would also make `current_node` and `visited` optional, default them to the root node and an empty set when they are `None`.

Or maybe we can keep `dfs()` as it is and add another higher-level method `search` that does what I say above...

PR #597, opened 2023-08-23
[ci] Add test for installation in multiple architectures

### LINKED ISSUE #160: ci: Check that `pip install frequenz-sdk` works in an arm64 environment
### What's needed?

A CI check to ensure that we are able to install the sdk both in amd64 (which we already have) and arm64 environments.

This became a problem with `frequenz-sdk==0.16.0`, which installs fine in amd64 `python:3.11-slim` docker image,  but because of a dependency `grpcio==1.51.1`, requires a C++ compiler when installing in an arm64 image.

Having a test allows us to be aware of such issues and look for alternatives,  or at least provide documentation on how to deal with the issue.

### Proposed solution

_No response_

### Use cases

_No response_

_No response_

### Additional context

_No response_

tiyash-basu-frequenz:
Users should be able to install the frequenz-sdk package in amd64 and arm64 systems. This commit adds a simple test to the CI to validate this.

The aforementioned test builds a multi-platform docker image with the generated wheel package. If the image gets successfully generated, then the test passes.

Additionally the updated docs are published only if this test is passed.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -146,7 +146,7 @@ jobs:
 
   publish-docs:
     name: Publish documentation website to GitHub pages
-    needs: ["nox", "build"]
+    needs: ["nox", "test-installation"]
     if: github.event_name == 'push'
     runs-on: ubuntu-20.04
     permissions:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -83,6 +83,31 @@ jobs:
           path: dist/
           if-no-files-found: error
 
+  test-installation:
+    name: Test package installation in different architectures
+    needs: ["build"]
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Fetch sources
+        uses: actions/checkout@v3
+      - name: Download package
+        uses: actions/download-artifact@v3
+        with:
+          name: dist-packages
+          path: dist
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      - name: Set up docker-buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Test Installation
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          file: .github/containers/test-installation/Dockerfile
+          platforms: linux/amd64,linux/arm64
+          tags: localhost/test-installation
+          push: false
+
   test-docs:
     name: Test documentation website generation
     if: github.event_name != 'push'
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -83,6 +83,31 @@ jobs:
...
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      - name: Set up docker-buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Test Installation
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          file: .github/containers/test-installation/Dockerfile
+          platforms: linux/amd64,linux/arm64
```
llucax, line 107: I guess an alternative to this would be to make this job a matrix. I guess the downside is we need to do the setup many times and the advantage is we parallelize the testing of different archs, right? Did you go for this solution for some particular reason or just because it is what we use somewhere else?

llucax (commented):
We were also a bit concerned about adding too much time to the CI with tests that usually don't break (like the related issues of testing in windows/mac/arm64), the idea was to only run them in the merge queue, or before releases. This test adds *only* 1 minute to the CI run, so I guess it is fine, so this is mostly a FYI, I think we can merge as is and change/improve in the future.

.github/workflows/ci.yaml
tiyash-basu-frequenz: >  I guess the downside is we need to do the setup many times and the advantage is we parallelize the testing of different archs

Correct. That's the trade-off. Given the amount of time this step takes is still less than the `nox` step, this still won't be a bottleneck.

But lemme know if you want this to be parallelized. I don't have a strong opinion about this.

tiyash-basu-frequenz:
> We were also a bit concerned about adding too much time to the CI with tests that usually don't break (like the related issues of testing in windows/mac/arm64), the idea was to only run them in the merge queue, or before releases. This test adds _only_ 1 minute to the CI run, so I guess it is fine, so this is mostly a FYI, I think we can merge as is and change/improve in the future.

I feel there is value in running this test for PRs in general. This test does not take long, is easier to fix at an early stage, prevents releases getting blocked, and is potentially time-consuming to fix when there are a lot of commits/PRs that could have caused this.

.github/workflows/ci.yaml
llucax: Let's leave it like this, I was mainly curious if you had consider it. Thanks!

llucax (approved):

.github/workflows/ci.yaml
tiyash-basu-frequenz: I did. But I chose the other option for now :)

PR #603, opened 2023-08-25
DataPipeline: Fix resampling/ds/power actors not started

Marenz:


Marenz:
no release notes as we haven't released the bug yet that this fixes

src/frequenz/sdk/microgrid/_data_pipeline.py
```diff
@@ -237,6 +239,7 @@ def _data_sourcing_request_sender(self) -> Sender[ComponentMetricRequest]:
                 registry=self._channel_registry,
             )
             self._data_sourcing_actor = _ActorInfo(actor, channel)
+            asyncio.create_task(self._data_sourcing_actor.actor.start())
```
llucax: This shouldn't be needed, as it is started in `_start()`.

src/frequenz/sdk/microgrid/_data_pipeline.py
```diff
@@ -262,6 +265,7 @@ def _resampling_request_sender(self) -> Sender[ComponentMetricRequest]:
                 config=self._resampler_config,
             )
             self._resampling_actor = _ActorInfo(actor, channel)
+            asyncio.create_task(self._resampling_actor.actor.start())
```
llucax: This shouldn't be needed, as it is started in `_start()`.

llucax (commented):
I guess this is OK. I'm not crazy about this async-non-awaitable initialization (like what we had with `new_receiver()`) but I guess it is not the time to revisit that now. Thanks for looking into it!

llucax:
I wonder if it is worth not starting the power distributing actor eagerly, it will make things easier and it only means having one task idling I guess. But again, not for this PR, just thinking out loud.

src/frequenz/sdk/microgrid/_data_pipeline.py
Marenz: No, this is the whole point of this PR, it's not started in `_start()` if you call `_start()` _before_ you call any function that uses `_data_sourcing_request_sender()` as the actor has not been created at that point.

So if you don't call `logical_meter()` or any similar method that indirectly uses this, it will never be started.

Marenz:
> I wonder if it is worth not starting the power distributing actor eagerly,

But we aren't starting it eagerly? We only start it once the `battery_pool()`  was requested

llucax:
> But we aren't starting it eagerly? We only start it once the `battery_pool()` was requested

The power distributing actor, not the resampler and data sourcing actor.

llucax:
The data pipeline should be a background service too, so whoever wants to create one, even if it is for testing, knows that it has to be started too.

src/frequenz/sdk/microgrid/_data_pipeline.py
llucax: Oh, the thing here is you are talking about a case where `microgrid.initialize()` isn't being called, right? That's the case in the new test at least.

llucax:
For the records, we decided to decouple starting with making sure the instance is ready, so we are going to make `start()` not `async` (`def start() -> None: pass`) and add `async def ready() -> None: pass` to `BackgroundService`. This way users can decide if they need to wait until an instance is ready or not. The fix should be re-done after this is implemented. @Marenz is going to do it.

FYI @shsms @matthias-wende-frequenz

shsms:
But for this PR,  just a call to the sync `start` is enough right?  Because we can just call `start`, then block on the output channels, until we start seeing something.  

And how would `ready` know the actor has started?  I mean it should block until it is ready,  right?  Else you wouldn't make is async, if it was just returning a boolean immediately.  It shouldn't poll,  does it use a condition variable instead, for example?

> The fix should be re-done after this is implemented. @Marenz is going to do it.

do you mean we take this PR as it is,  and @Marenz updates BackgroundService/Actor classes separately?

Marenz:
> And how would ready know the actor has started? 

It's a NOP by default and deriving classes can override it and implement whatever they see fitting for waiting.

> do you mean we take this PR as it is, and @Marenz updates BackgroundService/Actor classes separately?

No, I implemented the fix here now

src/frequenz/sdk/microgrid/_data_pipeline.py
Marenz: No no, it's even independent of that. `microgrid.initialize()` just calls `_start()`. Start will start datasoucring and resampling actors if they are set at that point.

They are set only on demand when requested however. So if you only do `.logical_meter()` after `initialize()` you will have the same problem.

shsms:
In the second commit's message,  I guess you don't mean "never", but "not until they're needed"?

But LGTM otherwise.

I guess we need cleanup before the async loop exits,  but that can be done by converting _DataPipeline into a background service, separately like @llucax suggests.

Marenz:
> In the second commit's message, I guess you don't mean "never", but "not until they're needed"?

no, I mean "never", but the confusing part may be that I am describing the situation that the commit fixes, not the situation after the fix...

shsms:
Maybe it should say "was never started at all", etc. to be clear.

Marenz:
Updated wording

shsms (dismissed):

llucax:
> But for this PR, just a call to the sync `start` is enough right? Because we can just call `start`, then block on the output channels, until we start seeing something.

Yes.

I've been thinking about this and I think @Marenz was right saying that I was right about what I said about https://github.com/frequenz-floss/frequenz-channels-python/pull/154 and we shouldn't have `ready()` at all :laughing: 

I think if we start allowing these synchronization points we can even end up in a deadlock easily is some actor A needs to wait for B to be `ready()` but then `B.ready()`, needs to wait until A is `ready()` (this could be transitional with quite a few actors in the middle so very hard to see at first sight).

If we don't have any use case for it, it is probably better not to encourage it by making it part of the API of `BackgroundService`.

Sorry for the noise :-/

src/frequenz/sdk/actor/_run_utils.py
```diff
@@ -19,8 +19,12 @@ async def run(*actors: Actor) -> None:
         actors: the actors to be awaited.
     """
     _logger.info("Starting %s actor(s)...", len(actors))
+
+    for actor in actors:
+        actor.start()
```
llucax: The log message about actors starting should be moved here. I was thinking it might be nice to also have an `if actor.is_running` and print a different message if the actor was already started than if we just started it, maybe it could be helpful when debugging if one didn't realize that one actor was started before the `run()` or something.

src/frequenz/sdk/microgrid/_data_pipeline.py
```diff
@@ -262,14 +264,17 @@ def _resampling_request_sender(self) -> Sender[ComponentMetricRequest]:
...
     async def _start(self) -> None:
         """Start the data pipeline actors."""
         if self._data_sourcing_actor:
-            await self._data_sourcing_actor.actor.start()
+            await self._data_sourcing_actor.actor.ready()
         if self._resampling_actor:
-            await self._resampling_actor.actor.start()
+            await self._resampling_actor.actor.ready()
+        if self._power_distributing_actor:
+            await self._power_distributing_actor.ready()
```
llucax: I still don't see why do we need to have the `_start()` if we are starting all actors elsewhere anyway. I think we either remove `_start()` or we leave the eager starting in `_start()` and make users of this class call `_start()` before using it.

As I said, I would make the date pipeline a background service. If we only want a *temporary* fix in this PR, then I would only start in `_start()`.

With the power distributing actor is different because it start lazily only in the battery pool is used.

llucax (commented):
A few comments besides suggesting back to remove `ready()` and just make `start()` non-async.

Also you are missing some `await actor.start()` that also needs to be fixed, like in the release notes. I suggest using `git grep 'await.*\.start()'` to search for them.

llucax (changes requested):
Blocking the merge to avoid accidental merges because there are things to discuss before merging.

src/frequenz/sdk/microgrid/_data_pipeline.py
Marenz: I am fine with removing `_start()` completely

llucax: I'm fine with that too, so go ahead and make everything lazy

Marenz: lazy is my specialty

src/frequenz/sdk/microgrid/_data_pipeline.py
```diff
@@ -262,17 +264,9 @@ def _resampling_request_sender(self) -> Sender[ComponentMetricRequest]:
...
 
-    async def _start(self) -> None:
-        """Start the data pipeline actors."""
-        if self._data_sourcing_actor:
-            await self._data_sourcing_actor.actor.start()
-        if self._resampling_actor:
-            await self._resampling_actor.actor.start()
-        # The power distributing actor is started lazily when the first battery pool is
-        # created.
-
```
llucax, line 275: IMHO this change should go to the second commit (fixing the start in data pipeline) not the first one (make start sync).

src/frequenz/sdk/microgrid/_data_pipeline.py
```diff
@@ -300,7 +294,6 @@ async def initialize(resampler_config: ResamplerConfig) -> None:
     if _DATA_PIPELINE is not None:
         raise RuntimeError("DataPipeline is already initialized.")
     _DATA_PIPELINE = _DataPipeline(resampler_config)
-    await _DATA_PIPELINE._start()  # pylint: disable=protected-access
```
llucax, line 303: Same.

src/frequenz/sdk/actor/_run_utils.py
```diff
@@ -19,26 +19,15 @@ async def run(*actors: Actor) -> None:
...
-        "running",
-        "finished",
-    )
-
-    _logger.info("All %s actor(s) finished.", len(actors))
+    for actor in actors:
+        if actor.is_running:
+            _logger.info("Actor %s: already running.", actor)
+        else:
+            actor.start()
```
llucax, line 28: Optional, but looking at it now I like the symmetry of having both messages, if it was really started or not.

```suggestion
            _logger.info("Actor %s: Starting...", actor)
            actor.start()
```

src/frequenz/sdk/actor/_run_utils.py
```diff
@@ -19,26 +19,15 @@ async def run(*actors: Actor) -> None:
...
-    await _wait_tasks(
-        set(asyncio.create_task(a.wait(), name=str(a)) for a in actors),
-        "running",
-        "finished",
-    )
-
-    _logger.info("All %s actor(s) finished.", len(actors))
+    for actor in actors:
+        if actor.is_running:
+            _logger.info("Actor %s: already running.", actor)
```
llucax: Nitpick, but all other message start with caps after the `:`.

```suggestion
            _logger.info("Actor %s: Already running, skipping start.", actor)
```

llucax (dismissed):
I would say all comments are optional, so approving.

llucax (approved):

PR #606, opened 2023-08-25
Add module to easily create component graphs

Marenz:
This adds a module that allows the easy creation of component graphs.
Here is a simple example:

```python
gen = GraphGenerator()

# Pre-create a battery component to refer to it later or to use it in multiple
# places in the graph.
special_bat = gen.component(ComponentCategory.BATTERY)

graph = gen.to_graph(
    (
        ComponentCategory.METER, # grid side meter
        [ # list of components connected to grid side meter
            (
                ComponentCategory.METER, # Meter in front of battery->inverter
                ( # A tuple, first is connected to parent, second is the child of the first
                    # Inverter in front of battery, type will be
                    # set to InverterType.BATTERY
                    ComponentCategory.INVERTER,
                    ComponentCategory.BATTERY,  # Battery
                ),
            ),
            (
                ComponentCategory.METER,
                (
                    # Inverter in front of battery, type will be
                    # set to InverterType.BATTERY
                    ComponentCategory.INVERTER,
                    special_bat,  # Pre-created battery
                ),
            ),
        ],
    )
)
```

it's not actually used yet, but I plan to do that for the n:m battery:inverter
issue.

It also updates the MockMicrogrid to accept such graphs as parameter.
I suspect that part will receive some more udpates and fixes as I use it more.

### DIFF
--- a/tests/timeseries/mock_microgrid.py
+++ b/tests/timeseries/mock_microgrid.py
@@ -16,6 +16,7 @@ from frequenz.sdk import microgrid
 from frequenz.sdk._internal._asyncio import cancel_and_await
 from frequenz.sdk.actor import ResamplerConfig
 from frequenz.sdk.microgrid import _data_pipeline
+from frequenz.sdk.microgrid._graph import _MicrogridComponentGraph
 from frequenz.sdk.microgrid.client import Connection
 from frequenz.sdk.microgrid.component import (
     Component,
@@ -53,16 +54,17 @@ class MockMicrogrid:  # pylint: disable=too-many-instance-attributes
 
     def __init__(  # pylint: disable=too-many-arguments
         self,
-        grid_meter: bool,
+        grid_meter: bool | None = None,
         api_client_streaming: bool = False,
         num_values: int = 2000,
         sample_rate_s: float = 0.01,
         num_namespaces: int = 1,
+        graph: _MicrogridComponentGraph | None = None,
     ):
         """Create a new instance.
 
         Args:
-            grid_meter: whether there is a meter successor of the GRID component.
+            grid_meter: optional, whether there is a meter successor of the GRID component.
             api_client_streaming: whether the mock client should be configured to stream
                 raw data from the API client.
             num_values: number of values to generate for each component.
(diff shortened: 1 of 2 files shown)


llucax:
Just the other day while looking if networkx had DFS, I noticed it has [graph generation](https://networkx.org/documentation/stable/reference/generators.html). I guess what this does is pretty different, but I just wanted to point it out so everybody knows it exists. Maybe it could be useful for fuzzy testing or something (?)

Marenz:
Yeah, this one is more targeted to our SDK classes & types and just tries to make manual generation more fun

shsms:
Very lisp-like, also nice that both methods can be used to with MockMicrogrid now.  Can I ask for one more test case, with a list at the top-level, i.e without a grid meter?

tests/timeseries/mock_microgrid.py
```diff
@@ -71,28 +73,76 @@ def __init__(  # pylint: disable=too-many-arguments
...
                 battery_pool, etc) are used, and the same metric is used by formulas in
                 different namespaces.
+            graph: optional, a graph of components to use instead of the default grid
+                layout. If specified, grid_meter must be None.
         """
-        self._components: Set[Component] = set(
-            [
-                Component(1, ComponentCategory.GRID),
-            ]
+        if (grid_meter is not None or grid_meter is True) and graph is not None:
```
shsms: In `(grid_meter is not None or grid_meter is True)`, the second condition appears to be redundant.  
```suggestion
        if grid_meter is not None and graph is not None:
```

tests/utils/graph_generator.py
```diff
@@ -0,0 +1,320 @@
...
+        def inverter_type(category: ComponentCategory) -> InverterType | None:
+            match category:
+                case ComponentCategory.BATTERY:
+                    return InverterType.BATTERY
+                case ComponentCategory.CHP:
+                    return InverterType.SOLAR
+                case ComponentCategory.EV_CHARGER:
+                    return InverterType.HYBRID
+                case _:
+                    return None
```
shsms: CHP and EV Chargers don't have inverters in front of them.  And we don't have a PV component type,  so we can guess this only if there's a Battery as a successor.  Actually even then not 100%,  because it could be a HYBRID inverter (which handle batteries and PV in the same inverter).  But we can default to Battery inverter in this case.

And for the other types of inverters, we can say they need to created manually, with the `gen.component()` method that you're using.

```suggestion
        def inverter_type(successor_category: ComponentCategory) -> InverterType | None:
            match successor_category:
                case ComponentCategory.BATTERY:
                    return InverterType.BATTERY
                case _:
                    raise TypeError(f"Found an inverter in front of a {successor_category}")
```

Marenz: > And for the other types of inverters, we can say they need to created manually, with the gen.component() method that you're using.

Right, but it's also valid to have a `None` type for the inverter, right?

shsms: I think people can use `InverterType.NONE` instead.

tests/utils/graph_generator.py
```diff
@@ -0,0 +1,315 @@
...
+        Returns:
+            a tuple containing the components and connections of the graph.
+        """
+        component: Component
+
+        # Handle both Component and ComponentCategory types
+        if isinstance(components[0], Component):
+            component = components[0]
+        elif isinstance(components[0], ComponentCategory):
+            component = self.component(components[0])
```
shsms: optional:  destructure to `parent, children = components`, and then use those instead of `components[0]`, etc.

But I'm fine to take this in as it is as well,  it is not complicated.

shsms (dismissed):

shsms:
Removing from queue just to give you a chance to read the comments, but feel free to add it back.

shsms (approved):

PR #610, opened 2023-08-28
Initialize a `Grid` object with a given `Fuse` limit

### LINKED ISSUE #360: Expose grid connection limits
### What's needed?

We need to expose the new [microgrid grid connection rated fuse maximum current](https://github.com/frequenz-floss/frequenz-api-microgrid/blob/2beb6745b2e0be82f12e8ef171bec0ed1c51362a/proto/frequenz/api/microgrid/grid.proto#L25).

For now this limit can be static (queried from the microgrid API and kept immutable). A change in the grid connection limit will need a SDK restart. In the future we want to support dynamically changing this limit.

### Low-level

Add a new `Component` that wraps the `Grid` object and its metadata, as it is done with other microgrid components.

### High-level

1. Add an object to represent the `GridConnection` and a read-only rated fuse current property
2. Add a global instance exposed via `frequenz.sdk.microgrid.grid_connection`
3. Initialize the global instance in the `initialize()` function
   a. Until the new version of the microgrid API is deployed, just fill the instance `fuse.current.rate` with `None`

We should probably also add some utility methods for checking and adjusting a current value to the accepted limits. We could accept multiple current values so it is easy to use even with 3-phase samples.

```py
class Current:

    rate: float | None

    def is_within_bounds(self, *current: float) -> bool:
        ...

    @overload
    def adjust(self, /, current: float) -> float:
        ...

    @overload
    def adjust(self, /, current1: float, current2: float, *extra_currents: float) -> list[float]:
        ...

    def adjust(self, *current: float) -> float | list[float]:
        """Truncate any current above the rated_fuse_current."""

class Fuse:

    current: Current

class GridConnection:

    fuse: Fuse
```

### Use cases

Probably the `PowerManager`/`PowerArbitrator` (#161) or the should check these limits when applying charge/discharge commands, but others might as well, like the `EVChargerPool` might want to check the limit before increasing the bounds to allow vehicles to charge with high power.

Example usage:

```python
from frequenz.sdk import microgrid

desired_current = 18.5
grid_fuse = microgrid.grid_connection.fuse
if not grid_fuse.current.is_within_bounds(desired_current):
    rate = grid_fuse.current.rate
    _logger.warning("The desired current ({desired_current}) is out of bounds (the fuse is rated {rate}). Will be adjusted.")
good_current = grid_fuse.current.adjust(desired_current)
_logger.info("Using current: {good_current}")
```

_No response_

### Additional context

Design discussion:
* https://github.com/frequenz-floss/frequenz-sdk-python/discussions/322

Future work:

- [ ] Initialize with the proper microgrid API calls after the new microgrid API exposing the limit is deployed
- [ ] Support dynamically changing this limit. There are 2 main approaches for this:

  * Streaming interface: This means we need a streaming interface from the microgrid API and we should make the `rated_fuse_current` a channel that streams changes in this property, like the `battery_pool()`, etc.
  * Polling: We just add a task to keep the `fuse.current.rate` updated by polling the microgrid API (or uses a streaming gRPC call if available).

  It seems like for now the polling interface should be enough.

tiyash-basu-frequenz:
Calling `microgrid.initialize()` now also initializes the microgrid's grid connection point as a singleton object of a newly added type `Grid`. This object can be obtained by calling `microgrid.grid.get()`. This object exposes the max current that can course through the grid connection point, which is useful for the power distribution algorithm. The max current is provided by the Microgrid API, and can be obtained by calling `microgrid.grid.get().fuse.max_current`.

Note that a microgrid is allowed to have zero or one grid connection point. Microgrids configured as islands will have zero grid connection points, and microgrids configured as grid-connected will have one grid connection point.

A new class `Fuse` has been added to represent fuses. This class has a member variable `max_current` which represents the maximum current that can course through the fuse. If the current flowing through a fuse is greater than this limit, then the fuse will break the circuit.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -102,6 +102,10 @@ This release replaces the `@actor` decorator with a new `Actor` class.
 
 - `Actor`: This new class inherits from `BackgroundService` and it replaces the `@actor` decorator.
 
+- Calling `microgrid.initialize()` now also initializes the microgrid's grid connection point as a singleton object of a newly added type `Grid`. This object can be obtained by calling `microgrid.grid.get()`. This object exposes the max current that can course through the grid connection point, which is useful for the power distribution algorithm. The max current is provided by the Microgrid API, and can be obtained by calling `microgrid.grid.get().max_current`.
+
+  Note that a microgrid is allowed to have zero or one grid connection point. Microgrids configured as islands will have zero grid connection points, and microgrids configured as grid-connected will have one grid connection point.
+
 ## Bug Fixes
 
 - Fixes a bug in the ring buffer updating the end timestamp of gaps when they are outdated.
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -102,9 +102,12 @@ This release replaces the `@actor` decorator with a new `Actor` class.
 
 - `Actor`: This new class inherits from `BackgroundService` and it replaces the `@actor` decorator.
 
-- Calling `microgrid.initialize()` now also initializes the microgrid's grid connection point as a singleton object of a newly added type `Grid`. This object can be obtained by calling `microgrid.grid.get()`. This object exposes the max current that can course through the grid connection point, which is useful for the power distribution algorithm. The max current is provided by the Microgrid API, and can be obtained by calling `microgrid.grid.get().max_current`.
+- Calling `microgrid.initialize()` now also initializes the microgrid's grid connection point as a singleton object of a newly added type `Grid`. This object can be obtained by calling `microgrid.grid.get()`. This object exposes the max current that can course through the grid connection point, which is useful for the power distribution algorithm. The max current is provided by the Microgrid API, and can be obtained by calling `microgrid.grid.get().fuse.max_current`.
 
   Note that a microgrid is allowed to have zero or one grid connection point. Microgrids configured as islands will have zero grid connection points, and microgrids configured as grid-connected will have one grid connection point.
+
+- A new class `Fuse` has been added to represent fuses. This class has a member variable `max_current` which represents the maximum current that can course through the fuse. If the current flowing through a fuse is greater than this limit, then the fuse will break the circuit.
+
 
 ## Bug Fixes
 
--- a/src/frequenz/sdk/microgrid/component/_component.py
+++ b/src/frequenz/sdk/microgrid/component/_component.py
@@ -9,6 +9,7 @@ from dataclasses import dataclass
 from enum import Enum
 
 import frequenz.api.common.components_pb2 as components_pb
+import frequenz.api.microgrid.grid_pb2 as grid_pb
 import frequenz.api.microgrid.inverter_pb2 as inverter_pb
 
 
@@ -106,12 +107,36 @@ def _component_category_from_protobuf(
 
 
 @dataclass(frozen=True)
+class ComponentMetadata:
+    """Base class for component metadata classes."""
+
+
+@dataclass(frozen=True)
+class GridMetadata(ComponentMetadata):
+    """Metadata for a grid connection point."""
+
+    max_current: float
+    """maximum current rating of the grid connection point in amps."""
+
+
+def _component_metadata_from_protobuf(
+    component_category: components_pb.ComponentCategory.ValueType,
+    component_metadata: grid_pb.Metadata,
+) -> GridMetadata | None:
+    if component_category == components_pb.ComponentCategory.COMPONENT_CATEGORY_GRID:
+        return GridMetadata(float(component_metadata.rated_fuse_current))
+
+    return None
+
+
+@dataclass(frozen=True)
 class Component:
     """Metadata for a single microgrid component."""
 
     component_id: int
     category: ComponentCategory
     type: ComponentType | None = None
+    metadata: ComponentMetadata | None = None
 
     def is_valid(self) -> bool:
         """Check if this instance contains valid data.
--- a/src/frequenz/sdk/microgrid/component/_component.py
+++ b/src/frequenz/sdk/microgrid/component/_component.py
@@ -7,7 +7,6 @@ from __future__ import annotations
 
 from dataclasses import dataclass
 from enum import Enum
-from typing import Optional
 
 import frequenz.api.common.components_pb2 as components_pb
 import frequenz.api.microgrid.inverter_pb2 as inverter_pb
@@ -32,7 +31,7 @@ class InverterType(ComponentType):
 def _component_type_from_protobuf(
     component_category: components_pb.ComponentCategory.ValueType,
     component_metadata: inverter_pb.Metadata,
-) -> Optional[ComponentType]:
+) -> ComponentType | None:
     """Convert a protobuf InverterType message to Component enum.
 
     For internal-only use by the `microgrid` package.
@@ -112,7 +111,7 @@ class Component:
 
     component_id: int
     category: ComponentCategory
-    type: Optional[ComponentType] = None
+    type: ComponentType | None = None
 
     def is_valid(self) -> bool:
         """Check if this instance contains valid data.
--- a/src/frequenz/sdk/microgrid/component/_component.py
+++ b/src/frequenz/sdk/microgrid/component/_component.py
@@ -12,6 +12,9 @@ import frequenz.api.common.components_pb2 as components_pb
 import frequenz.api.microgrid.grid_pb2 as grid_pb
 import frequenz.api.microgrid.inverter_pb2 as inverter_pb
 
+from ...timeseries import Current
+from ..fuse import Fuse
+
 
 class ComponentType(Enum):
     """A base class from which individual component types are derived."""
@@ -110,13 +113,13 @@ def _component_category_from_protobuf(
 class ComponentMetadata:
     """Base class for component metadata classes."""
 
+    fuse: Fuse | None = None
+    """The fuse at the grid connection point."""
+
 
 @dataclass(frozen=True)
 class GridMetadata(ComponentMetadata):
     """Metadata for a grid connection point."""
-
-    max_current: float
-    """maximum current rating of the grid connection point in amps."""
 
 
 def _component_metadata_from_protobuf(
@@ -124,7 +127,9 @@ def _component_metadata_from_protobuf(
     component_metadata: grid_pb.Metadata,
 ) -> GridMetadata | None:
     if component_category == components_pb.ComponentCategory.COMPONENT_CATEGORY_GRID:
-        return GridMetadata(float(component_metadata.rated_fuse_current))
+        max_current = Current.from_amperes(component_metadata.rated_fuse_current)
+        fuse = Fuse(max_current)
+        return GridMetadata(fuse)
 
     return None
 
--- a/src/frequenz/sdk/microgrid/component/_component.py
+++ b/src/frequenz/sdk/microgrid/component/_component.py
@@ -125,6 +125,14 @@ class Component:
             self.component_id > 0 and any(t == self.category for t in ComponentCategory)
         ) or (self.component_id == 0 and self.category == ComponentCategory.GRID)
 
+    def __hash__(self) -> int:
+        """Compute a hash of this instance, obtained by hashing the `component_id` field.
+
+        Returns:
+            Hash of this instance.
+        """
+        return hash(self.component_id)
+
 
 class ComponentMetricId(Enum):
     """An enum representing the various metrics available in the microgrid."""
--- a/src/frequenz/sdk/microgrid/fuse.py
+++ b/src/frequenz/sdk/microgrid/fuse.py
@@ -1,0 +1,16 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Fuse data class."""
+
+from dataclasses import dataclass
+
+from ..timeseries import Current
+
+
+@dataclass(frozen=True)
+class Fuse:
+    """Fuse data class."""
+
+    max_current: Current
+    """Rated current of the fuse."""
--- a/src/frequenz/sdk/microgrid/grid.py
+++ b/src/frequenz/sdk/microgrid/grid.py
@@ -11,7 +11,6 @@ import logging
 from collections.abc import Iterable
 from dataclasses import dataclass
 
-from ..timeseries import Current
 from .component import Component
 from .component._component import ComponentCategory
 from .fuse import Fuse
@@ -36,7 +35,8 @@ def initialize(components: Iterable[Component]) -> None:
 
     Raises:
         RuntimeError: If there is more than 1 grid connection point in the
-            microgrid.
+            microgrid, or if the grid connection point is not initialized,
+            or if the grid connection point does not have a fuse.
     """
     global _GRID  # pylint: disable=global-statement
 
@@ -57,8 +57,14 @@ def initialize(components: Iterable[Component]) -> None:
             f"Expected at most one grid connection, got {grid_connections_count}"
         )
     else:
-        max_current = grid_connections[0].metadata.max_current  # type: ignore
-        fuse = Fuse(Current.from_amperes(max_current))
+        if grid_connections[0].metadata is None:
+            raise RuntimeError("Grid metadata is None")
+
+        fuse = grid_connections[0].metadata.fuse
+
+        if fuse is None:
+            raise RuntimeError("Grid fuse is None")
+
         _GRID = Grid(fuse)
 
 
--- a/src/frequenz/sdk/microgrid/grid.py
+++ b/src/frequenz/sdk/microgrid/grid.py
@@ -1,0 +1,71 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Grid connection point.
+
+This module provides the `Grid` type, which represents a grid connection point
+in a microgrid.
+"""
+
+import logging
+from collections.abc import Iterable
+from dataclasses import dataclass
+
+from .component import Component
+from .component._component import ComponentCategory
+
+
+@dataclass(frozen=True)
+class Grid:
+    """A grid connection point."""
+
+    max_current: float
+    """The maximum current that can course through the grid connection point, in Amperes."""
+
+
+_GRID: Grid | None = None
+
+
+def initialize(components: Iterable[Component]) -> None:
+    """Initialize the grid connection.
+
+    Args:
+        components: The components in the microgrid.
+
+    Raises:
+        RuntimeError: If there is more than 1 grid connection point in the
+            microgrid.
+    """
+    global _GRID  # pylint: disable=global-statement
+
+    grid_connections = list(
+        component
+        for component in components
+        if component.category == ComponentCategory.GRID
+    )
+
+    grid_connections_count = len(grid_connections)
+
+    if grid_connections_count == 0:
+        logging.info(
+            "No grid connection found for this microgrid. This is normal for an islanded microgrid."
+        )
+    elif grid_connections_count > 1:
+        raise RuntimeError(
+            f"Expected at most one grid connection, got {grid_connections_count}"
+        )
+    else:
+        max_current = grid_connections[0].metadata.max_current  # type: ignore
+        _GRID = Grid(max_current)
+
+
+def get() -> Grid | None:
+    """Get the grid connection.
+
+    Note that a microgrid configured as an island will not have a grid
+    connection point. For such microgrids, this function will return `None`.
+
+    Returns:
+        The grid connection.
+    """
+    return _GRID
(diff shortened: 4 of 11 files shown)


src/frequenz/sdk/microgrid/component/_component.py
```diff
@@ -106,13 +108,64 @@ def _component_category_from_protobuf(
...
 
 
+class ComponentMetadata:
+    """Base class for component metadata classes."""
+
+
+class GridMetadata(ComponentMetadata):
+    """Metadata for a grid connection point."""
+
+    def __init__(self, max_current: float) -> None:
```
tiyash-basu-frequenz: It'd be nice if this were a dataclass. But changing it to a dataclass turns `Component.metadata` into a dict, causing a few comparison tests to fail. Has anyone here encountered such an issue before?

daniel-zullo-frequenz: Have you tried with `@dataclass(frozen=True)`?

src/frequenz/sdk/microgrid/fuse.py
```diff
@@ -0,0 +1,19 @@
...
+
+from ..timeseries import Current
+
+
+@dataclass(frozen=True)
+class Fuse:
+    """Fuse data class.
+
+    Attributes:
+        max_current: Rated current of the fuse, in Amperes.
```
daniel-zullo-frequenz: As `max_current` is `Current` type we don't need to explicitly document the unit.
And we usually document attributes below the attribute definition as follows:

```python
@dataclass(frozen=True)
class Fuse:
    """Fuse data class."""

    max_current: Current
    """Rated current of the fuse."""

File: src/frequenz/sdk/microgrid/grid.py [hunk 3]
```diff
@@ -0,0 +1,82 @@
...
+
+@dataclass(frozen=True)
+class Grid:
+    """A grid connection point.
+
+    Attributes:
+        fuse: The fuse protecting the grid connection point.
+    """
+
+    fuse: Fuse
```
daniel-zullo-frequenz:
```suggestion
    """A grid connection point."""

    fuse: Fuse
    """The fuse protecting the grid connection point."""
```

src/frequenz/sdk/microgrid/grid.py
```diff
@@ -0,0 +1,82 @@
...
+
+    grid_connections_count = len(grid_connections)
+
+    if grid_connections_count == 0:
+        logging.info(
+            "No grid connection found for this microgrid. This is normal for an islanded microgrid."
+        )
+    elif grid_connections_count > 1:
+        raise RuntimeError(
+            f"Expected at most one grid connection, got {len(grid_connections)}"
```
daniel-zullo-frequenz: Alternatively you can use the dedicated variable `grid_connections_count`

src/frequenz/sdk/microgrid/grid.py
```diff
@@ -0,0 +1,82 @@
...
+
+def initialize(components: Iterable[Component]) -> None:
+    """Initialize the grid connection.
+
+    Args:
+        components: The components in the microgrid.
+
+    Raises:
+        RuntimeError: If there is more than 1 grid connection point in the
+            microgrid.
```
daniel-zullo-frequenz: Actually there are 3 cases where `RuntimeError` is raised that you might want briefly to mention them

src/frequenz/sdk/microgrid/grid.py
```diff
@@ -0,0 +1,82 @@
...
+        logging.info(
+            "No grid connection found for this microgrid. This is normal for an islanded microgrid."
+        )
+    elif grid_connections_count > 1:
+        raise RuntimeError(
+            f"Expected at most one grid connection, got {len(grid_connections)}"
+        )
+    else:
+        if grid_connections[0].metadata is None:
+            raise RuntimeError("Grid metadata is None")
```
daniel-zullo-frequenz, line 61: Optionally a more "pythonic" way to do it is `if not grid_connections[0].metadata:`

src/frequenz/sdk/microgrid/grid.py
```diff
@@ -0,0 +1,82 @@
...
+
+        fuse = grid_connections[0].metadata.fuse
+
+        if fuse is None:
```
daniel-zullo-frequenz, line 65:
```suggestion
        if not fuse:
```

tests/microgrid/mock_api.py
```diff
@@ -100,6 +102,17 @@ def add_component(
                     inverter=InverterMetadata(type=inverter_type),
                 )
             )
+        elif (
+            component_category == ComponentCategory.COMPONENT_CATEGORY_GRID
+            and max_current is not None
```
daniel-zullo-frequenz, line 109: if `max_current` should be always different than zero then I'd suggest to use `and max_current` instead as this covers the check for not `None` and not `0`

daniel-zullo-frequenz (commented):
I have a few minor suggestions to check for and LGTM otherwise

src/frequenz/sdk/microgrid/component/_component.py
matthias-wende-frequenz: Is there a reason why `max_current` is not of type `Current`?

tiyash-basu-frequenz: >Have you tried with @dataclass(frozen=True)?

yes, but it did not work.

tiyash-basu-frequenz: >Is there a reason why max_current is not of type Current?

Because that's how I started, and it did not make sense to change it to `Current` because this gets changed to `Fuse` anyway.

tests/microgrid/mock_api.py
```diff
@@ -89,6 +90,7 @@ def add_component(
         self,
         component_id: int,
         component_category: ComponentCategory.V,
+        max_current: Optional[int] = None,
```
daniel-zullo-frequenz: Also I wonder if `max_current` should be `Optional[Current]` type instead

src/frequenz/sdk/microgrid/grid.py
tiyash-basu-frequenz: Oof that's a weird one. It feels like casting fuse to a `bool` type.

tests/microgrid/mock_api.py
tiyash-basu-frequenz: This gets changed to `Fuse` a few commits later, so this change is probably not worth the effort.

src/frequenz/sdk/microgrid/grid.py
daniel-zullo-frequenz: yeah feel free to keep it as it currently is

tests/microgrid/mock_api.py
daniel-zullo-frequenz: Are you sure? I'm checking all files changes (not per commit) and this is still there :/ Maybe it's an issue in my browser :thinking:

src/frequenz/sdk/microgrid/grid.py
tiyash-basu-frequenz: Same: https://github.com/frequenz-floss/frequenz-sdk-python/pull/610#discussion_r1307389436

> Oof that's a weird one. It feels like casting fuse to a bool type.

tests/microgrid/mock_api.py
tiyash-basu-frequenz: > if `max_current` should be always different than zero

I did not make that assumption here. I think a mock should not have conditions on data values. The functional implementation should have it instead.

But it raises a good point. In this case, would it make sense to raise an error in the `Grid` module if `fuse.max_current == 0`, given the low-level layer it exists in? Or should it be handled by the power distributor?

src/frequenz/sdk/microgrid/fuse.py
tiyash-basu-frequenz: Done! Thanks for the suggestion. Do we want to enforce this style in doc tests?

src/frequenz/sdk/microgrid/grid.py
tiyash-basu-frequenz: Done

src/frequenz/sdk/microgrid/grid.py
tiyash-basu-frequenz: Done

src/frequenz/sdk/microgrid/grid.py
tiyash-basu-frequenz: Done

tests/microgrid/mock_api.py
tiyash-basu-frequenz: Oh this is the `mock_api` class. Sorry, you are right, I misread the filename. Will change it to `Current`.

tiyash-basu-frequenz: Done

src/frequenz/sdk/microgrid/fuse.py
daniel-zullo-frequenz: No, not really. Test might not need documentation except some special cases, it is up to the developer

tests/microgrid/mock_api.py
daniel-zullo-frequenz: > I did not make that assumption here. I think a mock should not have conditions on data values. The functional implementation should have it instead.

You're right, agreed.

> would it make sense to raise an error in the Grid module if fuse.max_current == 0, given the low-level layer it exists in? Or should it be handled by the power distributor?

I think It does make sense to do it in the Grid module to prevent programming or setups errors. And additionally can be done in the power distributor. Maybe @matthias-wende-frequenz has another opinion on this matter?

src/frequenz/sdk/microgrid/component/_component.py
tiyash-basu-frequenz: If there are no further suggestions, then I'll mark this resolved.

tests/microgrid/test_graph.py
```diff
@@ -816,8 +823,11 @@ async def test_refresh_from_api(self) -> None:
         servicer.set_connections([(101, 111), (111, 131)])
         await graph.refresh_from_api(client)
 
+        grid_max_current = Current.from_amperes(0.0)
```
daniel-zullo-frequenz: We can use the quantity zero constructor instead as follows:
```suggestion
        grid_max_current = Current.zero()
```

src/frequenz/sdk/microgrid/grid.py
```diff
@@ -0,0 +1,75 @@
...
+@dataclass(frozen=True)
+class Grid:
+    """A grid connection point.
+
+    Attributes:
+        max_current: The maximum current that can course through the grid
+            connection point, in Amperes.
+    """
+
+    max_current: float
```
matthias-wende-frequenz: This should be of type `Current`.

src/frequenz/sdk/microgrid/component/_component.py
llucax: > yes, but it did not work.

How it didn't work? Was there some error? It would be nice to keep the same style as other classes and besides immutability, dataclass will also take care of implementing `__eq__` for example.

RELEASE_NOTES.md
```diff
@@ -92,6 +92,12 @@ This release replaces the `@actor` decorator with a new `Actor` class.
 
 - `Actor`: This new class inherits from `BackgroundService` and it replaces the `@actor` decorator.
 
+- Calling `microgrid.initialize()` now also initializes the microgrid's grid connection point as a singleton object of a newly added type `Grid`. This object can be obtained by calling `microgrid.grid.get()`. This object exposes the max current that can course through the grid connection point, which is useful for the power distribution algorithm. The max current is provided by the Microgrid API, and can be obtained by calling `microgrid.grid.get().fuse.max_current`.
+  Note that a microgrid is allowed to have zero or one grid connection point. Microgrids configured as islands will have zero grid connection points, and microgrids configured as grid-connected will have one grid connection point.
```
llucax, line 107: For proper formatting, an empty line should be left if you want both paragraphs to be really rendered as paragraphs.

```suggestion

  Note that a microgrid is allowed to have zero or one grid connection point. Microgrids configured as islands will have zero grid connection points, and microgrids configured as grid-connected will have one grid connection point.
```

src/frequenz/sdk/microgrid/component/_component.py
```diff
@@ -106,13 +112,78 @@ def _component_category_from_protobuf(
...
+    """Base class for component metadata classes."""
+
+    @property
+    @abstractmethod
+    def fuse(self) -> Optional[Fuse]:
+        """Get the fuse associated with this component.
+
+        Returns:
+            The fuse associated with this component.
+        """
```
llucax: Any reason not to provide a default implementation returning `None` here if having a `fuse` is optional? (so `fuse` won't be an abstract method). Even more, it could be a dataclass with just an optional property fuse:

```py
@dataclass(froze=True)
class ComponentMetadata(ABC):
    """Base class for component metadata classes."""

    fuse: Fuse | None = None
        """The fuse associated with this component."""
```

Anyone can override it if they want.

src/frequenz/sdk/microgrid/component/_component.py
```diff
@@ -106,13 +112,78 @@ def _component_category_from_protobuf(
...
+    def __init__(self, fuse: Fuse) -> None:
+        """Create a new instance.
+
+        Args:
+            fuse: The fuse at the grid connection point.
+        """
+        self._fuse = fuse
+
+    @property
+    def fuse(self) -> Optional[Fuse]:
```
llucax: Nitpick: *modern* Python uses `Fuse | None` (but we are trying to modernize the code base, so it would be nice if new code uses the new style, although if the file in general is using the old style it is OK to leave it and make the modernization of the whole file in another commit/PR)

```suggestion
    def fuse(self) -> Fuse | None:
```

tests/microgrid/mock_api.py
matthias-wende-frequenz: > has another opinion on this matter?

No other opinion.

tiyash-basu-frequenz: Okay, then I think we can resolve this discussion here, and pass this on to the power distributor.

tests/microgrid/test_graph.py
tiyash-basu-frequenz: Nice! Done

RELEASE_NOTES.md
tiyash-basu-frequenz: Done

src/frequenz/sdk/microgrid/component/_component.py
tiyash-basu-frequenz: Added a default that returns `None`

src/frequenz/sdk/microgrid/component/_component.py
tiyash-basu-frequenz: Amazing! I inserted a commit to modernize the declarations, and then changed my commits to use this style.

src/frequenz/sdk/microgrid/component/_component.py
tiyash-basu-frequenz: Here is an example with `fuse`:
```
tests/microgrid/test_graph.py::Test_MicrogridComponentGraph::test_refresh_from_api <actor_restart_limit> Changing the restart limit from None to 0
FAILED<actor_restart_limit> Resetting restart limit to None
...
E       AssertionError: assert
{Component(component_id=131, category=<ComponentCategory.EV_CHARGER: 6>, type=None, metadata=None), Component(component_id=101, category=<ComponentCategory.GRID: 1>, type=None, metadata={'fuse': {'max_current': Current(value=0, exponent=0)}}), Component(component_id=111, category=<ComponentCategory.METER: 2>, type=None, metadata=None)}
== {Component(component_id=131, category=<ComponentCategory.EV_CHARGER: 6>, type=None, metadata=None), Component(component_id=101, category=<ComponentCategory.GRID: 1>, type=None, metadata=GridMetadata(fuse=Fuse(max_current=Current(value=0, exponent=0)))), Component(component_id=111, category=<ComponentCategory.METER: 2>, type=None, metadata=None)}
```

The LHS has a dict, but the RHS has an object. The method of constructing these two look the same to me, but I am definitely missing something.

tiyash-basu-frequenz: Found the culprit: https://github.com/frequenz-floss/frequenz-sdk-python/blob/1fe5cb8c0103f37580988fa03da3a345a3079290/src/frequenz/sdk/microgrid/_graph.py#L476

The graph module is converting components to dicts.

Does anyone know why this conversion to a dict is required?
Edit: because the `add_node` method takes only one positional argument, which is the component ID. Everything after it therefore has to be expanded.

tiyash-basu-frequenz: Will push changes today to get this as a dataclass.

tiyash-basu-frequenz: Done. Ready for a look.

tiyash-basu-frequenz:
Rebased on latest commit in v0.x.x.

src/frequenz/sdk/microgrid/grid.py
```diff
@@ -0,0 +1,80 @@
...
+
+@dataclass(frozen=True)
+class Grid:
+    """A grid connection point."""
+
+    fuse: Fuse
+    """The fuse protecting the grid connection point."""
+
+
+_GRID: Optional[Grid] = None
```
llucax:
```suggestion
_GRID: Grid | None = None
```

src/frequenz/sdk/microgrid/grid.py
```diff
@@ -0,0 +1,80 @@
...
+
+"""Grid connection point.
+
+This module provides the `Grid` type, which represents a grid connection point
+in a microgrid.
+"""
+
+import logging
+from dataclasses import dataclass
+from typing import Iterable, Optional
```
llucax: New import location for `Iterable` (and `Optional` is not necessary)

```suggestion
from collections.abc import Iterable
```

src/frequenz/sdk/microgrid/grid.py
```diff
@@ -0,0 +1,80 @@
...
+
+        fuse = grid_connections[0].metadata.fuse
+
+        if fuse is None:
+            raise RuntimeError("Grid fuse is None")
+
+        _GRID = Grid(fuse)
+
+
+def get() -> Optional[Grid]:
```
llucax:
```suggestion
def get() -> Grid | None:
```

tests/microgrid/mock_api.py
```diff
@@ -89,6 +92,7 @@ def add_component(
         self,
         component_id: int,
         component_category: ComponentCategory.V,
+        max_current: Optional[Current] = None,
```
llucax, line 95:
```suggestion
        max_current: Current | None = None,
```

llucax (dismissed):
A few more optional (no pun intended) comments. Maybe @shsms should have a quick look  too as he is probably the most familiar with the microgrid API code.

src/frequenz/sdk/microgrid/grid.py
tiyash-basu-frequenz: Done

src/frequenz/sdk/microgrid/grid.py
tiyash-basu-frequenz: Done

src/frequenz/sdk/microgrid/grid.py
tiyash-basu-frequenz: ~I think it's better if we stay consistent in this file and use the `Optional[X]` format here for now. We should, however, update occurrences of this format everywhere in the SDK. Although, that sounds out-of-scope for this PR to me.~

Sorry, please ignore the previous text. Done

src/frequenz/sdk/microgrid/grid.py
matthias-wende-frequenz: ~isn't typing the new location?~ Forget it. Luca was right and chatgpt was wrong :D

tests/microgrid/mock_api.py
tiyash-basu-frequenz: I think it's better if we stay consistent in this file and use the `Optional[X]` format here for now. We should, however, update occurrences of this format everywhere in the SDK. Although, that sounds out-of-scope for this PR to me.

src/frequenz/sdk/microgrid/grid.py
tiyash-basu-frequenz: Lemme see what pylance suggests.

tiyash-basu-frequenz: pylance suggests `typing`. I changed it to `typing`.

llucax: Sorry, but `pylance` is wrong :)

Python provided the stuff in `collections.abc` but at first it could just use regular types as type hints, so they needed to add special types to the `typing` module (there was even a `typing.Int`), but then it got the ability to use regular types, the use of types in `typing` become deprecated.

tiyash-basu-frequenz: reverted

llucax: Thanks! Just for more context:

image

llucax (approved):

PR #612, opened 2023-08-29
Don't return `None` from battery pool methods

### LINKED ISSUE #545: `None` values from BatteryPool's SoC/Capacity methods
When a BatteryPool has no usable batteries,  it sends out a `None` value for `SoC`, `Capacity` and `PowerBounds`,  indicating that these values are not available, in the absence of working batteries.

We need to figure out how this information can be more clearly represented.

One option is in the case of `SoC` and `Capacity`, which stream `Sample` objects,  we can just set the `Sample`s value to `None` instead.  And provide a similar mechanism for the `PowerBounds` as well.

stefan-brus-frequenz:
In the case when no battery is available, the internal values of the `Sample` or `PowerMetric` will be set to `None` instead.

### DIFF
--- a/src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
+++ b/src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
@@ -124,7 +124,7 @@ class MetricCalculator(ABC, Generic[T]):
         self,
         metrics_data: dict[int, ComponentMetricsData],
         working_batteries: set[int],
-    ) -> T | None:
+    ) -> T:
         """Aggregate the metrics_data and calculate high level metric.
 
         Missing components will be ignored. Formula will be calculated for all
@@ -191,7 +191,7 @@ class CapacityCalculator(MetricCalculator[Sample[Energy]]):
         self,
         metrics_data: dict[int, ComponentMetricsData],
         working_batteries: set[int],
-    ) -> Sample[Energy] | None:
+    ) -> Sample[Energy]:
         """Aggregate the metrics_data and calculate high level metric.
 
         Missing components will be ignored. Formula will be calculated for all
@@ -229,7 +229,7 @@ class CapacityCalculator(MetricCalculator[Sample[Energy]]):
             total_capacity += usable_capacity
 
         return (
-            None
+            Sample(datetime.now(tz=timezone.utc), None)
             if timestamp == _MIN_TIMESTAMP
             else Sample[Energy](timestamp, Energy.from_watt_hours(total_capacity))
         )
@@ -281,7 +281,7 @@ class TemperatureCalculator(MetricCalculator[Sample[Temperature]]):
         self,
         metrics_data: dict[int, ComponentMetricsData],
         working_batteries: set[int],
-    ) -> Sample[Temperature] | None:
+    ) -> Sample[Temperature]:
         """Aggregate the metrics_data and calculate high level metric for temperature.
 
         Missing components will be ignored. Formula will be calculated for all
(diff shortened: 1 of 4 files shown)


src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
```diff
@@ -60,7 +60,7 @@ def battery_inverter_mapping(batteries: Iterable[int]) -> dict[int, int]:
 
 # Formula output types class have no common interface
 # Print all possible types here.
-T = TypeVar("T", Sample[Percentage], Sample[Energy], PowerMetrics, Sample[Temperature])
+T = TypeVar("T", Sample[Percentage | None], Sample[Energy | None], PowerMetrics, Sample[Temperature | None])
```
shsms: This change is not necessary, because `Sample` already has typing for `None` inside it.  Same in the return types below: https://github.com/frequenz-floss/frequenz-sdk-python/blob/3551cd3052efc9f60aa71b0acd2e3174c5c81a8a/src/frequenz/sdk/timeseries/_base_types.py#L18-L30

src/frequenz/sdk/timeseries/battery_pool/_metric_calculator.py
```diff
@@ -229,7 +229,7 @@ def calculate(
             total_capacity += usable_capacity
 
         return (
-            None
+            Sample(timestamp, None)
             if timestamp == _MIN_TIMESTAMP
```
shsms: I think in all these places, we have to use `datetime.now(tz=timezone.utc)`, because otherwise, we'll send out the `_MIN_TIMESTAMP`, which is year 1 I think.

shsms (commented):
Just a couple of comments, LGTM otherwise.

tests/timeseries/_battery_pool/test_battery_pool.py
```diff
@@ -438,7 +438,11 @@ async def run_test_battery_status_channel(  # pylint: disable=too-many-arguments
     msg = await asyncio.wait_for(
         battery_pool_metric_receiver.receive(), timeout=waiting_time_sec
     )
-    assert msg is None
+    if isinstance(msg, PowerMetrics):
+        assert msg.inclusion_bounds is None
+        assert msg.exclusion_bounds is None
+    elif isinstance(msg, Sample):
+        assert msg.value is None
```
shsms, line 448: does just saying `else:` also work here?  you can test locally with `nox -R -e mypy`.

If not,  we'll probably have to add a separate `else:` that always raises,  so that future changes to `T` don't pass through the gaps.

tests/timeseries/_battery_pool/test_battery_pool.py
```diff
@@ -662,7 +666,8 @@ async def run_capacity_test(setup_args: SetupArgs) -> None:
     await streamer.stop_streaming(batteries_in_pool[0])
     await asyncio.sleep(MAX_BATTERY_DATA_AGE_SEC + 0.2)
     msg = await asyncio.wait_for(capacity_receiver.receive(), timeout=waiting_time_sec)
-    assert msg is None
+    if isinstance(msg, Sample):
+        assert msg.value is None
```
shsms: I wonder if we can say: `assert isinstance(msg, Sample) and msg.value is None`, so we always have an assert.  Same for below.

tests/timeseries/_battery_pool/test_battery_pool.py
stefan-brus-frequenz: > does just saying else: also work here? you can test locally with nox -R -e mypy.

It does not, because then it does not know what the type of `msg` is, and whether or not it has the `value` field. I will add the additional `else` statement.

shsms (dismissed):

shsms:
I'm guessing this needs to be rebased, for the old requirements to go away.

llucax:
> I'm guessing this needs to be rebased, for the old requirements to go away.

YES! (But it is more "for the new requirements to be met)

stefan-brus-frequenz:
rebase'd

shsms (approved):

Issue #436, opened 2023-06-14
Option to print formulas with metadata, for debugging


shsms:
### What's needed?

FormulaEngines don't have a custom `__str__` method.  They have the formula steps buried deep in some private attributes that can be printed,  but those are not very readable.

A better way to see and verify the formulas is needed.

### Proposed solution

Implement a `__str__` method for FormulaEngines,  that would:

1. Display the formulas in in-fix form,
2. Include component types for components that are part of the formula.

PR #613, opened 2023-08-29
Improve formula formatting to use infix notation

christianparpart:
refs #436

Okay, I think I'd like to split #436 into TWO PRs, because the second part of that ticket implies working on other parts of the SDK that I am not sure how to deal with just yet and that will be more intrusive as well.

The first part is considered done.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -24,7 +24,7 @@
 
 - NaN values are treated as missing when gaps are determined in the `OrderedRingBuffer`.
 
-- A new class `FormulaFormatter` has been added to format formulas from `FormulaEngine`. For convenience, `FormulaEngine` now provides an `__str__` override that makes use of this to make formulas more intuitively introspectible.
+- Now when printing `FormulaEngine` for debugging purposes the the formula will be shown in infix notation, which should be easier to read.
 
 ## Bug Fixes
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -24,6 +24,8 @@
 
 - NaN values are treated as missing when gaps are determined in the `OrderedRingBuffer`.
 
+- A new class `FormulaFormatter` has been added to format formulas from `FormulaEngine`. For convenience, `FormulaEngine` now provides an `__str__` override that makes use of this to make formulas more intuitively introspectible.
+
 ## Bug Fixes
 
 <!-- Here goes notable bug fixes that are worth a special mention or explanation -->
--- a/src/frequenz/sdk/timeseries/_formula_engine/__init__.py
+++ b/src/frequenz/sdk/timeseries/_formula_engine/__init__.py
@@ -4,11 +4,13 @@
 """A formula engine for applying formulas."""
 from ._formula_engine import FormulaEngine, FormulaEngine3Phase
 from ._formula_engine_pool import FormulaEnginePool
+from ._formula_formatter import FormulaFormatter
 from ._resampled_formula_builder import ResampledFormulaBuilder
 
 __all__ = [
     "FormulaEngine",
     "FormulaEngine3Phase",
     "FormulaEnginePool",
+    "FormulaFormatter",
     "ResampledFormulaBuilder",
 ]
--- a/src/frequenz/sdk/timeseries/_formula_engine/__init__.py
+++ b/src/frequenz/sdk/timeseries/_formula_engine/__init__.py
@@ -4,13 +4,11 @@
 """A formula engine for applying formulas."""
 from ._formula_engine import FormulaEngine, FormulaEngine3Phase
 from ._formula_engine_pool import FormulaEnginePool
-from ._formula_formatter import FormulaFormatter
 from ._resampled_formula_builder import ResampledFormulaBuilder
 
 __all__ = [
     "FormulaEngine",
     "FormulaEngine3Phase",
     "FormulaEnginePool",
-    "FormulaFormatter",
     "ResampledFormulaBuilder",
 ]
--- a/src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
+++ b/src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
@@ -1,6 +1,8 @@
 # License: MIT
 # Copyright © 2022 Frequenz Energy-as-a-Service GmbH
 
+# pylint: disable=too-many-lines
+
 """A formula engine that can apply formulas on streaming data."""
 
 from __future__ import annotations
@@ -18,6 +20,7 @@ from ..._internal._asyncio import cancel_and_await
 from .. import Sample, Sample3Phase
 from .._quantities import Quantity, QuantityT
 from ._formula_evaluator import FormulaEvaluator
+from ._formula_formatter import FormulaFormatter
 from ._formula_steps import (
     Adder,
     Averager,
@@ -303,6 +306,19 @@ class FormulaEngine(
                 )
             else:
                 await sender.send(msg)
+
+    def __str__(self) -> str:
+        """Return a string representation of the formula.
+
+        Returns:
+            A string representation of the formula.
+        """
+        steps = (
+            self._builder._build_stack
+            if len(self._builder._build_stack) > 0
+            else self._builder._steps
+        )
+        return FormulaFormatter.format(steps)
 
     def new_receiver(
         self, name: str | None = None, max_size: int = 50
@@ -618,6 +634,15 @@ class FormulaBuilder(Generic[QuantityT]):
             self._steps.append(self._build_stack.pop())
 
         return self._steps, self._metric_fetchers
+
+    def __str__(self) -> str:
+        """Return a string representation of the formula.
+
+        Returns:
+            A string representation of the formula.
+        """
+        steps = self._steps if len(self._steps) > 0 else self._build_stack
+        return FormulaFormatter.format(steps)
 
     def build(self) -> FormulaEngine[QuantityT]:
         """Create a formula engine with the steps and fetchers that have been pushed.
--- a/src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
+++ b/src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
@@ -20,7 +20,7 @@ from ..._internal._asyncio import cancel_and_await
 from .. import Sample, Sample3Phase
 from .._quantities import Quantity, QuantityT
 from ._formula_evaluator import FormulaEvaluator
-from ._formula_formatter import FormulaFormatter
+from ._formula_formatter import format_formula
 from ._formula_steps import (
     Adder,
     Averager,
@@ -318,7 +318,7 @@ class FormulaEngine(
             if len(self._builder._build_stack) > 0
             else self._builder._steps
         )
-        return FormulaFormatter.format(steps)
+        return format_formula(steps)
 
     def new_receiver(
         self, name: str | None = None, max_size: int = 50
@@ -642,7 +642,7 @@ class FormulaBuilder(Generic[QuantityT]):
             A string representation of the formula.
         """
         steps = self._steps if len(self._steps) > 0 else self._build_stack
-        return FormulaFormatter.format(steps)
+        return format_formula(steps)
 
     def build(self) -> FormulaEngine[QuantityT]:
         """Create a formula engine with the steps and fetchers that have been pushed.
--- a/src/frequenz/sdk/timeseries/_formula_engine/_formula_formatter.py
+++ b/src/frequenz/sdk/timeseries/_formula_engine/_formula_formatter.py
@@ -244,7 +244,9 @@ class FormulaFormatter:
                 case OpenParen():
                     pass  # We gently ignore this one.
 
-        assert len(self._stack) == 1
+        assert (
+            len(self._stack) == 1
+        ), f"The formula {postfix_expr} is not valid. Evaluation stack left-over: {self._stack}"
         return self._stack[0].value
 
     def _format_binary(self, operator: Operator) -> None:
--- a/src/frequenz/sdk/timeseries/_formula_engine/_formula_formatter.py
+++ b/src/frequenz/sdk/timeseries/_formula_engine/_formula_formatter.py
@@ -1,0 +1,239 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Formatter for the formula."""
+
+from __future__ import annotations
+
+import enum
+
+from ._formula_steps import (
+    Adder,
+    Averager,
+    Clipper,
+    ConstantValue,
+    Divider,
+    FormulaStep,
+    Maximizer,
+    MetricFetcher,
+    Minimizer,
+    Multiplier,
+    OpenParen,
+    Subtractor,
+)
+
+
+class OperatorPrecedence(enum.Enum):
+    """The precedence of an operator."""
+
+    SUBTRACTION = 1
+    ADDITION = 2
+    DIVISION = 3
+    MULTIPLICATION = 4
+    PRIMARY = 9
+
+    def __lt__(self, other: OperatorPrecedence) -> bool:
+        """Test the precedence of this operator is less than the precedence of the other operator.
+
+        Args:
+            other: The other operator (on the right-hand side).
+
+        Returns:
+            Whether the precedence of this operator is less than the other operator.
+        """
+        return self.value < other.value
+
+    def __str__(self) -> str:
+        """Return the string representation of the operator precedence.
+
+        Returns:
+            The string representation of the operator precedence.
+        """
+        match self:
+            case OperatorPrecedence.SUBTRACTION:
+                return "-"
+            case OperatorPrecedence.ADDITION:
+                return "+"
+            case OperatorPrecedence.DIVISION:
+                return "/"
+            case OperatorPrecedence.MULTIPLICATION:
+                return "*"
+            case OperatorPrecedence.PRIMARY:
+                return "primary"
+
+
+class StackItem:
+    """Stack item for the formula formatter."""
+
+    def __init__(self, value: str, precedence: OperatorPrecedence):
+        """Initialize the StackItem.
+
+        Args:
+            value: The value of the stack item.
+            precedence: The precedence of the stack item.
+        """
+        self.value = value
+        self.precedence = precedence
+
+    def __str__(self) -> str:
+        """Return the string representation of the stack item.
+
+        This is used for debugging purposes.
+
+        Returns:
+            str: The string representation of the stack item.
+        """
+        return f'("{self.value}", {self.precedence})'
+
+    def as_left_value(self, outer_precedence: OperatorPrecedence) -> str:
+        """Return the value of the stack item with parentheses if necessary.
+
+        Args:
+            outer_precedence: The precedence of the outer stack item.
+
+        Returns:
+            str: The value of the stack item with parentheses if necessary.
+        """
+        return f"({self.value})" if self.precedence < outer_precedence else self.value
+
+    def as_right_value(self, outer_precedence: OperatorPrecedence) -> str:
+        """Return the value of the stack item with parentheses if necessary.
+
+        Args:
+            outer_precedence: The precedence of the outer stack item.
+
+        Returns:
+            str: The value of the stack item with parentheses if necessary.
+        """
+        return f"({self.value})" if self.precedence < outer_precedence else self.value
+
+    @staticmethod
+    def create_binary(
+        lhs: StackItem, operator: OperatorPrecedence, rhs: StackItem
+    ) -> StackItem:
+        """Create a binary stack item.
+
+        Args:
+            lhs: The left-hand side of the binary operation.
+            operator: The operator of the binary operation.
+            rhs: The right-hand side of the binary operation.
+
+        Returns:
+            StackItem: The binary stack item.
+        """
+        pred = OperatorPrecedence(operator)
+        return StackItem(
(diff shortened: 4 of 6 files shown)


src/frequenz/sdk/timeseries/_formula_engine/_formula_printer.py
```diff
@@ -0,0 +1,220 @@
...
+    Divider,
+    FormulaStep,
+    MetricFetcher,
+    Multiplier,
+    OpenParen,
+    Subtractor,
+)
+
+class OperatorPrecedence(enum.Enum):
+    """Represents the precedence of an operator."""
```
llucax: Nitpick
```suggestion
    """The precedence of an operator."""
```

src/frequenz/sdk/timeseries/_formula_engine/_formula_printer.py
```diff
@@ -0,0 +1,220 @@
...
+
+    Subtraction = 1
+    Addition = 2
+    Division = 3
+    Multiplication = 4
+    Primary = 9
```
llucax: The convention for enums in Python is `ALL_CAPS`.

src/frequenz/sdk/timeseries/_formula_engine/_formula_printer.py
```diff
@@ -0,0 +1,220 @@
...
+            case "+":
+                return OperatorPrecedence.Addition
+            case "-":
+                return OperatorPrecedence.Subtraction
+            case "*":
+                return OperatorPrecedence.Multiplication
+            case "/":
+                return OperatorPrecedence.Division
+            case _:
+                return OperatorPrecedence.Primary
```
llucax: I would name this `from_str()`, as we are using that convention for quantities for example. `map` in Python means mapping using a function.

src/frequenz/sdk/timeseries/_formula_engine/_formula_printer.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+class StackItem:
+    """Represents a stack item for the formula printer."""
```
llucax:
```suggestion
    """A stack item for the formula printer."""
```

src/frequenz/sdk/timeseries/_formula_engine/_formula_printer.py
```diff
@@ -0,0 +1,220 @@
...
+        """Return the string representation of the stack item.
+
+        This is used for debugging purposes.
+
+        Returns:
+            str: The string representation of the stack item.
+        """
+        return f"(\"{self.value}\", {self.precedence})"
+
+    def plvalue(self, outer_precedence: OperatorPrecedence) -> str:
```
llucax: What is `pl` and `pr`? This needs a more readable name.

src/frequenz/sdk/timeseries/_formula_engine/_formula_printer.py
```diff
@@ -0,0 +1,220 @@
...
+
+        Args:
+            outer_precedence: The precedence of the outer stack item.
+
+        Returns:
+            str: The value of the stack item with parentheses if necessary.
+        """
+        return f"({self.value})" if self.precedence < outer_precedence else self.value
+
+    def prvalue(self, outer_precedence: OperatorPrecedence) -> str:
```
llucax: Same here.

src/frequenz/sdk/timeseries/_formula_engine/_formula_printer.py
```diff
@@ -0,0 +1,220 @@
...
+        """
+        return StackItem(str(value), OperatorPrecedence.Primary)
+
+
+class FormulaPrinter:
+    """Prints a formula."""
+
+    def __init__(self) -> None:
+        """Initialize the FormulaPrinter."""
+        self._stack = list[StackItem]()
```
llucax: I guess this will be small enough to not matter, but it is more suitable to use a `collections.deque` as a stack instead of a `list`.

src/frequenz/sdk/timeseries/_formula_engine/_formula_printer.py
```diff
@@ -0,0 +1,220 @@
...
+            value: The value of the literal.
+
+        Returns:
+            StackItem: The literal stack item.
+        """
+        return StackItem(str(value), OperatorPrecedence.Primary)
+
+
+class FormulaPrinter:
+    """Prints a formula."""
```
llucax: This is not really printing, it is formatting, so I would name it `FormulaFormatter`.

src/frequenz/sdk/timeseries/_formula_engine/_formula_printer.py
```diff
@@ -0,0 +1,220 @@
...
+        """Return the formula as a string in infix notation.
+
+        Args:
+            postfix_expr: The steps of the formula in postfix notation order.
+
+        Returns:
+            str: The formula in infix notation.
+        """
+        printer = FormulaPrinter()
+        return printer._transform(postfix_expr)
```
llucax: I wouldn't have a static method instantiating the class on each call for a couple of reasons:

1. There is no need to instantiate over and over again.
2. The class is more clumsy, as this is at the end just a plain function call, but you need to add some class name.

I think it would be more Pythonic to just make this a function and cache the instance in a global:

```python
_global_instance = FormulaFormatter()

def format(postfix_expr: list[FormulaStep]) -> str:
    return _global_instance.format(postfix_expr)
```

src/frequenz/sdk/timeseries/_formula_engine/_formula_printer.py
```diff
@@ -0,0 +1,220 @@
...
+        Pops the arguments of the binary expression from the stack
+        and pushes the string representation of the binary operation to the stack.
+
+        Args:
+            operator: The operator of the binary operation.
+        """
+        left, right = self._pop_two_from_stack()
+        self._stack.append(StackItem.create_binary(left, operator, right))
+
+    def _transform(self, postfix_expr: list[FormulaStep]) -> str:
```
llucax:
```suggestion
    def format(self, postfix_expr: list[FormulaStep]) -> str:
```

src/frequenz/sdk/timeseries/_formula_engine/_formula_printer.py
```diff
@@ -0,0 +1,220 @@
...
+    Primary = 9
+
+    def __gt__(self, other: OperatorPrecedence) -> bool:
+        """Return whether the precedence of this operator is greater than the precedence of the other operator.
+
+        Args:
+            other: The other operator (on the right-hand side).
+
+        Returns:
+            bool: Whether the precedence of this operator is greater than the precedence of the other operator.
```
llucax: Please don't specify types in the docs, the tooling is able to get it from the code anyways.

src/frequenz/sdk/timeseries/_formula_engine/_formula_printer.py
```diff
@@ -0,0 +1,220 @@
...
+
+        Args:
+            lhs: The left-hand side of the binary operation.
+            op: The operator of the binary operation.
+            rhs: The right-hand side of the binary operation.
+
+        Returns:
+            StackItem: The binary stack item.
+        """
+        pred = OperatorPrecedence.map(op)
```
llucax: If `map` is only used here, where I assume you know already this is a binary op, then another alternative could just be:

```suggestion
        pred = OperatorPrecedence(op)
```

And you can remove `map()` completely.

src/frequenz/sdk/timeseries/_formula_engine/_formula_steps.py
```diff
@@ -207,6 +207,14 @@ def __init__(self, fetchers: List[MetricFetcher[QuantityT]]) -> None:
         """
         self._fetchers: list[MetricFetcher[QuantityT]] = fetchers
 
+    def fetchers(self) -> list[MetricFetcher[QuantityT]]:
```
llucax:
```suggestion
    @property
    def fetchers(self) -> list[MetricFetcher[QuantityT]]:
```

tests/timeseries/test_formula_printer.py
```diff
@@ -0,0 +1,92 @@
...
+
+from frequenz.sdk.timeseries._formula_engine._tokenizer import (
+    Tokenizer,
+    TokenType,
+)
+
+import logging
+
+
+LOGGER = logging.getLogger(__name__)
```
llucax: I would just use `print()` in tests, otherwise you need to pass some options to `pytest` to be able to see the log output.

llucax (commented):
I guess if you created a draft PR is because you wanted some early feedback, so here it is :)

src/frequenz/sdk/timeseries/_formula_engine/_formula_printer.py
christianparpart: lol @llucax ! :-P

src/frequenz/sdk/timeseries/_formula_engine/_formula_printer.py
christianparpart: by reading the function documentation, it should be clear that l=left, r=right, p=parenthesis. But I agree, longer names are always quicker to understand. :)

src/frequenz/sdk/timeseries/_formula_engine/_formula_printer.py
christianparpart: I think it does not matter, because we neither need index-access, nor do we manipulate the front of the stack. What do you think?

src/frequenz/sdk/timeseries/_formula_engine/_formula_printer.py
christianparpart: We are having a member in that class (it might become more soon). I agree, I could just use a global here, but I am inherently against shared global state (!). I may as well just pass the members via call params, but that'll bloat up caller site.

I have renamed the (non-static) method to `_format` and and use this internally, which will also reduce the amount objects to be constructed.

src/frequenz/sdk/timeseries/_formula_engine/_formula_printer.py
christianparpart: @llucax I've done that. But all I won was getting rid of `map` (which I renamed to `from_str`) but I had to add `__str__` instead, because I needed to convert back to a string representation, as `LHS op RHS` I need to print a proper string for `op` :).

src/frequenz/sdk/timeseries/_formula_engine/_formula_steps.py
christianparpart: Thanks. I am not yet sure we will keep that code. We can leave it in, for sure (I was wondering why it was missing). I was trying to access somehow the type information about a fetcher here. Not yet done nor succeeded though :)

tests/timeseries/test_formula_printer.py
christianparpart: @llucax but `print()` calls weren't printed during `nox` call. Do you know how I can get print() being printed when calling `nox ... pytest ...`?

src/frequenz/sdk/timeseries/_formula_engine/_formula_printer.py
llucax: I thought about that, but still couldn't understand what this is about, so maybe some documentation could also help. I can imagine this might be a common term if compilers development or when working with parsers, but from an external observers POV it is not obvious. Also the implementation of both are the same, which adds to the confusion.  :D Is this a bug? You have having the same function with 2 names for some other reason?

src/frequenz/sdk/timeseries/_formula_engine/_formula_printer.py
llucax: Confusingly enough, `list` is actually a C array in Python, so needing indexing is actually a reason to use `list`, not to use it :P

A `deque` is a list of arrays, so indexing is potentially O(n), but pushing and popping from any end is amortized constant time.

Again, if we are talking about less than hundreds or items, probably `list` and `deque` are exactly the same. So it was more of an informational comment, because I guess we'll be way below that.

All of this IIRC.

christianparpart: Python is sooo confusing.
However, I do not need any indexing. I just extend at the end (which ever end, just happens to be the right end), and pop off at that. I expect Python's `list` implementation to be intelligent enough to not resize upon every append and pop, but instead having an internal `used_elements` counter that tells how many elements are in that list, effectively making a `pop()` O(1) all the times (not so good for first pushes). But I do not see how such kind of optimization is actually useful in this context, because this functionality is most likely only being used for debugging / logging / introspection, and least likely in production hot paths.

llucax: Yeah, `pop()`ing from the end should be amortized `O(1)` too (I guess it might decide to resize if the array becomes too small). :+1:

src/frequenz/sdk/timeseries/_formula_engine/_formula_steps.py
llucax: I would remove it or make it a property to follow the pattern of other similar methods that are just accessing an attribute (effectively making it read-only to the user). Actually I would probably return `Sequence[...]` instead of `list` so it's a bit more read-only (the type checker won't allow users to mutate the returned `list` if you use `Sequence`.

src/frequenz/sdk/timeseries/_formula_engine/_formula_printer.py
llucax: For for the fun of it, a shorter (but slightly slower) implementation. Too bad that the sorting of the string doesn't match the sort we need :P

```python
import functools

class OperatorPrecedence(enum.Enum):
    """The precedence of an operator."""

    SUBTRACTION = "-"
    ADDITION = "+"
    DIVISION = "/"
    MULTIPLICATION = "*"
    PRIMARY = "<primary>"

    @functools.cache
    def __lt__(self, other: OperatorPrecedence) -> bool:
        """Test the precedence of this operator is less than the precedence of the other operator.

        Args:
            other: The other operator (on the right-hand side).

        Returns:
            Whether the precedence of this operator is less than the other operator.
        """
        precedence = list(OperatorPrecedence)
        return precedence.index(self) < precedence.index(other)
```

src/frequenz/sdk/timeseries/_formula_engine/_formula_printer.py
llucax: > We are having a member in that class (it might become more soon).

We can add more functions that forward to the global formatter then, if it makes sense.

> I agree, I could just use a global here, but I am inherently against shared global state (!). I may as well just pass the members via call params, but that'll bloat up caller site.

I think this is unjustified in Python and in the context of the SDK, which is not multi-threaded, we use globals in many cases, so just "I don't like shared global state" doesn't feel like a good justification against it by itself. Using the class name just as a namespace doesn't make any sense (and then the pointless extra instantiation too each time this is called). Even Python itself does it all the time, for example [`pprint`](https://docs.python.org/3/library/pprint.html) is implemented as a class, but offers a global function for convenience (actually in this case they [instantiate the class on every call](https://github.com/python/cpython/blob/2ef888d25046963a1c7572f7204c26047c82f6a2/Lib/pprint.py#L48-L55) :scream:, but I guess that's because the Python stdlib can't assume the code is not multi-threaded, I would be OK with that too, again my main issue against the `staticmethod` is the pointless usage of the class name to call a simple function).

llucax (commented):
Except for resisting against the global and the `fetcher` not being a property, LGTM. I would also add a release note for this, as users are affected (they get more useful information when printing a formula).

tests/timeseries/test_formula_printer.py
llucax: They should be printed when test fail automatically though. To print them unconditionally: `nox -e pytest_max -- -s` (or `pytest -s`). While to see the logs you need to pass `--log-cli-level=DEBUG` to see the log output even for failing tests.

llucax: (BTW, please don't mark ongoing conversations as resolved, I missed this one, only found it because I explicitly expanded the closed conversation)

src/frequenz/sdk/timeseries/_formula_engine/_formula_printer.py
shsms: This is not part of the hot-path, people will print formulas once at the start of the application for example,  so there's no perf gain to be had from having a global variable, and it is not like in the other places where it was really necessary.

So I'm in favour of avoiding unnecessary globals as well.  And it is a class method now.

shsms:
Now that we have support for nested formulas, I think this looks good in terms of functionality.

But I'll let @llucax approve, because he has a few open discussions.

tests/timeseries/test_formula_printer.py
christianparpart: @llucax I only mark conversations as resolved if I consider them resolved.

llucax: But you asked me a question! How it can be resolved? :D

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
```diff
@@ -28,6 +30,7 @@
 from .. import Sample, Sample3Phase
 from .._quantities import Quantity, QuantityT
 from ._formula_evaluator import FormulaEvaluator
+from ._formula_formatter import FormulaFormatter
```
llucax:
```suggestion
from ._formula_formatter import format_formula
```

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
```diff
@@ -314,6 +317,19 @@ async def _run(self) -> None:
...
+
+        Returns:
+            A string representation of the formula.
+        """
+        steps = (
+            self._builder._build_stack
+            if len(self._builder._build_stack) > 0
+            else self._builder._steps
+        )
+        return FormulaFormatter.format(steps)
```
llucax:
```suggestion
        return format_formula(steps)
```

src/frequenz/sdk/timeseries/_formula_engine/_formula_engine.py
```diff
@@ -629,6 +645,15 @@ def finalize(
...
         return self._steps, self._metric_fetchers
 
+    def __str__(self) -> str:
+        """Return a string representation of the formula.
+
+        Returns:
+            A string representation of the formula.
+        """
+        steps = self._steps if len(self._steps) > 0 else self._build_stack
+        return FormulaFormatter.format(steps)
```
llucax:
```suggestion
        return format_formula(steps)
```

src/frequenz/sdk/timeseries/_formula_engine/_formula_formatter.py
```diff
@@ -0,0 +1,272 @@
...
+        """Return the formula as a string in infix notation.
+
+        Args:
+            postfix_expr: The steps of the formula in postfix notation order.
+
+        Returns:
+            str: The formula in infix notation.
+        """
+        formatter = FormulaFormatter()
+        return formatter._format(postfix_expr)
```
llucax:
```suggestion
```

src/frequenz/sdk/timeseries/_formula_engine/_formula_formatter.py
```diff
@@ -0,0 +1,272 @@
...
+
+    def _pop_two_from_stack(self) -> tuple[StackItem, StackItem]:
+        """Pop two items from the stack.
+
+        Returns:
+            The two items popped from the stack.
+        """
+        right = self._stack.pop()
+        left = self._stack.pop()
+        return left, right
```
llucax, line 259:
```suggestion
        return left, right


def format_formula(postfix_expr: list[FormulaStep]) -> str:
        """Return a representation of the formula as a string in infix notation.

        Args:
            postfix_expr: The steps of the formula in postfix notation order.

        Returns:
            str: The formula in infix notation.
        """
        formatter = FormulaFormatter()
        return formatter._format(postfix_expr)
```

Not using a global since it was 2 against 1 :P

tests/timeseries/test_formula_printer.py
christianparpart: Sometimes I don't see very good. :P

src/frequenz/sdk/timeseries/_formula_engine/__init__.py
```diff
@@ -4,11 +4,13 @@
...
 from ._formula_engine_pool import FormulaEnginePool
+from ._formula_formatter import format_formula
 from ._resampled_formula_builder import ResampledFormulaBuilder
 
 __all__ = [
     "FormulaEngine",
     "FormulaEngine3Phase",
     "FormulaEnginePool",
     "ResampledFormulaBuilder",
+    "format_formula",
```
llucax:
```suggestion
```

There is no need to export here, these are all private modules and internal imports should be done using the original location to avoid potential circular dependencies.

RELEASE_NOTES.md
```diff
@@ -24,6 +24,8 @@
 
 - NaN values are treated as missing when gaps are determined in the `OrderedRingBuffer`.
 
+- A new class `format_formula` has been added to format formulas from `FormulaEngine`. For convenience, `FormulaEngine` now provides an `__str__` override that makes use of this to make formulas more intuitively introspectible.
+
```
llucax: I wouldn't enter into that much detail, for the user the only visible change is when printing formulas the output was improved to use infix notation. For example something like:

```suggestion
- Now when printing `FormulaEngine` for debugging purposes the the formula will be shown in infix notation, which should be easier to read.
```

llucax (commented):
2 more small comments and we are set!

llucax (approved):
Really nice to see some pretty printing in the formulas. Thanks!

PR #617, opened 2023-08-30
Clear release notes

llucax:


llucax:
Force-merged as it is trivial, and should be merged before anything else.

PR #622, opened 2023-08-30
[ci] Add tests for non-native architectures

### LINKED ISSUE #391: Enable CI for ARM
### What's needed?

We should make sure the SDK works in ARM too, as this is the most typical architecture used by controllers.

### Proposed solution

Make test (and only tests, we don't need to run all lints) in ARM (in all supported platforms) in the `merge_group` event, so reviewing and iterating over PRs is still fast.

tiyash-basu-frequenz:
This change introduces a new job to the CI to test the SDK on non-native architectures. Github CI ATM provides amd64 runners by default, so the new job targets testing on arch64 for now.

The tests are run in containers running on the target platform(s). The container definitions are supposed to exist in the directory .github/containers/nox-cross-arch. A definition for an arm64-ubuntu20.04-python3.11 container can be found there.

Since the tests take quite a while to run, given that they run on emulations, this step runs only on the `merge_group` event.


tiyash-basu-frequenz:
It is a false positive. The warning reported in the CI (check [here](https://github.com/frequenz-floss/frequenz-sdk-python/actions/runs/6026349126/job/16467161486#step:5:519)) will disappear in future CI runs once the migration to BackgroundService is completed.

For now, we can ignore the CI issue here.

.github/containers/nox-cross-arch/arm64-ubuntu-20.04-python-3.11.Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM docker.io/library/ubuntu:20.04
```
llucax, line 1: Any reason not to use the official Python image to save some setup and depending on some PPA?

https://hub.docker.com/_/python

.github/workflows/ci.yaml
```diff
@@ -57,6 +57,75 @@ jobs:
         run: nox -e ci_checks_max pytest_min
         timeout-minutes: 10
 
+  nox-cross-arch:
+    name: Cross-arch tests with nox
+    needs: ["nox"]
```
llucax: Any reason not to run it in parallel with `nox` instead? I guess you did it to save some CPU in case the "no cross-arch version" fails, but serializing it means a much slower CI in the success case (which hopefully is the majority of cases :D).

.github/workflows/ci.yaml
```diff
@@ -57,6 +57,75 @@ jobs:
...
+      # exists for the combination of items in the matrix.
+      # Refer to .github/containers/nox-cross-arch/README.md to learn how to
+      # add and name new dockerfiles.
+      matrix:
+        arch:
+          - arm64
+        os:
+          - ubuntu-20.04
+        python:
+          - "3.11"
```
llucax, line 97: You probably want to add now the `nox` *session* here too, it brought some nice speedups when doing it for the *normal* `nox` job.

.github/workflows/ci.yaml
```diff
@@ -57,6 +57,75 @@ jobs:
         run: nox -e ci_checks_max pytest_min
         timeout-minutes: 10
 
+  nox-cross-arch:
+    name: Cross-arch tests with nox
+    needs: ["nox"]
+    if: github.event_name == 'merge_group'
```
llucax, line 84: Ah, nice! probably it is worth mentioning in the PR description/commit that this only runs in the merge queue, right?

.github/containers/nox-cross-arch/README.md
```diff
@@ -0,0 +1,21 @@
+# Cross-Arch Testing Containers
```
llucax: Maybe mention this, add a link to this file or something, in the  `CONTRIBUTING.md` file? And maybe also in the `README.md` we should mention that now ARM is officially supported or something?

I feel this will have very little visibility after merged, but maybe it is OK, maybe people will find out where it works or doesn't by just trying to install it. Anyway, mentioning it in case someone has a good idea :)

.github/workflows/ci.yaml
```diff
@@ -57,6 +57,75 @@ jobs:
...
+        run: |
+          docker run \
+            --rm \
+            -v $(pwd):/${{ github.workspace }} \
+            -w ${{ github.workspace }} \
+            --net=host \
+            --platform linux/${{ matrix.arch }} \
+            localhost/nox-cross-arch:latest \
+            nox -e pytest_min
+        timeout-minutes: 30
```
llucax: I would also split these steps to match what we now have in the `nox` job, so we get the output of `pip freeze` to know exactly which deps are being installed for debugging purposes.

llucax (commented):
Awesome! Just a few comments and questions...

Also, wish for the future: Have our own GitHub ARM runners (https://actuated.dev/blog/native-arm64-for-github-actions)

.github/containers/nox-cross-arch/arm64-ubuntu-20.04-python-3.11.Dockerfile
tiyash-basu-frequenz: The other tests in the CI use Ubuntu 20.04. It's better to be consistent here, and use the same base image instead of changing it to a different one.

Also, the test matrix should cover _both_ OS and python version. In case we get Ubuntu 22.04 controller, we need to be able to extend this systematically.

.github/workflows/ci.yaml
tiyash-basu-frequenz: No reason at all. I can remove the dependency.

.github/containers/nox-cross-arch/README.md
tiyash-basu-frequenz: Sure, I can extend the other docs too.

.github/workflows/ci.yaml
tiyash-basu-frequenz: > I would also split these steps to match what we now have in the nox job

So, run the other step before the docker run command? What happens if some of the dependencies are architecture dependent? Is that even possible for the SDK? Should the tests fail then? (that could be a legit requirement)

tiyash-basu-frequenz: I could try splitting the steps, but run those in docker, and see what happens.

.github/containers/nox-cross-arch/arm64-ubuntu-20.04-python-3.11.Dockerfile
tiyash-basu-frequenz: Btw, there is no official image using ubuntu 20.04 with python 3.11. That's why this needs to be setup manually.

.github/workflows/ci.yaml
llucax: Yeah, I think we should try to parallelize as much as possible to the total CI time is as low as possible in the green case.

.github/workflows/ci.yaml
llucax: Ah, right. This is more complicated because you are running `docker` manually. This might work if we create a virtualenv in the current directory (so the `pip install` shouldn't be done in the `Dockerfile` entrypoint, it should be done in the GH workflow), then both venvs, then one we created and the ones created by nox will be mounted as a volume and preserved across docker runs.

It would be ideal to just be able to [run the whole job in the container directly](https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container), but I guess we can't because of the QEMU part. Is QEMU needed for building the image, running it or both?

Another way that might be easier is to run a step with `docker run -d ...`, add a final step with `if: always()` that will kill the container we just ran as a cleanup step (to be nice), and then to split the intermediate steps use `docker exec` on the daemonized container running in the background.

Anyway, I thought this was going to be much trivial, I didn't have in mind the explicit docker part, I'm not sure if any of the above work is worth it just to get the extra debug info ~and a some speedup~ [^1], I guess it depends on how much keen you are in trying one of the approaches :)

[^1]: The speedup is actually achieved by adding the `nox` sessions to the matrix, which is independent from the splitting, but also if all the QEMU setup and image building is too slow it might actually slow things up instead of speeding them up

tiyash-basu-frequenz: > Is QEMU needed for building the image, running it or both?

QEMU is required for running the image. We need to emulate an arm64 platform here, and that's what QEMU does for us.

tiyash-basu-frequenz: > This might work if we create a virtualenv in the current directory (so the pip install shouldn't be done in the Dockerfile entrypoint, it should be done in the GH workflow),

Right, that's what I am trying now. The venv installation also needs to be containerized, because there could be arch/platform-dependent python dependencies.

tiyash-basu-frequenz: The image building is cached btw, and took less than 30s in my tests.

tiyash-basu-frequenz: So, after a few (lengthy) tests and trials, it turns out that caching dependencies from pyproject.toml is non-trivial. It is not possible to install just the dependencies [yet](https://github.com/pypa/pip/issues/11440). So, just installing the dependencies in the image doesn't work that way. You could `COPY` the whole project into the image and run it, but then the layer gets rebuilt every time, defeating the purpose.

The next thing that could be cached is the `.nox` directory, which is trivial, and has been added now.

Once the docker image and the .nox directory is in cache, the job takes about 13-15 minutes to run. Although a lot of time is taken by the last tests (`src/frequenz/.../*.py`).

tiyash-basu-frequenz: >Another way that might be easier is to run a step with docker run -d ..., add a final step with if: always() that will kill the container we just ran as a cleanup step (to be nice), and then to split the intermediate steps use docker exec on the daemonized container running in the background.

While running commands in the docker container technically works, it does not help with caching cross-arch packages. Maybe I misunderstood something here?

.github/workflows/ci.yaml
tiyash-basu-frequenz: Done

.github/workflows/ci.yaml
tiyash-basu-frequenz: Done. Now multiple `nox` sessions can be run in parallel, if needed.

.github/containers/nox-cross-arch/README.md
tiyash-basu-frequenz: I extended the contribution docs, and added a release note.

Is there a place where we specify the officially supported setups? E.g., Ubuntu 20.04, python 3.11, amd64 (and now arm64)? I could not find it in the README, and it could make sense to add one there.

tiyash-basu-frequenz:
The CI complains:
```
WARNING -  Doc file 'CONTRIBUTING.md' contains a relative link '.github/containers/nox-cross-arch/README.md', but the target is not found among documentation files.
```

but the file exists, and I have tested the link in my local environment.
Looking into it.

Edit: Fixed. Needed an entry in the `docs` directory.

.github/workflows/ci.yaml
llucax: Oh, I didn't notice you were only running the `pytest_min`, I guess the reason is not to test the linting and checks? If that's the case, which is reasonable, you can use the sessions `pytest_min` and `pytest_max`, but I think we should check for both, as we want to make sure our code still works with new dependency versions (and new versions of dependencies are also available for ARM).

tiyash-basu-frequenz: >I guess the reason is not to test the linting and checks? 

That is indeed the reason.

tiyash-basu-frequenz: I'll add `pytest_max`.

.github/workflows/ci.yaml
llucax: Mmm. We should probably keep an eye on this, I was thinking that since we are not adding a branch protection rule to block merges that fail this check (because it will not be run in the PR, so it would be impossible to queue in this case), if the check fails in the merge queue, it will be merged anyway (not 100% sure but that's my theory) and nobody will see that this check failed, as it is hidden in the depths of the merge queue.

We probably should run it on `push` too, specially for `push` matching a tag, to make sure at least tags are tested. Ideally if this step fails, the publishing of the release should be blocked, which is also not going to happen in the current state.

I think we might try to run it always and see how slow it is. If it is too slow, we can think about moving it back to the merge queue (we could have a trick to run it in the PR but make it a NOP), or decide if we do it on any `push`, or only tag pushes, and how we make sure releases don't progress if it fails in that case.

Sorry for the back and forth, as I said before this had some tricky points.

.github/containers/nox-cross-arch/README.md
llucax: I don't think so, at some point I think we had something in the readme, but it got removed. I think the readme should be a good place to have a small section mentioning supported archs and python versions. I wonder if we could add some badge that actually gets it from the CI, but maybe for now we are fine with a static version.

docs/.github/containers/nox-cross-arch/README.md
llucax, line 1: I would remove this file, `docs/` is for generating the docs website, and I'm not sure if a directory with `.` will be included at all. The documentation website will have a link to this document anyway via the contributing guide, which is already included.

.github/workflows/ci.yaml
```diff
@@ -79,6 +79,86 @@ jobs:
...
+      - name: Run nox
+        run: |
+          docker run \
+            --rm \
+            -v $(pwd):/${{ github.workspace }} \
+            -w ${{ github.workspace }} \
+            --net=host \
+            --platform linux/${{ matrix.arch }} \
+            localhost/nox-cross-arch:latest \
+            nox -R -e ${{ matrix.nox-session }}
```
llucax: The `-R` means to not create the `venv`, but this only makes sense if the `venv` was created before. Also for the CI we need to make sure the `venv` is (re)generated in case dependencies changed.

```suggestion
            nox -e ${{ matrix.nox-session }}
```

.github/workflows/ci.yaml
```diff
@@ -79,6 +79,86 @@ jobs:
...
+          rm -rf /tmp/.buildx-cache
+          mv /tmp/.buildx-cache-new /tmp/.buildx-cache
+
+      # Cache nox downloads
+      - name: Cache nox downloads
+        uses: actions/cache@v3
+        with:
+          path: |
+            .nox/${{ matrix.nox-session }}
+          key: nox-${{ matrix.nox-session }}-${{ matrix.arch }}-${{ matrix.os }}-${{ matrix.python }}-${{ hashFiles('**/pyproject.toml') }}
```
llucax: This is probably not worth keeping, as we need to regenerate the nox venvs to make sure new dependencies are accounted for when creating the venvs. If you want to cache the downloaded packages, you could do that by saving `$(pip cache dir)` (usually `~/.cache/pip`).

.github/workflows/ci.yaml
llucax: I just saw your comment about this taking ~15m (I would say over 5m is too much). I think that's a bit too much to run on every PR. I guess the best next option would be to run it always, but inside the `run` check if it is a PR and if it is, just make it a NOP (ideally mark it as skipped but I'm not sure if a branch protection rule is considered fulfilled if the check was skipped).

This way we can require for the check to pass, and it will only run on the merge queue for PRs, and will be rejected from the queue if it fails. It will also run for pushes, and if we add it as a dependency for the `publish-docs` job, then a release will be blocked if the ARM tests fail.

.github/workflows/ci.yaml
tiyash-basu-frequenz: Done

.github/workflows/ci.yaml
tiyash-basu-frequenz: I was under the same impression, but I was wrong. In my tests, the env gets created if it does not exist.

tiyash-basu-frequenz: Check this for an example: https://github.com/tiyash-basu-frequenz/frequenz-sdk-python/actions/runs/6082366114/job/16499995712

.github/workflows/ci.yaml
tiyash-basu-frequenz: No, as mentioned above, this works.

tiyash-basu-frequenz: I could also add `~/.cache/pip` to the cache.

llucax:
@tiyash-basu-frequenz 

> The CI complains:
> 
> ```
> WARNING -  Doc file 'CONTRIBUTING.md' contains a relative link '.github/containers/nox-cross-arch/README.md', but the target is not found among documentation files.
> ```
> 
> but the file exists, and I have tested the link in my local environment. Looking into it.
> 
> Edit: Fixed. Needed an entry in the `docs` directory.

There are still issues with this:

> INFO    -  The following pages are being built only for the preview but will be excluded from `mkdocs build` per `exclude_docs`:
>             - http://127.0.0.1:8000/.github/containers/nox-cross-arch/

Also the URL looks very weird. I think the easiest solution for this would be to move the contents of `.github/containers/nox-cross-arch/README.md` to `CONTRIBUTING.md` and maybe leave only there a comment saying "Look at `CONTRIBUTING.md` for details.

.github/workflows/ci.yaml
tiyash-basu-frequenz: > Also for the CI we need to make sure the venv is (re)generated in case dependencies changed.

This requirement has also been addressed here. If the dependencies change, then `pyproject.toml` changes, which invalidates the cache. Then `nox` downloads everything all over again.

llucax: A way to still get the debug info about dependencies would be something like `/bin/sh -c "nox --install-only -e ${{ matrix.nox-session }}; pip freeze; nox -e ${{ matrix.nox-session }}`.

llucax (commented):
Sorry for the messy review. So as a summary, I would:

* Try the NOP approach for running the job in PRs (and running on `push` too) so we can have a branch protection rule.
* Add the `pytest_max` `nox` session to the matrix.
* Remove `-R` from `nox` invocation.

With this I think we are ready to merge (unless the CI is unbearably slow and we need to address when to run it before merging).

llucax:
One more comment, if you can maybe let the job run once in this PR it might be interesting to know how much time does it get to run in GitHub Actions.

tiyash-basu-frequenz:
> Also the URL looks very weird. I think the easiest solution for this would be to move the contents of .github/containers/nox-cross-arch/README.md to CONTRIBUTING.md and maybe leave only there a comment saying "Look at CONTRIBUTING.md for details.

Done

tiyash-basu-frequenz:
>* Try the NOP approach for running the job in PRs (and running on push too) so we can have a branch protection rule.
> * Add the pytest_max nox session to the matrix.
> * Remove -R from nox invocation.


As mentioned in threads, `pytest_max` has been added already, and, also explained in thread, there is no need to remove `-R` from nox.

.github/workflows/ci.yaml
llucax: > Yeah, I wasn't talking about caching at all, it was just a small thing about having a separate step to be able to know which dependencies were used exactly when running the tests, to be able to debug weird cases where stuff work locally but not in the CI or viceversa.

Oh, right, and this doesn't need to be checked more than once at all. There is too much mess to be cleaned here :(

As a workaround, if this saves a lot of time, you could try running `nox -e pytest_xxx -- tests`, this should not run the testing of examples in `src/` (which are just checked by `pylint`, they are not really run or anything, so there is no point in doing it more than once per-CI run).

llucax: > While running commands in the docker container technically works, it does not help with caching cross-arch packages. Maybe I misunderstood something here?

The main point about that was not caching actually, but just printing the exact version of installed deps, for debugging purposes only, but I just realized that we can do that in only once step too, it is just a bit more annoying to go look for the output. I put the suggestion in the review.

.github/workflows/ci.yaml
llucax: Really? The help says:

```
  -r, --reuse-existing-virtualenvs
                        Re-use existing virtualenvs instead of recreating them.
...
-R                    Re-use existing virtualenvs and skip package re-installation. This is an alias for '--reuse-existing-virtualenvs --no-install'.
...
  --no-install          Skip invocations of session methods for installing packages (session.install, session.conda_install, session.run_always) when a virtualenv is being reused.
```

In any case, we do really want to re-install on each run, to make sure we get the latest dependencies, at least for `pytest_max`. If we don't, I don't think dependencies will be upgraded, to do so `pip` needs to be invoked with `-U` and `nox` doesn't seem to do that.

llucax: > This requirement has also been addressed here. If the dependencies change, then `pyproject.toml` changes, which invalidates the cache. Then `nox` downloads everything all over again.

No, this is not exactly the case, if we depend on `xxx >= 1.1.1, < 2` for example, then if the venv had installed 1.1.1 and then a new 1.2.0 is out, the package will not be upgraded by `nox` and the `pyproject.toml` file won't change either. That works for pinned dependencies, but since this is a library we want to support the widest range of dependencies possible.

.github/workflows/ci.yaml
llucax: Lets keep the discussion in the other thread to avoid repetition.

.github/containers/nox-cross-arch/README.md
tiyash-basu-frequenz: Okay, so how do we proceed here? We add a `Supported Platforms` sections to the README?

.github/workflows/ci.yaml
tiyash-basu-frequenz: okay, I'll add `pip freeze` to the commands in the container.

.github/workflows/ci.yaml
tiyash-basu-frequenz: okay I'll remove `-R`. I'll also remove the caching step then because it would not make sense anymore.

.github/containers/nox-cross-arch/README.md
llucax: Yeah, do it. Then we see if we find a better way.

.github/workflows/ci.yaml
tiyash-basu-frequenz: As discussed, the new step continues to run on the `merge_group` event. It is however commented out now for running a few tests.

>and if we add it as a dependency for the publish-docs job, then a release will be blocked if the ARM tests fail.

Yep. Added a dependency to `publish-docs` for `nox-cross-arch`.

.github/containers/nox-cross-arch/README.md
tiyash-basu-frequenz: Done.

.github/workflows/ci.yaml
tiyash-basu-frequenz: Done

.github/workflows/ci.yaml
tiyash-basu-frequenz: Done. Also, added steps to cache pip downloads.

tiyash-basu-frequenz: Makes sense. Done

tiyash-basu-frequenz:
I removed the filter to run it only on merge_group events, to get some test results.
Here is a run without caches: https://github.com/frequenz-floss/frequenz-sdk-python/actions/runs/6096928673/job/16543523647?pr=622
Here is a reattempt of it, with caches: https://github.com/frequenz-floss/frequenz-sdk-python/actions/runs/6096928673/job/16546405568?pr=622

CONTRIBUTING.md
```diff
@@ -154,3 +154,27 @@ These are the steps to create a new release:
    eventually too).
 
 7. Celebrate!
+
+##  Cross-Arch Testing
+
+This project has built-in support for testing across multiple architectures. Currently, our CI conducts tests on `aarch64` machines using QEMU emulation. We also have the flexibility to expand this support to include additional architectures in the future.
```
llucax: Nitpick: I would use `arm64` since this is what's used in the code, and according to [Wikipedia](https://en.wikipedia.org/wiki/AArch64) they are synonims. I think it makes things easier to understand too, as for people that didn't work much with the arch it is not that clear what aarch64 but arm64 is obvious.
```suggestion
This project has built-in support for testing across multiple architectures. Currently, our CI conducts tests on `arm64` machines using QEMU emulation. We also have the flexibility to expand this support to include additional architectures in the future.
```

CONTRIBUTING.md
```diff
@@ -154,3 +154,27 @@ These are the steps to create a new release:
...
+
+This project containers dockerfiles that can be used in the CI to test the
+python package in non-native machine architectures, e.g., `aarch64`.
```
llucax:
```suggestion
python package in non-native machine architectures, e.g., `arm64`.
```

RELEASE_NOTES.md
```diff
@@ -24,6 +24,8 @@
 
 - NaN values are treated as missing when gaps are determined in the `OrderedRingBuffer`.
 
+- The CI now runs cross-arch tests on `aarch64` architectures.
```
llucax:
```suggestion
- The CI now runs cross-arch tests on `arm64` architectures.
```

.github/workflows/ci.yaml
```diff
@@ -79,6 +79,98 @@ jobs:
...
+            -w ${{ github.workspace }} \
+            --net=host \
+            --platform linux/${{ matrix.arch }} \
+            localhost/nox-cross-arch:latest \
+            bash -c "pip install -e .[dev-noxfile]; nox --install-only -e ${{ matrix.nox-session }}; pip freeze; nox -e ${{ matrix.nox-session }}"
+        timeout-minutes: 30
+
+      # This ensures that the runner has access to the pip cache.
+      - name: Reset pip cache ownership
+        run: sudo chown -R $USER:$USER /tmp/pip-cache
```
llucax: If the previous step fails, then the ownership reset won't be done, right? Shouldn't this use [`if: always()`](https://docs.github.com/en/actions/learn-github-actions/expressions#always)?
```suggestion
      # This ensures that the runner has access to the pip cache.
      - name: Reset pip cache ownership
        if: always()
        run: sudo chown -R $USER:$USER /tmp/pip-cache
```

llucax (commented):
Just 2 more comments, one small and one that might be important.

llucax:
> I removed the filter to run it only on merge_group events, to get some test results.
> Here is a run without caches: [frequenz-floss/frequenz-sdk-python/actions/runs/6096928673/job/16543523647?pr=622](https://github.com/frequenz-floss/frequenz-sdk-python/actions/runs/6096928673/job/16543523647?pr=622)
> Here is a reattempt of it, with caches: [frequenz-floss/frequenz-sdk-python/actions/runs/6096928673/job/16546405568?pr=622](https://github.com/frequenz-floss/frequenz-sdk-python/actions/runs/6096928673/job/16546405568?pr=622)

Great to see it green! So the cache makes a big difference but still with cache it is in the 15+m so I think we should proceed with only running on the merge queue. As we talked about, push is not necessary as the exact same commit tested in the merge queue will be pushed, so we'll be testing the exact same code twice.

.github/workflows/ci.yaml
tiyash-basu-frequenz: True, makes sense. Done

CONTRIBUTING.md
tiyash-basu-frequenz: Happy to use `arm64` if it improves consistency.
Done.

CONTRIBUTING.md
tiyash-basu-frequenz: Done

RELEASE_NOTES.md
tiyash-basu-frequenz: Done

llucax:
OK, this is good to go from my side, it only needs a rebase and :rocket: :tada: :shipit:

llucax:
I will enable the branch protection rule to check for cross-arch too to see if a skipped test is counted as passed.

tiyash-basu-frequenz:
Rebased.

llucax:
Mmm, it seems the branch protection doesn't work because the matrix is not "expanded". I guess if I add the not expanded version, then the merge queue will not have it, but I will try it out.

llucax (dismissed):

llucax:
Let's see if it goes through the queue :crossed_fingers:

tiyash-basu-frequenz:
> Let's see if it goes through the queue 🤞

:crossed_fingers:

llucax:
It doesn't look like it will, it keeps waiting for the non-matrix job name to complete.

image

Let's wait until the matrix jobs complete just in case.

Otherwise maybe we need to move the `if:` to each step or something...

llucax:
Known issue with no solution in sight, other users are using the workaround I suggest above. It is 

* https://github.com/actions/runner/issues/952
* https://github.com/orgs/community/discussions/9141

I will be misleading to see a green when it is not actually executed, specially if it then fails in the queue, but I guess is the best we can do :cry:

llucax:
Yeah, it doesn't work, now all jobs finished:

image

But the queue is still waiting for the non-matrix job in amber...

tiyash-basu-frequenz:
Pushing the workaround in a bit.

llucax:
OK, another option I guess it would be to create a fake non matrix job with the same name, maybe that's easier and less confusing as it could still be marked as skipped

llucax:
I'm thinking about:

```yaml
  # TODO: Add comment explaining what is this for
  nox-cross-arch-dummy:
    name: Cross-arch tests with nox
    if: never()
    runs-on: ${{ matrix.os }}

    steps:
      - name: Dummy step
        run: true

  nox-cross-arch:
    name: Cross-arch tests with nox
    if: github.event_name == 'merge_group'
    strategy:
      fail-fast: false
      # Before adding new items to this matrix, make sure that a dockerfile
      # exists for the combination of items in the matrix.
      # Refer to .github/containers/nox-cross-arch/README.md to learn how to
      # add and name new dockerfiles.
      matrix:
        arch:
          - arm64
        os:
          - ubuntu-20.04
        python:
          - "3.11"
        nox-session:
          - "pytest_min"
          - "pytest_max"
    runs-on: ${{ matrix.os }}

    steps:
      - name: Fetch sources
        uses: actions/checkout@v3
    #...
```

llucax:
Oh, no, we should fake the ones that run in the queue, so it is more boilerplate, as we need 2 dummy jobs (actually one per matrix item in the future).

tiyash-basu-frequenz:
Another workaround would be to create a dummy job that runs only on merge_group events, and marking it as a dependency for the nox-cross-arch job.

I need to read a bit more about the issue. Maybe the CI handles `if` and `needs` in different ways.

llucax:
Not sure if it is worth it in this case, but just in case:

```yaml
  # TODO: Add comment explaining what is this for
  nox-cross-arch-dummy-min:
    name: Cross-arch tests with nox (arm64, ubuntu-20.04, 3.11, pytest_min)
    if: never()
    runs-on: ${{ matrix.os }}

    steps:
      - name: Dummy step
        run: true

  nox-cross-arch-dummy-max:
    name: Cross-arch tests with nox (arm64, ubuntu-20.04, 3.11, pytest_max)
    if: never()
    runs-on: ${{ matrix.os }}

    steps:
      - name: Dummy step
        run: true

  nox-cross-arch:
    name: Cross-arch tests with nox
    if: github.event_name == 'merge_group'
    strategy:
      fail-fast: false
      # Before adding new items to this matrix, make sure that a dockerfile
      # exists for the combination of items in the matrix.
      # Refer to .github/containers/nox-cross-arch/README.md to learn how to
      # add and name new dockerfiles.
      matrix:
        arch:
          - arm64
        os:
          - ubuntu-20.04
        python:
          - "3.11"
        nox-session:
          - "pytest_min"
          - "pytest_max"
    runs-on: ${{ matrix.os }}

    steps:
      - name: Fetch sources
        uses: actions/checkout@v3
    #...
```

llucax:
Ah, yeah, that's not bad to have a job that "merges" all the matrix jobs for cross-arch and we add that as the branch protection :+1:

llucax:
```yaml
  # TODO: Add comment explaining what is this for
  nox-cross-arch-all:
    needs: ["nox-cross-arch"]
    name: Cross-arch tests with nox
    if: github.event_name == 'merge_group'
    runs-on: ubuntu-20.04

    steps:
      - name: Dummy step
        run: true

  nox-cross-arch:
    name: Cross-arch tests with nox
    if: github.event_name == 'merge_group'
    strategy:
      fail-fast: false
      # Before adding new items to this matrix, make sure that a dockerfile
      # exists for the combination of items in the matrix.
      # Refer to .github/containers/nox-cross-arch/README.md to learn how to
      # add and name new dockerfiles.
      matrix:
        arch:
          - arm64
        os:
          - ubuntu-20.04
        python:
          - "3.11"
        nox-session:
          - "pytest_min"
          - "pytest_max"
    runs-on: ${{ matrix.os }}

    steps:
      - name: Fetch sources
        uses: actions/checkout@v3
    #...
```

I think this should work. For merge queues `nox-cross-arch-all` will depend on `nox-cross-arch-all` and run, so when used to protect the branch merge queues should be merged.

For PRs `nox-cross-arch-all` should be skipped, also fulfilling the branch protection rule.

tiyash-basu-frequenz:
Yep, that's exactly what I was thinking of.

tiyash-basu-frequenz:
Added the workaround.

.github/workflows/ci.yaml
```diff
@@ -79,6 +79,114 @@ jobs:
...
+
+  # This job runs if `nox-cross-arch` ran and succeeded.
+  # This is required because, when the `nox-cross-arch` job is skipped, its
+  # inner matrix is not expanded, and branch protection rules on the
+  # inner-matrix jobs get stuck. So instead of `nox-cross-arch`, this job can be
+  # used in the branch protection rules.
+  # At the time of writing this, there is an ongoing discussion about this here:
+  # https://github.com/orgs/community/discussions/9141
+  nox-cross-arch-all:
+    name: A dummy non-matrix job that only passes if the nox-cross-arch job passed.
```
llucax: I would name this exactly the same as `nox-cross-arch`, so the hack gets hidden to the users, the branch protecting rules work on the name, and the name can be duplicated (we use that trick to make DCO work with merge queues).

```suggestion
    name: Cross-arch tests with nox
```

tiyash-basu-frequenz: Sure, done.

llucax (dismissed):

llucax:
Jobs are not being created, it seems like something is wrong:

> Invalid workflow file: .github/workflows/ci.yaml#L183
> The workflow is not valid. .github/workflows/ci.yaml (Line: 183, Col: 5): Required property is missing: runs-on

.github/workflows/ci.yaml
```diff
@@ -176,7 +284,7 @@ jobs:
 
   publish-docs:
     name: Publish documentation website to GitHub pages
-    needs: ["nox", "test-installation"]
+    needs: ["nox", "nox-cross-arch", "test-installation"]
```
llucax: Also this, just to make the jobs graph look prettier :P

```suggestion
    needs: ["nox", "nox-cross-arch-all", "test-installation"]
```

.github/workflows/ci.yaml
```diff
@@ -79,6 +79,114 @@ jobs:
...
+  # This job runs if `nox-cross-arch` ran and succeeded.
+  # This is required because, when the `nox-cross-arch` job is skipped, its
+  # inner matrix is not expanded, and branch protection rules on the
+  # inner-matrix jobs get stuck. So instead of `nox-cross-arch`, this job can be
+  # used in the branch protection rules.
+  # At the time of writing this, there is an ongoing discussion about this here:
+  # https://github.com/orgs/community/discussions/9141
+  nox-cross-arch-all:
+    name: Cross-arch tests with nox
+    needs: ["nox-cross-arch"]
```
llucax, line 184:
```suggestion
    needs: ["nox-cross-arch"]
    runs-on: ubuntu-20.04
```

tiyash-basu-frequenz:
> Jobs are not being created, it seems like something is wrong:
> 
> > Invalid workflow file: .github/workflows/ci.yaml#L183
> > The workflow is not valid. .github/workflows/ci.yaml (Line: 183, Col: 5): Required property is missing: runs-on

Oops that looks like a bug. Checking.

tiyash-basu-frequenz:
Fixed.

llucax (approved):

PR #623, opened 2023-08-30
Add more formula tests for min and max functions

matthias-wende-frequenz:
Also we make the interface public as it's considered ready to use.

### DIFF
--- a/tests/timeseries/_formula_engine/test_formula_composition.py
+++ b/tests/timeseries/_formula_engine/test_formula_composition.py
@@ -171,18 +171,68 @@ class TestFormulaComposition:
         assert count == 10
 
     async def test_formula_composition_min_max(self, mocker: MockerFixture) -> None:
-        """Test the composition of formulas with min/max values."""
+        """Test the composition of formulas with the min and max."""
+        mockgrid = MockMicrogrid(grid_meter=True)
+        mockgrid.add_chps(1)
+        await mockgrid.start(mocker)
+
+        logical_meter = microgrid.logical_meter()
+        engine_min = logical_meter.grid_power.min(logical_meter.chp_power).build(
+            "grid_power_min"
+        )
+        engine_min_rx = engine_min.new_receiver()
+        engine_max = logical_meter.grid_power.max(logical_meter.chp_power).build(
+            "grid_power_max"
+        )
+        engine_max_rx = engine_max.new_receiver()
+
+        await mockgrid.mock_resampler.send_meter_power([100.0, 200.0])
+
+        # Test min
+        min_pow = await engine_min_rx.receive()
+        assert (
+            min_pow and min_pow.value and min_pow.value.isclose(Power.from_watts(100.0))
+        )
+
+        # Test max
+        max_pow = await engine_max_rx.receive()
+        assert (
+            max_pow and max_pow.value and max_pow.value.isclose(Power.from_watts(200.0))
+        )
+
+        await mockgrid.mock_resampler.send_meter_power([-100.0, -200.0])
+
+        # Test min
+        min_pow = await engine_min_rx.receive()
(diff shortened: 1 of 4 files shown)


shsms:
In the tests, could you add additional inputs that test both sides of max/min for each case?  Looks like all cases are just testing just one side.

It would be nice to have some `-`, `/` as well,  because those are the trickier parts of formulas.  

Also it would be nice to test cases that have basic operations inside and to the right of min/max.  Like: `a - b.min(c / d) - (c - e)`, etc.

shsms (dismissed):
I did some tests to make sure that this works in strange cases  and I'm now convinced that this works.  So I have no objections to taking this in as it is now.

tests/timeseries/_formula_engine/test_formula_composition.py
```diff
@@ -171,16 +171,70 @@ async def test_formula_composition_missing_bat(self, mocker: MockerFixture) -> N
...
     async def test_formula_composition_min_max(self, mocker: MockerFixture) -> None:
-        """Test the composition of formulas with min/max values."""
+        """Test the composition of formulas with the min and max"""
         mockgrid = MockMicrogrid(grid_meter=True)
+        mockgrid.add_chps(1)
         await mockgrid.start(mocker)
 
         logical_meter = microgrid.logical_meter()
-        engine_min = logical_meter.grid_power._min(  # pylint: disable=protected-access
+        engine_min = logical_meter.grid_power.min(  # pylint: disable=protected-access
```
daniel-zullo-frequenz: do you know why this still needs disable protected-access?

daniel-zullo-frequenz (commented):
Only a question/comment to check for and LGTM

tests/timeseries/_formula_engine/test_formula_composition.py
matthias-wende-frequenz: I'm pretty sure it's not needed. Good catch!

tests/timeseries/_formula_engine/test_formula_composition.py
```diff
@@ -171,16 +171,70 @@ async def test_formula_composition_missing_bat(self, mocker: MockerFixture) -> N
...
+        mockgrid = MockMicrogrid(grid_meter=True)
+        await mockgrid.start(mocker)
+
+        logical_meter = microgrid.logical_meter()
+        engine_min = logical_meter.grid_power.min(  # pylint: disable=protected-access
             Power.zero()
         ).build("grid_power_min")
         engine_min_rx = engine_min.new_receiver()
-        engine_max = logical_meter.grid_power._max(  # pylint: disable=protected-access
+        engine_max = logical_meter.grid_power.max(  # pylint: disable=protected-access
```
daniel-zullo-frequenz: Then you might want to remove this one too (if it works)
```suggestion
        engine_max = logical_meter.grid_power.max(
```

tests/timeseries/_formula_engine/test_formula_composition.py
```diff
@@ -171,16 +171,70 @@ async def test_formula_composition_missing_bat(self, mocker: MockerFixture) -> N
...
+
+    async def test_formula_composition_min_max_const(
+        self, mocker: MockerFixture
+    ) -> None:
+        """Test the compositing formulas and constants with the min and max functions."""
+        mockgrid = MockMicrogrid(grid_meter=True)
+        await mockgrid.start(mocker)
+
+        logical_meter = microgrid.logical_meter()
+        engine_min = logical_meter.grid_power.min(  # pylint: disable=protected-access
```
daniel-zullo-frequenz:
```suggestion
        engine_min = logical_meter.grid_power.min(
```

daniel-zullo-frequenz (approved):

PR #638, opened 2023-09-04
Fixes on copy behavior in ring buffer window method

cwasicki:
This PR introduces several copy-related updates to the ring buffer window functionality

Changes:
* Default Copy Behavior: Changed the default behavior to copy data in the ring buffer window. This is to prioritize data integrity over performance, which is a minor concern for most expected use-cases. Made `force_copy` as a keyword argument.
* Moved the logic for extracting wrapped buffers into a separate method to fix a bug and allow for better testing and future maintainability.
* Handling None Values: Fixed an issue where None values were being forcibly copied in the ring buffer window. The decision to enforce a copy is now left to the user, even when the data contains None values.
* Test Coverage: Added tests for the window method in the ring buffer, including tests for expected copy behavior on missing values.
* Fix to return an empty array in case start = end in ring buffer window.
* Expose restricted ring buffer window by moving window.

Part of https://github.com/frequenz-floss/frequenz-sdk-python/issues/214

### DIFF
--- a/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
+++ b/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
@@ -292,6 +292,9 @@ class OrderedRingBuffer(Generic[FloatArray]):
                 f"end parameter {end} has to predate start parameter {start}"
             )
 
+        if start == end:
+            return np.array([]) if isinstance(self._buffer, np.ndarray) else []
+
         start_index = self.datetime_to_index(start)
         end_index = self.datetime_to_index(end)
 
--- a/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
+++ b/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
@@ -301,7 +301,10 @@ class OrderedRingBuffer(Generic[FloatArray]):
 
     @staticmethod
     def _wrapped_buffer_window(
-        buffer: FloatArray, start_pos: int, end_pos: int, force_copy: bool = True
+        buffer: FloatArray,
+        start_pos: int,
+        end_pos: int,
+        force_copy: bool = True,
     ) -> FloatArray:
         """Get a wrapped window from the given buffer.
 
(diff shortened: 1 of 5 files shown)


cwasicki:
Rebased @matthias-wende-frequenz

cwasicki:
@matthias-wende-frequenz Ready for review

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -295,22 +295,43 @@ def window(
         start_index = self.datetime_to_index(start)
         end_index = self.datetime_to_index(end)
 
+        return self._wrapped_buffer_window(
+            self._buffer, start_index, end_index, force_copy
+        )
+
+    @staticmethod
+    def _wrapped_buffer_window(
```
matthias-wende-frequenz, line 306: Does this have to be a static method? I'd suggest to make it a class method as long it's not used outside the class context.

tests/timeseries/test_moving_window.py
```diff
@@ -66,6 +66,18 @@ def init_moving_window(
...
+    Args:
+        i: Index to create datetime from.
+
+    Returns:
+        Datetime object.
+    """
+    return datetime.fromtimestamp(i, tz=timezone.utc)
+
+
 async def test_access_window_by_index() -> None:
```
matthias-wende-frequenz, line 81: I wonder if that's the correct function to add the tests. This was supposed to access the window by integer index, but the tests are accessing by datetime object.

matthias-wende-frequenz (commented):
Just two small comments. Looks good otherwise!
Also needs a rebase.

tests/timeseries/test_moving_window.py
cwasicki: The newly added datetime slice tests were added in `test_access_window_by_ts_slice`. Then a single failure test to for int slices in `test_access_window_by_int_slice`

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
cwasicki: Why do you think so, or what is the advantage of a classmethod? I made it static since it's an isolated algorithmic piece which does not need any internal state of the class. Only the buffer argument would change here. 

The main motivation was testing purpose (actually contained multiple bugs before), i.e. separate wrapping algorithm from index logic (which will become more tricky in follow-ups). But it can also easily be stripped out when needed elsewhere.

cwasicki: One example for outside usage I could imagine is the periodic feature extractor, which I assume is using similar code.

matthias-wende-frequenz: You are right. It doesn't depend on the internal state. Let's keep it as is.

tests/timeseries/test_moving_window.py
matthias-wende-frequenz: Sorry I missed that. All good then ;).

matthias-wende-frequenz (approved):

PR #639, opened 2023-09-05
Move "fake_time" fixture to conftest.py

### LINKED ISSUE #223: Create a util module in tests to share common test functionality
### What's needed?

The following function is duplicated in a couple of test modules

```python
@pytest.fixture
def fake_time() -> Iterator[time_machine.Coordinates]:
    """Replace real time with a time machine that doesn't automatically tick."""
    with time_machine.travel(0, tick=False) as traveller:
        yield traveller
```

### Proposed solution

Move that common/shared functionality into a new module so tests that needed can use it rather than duplicating it.

### Use cases

This is just to avoid code redundancy/duplication in tests.

_No response_

### Additional context

_No response_

stefan-brus-frequenz:
"conftest.py" contains fixtures that are shared among multiple test cases.

Fixes #223

### DIFF
--- a/tests/actor/test_resampling.py
+++ b/tests/actor/test_resampling.py
@@ -35,13 +35,6 @@ def event_loop() -> Iterator[async_solipsism.EventLoop]:
     loop.close()
 
 
-@pytest.fixture
-def fake_time() -> Iterator[time_machine.Coordinates]:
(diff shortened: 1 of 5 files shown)


llucax (approved):

llucax:
If there are no further comments about this, I will queue for merge.

PR #641, opened 2023-09-05
Treat NaN as gap in ring buffer

### LINKED ISSUE #601: Treat NaN as gap in ring buffer
### What's needed?

Currently a gap is defined as one or multiple consecutive samples with `None` values. If the value is `NaN` or `inf`, these are not treated as gaps. 

### Proposed solution

Include `NaN` in the definition of a gap.

### Use cases

Using formula engine, `NaN` or `inf` are converted to `None` (see https://github.com/frequenz-floss/frequenz-sdk-python/blob/cac547386533e7f8fbf0c08ffc5ce5d522b1a011/src/frequenz/sdk/timeseries/_formula_engine/_formula_evaluator.py#L130-L131). In this case, which I assume covers most of our current applications, the behavior is not altered.

When data from the microgrid API is directly injected into the moving window, this behavior is not ensured and would be inconsistent. Also if data from other sources (e.g. trading data) is injected, `NaN` would not be treated in the gap. 

* Leave it to the user to check.
* Convert `NaN` to `None` when adding to the ring buffer.

### Additional context

There was the discussion what to do with `inf` values. To keep it consistent with the logical meter these would also have to be converted. The same is even true for `NaN`, since they arise in different situations than `None`s. However, `inf` values could be reasonable in certain context and transfer more information over `None` than `NaN`s do.

cwasicki:
Fixes https://github.com/frequenz-floss/frequenz-sdk-python/issues/601

### DIFF
--- a/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
+++ b/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
@@ -108,6 +108,17 @@ class OrderedRingBuffer(Generic[FloatArray]):
         """
         return self._gaps
 
+    def has_value(self, sample: Sample[QuantityT]) -> bool:
+        """Check if a sample has a value and it's not NaN.
+
+        Args:
+            sample: sample to check.
+
+        Returns:
+            True if the sample has a value and it's not NaN.
+        """
+        return not (sample.value is None or sample.value.isnan())
+
     @property
     def maxlen(self) -> int:
         """Get the max length.
@@ -154,10 +165,14 @@ class OrderedRingBuffer(Generic[FloatArray]):
         )
 
         # Update data
-        value: float = np.nan if sample.value is None else sample.value.base_value
+        if self.has_value(sample):
+            assert sample.value is not None
+            value = sample.value.base_value
+        else:
+            value = np.nan
         self._buffer[self.datetime_to_index(timestamp)] = value
 
-        self._update_gaps(timestamp, prev_newest, sample.value is None)
+        self._update_gaps(timestamp, prev_newest, not self.has_value(sample))
 
     @property
     def time_bound_oldest(self) -> datetime:
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -20,6 +20,7 @@
 
 - A new class `Fuse` has been added to represent fuses. This class has a member variable `max_current` which represents the maximum current that can course through the fuse. If the current flowing through a fuse is greater than this limit, then the fuse will break the circuit.
 
+- NaN values are treated as missing when gaps are determined in the `OrderedRingBuffer`.
 
 ## Bug Fixes
 
--- a/tests/timeseries/test_ringbuffer.py
+++ b/tests/timeseries/test_ringbuffer.py
@@ -192,6 +192,67 @@ def test_timestamp_ringbuffer_missing_parameter(
     assert len(buffer.gaps) == 1
 
 
+def dt(i: int) -> datetime:  # pylint: disable=invalid-name
+    """Create datetime objects from indices.
+
+    Args:
+        i: Index to create datetime from.
+
+    Returns:
+        Datetime object.
+    """
+    return datetime.fromtimestamp(i, tz=timezone.utc)
+
+
+def test_gaps() -> None:
+    """Test gap treatment in ordered ring buffer."""
+    buffer = OrderedRingBuffer([0.0] * 5, ONE_SECOND)
+    assert len(buffer) == 0
+    assert len(buffer.gaps) == 0
+
+    buffer.update(Sample(dt(0), Quantity(0)))
+    assert len(buffer) == 1
+    assert len(buffer.gaps) == 1
+
+    buffer.update(Sample(dt(6), Quantity(0)))
+    assert len(buffer) == 1
+    assert len(buffer.gaps) == 1
+
+    buffer.update(Sample(dt(2), Quantity(2)))
+    buffer.update(Sample(dt(3), Quantity(3)))
+    buffer.update(Sample(dt(4), Quantity(4)))
+    assert len(buffer) == 4
+    assert len(buffer.gaps) == 1
+
+    buffer.update(Sample(dt(3), None))
+    assert len(buffer) == 3
+    assert len(buffer.gaps) == 2
+
+    buffer.update(Sample(dt(3), Quantity(np.nan)))
+    assert len(buffer) == 4  # should be 3
+    assert len(buffer.gaps) == 1  # should be 2
+
+    buffer.update(Sample(dt(2), Quantity(np.nan)))
+    assert len(buffer) == 4  # should be 2
+    assert len(buffer.gaps) == 1  # should be 2
+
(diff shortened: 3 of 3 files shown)


matthias-wende-frequenz (dismissed):

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -108,6 +108,17 @@ def gaps(self) -> List[Gap]:
         """
         return self._gaps
 
+    def is_missing_value(self, value: QuantityT | None) -> bool:
```
llucax: Nitpick:
```suggestion
    def is_value_missing(self, value: QuantityT | None) -> bool:
```

Or simply `is_missing()`. If you take a `Sample` instead of a `QuantityT`, I think `is_value_missing()` makes more sense.

llucax: Also, using a negative as part of a function name is usually not recommended, because you can end up with a double negation, which is harder to reason about. What about:

```suggestion
    def has_value(self, sample: Sample) -> bool:
```

But I'm not 100% sure in this case, because I guess here *missing* is important, as you are talking about gaps. Another alternative could be `is_value_present(sample)`. So just food for thought, I'm not convinced either way :P

Also, infinite is considered as not missing, right?

llucax (dismissed):
Just a couple of bike shedding naming comments, but LGTM in general!

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
cwasicki: I can make it taking a sample instead. Do you think it can be moved to a method of `Sample` at some point?

Will change it to `has_value`. There is already a `is_missing`, which is about timestamps. I guess another candidate to improve naming.

cwasicki: > Also, infinite is considered as not missing, right?

So far yes, on the typical microgrid data we work with infinities will already be turned into NaNs and for other use-cases I did not want to restrict it too much.

matthias-wende-frequenz (dismissed):

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
llucax: > I can make it taking a sample instead. Do you think it can be moved to a method of Sample at some point?

I'm not sure, because again what's a present value could depend on the application, for example is `infinite` a value or not? I would be in favor of adding more common methods from `math` that would also check for `None`, like `is_finite()`, but that is `not isnan() and not isinf()`.

> There is already a `is_missing`, which is about timestamps.  I guess another candidate to improve naming.

Oh! Yes :)

llucax: Another alternative could be `is_gap()` which ties it up pretty well with the concept.

But I was thinking if maybe `is_missing(timestamp)` can be unified into a `is_gap(sample)`, then `update()` could already pass if a timestamp is missing to `self._update_gaps()`. But maybe this is too much of a refactor to be on the scope of this PR.

cwasicki:
@llucax Updated

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
cwasicki: Not sure about `is_gap(sample)`: `is_missing` checks whether a sample is already in our gap list, `has_value` checks whether a sample can be considered a gap.

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -108,6 +108,17 @@ def gaps(self) -> List[Gap]:
...
+    def has_value(self, sample: Sample[QuantityT]) -> bool:
+        """Check if a sample has a value and it's not NaN.
+
+        Args:
+            sample: sample to check.
+
+        Returns:
+            True if the sample has a value and it's not NaN.
+        """
+        return not ((sample.value is None) or sample.value.isnan())
```
llucax: If you want to save some parenthesis (which might make it easier to read):

```suggestion
        return sample.value is not None and not sample.value.isnan()
```

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -154,10 +165,14 @@ def update(self, sample: Sample[QuantityT]) -> None:
         )
 
         # Update data
-        value: float = np.nan if sample.value is None else sample.value.base_value
+        if self.has_value(sample):
+            assert sample.value is not None
```
llucax, line 169: I just noticed this, too bad that this is needed, because it was just checked by `has_value()`.

You could avoid this by making `has_value` a function instead and using `TypeGuard`, but that would also need to take it back to operate on `float | None`:

```py
from typing import TypeGuard

def has_value(sample: float | None) -> TypeGuard[float]:
    """Check if a sample has a value and it's not NaN.
    Args:
        sample: sample to check.
    Returns:
        True if the sample has a value and it's not NaN.
    """
    return value is not None and not value.isnan()

if has_value(sample):  # If this returns True, then the type checker narrows the type to `float`
    value = sample.value.base_value  # this is now fine without the assert
else:
    value = np.nan
```

I thought about alternatives and all seem awful, so I guess we'll have to live with the assert :shrug:

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -154,10 +165,14 @@ def update(self, sample: Sample[QuantityT]) -> None:
...
+            value = sample.value.base_value
+        else:
+            value = np.nan
```
llucax, line 172: Or.... :exclamation::exclamation::exclamation: just not use the function in this context... The original code is pretty simple and understandable (and works the same, as if it is NaN we'll use `np.nan` anyway, or is it `np.nan` different from `float('nan')`?) without it too:

```py
        valueL = np.nan if sample.value is None else sample.value
        self._buffer[self.datetime_to_index(timestamp)] = value
        self._update_gaps(timestamp, prev_newest, not self.has_value(sample))
```

llucax (dismissed):
All optional comments, LGTM as it is too.

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
cwasicki: I would like to leave it as it is required in both these places. So you have one location where you define what a missing value is and what not (e.g. if we want to change it to include `inf`s at some point.)

Actually I find it not so bad, even more readable. But yea, the assertion is annoying.

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
cwasicki: Was thinking about that and found it even less readable. But removed one level.

cwasicki:
@llucax @matthias-wende-frequenz Updated, but I am not authorized to merge.

matthias-wende-frequenz (approved):

PR #643, opened 2023-09-06
Add duration to gaps and expose gaps in moving window

### LINKED ISSUE #558: Expose gaps in moving window
### What's needed?

Implement method to return the maximum data gap in a `MovingWindow` given timestamp or index ranges representing a window size

### Proposed solution

Compute the maximum data gap size iterating through the `MovingWindow` data gaps (underlying buffer)  for the given timestamp or index ranges representing a window size

### Use cases

Actors like peak-shaving needs this functionality to avoid implementing it on their ends.

### Update
Note the scope of the issue was updated as follows:
The `MovingWindow` will provide access to a list of all gaps in a defined subset of the moving window. Applications can get the maximum gap or the sum of gaps by using something like `max(g.duration for g in gaps) or sum(...)`, respectively.

cwasicki:
Still WIP

Fixes https://github.com/frequenz-floss/frequenz-sdk-python/issues/558

### DIFF
--- a/src/frequenz/sdk/timeseries/_moving_window.py
+++ b/src/frequenz/sdk/timeseries/_moving_window.py
@@ -246,6 +246,15 @@ class MovingWindow(BackgroundService):
             asyncio.create_task(self._resampler.resample(), name="resample")
         )
 
+    def gaps(self) -> list[Gap]:
+        """
+        Return a list of gaps in the `MovingWindow`.
+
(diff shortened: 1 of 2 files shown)


cwasicki:
Not needed anymore (https://github.com/frequenz-floss/frequenz-sdk-python/issues/558#issuecomment-1743187358)

PR #647, opened 2023-09-07
Add a glossary to the documentation

### LINKED ISSUE #142: Move the glossary to the documentation
### What's needed?

Now that we are generating the documentation as a website we should include the glossary there.

### Proposed solution

Move https://github.com/frequenz-floss/frequenz-sdk-python/wiki/Glossary to `docs/glossary.md` or similar. It should probably be a top level page so it appears in the navigation bar on the top.

### Use cases

_No response_

_No response_

### Additional context

_No response_

llucax:
Based on https://github.com/frequenz-floss/frequenz-sdk-python/wiki/Glossary, with a lot of improvements and additions.

Fixes #142.


llucax:
After this is approved, I will make the glossary text fit 80 columns to match other documents. I forgot to do it before any other changes and now it would mean a lot of work to change it. Avoiding adding the final commit now because if there are correction and I need to rebase, the rebasing also breaks the 80 columns fitting, so better to do it right before merging, as there are no content changes there so there should be nothing else to be changed.

docs/intro/glossary.md
```diff
@@ -0,0 +1,255 @@
...
+
+### PV
+
+Photovoltaic. See the [Wikipedia article](https://en.wikipedia.org/wiki/Photovoltaics) for more details.
+
+In the SDK it is normally used as a synonym for [solar panel](#solar-panel).
+
+### SoC
+
+[State of charge](#state-of-charge). See the [Wikipedia article](https://en.wikipedia.org/wiki/State_of_charge) for more details.
```
tiyash-basu-frequenz, line 40: I'd suggest adding a line about what it does (The amount of energy left in a storage w.r.t it's full capacity).

docs/intro/glossary.md
```diff
@@ -0,0 +1,255 @@
...
+
+### SoP
+
+[State of power](#state-of-power).
```
tiyash-basu-frequenz, line 44: I'd suggest adding a line about what it does (Upper limits on charge and discharge power).

docs/intro/glossary.md
```diff
@@ -0,0 +1,255 @@
...
+
+## Microgrid
+
+A local electrical grid that connects a set of different [types of components](#component-category) together. It can be connected to the main grid, or be completely isolated.
```
tiyash-basu-frequenz: `main grid` => `public grid`?

tiyash-basu-frequenz: `or be completely isolated.` => `or be completely isolated, in which case it is known as an island.`

docs/intro/glossary.md
```diff
@@ -0,0 +1,255 @@
...
+
+### Component Category
+
+The different types of [components](#component) a [microgrid](#microgrid) can have, such as an inverter, battery, meter, etc.
```
tiyash-basu-frequenz: `type` is a bit tricky to use here, because in our terminology, types refer to the level-2 hierarchy. E.g., ComponentCategory could be Inverter, and InverterType could be PV.

Maybe we rephrase it as `The category a component can belong to, such as an inverter, ...`

docs/intro/glossary.md
```diff
@@ -0,0 +1,255 @@
...
+
+Aggregates [components](#component) not covered above. Under typical circumstances, this corresponds to the site's [gross consumption](#gross-consumption), excluding active parts and the [battery](#battery). The term 'consumer' implies that this component predominantly consumes electric power. It can support negative values for exotic site topologies or short-term effects.
+
+### `{component}_consumption_power`
+
+Positive power value if `{component}_power` is positive, otherwise zero.
+
+### `{component}_production_power`
+
+Positive power value if `{component}_power` is negative, otherwise zero.
```
llucax: Specially for @cwasicki as you wrote this:

I don't think these belong in here to be honest, but I didn't know what to do with them.

Eventually we might need to define all important metrics in the **Metrics** section, maybe even per component category (or mentioning which component category uses that metric). But defining symbols on the SDK code IMHO should be only done in the auto-generated API documentation, otherwise things will get outdated and out of sync very quickly.

Maybe we can still have in here the **Microgrid Power Terminology** text but structure it differently, as definitions, like defining what do we understand with **Consumption** and **Production** and its relationship with **passive sign convention** (we should define that too :D). I might give this a try if you agree.

Also cc @matthias-wende-frequenz.

docs/intro/glossary.md
```diff
@@ -0,0 +1,255 @@
...
+
+#### Converter
+
+A [DC](#dc)-DC converter.
+
+#### EV Charger
+
+A station for charging [EVs](#ev).
+
+#### Electrolizer
```
tiyash-basu-frequenz:
```suggestion
#### Electrolyzer
```

docs/intro/glossary.md
```diff
@@ -0,0 +1,255 @@
...
+
+An device that converts water into hydrogen and oxygen.
+
+#### Grid
+
+A point where the local [microgrid](#microgrid) is connected to the main electricity grid.
```
tiyash-basu-frequenz: `Grid` actually refers to the public electricity infrastructure. Maybe worth pointing out the distinction here.

docs/intro/glossary.md
```diff
@@ -0,0 +1,255 @@
...
+
+This glossary provides definitions for common terminology used in the SDK,
+focusing on microgrid components, metrics, measurements, and power-related
+terms.
+
+## Common Acronyms
+
+### AC
+
+Alternating current. See the [Wikipedia article](https://en.wikipedia.org/wiki/Alternating_current) for more details.
```
tiyash-basu-frequenz, line 10: Maybe also worth adding active, reactive, and apparent power below this?

docs/intro/glossary.md
```diff
@@ -0,0 +1,255 @@
...
+
+An individual numeric value obtained from a [metric](#metric) of a [component](#component) instance. It is consistently of type `float`, but it is often expressed in specific units.
+
+In the context of a sample, this is commonly referred to as a *sample value*.
+
+For example, a measurement of the capacity of a battery with component ID 5 can be **400**, typically measured in Watt-hours (Wh).
+
+### Timestamp
+
+A specific point in time, always represented as a `datetime` with a `timezone` attached.
```
tiyash-basu-frequenz, line 156: Since these are unix timestamps, maybe it's worth mentioning that it is counted from 01.01.1970.

docs/intro/glossary.md
```diff
@@ -0,0 +1,255 @@
...
+
+In [microgrid](#microgrid) context sometimes used as the remaining difference between the [net consumption](#net-consumption) and the [battery](#battery) power, i.e. what we define as [grid power](#grid_power).
+
+### Residual Load
+
+Same as [residual consumption](#residual-consumption).
+
+### State of Charge
+
+The level of charge of a [battery](#battery) relative to its capacity, expressed in percentage points. Calculated as the ratio between the remaining energy in the battery at a given time and the maximum possible energy under similar health conditions. [Source](https://epicpower.es/wp-content/uploads/2020/08/AN028_SoC-SoH-SoP-definitions_v3.pdf)
```
tiyash-basu-frequenz, line 291: This whole text could move above, where `SoC` is mentioned earlier, and here we could just add a link.

docs/intro/glossary.md
```diff
@@ -0,0 +1,255 @@
...
+
+### State of Power
+
+The ratio of peak power to nominal power. Peak power is the maximum power that can be sustained for a specific duration without violating preset operational design limits on battery voltage, SsC, power, or current.
```
tiyash-basu-frequenz: This whole text could move above, where `SoP` is mentioned earlier, and here we could just add a link.

docs/intro/glossary.md
```diff
@@ -0,0 +1,255 @@
...
+
+Can be positive (indicating [battery](#battery) charging) or negative (indicating battery discharging). This also applies to [EV charging](#ev-charger) stations.
+
+### `{pv,chp,wind}_power`
+
+Pertaining to active [components](#component), this represents power generated from local (renewable) sources. It is negative when electricity is generated, otherwise zero or positive in case of self-consumption. Alternative terms include generation, supply, and source.
+
+### `consumer_power`
+
+Aggregates [components](#component) not covered above. Under typical circumstances, this corresponds to the site's [gross consumption](#gross-consumption), excluding active parts and the [battery](#battery). The term 'consumer' implies that this component predominantly consumes electric power. It can support negative values for exotic site topologies or short-term effects.
```
tiyash-basu-frequenz: `Aggregates` => `Covers`? (aggregate could mean something else if the reader is a data scientist, e.g.).

tiyash-basu-frequenz: Nitpick: `short-term effects.` => `short durations.`

tiyash-basu-frequenz (commented):
A few suggestions, but looks good in general. Thanks for the effort! Maybe this evolves into a company-wide glossary in the future.

docs/intro/glossary.md
tiyash-basu-frequenz: I find `*_power` a bit redundant in this format. We already mention Grid and Net Consumption power above. These terms could go in that section, agree.

tiyash-basu-frequenz: > passive sign convention (we should define that too :D)

Yes!

docs/intro/glossary.md
llucax: We should definitely add them, but these should go to the `Metrics` section I would say, this section is about acronyms.

docs/intro/glossary.md
llucax: Have you seen the definition of "State of charge"? Again, the pure definition (expansion) of acronyms is separate from the definition of metrics. If what's in the metrics section doesn't cover it, could you suggest an exact text to add?

docs/intro/glossary.md
llucax: Also probably in the metrics section, right? If what's in the metrics section doesn't cover it, could you suggest an exact text to add?

docs/intro/glossary.md
llucax: Ah, yeah, I was struggling with that, at first I used "wide" grid, so "main" grid (which is present in the microgrid API) was an improvement, but "public" grid sounds better :D

docs/intro/glossary.md
tiyash-basu-frequenz: I missed earlier that we are just mentioning acronyms in this section.

docs/intro/glossary.md
tiyash-basu-frequenz: ~I missed earlier that we are just mentioning acronyms in this section.~ Sorry wrong thread, please ignore.

llucax: I also added a definition for "Island", as I guess it might not be obvious if read out of context.

docs/intro/glossary.md
```diff
@@ -0,0 +1,237 @@
...
+
+The level of charge of a [battery](#battery) relative to its capacity, expressed in percentage points. Calculated as the ratio between the remaining energy in the battery at a given time and the maximum possible energy under similar health conditions. [Source](https://epicpower.es/wp-content/uploads/2020/08/AN028_SoC-SoH-SoP-definitions_v3.pdf)
+
+### State of Power
+
+The ratio of peak power to nominal power. Peak power is the maximum power that can be sustained for a specific duration without violating preset operational design limits on battery voltage, SsC, power, or current.
+
+This indicator is crucial to ensure that charge or discharge power remains within specific limits, optimizing the battery's lifespan. It is particularly useful in peak power applications to define battery conditions for substantial charges or discharges.
+
+The state of power depends on the state of charge, battery capacity, initial characteristics, chemistry, and battery voltage. [Source](https://epicpower.es/wp-content/uploads/2020/08/AN028_SoC-SoH-SoP-definitions_v3.pdf)
```
daniel-zullo-frequenz: I think external factors like temperature should be also mentioned, as they can significantly impact the state of power

docs/intro/index.md
```diff
@@ -0,0 +1,14 @@
+# Introduction
+
+There are several fundamental concepts that one should comprehend before
+diving into the Frequenz SDK. These concepts encompass both general notions
+within the energy sector or microgrids and specific details regarding the SDK
+itself, such as asynchronous programming, use of channels for communication,
+the actor model, data flow, and the composition of a component graph.
```
daniel-zullo-frequenz: Optionally:
```suggestion
the actor model, data pipelines, and the composition of a components graph.
```

docs/intro/glossary.md
llucax: OK, makes sense. I also removed the "such as", as we have the full list just below this now.

docs/intro/glossary.md
```diff
@@ -0,0 +1,237 @@
...
+
+Alternating current. See the [Wikipedia article](https://en.wikipedia.org/wiki/Alternating_current) for more details.
+
+### BMS
+
+Battery management system. See the [Wikipedia article](https://en.wikipedia.org/wiki/Battery_management_system) for more details.
+
+### CHP
+
+Combined heat and power. See the [Wikipedia article](https://en.wikipedia.org/wiki/Combined_heat_and_power) for more details.
```
daniel-zullo-frequenz, line 18: Just a global and optional suggestion to consider, I'd expand the acronyms by capitalizing the words that match the acronym stands for, e.g.
```suggestion
Combined Heat and Power. See the [Wikipedia article](https://en.wikipedia.org/wiki/Combined_heat_and_power) for more details.
```

docs/intro/glossary.md
llucax: This is taken from the common API. It's even in the specs (`COMPONENT_CATEGORY_ELECTROLYZER`).  :grimacing: 

Doing a quick google search it seems like Electrolyzer is correct. :thinking:

docs/intro/glossary.md
llucax: This was also taken from the common API docs. I now replaced `main` with `public`, do you think that's enough or do you think it needs further clarification?

docs/intro/glossary.md
```diff
@@ -0,0 +1,237 @@
...
+
+A storage system for electrical energy.
+
+#### CHP Plant
+
+A generator that produces combined heat and power ([CHP](#chp)). Usually powered via combustion of some form of fuel.
+
+#### Converter
+
+A [DC](#dc)-DC converter.
```
daniel-zullo-frequenz: hmm DC to DC converter? :thinking: 
To avoid confusion I'd suggest to write it as follows:
```suggestion
A [DC](#dc) converter.
```

docs/intro/glossary.md
llucax: Not in the SDK, in the SDK they are `datetime`. I think being a timestamp is a microgrid API implementation detail, isn't it? In any case I think this also doesn't belong to the glossary (arguably not even the Python type is relevant here), it belongs to the code documentation.

docs/intro/glossary.md
```diff
@@ -0,0 +1,237 @@
...
+
+#### EV Charger
+
+A station for charging [EVs](#ev).
+
+#### Electrolizer
+
+An device that converts water into hydrogen and oxygen.
```
daniel-zullo-frequenz:
```suggestion
A device that converts water into hydrogen and oxygen.
```

docs/intro/glossary.md
llucax: I prefer to keep acronyms just as an expansion for what they stand for.

docs/intro/glossary.md
llucax: Same. I will close this comment to avoid duplicated discussions. If we decide to change it, then I can update both.

docs/intro/glossary.md
tiyash-basu-frequenz: I missed earlier that we are just mentioning acronyms in this section.

docs/intro/glossary.md
```diff
@@ -0,0 +1,237 @@
...
+
+### Component
+
+A device (of a particular [category](#component-category)) within a [microgrid](#microgrid).
+
+### Component ID
+
+A numeric identifier uniquely representing an instance of a [component](#component) in a particular[microgrid](#microgrid). It is always of type `int`.
+
+For example, a battery an have a component ID **5**.
```
daniel-zullo-frequenz: Maybe:
```suggestion
For example, a battery has a component ID **5**.
```

docs/intro/glossary.md
```diff
@@ -0,0 +1,237 @@
...
+
+### SoC
+
+[State of charge](#state-of-charge). See the [Wikipedia article](https://en.wikipedia.org/wiki/State_of_charge) for more details.
+
+### SoP
+
+[State of power](#state-of-power).
+
+## Microgrid
```
tiyash-basu-frequenz, line 46: `Microgrid` is not an acronym. There should be a different header above this line, maybe `Commonly Used Terms`.

docs/intro/glossary.md
llucax: Done.

docs/intro/glossary.md
llucax: Can you make a code suggestion?

llucax:
Addressed @tiyash-basu-frequenz comments. Will do @daniel-zullo-frequenz tomorrow, but I think I will start accumulating new commits, because rebasing is just too much work for plain text. :(

docs/intro/glossary.md
```diff
@@ -0,0 +1,241 @@
...
+
+A numeric identifier uniquely representing an instance of a [component](#component) in a particular[microgrid](#microgrid). It is always of type `int`.
+
+For example, a battery an have a component ID **5**.
+
+### Island
+
+A [microgrid](#microgrid) that is not connected to the public electricity [grid](#grid).
+
+## Component Data
```
tiyash-basu-frequenz, line 138: Does it make sense to categorise non-acronyms further? I think it would be easier to find words if it were just an alphabetically ordered list.

docs/intro/glossary.md
tiyash-basu-frequenz: I assume you meant `can have`?

docs/intro/glossary.md
```diff
@@ -0,0 +1,241 @@
...
+
+Direct current. See the [Wikipedia article](https://en.wikipedia.org/wiki/Direct_current) for more details.
+
+### EV
+
+Electric vehicle. See the [Wikipedia article](https://en.wikipedia.org/wiki/Electric_vehicle) for more details.
+
+### PSC
+
+Passive sign convention. See the [Wikipedia article](https://en.wikipedia.org/wiki/Passive_sign_convention) for more details.
```
tiyash-basu-frequenz, line 30: We should add "Passive sign convention" to the glossary too.

docs/intro/glossary.md
daniel-zullo-frequenz:
```suggestion
The state of power depends on the state of charge, battery capacity, initial characteristics, chemistry, and battery voltage, as well as external factors like temperature and humidity, which can also have a significant impact on it. [Source](https://epicpower.es/wp-content/uploads/2020/08/AN028_SoC-SoH-SoP-definitions_v3.pdf)
```

docs/intro/glossary.md
```diff
@@ -0,0 +1,241 @@
...
+
+This glossary provides definitions for common terminology used in the SDK,
+focusing on microgrid components, metrics, measurements, and power-related
+terms.
+
+## Common Acronyms
+
+### AC
+
+Alternating current. See the [Wikipedia article](https://en.wikipedia.org/wiki/Alternating_current) for more details.
```
daniel-zullo-frequenz, line 10:
```suggestion
Alternating Current. See the [Wikipedia article](https://en.wikipedia.org/wiki/Alternating_current) for more details.
```

docs/intro/glossary.md
```diff
@@ -0,0 +1,241 @@
...
+
+### BMS
+
+Battery management system. See the [Wikipedia article](https://en.wikipedia.org/wiki/Battery_management_system) for more details.
```
daniel-zullo-frequenz, line 14:
```suggestion
Battery Management System. See the [Wikipedia article](https://en.wikipedia.org/wiki/Battery_management_system) for more details.
```

docs/intro/glossary.md
tiyash-basu-frequenz: Oh I see that this is at the same level as `## Common Acronyms`. This is a bit confusing IMO.

docs/intro/glossary.md
```diff
@@ -0,0 +1,241 @@
...
+
+### Component ID
+
+A numeric identifier uniquely representing an instance of a [component](#component) in a particular[microgrid](#microgrid). It is always of type `int`.
+
+For example, a battery an have a component ID **5**.
+
+### Island
+
+A [microgrid](#microgrid) that is not connected to the public electricity [grid](#grid).
```
cwasicki, line 132: The definition of `consumer` is missing

docs/intro/glossary.md
```diff
@@ -0,0 +1,241 @@
...
+[State of charge](#state-of-charge). See the [Wikipedia article](https://en.wikipedia.org/wiki/State_of_charge) for more details.
+
+### SoP
+
+[State of power](#state-of-power).
+
+## Microgrid
+
+A local electrical grid that connects a set of different [types of components](#component-category) together. It can be connected to the public [grid](#grid), or be completely isolated, in which case it is known as an island.
+
```
cwasicki, line 49: I think there should be a distinction between assets (battery, PV, consumer, grid, etc) and maybe devices (inverter, meter, relay, etc) as they have a quite different role in the microgrid.

cwasicki (changes requested):

docs/intro/glossary.md
```diff
@@ -0,0 +1,241 @@
...
+For example, the metric ID for the capacity of a [battery](#battery) is simply **`capacity`**.
+
+### Time Series ID
+
+An identifier for a [time series](#time-series) originating from a [metric](#metric) of a specific component. Typically a string (`str`) derived from the tuple **([component ID](#component-id), [metric ID](#metric-id))** for [components](#component).
+
+For example, a time series for the capacity of a battery with component ID 5 has the ID **(component_id, metric_id)** (or **`f"{component_id}_{metric_id}"`**).
+
+### Timeseries ID
```
cwasicki, line 202: I guess that's obvious.

docs/intro/glossary.md
```diff
@@ -0,0 +1,241 @@
...
+(2022-01-01 22:00:00.000 UTC, 400)
+(2022-01-01 22:00:01.000 UTC, 401)
+(2022-01-01 22:00:02.000 UTC, 403)
+(2022-01-01 22:00:03.000 UTC, 402)
+(2022-01-01 22:00:04.000 UTC, 403)
+```
+
+### Timeseries
+
+Same as [time series](#time-series).
```
cwasicki, line 186: obvious.

docs/intro/glossary.md
tiyash-basu-frequenz: That's right. Here it is spelled with an `i` instead of a `y`.

docs/intro/glossary.md
tiyash-basu-frequenz: I gave it a bit of thought.
When we say `Grid`, we mean an abstraction over the whole grid, encapsulated into a single component in the microgrid. We do not actually mean the grid connection point anymore. This could certainly have effect on the documentation on our APIs too.

I think it's fine to leave it like this for now. We can discuss the terminology in a separate issue/discussion.

docs/intro/glossary.md
llucax: Oh, damn, I'm colorblind :facepalm:

docs/intro/glossary.md
llucax: Done. It was much more complicated than expected, is hard to explain without going into a lot of details about AC :(

docs/intro/glossary.md
llucax: Thanks!

docs/intro/glossary.md
llucax: Why the caps here (and in the other acronym definitions)?

docs/intro/glossary.md
llucax: True, I just don't know where to put it, as it doesn't fit any section for now. I guess I will create a `Electrical terms` section or something...

docs/intro/glossary.md
llucax: In raw, yes, not that much when it is rendered, as there is also a navigation bar with the hierarchy.

I'm not saying is the best organization ever, it is just what we had in the wiki, I started with that. We could rethink the structure, but I consider this out of the scope of this PR.

docs/intro/glossary.md
llucax: Can you do that in a followup PR? You can also create an issue so we don't forget. I want to start closing up this PR, because if we want to have a complete glossary before merging, it will be never merged :P

docs/intro/glossary.md
llucax: I think DC to DC (or DC-DC) is what's correct, @tiyash-basu-frequenz ?

llucax: I went with wikipedia: https://en.wikipedia.org/wiki/DC-to-DC_converter

Added a definition for `DC-to-DC converter` and said that `Converter` usually refers to that.

docs/intro/glossary.md
llucax: I think I will rephrase to `a battery with a component ID of **5**.`. Sounds good?

docs/intro/glossary.md
llucax: It wasn't in the wiki, I would also leave it for a separate PR.

docs/intro/glossary.md
llucax: This also comes from the wiki structure, this was started when we defined the pipeline, so the structure some from defining what a sample, metric, component, timeseries, etc. is. The idea behind the structure was to have definition for metrics, etc. I'm not sure what's best for discovering things, but you can always use the search bar if you are searching for a particular term, which is pretty good.

Again, I would leave changes to the structure for a separate PR, because this PR is just growing too much, and the goal just just to move the wiki to the web, and it's becoming an endless monster :)

docs/intro/glossary.md
llucax: Is for searching. If you search for "time series" in the search bar, "timeseries" won't be shown as a result.

docs/intro/glossary.md
llucax: Same.

docs/intro/glossary.md
llucax: Although not sure what to do with the source now, as it is not a literal copy of the text in the source. I guess we can leave it anyways.

docs/intro/glossary.md
tiyash-basu-frequenz: DC-DC converter is actually the term used in the industry.

daniel-zullo-frequenz: oh I didn't know. I admit it sounded confusing to me.

llucax (commented):
Updated the PR, I would like to merge it as is, because it is getting out of hands. Please try to focus only on obvious errors in the current text for the review and create new issues for any other stuff you want to see eventually changed.

docs/intro/glossary.md
```diff
@@ -0,0 +1,296 @@
...
+
+The power that is actually consumed by the [load](#load). It is the real part of the [apparent power](#apparent-power), P in the [power triangle](#power-triangle).
+
+#### Apparent Power
+
+The power that is actually supplied to the [load](#load). The magnitude of the vector sum of the [active](#active-power) and [reactive](#reactive-power) power, |S| in the [power-triangle](#power-triangle).
+
+#### Phase
+
+The angle of difference (in degrees) between current and voltage in an [AC](#ac) circuit, φ in the [power triangle](#power-triangle).
```
tiyash-basu-frequenz, line 263: Can you cite a source for this definition? It sounds incorrect.

Conventionally,
* `φ` is the angle between active and reactive power.
* `Phase` is a synonym for Line, as in, the 3 AC phases.

docs/intro/glossary.md
tiyash-basu-frequenz: I can imagine. Thanks for adding those definitions. I will take a look at those.

docs/intro/glossary.md
llucax: Wikipedia. https://en.wikipedia.org/wiki/AC_power

llucax: Well, it says "Phase of voltage relative to current", I read "Phase **or** voltage relative to current" when I was writing it, maybe that makes a difference.

tiyash-basu-frequenz: But in wikipedia it is unambiguous. There it is called `Phase of voltage relative to current`. Here it is ambiguous.

docs/intro/glossary.md
llucax: Done in the Microgrid section, as I didn't want to start a new section if we are not sure about the structure yet.

docs/intro/glossary.md
tiyash-basu-frequenz: Also, conventionally, systems use [Power Factor](https://en.wikipedia.org/wiki/Power_factor), instead of `φ`. Maybe that's a reason why wikipedia `φ` has an elaborate name (it's not used as often, so no one gave it a cool name).

docs/intro/glossary.md
tiyash-basu-frequenz: Fair enough.

docs/intro/glossary.md
tiyash-basu-frequenz: >I would leave changes to the structure for a separate PR

Sure

docs/intro/glossary.md
daniel-zullo-frequenz: Because I wanted to somehow highlight what the acronym stands for. I initially made an optional global comment to check if we want to go that way, then I made suggestions where we can possibly updated if the global suggestion was taken. So feel free to discard this and similar suggestions if they are not suitable

docs/intro/glossary.md
```diff
@@ -0,0 +1,296 @@
...
+
+A device that converts the wind's kinetic energy into electrical energy.
+
+### Component
+
+A device (of a particular [category](#component-category)) within a [microgrid](#microgrid).
+
+### Component ID
+
+A numeric identifier uniquely representing an instance of a [component](#component) in a particular[microgrid](#microgrid). It is always of type `int`.
```
daniel-zullo-frequenz:
```suggestion
A numeric identifier uniquely representing an instance of a [component](#component) in a particular [microgrid](#microgrid). It is always of type `int`.
```

docs/intro/glossary.md
```diff
@@ -0,0 +1,296 @@
...
+
+Same as [residual consumption](#residual-consumption).
+
+### State of Charge
+
+The level of charge of a [battery](#battery) relative to its capacity, expressed in percentage points. Calculated as the ratio between the remaining energy in the battery at a given time and the maximum possible energy under similar health conditions. [Source](https://epicpower.es/wp-content/uploads/2020/08/AN028_SoC-SoH-SoP-definitions_v3.pdf)
+
+### State of Power
+
+The ratio of peak power to nominal power. Peak power is the maximum power that can be sustained for a specific duration without violating preset operational design limits on battery voltage, SsC, power, or current.
```
daniel-zullo-frequenz: maybe `SoC`?
```suggestion
The ratio of peak power to nominal power. Peak power is the maximum power that can be sustained for a specific duration without violating preset operational design limits on battery voltage, SoC, power, or current.
```

docs/intro/glossary.md
```diff
@@ -0,0 +1,296 @@
+# Glossary
+
+This glossary provides definitions for common terminology used in the SDK,
```
daniel-zullo-frequenz:
```suggestion
This glossary provides definitions for common terminology used in the Frequenz SDK,
```

docs/intro/glossary.md
```diff
@@ -0,0 +1,296 @@
...
+focusing on microgrid components, metrics, measurements, and power-related
+terms.
```
daniel-zullo-frequenz:
```suggestion
focusing on microgrid components, metrics, measurements, and power-related terms.
```

daniel-zullo-frequenz (commented):
A couple of cosmetic suggestions and LGTM

docs/intro/glossary.md
cwasicki: It is, 4th bullet. https://github.com/frequenz-floss/frequenz-sdk-python/wiki/Glossary#microgrid-power-terminology

docs/intro/glossary.md
cwasicki: https://github.com/frequenz-floss/frequenz-sdk-python/issues/650

docs/intro/glossary.md
llucax: That's `consumer_power`, a terminology thing, IMHO glossary-wise it is covered by the terms **Power** and **Consumption**. Please create an follow-up PR (or an issue including the exact term and definition you want to include) if you think something else should be added.

docs/intro/glossary.md
llucax: Could you write a code suggestion on how to change this? I'm not very familiar with the term, so if I do it I think there is a high risk that I get it wrong again :worried:

llucax: Or maybe I just remove it if it's not very relevant, I just added to make the several components of AC power more complete.

tiyash-basu-frequenz: Sure, on it.

tiyash-basu-frequenz: Since I started the review on just one line, I cannot add multi-line suggestion I think :(
I have typed up a suggestion here, if it helps:

```md
#### Phase of Voltage Relative to Current

The angle of difference (in degrees) between current and voltage in an [AC](#ac) circuit, φ in the [power triangle](#power-triangle). Applicable to AC electricity only.

#### Power Factor

The ratio of real power to apparent power. Applicable to AC electricity only. Can be obtained by computing the cosine of  [φ](#phase-of-voltage-relative-to-current), in the [power triangle](#power-triangle). See the [Wikipedia article](https://en.wikipedia.org/wiki/Power_factor) for more details.
```

tiyash-basu-frequenz: >Or maybe I just remove it if it's not very relevant, I just added to make the several components of AC power more complete.

IMO it makes sense to keep it, since it's used in the diagram. But your call.

llucax: Great, thanks! I applied it but changed the `Applicable to AC electricity only` in **Power Factor** to ` in an [AC](#ac) circuit` as it is shorter and matches **Phase of Voltage Relative to Current**.

I also added AC to other AC definitions.

Please see:
* 63a5fbfc059e4bfe7b118d5d22089b296d66e9dc

docs/intro/glossary.md
cwasicki: Also update the terminology in the wiki replacing `component` with `asset_type` to stress the distinction.

docs/intro/glossary.md
cwasicki: I don't think it should be omitted in the glossary so added it also here https://github.com/frequenz-floss/frequenz-sdk-python/issues/650

llucax: Great, thanks!

daniel-zullo-frequenz (approved):

PR #651, opened 2023-09-08
Merge v0.25.x into v0.x.x

llucax:
This merges the following fixes released with v0.25.1:

- Replace the `section-index` plugin with `navigation.indexes`
- Fix consumer power formula
- Add test for consumer power formula


llucax:
Oh, it seems like the direct merge doesn't work because we need to fix the conflicts. Will create a new PR with a merge for the manual (resolved) merge.

PR #654, opened 2023-09-08
Add OS/X to CI

### LINKED ISSUE #1: Enable CI for Mac OS

christianparpart:
Closes #1 

the all so famous issue number #1

Mind, this is a draft PR because I will need to keep re-iterating until GHCI is happy

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -30,6 +30,7 @@ jobs:
       matrix:
         os:
           - ubuntu-20.04
+          - macos-latest
         python:
           - "3.11"
         nox-session:
@@ -190,7 +191,13 @@ jobs:
 
   build:
     name: Build distribution packages
-    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-20.04
+          - macos-latest
+    runs-on: ${{ matrix.os }}
     steps:
       - name: Fetch sources
         uses: actions/checkout@v3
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yaml
```diff
@@ -30,6 +30,7 @@ jobs:
       matrix:
         os:
           - ubuntu-20.04
+          - macos-latest
```
llucax, line 33: It is tricky to enable macos support like this, because we did that once but the macos runners were just too slow. Ideally we should only run it in the merge queue and pushes, as we do with arm64 now, but then it can't be part of the matrix, and would need a separate job I guess. Not 100% sure if there is a way to have `if` rules for specific items in the matrix.

.github/workflows/ci.yaml
```diff
@@ -212,7 +219,9 @@ jobs:
       - name: Build the source and binary distribution
         run: python -m build
 
+        # It is sufficient to upload only once
       - name: Upload distribution files
+        if: ${{ matrix.os == 'ubuntu-20.04' }}
```
llucax, line 224: This is not strictly true, we should also test that the built package is installable (we do it in the `test-installation` job. Not sure if we really need to do this using QEMU and docker though, for archs directly supported by GitHub it might be much faster to just do it installing in a venv.

.github/workflows/ci.yaml
llucax, line 33: Also if were going to go for this, we should use a specific macos release, otherwise the builds will not be reproducible in the future (we already had issues trying to do patch releases for old versions and the CI failing because its pulling newer dependencies that don't work with the old code).

llucax (commented):
My personal opinion is this is too much work to do it properly to be worth it right now. IMHO the way to go (unless there are `if` conditions for matrix items) would be to have a reusable action/workflow in repo config so we can create a separate job to run it only on push/queue without having to repeat the whole job, but if we want to repeat the whole job as an intermediate solution I will not oppose either.

PR #657, opened 2023-09-11
ci: Fix publishing of documentation on push

llucax:
Enable `nox-cross-arch` test for push events because if we don't do this, the `publish-docs` will not run on pushes, so the in-development documentation will not be published.

While cross-arch tests are generally executed in the merge queue on the identical commit, activating them for push events provides an added layer of security. This is especially important in scenarios where emergency pushes might bypass the merge queue.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -81,7 +81,7 @@ jobs:
 
   nox-cross-arch:
     name: Cross-arch tests with nox
-    if: github.event_name == 'merge_group'
+    if: github.event_name != 'pull_request'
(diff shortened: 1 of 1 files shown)


llucax:
FYI @tiyash-basu-frequenz

tiyash-basu-frequenz (approved):

llucax:
At some point we might check if a job already ran in the merge queue and skip it in the push to save CPU cycles. There is always so much that can be improved in the CI :scream:

daniel-zullo-frequenz (approved):

Issue #501, opened 2023-07-11
Support component graph configs with m:n battery-inverter connections


shsms:
### What's needed?

Currently multiple parts of the SDK expect that each battery is connected to a single inverter.  This needs to be generalized, such that there can be configs like 

- multiple inverters connected to a single battery
- multiple batteries connected to a single inverter
- ~multiple batteries connected to multiple inverters~

### Proposed solution

This needs to be implemented and/or tested in these modules:

- [x] #663
- [x] BatteryPool should support m:n battery-inverter connections
- [x] https://github.com/frequenz-floss/frequenz-sdk-python/pull/730
- [ ] PowerDistributingActor's `BatteryStatusTracker` needs to be updated
- [ ] Test against the sandbox

### Notes

Taken from the discussion below:
We can always assume that all inverters are equal, e.g. it doesn't matter in which part of the microgrid they release power.
The currently discussed and preferred solution is to stack inverters to max. efficiency (lower power -> lower efficiency)

Marenz:
How do we want to deal with multiple inverters connected to one (or multiple) batteries? What inverters should a charge/discharge command affect? @thomas-nicolai-frequenz @tiyash-basu-frequenz

shsms:
Just for context, when the battery pool is used to set power,  we only sent the battery ids to the power manager/distributor.  And they have to find the corresponding inverter.  Do we just distribute power equally if there are multiple inverters for a single battery?

thomas-nicolai-frequenz:
> Do we just distribute power equally if there are multiple inverters for a single battery?

As a default I'd say yes but we should support multiple strategies in the future as the most efficient way is to run inverters close to their max instead of within lower power ranges. This is when they become less efficient by default. That means the best case would be to enable just one and start enabling more if the power increases beyond a single inverter. However, keep in mind that we might also need to look into failover. Once inverter might fail but the second one still work. Thats something we do not have in a 1:1 setup.

@tiyash-basu-frequenz any thoughts on that?

thomas-nicolai-frequenz:
> one inverter to multiple batteries

is a different case.

Marenz:
> > one inverter to multiple batteries
>
> is a different case.

but that's exactly the case we're currently want to discuss here?

thomas-nicolai-frequenz:
> but that's exactly the case we're currently want to discuss here?

[...]

PR #660, opened 2023-09-11
PowerDistributing: Add support for n:m relations in invs:bats

### LINKED ISSUE #663: PowerDistributor should support m:n battery-inverter connections
Details in parent issue: https://github.com/frequenz-floss/frequenz-sdk-python/issues/501

Marenz:
- Allow component graph to be a DAG instead of a tree
- Fix comments in tests
- Don't guess missing power bounds from adjacent components
- GraphGenerator: Add function to generate list of components
- PowerDistributing: Add n:m support for inverters:batteries

refs #501

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -32,6 +32,11 @@
 
 
 
+- The PowerDistributingActor now supports n:m relations between inverters and
+  batteries.
+  This means that one or more inverters can be connected to one or more batteries.
+
+
 ## Bug Fixes
 
 - Fix rendering of diagrams in the documentation.
--- a/src/frequenz/sdk/actor/power_distributing/_distribution_algorithm/_distribution_algorithm.py
+++ b/src/frequenz/sdk/actor/power_distributing/_distribution_algorithm/_distribution_algorithm.py
@@ -14,14 +14,144 @@ from ....microgrid.component import BatteryData, InverterData
 _logger = logging.getLogger(__name__)
 
 
+@dataclass()
+class AggregatedBatteryData:
+    """Aggregated battery data."""
+
+    component_id: int
+    """The component ID of the first battery.
+
+    This is only used to identify the pair of battery and inverter.
+    """
+
+    soc: float
+    """The aggregated SoC of the batteries."""
+
+    capacity: float
+    """The aggregated capacity of the batteries."""
+
+    soc_upper_bound: float
+    """The aggregated upper SoC bound of the batteries."""
+
+    soc_lower_bound: float
+    """The aggregated lower SoC bound of the batteries."""
+
+    power_inclusion_upper_bound: float
+    """The aggregated upper power inclusion bound of the batteries."""
+
+    power_inclusion_lower_bound: float
+    """The aggregated lower power inclusion bound of the batteries."""
+
+    power_exclusion_upper_bound: float
+    """The aggregated upper power exclusion bound of the batteries."""
+
+    power_exclusion_lower_bound: float
+    """The aggregated lower power exclusion bound of the batteries."""
+
+    def __init__(self, batteries: list[BatteryData]) -> None:
+        """Create DistBatteryData from BatteryData.
+
+        Aggregates the data of the batteries:
+
+        * Capacity: Sum of capacities of all batteries.
+        * SoC: Weighted average of SoCs of all batteries.
+        * SoC bounds: Weighted average of SoC bounds of all batteries.
+        * Power inclusion bounds: Sum of power inclusion bounds of all batteries.
+        * Power exclusion bounds: Largest power exclusion bound multiplied by
+            the number of batteries.
+
+        Args:
+            batteries: The batteries to aggregate.
+        """
+        assert len(batteries) > 0, "No batteries given."
+
+        # We need only one component ID for DistBatteryData to be able to
+        # identify the pair
+        self.component_id = batteries[0].component_id
+
+        self.capacity = sum(b.capacity for b in batteries)
+
+        if self.capacity != 0.0:
+            self.soc = sum(b.soc * b.capacity for b in batteries) / self.capacity
+            self.soc_upper_bound = (
+                sum(b.soc_upper_bound * b.capacity for b in batteries) / self.capacity
+            )
+            self.soc_lower_bound = (
+                sum(b.soc_lower_bound * b.capacity for b in batteries) / self.capacity
+            )
+        else:
+            self.soc = math.nan
+            self.soc_upper_bound = math.nan
+            self.soc_lower_bound = math.nan
+
+        self.power_inclusion_upper_bound = sum(
+            b.power_inclusion_upper_bound for b in batteries
+        )
+        self.power_inclusion_lower_bound = sum(
+            b.power_inclusion_lower_bound for b in batteries
+        )
+        # To satisfy the largest exclusion bounds in the set we need to
+        # provide the power defined by the largest bounds multiplied by the
+        # number of batteries in the set.
+        self.power_exclusion_upper_bound = max(
+            b.power_exclusion_upper_bound for b in batteries
+        ) * len(batteries)
+        self.power_exclusion_lower_bound = min(
+            b.power_exclusion_lower_bound for b in batteries
+        ) * len(batteries)
+
+
 class InvBatPair(NamedTuple):
     """InvBatPair with inverter and adjacent battery data."""
 
-    battery: BatteryData
+    battery: AggregatedBatteryData
     """The battery data."""
 
-    inverter: InverterData
+    inverter: list[InverterData]
     """The inverter data."""
+
+
+@dataclass
+class AvailabilityRatio:
+    """Availability ratio for a battery-inverter pair."""
+
+    battery_id: int
+    """The battery ID."""
+
+    inverter_ids: list[int]
+    """The inverter IDs."""
+
+    ratio: float
+    """The availability ratio."""
+
+    min_power: float
+    """The minimum power that can be set for the battery-inverters pair."""
+
+
+@dataclass
+class _Power:
+    """Helper class for distribution algorithm."""
+
+    upper_bound: float
+    """The upper bound of the power that can be set for the battery."""
+
+    power: float
+    """The power to be set for the inverter."""
+
+
+_InverterSet = frozenset[int]
+"""A set of inverter IDs."""
+
+
+@dataclass
+class _Allocation:
+    """Helper class for distribution algorithm."""
+
+    inverter_ids: _InverterSet
+    """The IDs of the inverters."""
+
+    power: float
+    """The power to be set for the inverters."""
 
 
 @dataclass
@@ -227,7 +357,7 @@ class DistributionAlgorithm:
         components: list[InvBatPair],
         available_soc: dict[int, float],
         excl_bounds: dict[int, float],
-    ) -> tuple[list[tuple[InvBatPair, float, float]], float]:
+    ) -> tuple[list[AvailabilityRatio], float]:
         r"""Compute battery ratio and the total sum of all of them.
 
         battery_availability_ratio = capacity_ratio[i] * available_soc[i]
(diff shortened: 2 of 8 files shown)


Marenz:
The only thing that needs some looking at and better test is the scenario with 1 inverter -> 2 batteries, but one broken

src/frequenz/sdk/actor/power_distributing/_distribution_algorithm/_distribution_algorithm.py
```diff
@@ -249,24 +327,46 @@ def _compute_battery_availability_ratio(
...
+                # to handle multiple inverters can be implemented.
+                per_inverter_ratio = ratio / len(inverters)
+                for inverter in inverters:
+                    battery_availability_ratio.append(
+                        AvailabilityRatio(
+                            inverter.component_id,
+                            excl_bounds[inverter.component_id],
+                            per_inverter_ratio,
+                        )
+                    )
```
shsms: wasn't the consensus to use each inverter to its max power, before using the next one?  I think people on the ticket said otherwise, we lose power to inverter inefficiencies and self-consumptions.  

Also, it maybe easier to consider it a single device for the first pass,  so that you don't have to divide the exlusion bounds (if any) into several inverters, or try to give minimum power for all inverters.  

And after the stage as a single target power is done, you can just redistribute it to the inverters, filling up each one before going to the next.

tests/actor/power_distributing/test_distribution_algorithm.py
```diff
@@ -218,6 +237,31 @@ def test_distribute_power_two_batteries_2(self) -> None:
...
+        }
+
+        algorithm = DistributionAlgorithm(distributor_exponent=1)
+        result = algorithm._distribute_power(  # pylint: disable=protected-access
+            components, 600, available_soc, incl_bounds, excl_bounds
+        )
+
+        assert result.distribution == approx({1: 500})
+        assert result.remaining_power == approx(100.0)
+
```
shsms, line 262: I guess you need some tests for the distribution algorithm with exclusion bounds as well,  there's a class for those tests at the end of this file I think.

shsms (commented):
Still need to review the changes to the actor, in the last commit.  will do tomorrow.

src/frequenz/sdk/actor/power_distributing/_distribution_algorithm/_distribution_algorithm.py
Marenz: Yes, but after talking with a few more ppl we agreed that for a first version we take the most naive approach

RELEASE_NOTES.md
```diff
@@ -39,6 +39,11 @@
 
 - Drop `Averager` from `FormulaEngine`.
 
+- The PowerDistributingActor now supports n:m relations between inverters and
+  batteries.
+  This means that one or more inverters can be connected to one or more batteries.
```
llucax, line 37: I would also mention this affects the battery pool, as most users don't user the power distributing actor directly, and not use 80-chars limit, as it doesn't get properly rendered in GitHub release notes otherwise.

```suggestion
- The `PowerDistributingActor` (and thus the battery pools too) now supports *n:m* relations between inverters and batteries.

  This means that one or more inverters can be connected to one or more batteries.
```

llucax (commented):
Can you also update the PR description? It is outdated (it says there are stuff missing) and will become a commit message, so it should be an appropriate message for a merge commit.

I just gave this a quick look but didn't check the logic, I'm leaving that to @shsms.

src/frequenz/sdk/actor/power_distributing/_distribution_algorithm/_distribution_algorithm.py
shsms: Have you tried doing some tests with exclusion bounds?  I still feel this is the way to go:

> Also, it maybe easier to consider it a single device for the first pass, so that you don't have to divide the exlusion bounds (if any) into several inverters, or try to give minimum power for all inverters.
> 
> And after the stage as a single target power is done, you can just redistribute it to the inverters, filling up each one before going to the next.

RELEASE_NOTES.md
shsms: It is just the "control" methods in the battery pool.  Not all of it yet.

src/frequenz/sdk/actor/power_distributing/_distribution_algorithm/_distribution_algorithm.py
Marenz: Yeah, I am right now working on that exclusion part, after I added some tests I noticed that that part needs some work.

It is a tougher problem with those, indeed

src/frequenz/sdk/actor/power_distributing/_distribution_algorithm/_distribution_algorithm.py
```diff
@@ -304,113 +421,192 @@ def _distribute_power(  # pylint: disable=too-many-arguments
...
-            return DistributionResult(distribution, power_w)
-
+            final_distribution = {
+                inverter.component_id: 0.0
+                for _, inverters in components
+                for inverter in inverters
+            }
+            return DistributionResult(final_distribution, power_w)
+
+        # key: inverter_ids, value: (battery_id, power)
```
Marenz: This comment is wrong, fixing..

src/frequenz/sdk/actor/power_distributing/_distribution_algorithm/_distribution_algorithm.py
```diff
@@ -14,16 +14,117 @@
...
+                    / self.capacity,
+                )
+
+        for field in (
+            "power_inclusion_upper_bound",
+            "power_inclusion_lower_bound",
+            "power_exclusion_upper_bound",
+            "power_exclusion_lower_bound",
+        ):
+            setattr(self, field, sum(getattr(b, field) for b in batteries))
```
shsms: I think `getattr/setattr` are slower than using the `.` syntax, because it can be optimized better in the bytecode, rather than a string lookup at each step.  Especially because this is part of the hot loop.

Also using `setattr/getattr` bypass mypy or any other code analysis.

And I think it would be more readable to actually just spell it all out.

src/frequenz/sdk/actor/power_distributing/_distribution_algorithm/_distribution_algorithm.py
```diff
@@ -227,7 +328,7 @@ def _compute_battery_availability_ratio(
         components: list[InvBatPair],
         available_soc: dict[int, float],
         excl_bounds: dict[int, float],
-    ) -> tuple[list[tuple[InvBatPair, float, float]], float]:
+    ) -> tuple[list[AvailabilityRatio], float]:
```
shsms, line 360: It would be nice to have many smaller commits with one isolated thing per commit, instead of one large one, to make reviewing easier.  Else it is too many things to track in our heads at the same time.  

It is a bit hard for changes like this,  but we no longer care about individual commits passing the tests, if that helps with readability.

For example a commit just to migrate to `AvailabilityRatio`,  another commit for adding the new field in it, etc. is still something.

But for the future I guess.

src/frequenz/sdk/actor/power_distributing/_distribution_algorithm/_distribution_algorithm.py
```diff
@@ -304,113 +421,186 @@ def _distribute_power(  # pylint: disable=too-many-arguments
...
+                    incl_bounds[inverter_id] for inverter_id in ratio_data.inverter_ids
+                ),
+                incl_bounds[ratio_data.battery_id],
+            )
             if calculated_power > incl_bound:
-                excess_reserved[inverter.component_id] = incl_bound - excl_bound
-                # # Distribute between remaining batteries
-            elif calculated_power < excl_bound:
-                deficits[inverter.component_id] = calculated_power - excl_bound
+                excess_reserved[frozenset(ratio_data.inverter_ids)] = (
```
shsms: Maybe convert to `InverterSet` once, directly when creating the `AvailabilityRatio` instance, so you don't have to keep typing it everywhere?  

This needs to be done in the actor and battery pool also, because it looks like we're using them as dict keys in many places, including the power manager, and doing a lot of unnecessary allocations every time converting to a frozenset,  but I guess that can be done separately.

src/frequenz/sdk/actor/power_distributing/_distribution_algorithm/_distribution_algorithm.py
```diff
@@ -304,113 +421,186 @@ def _distribute_power(  # pylint: disable=too-many-arguments
...
+            distribution[InverterSet(ratio_data.inverter_ids)] = _Power(
+                upper_bound=incl_bound,
+                power=ratio_data.min_power,
+            )
 
-            distributed_power += excl_bound
-            distribution[inverter.component_id] = excl_bound
+        make_largest = make_dataclass(
+            "make_largest", [("inverter_ids", frozenset), ("power", float)]
+        )
```
shsms: Does mypy understand this syntax?  Why not just the normal one?

Also I think we shouldn't do this as part of the hot-path, better to move it outside the function.  Because it is still a runtime step in Python.

```suggestion
@dataclass
class _Allocation:
    inverter_ids: InverterSet
    power: float

File: src/frequenz/sdk/actor/power_distributing/_distribution_algorithm/_distribution_algorithm.py [hunk 9]
```diff
@@ -304,113 +421,186 @@ def _distribute_power(  # pylint: disable=too-many-arguments
...
                     distributed_power += left_over
 
-        for inverter_id, excess in excess_reserved.items():
-            distribution[inverter_id] += excess
+        for inverter_ids, excess in excess_reserved.items():
             distributed_power += excess
+            batter_power = distribution[inverter_ids]
+            batter_power.power += excess
+            # Add excess power to the inverter set
+            distribution[inverter_ids] = batter_power
```
shsms:
``` suggestion
            battery_power = distribution[inverter_ids]
            battery_power.power += excess
            # Add excess power to the inverter set
            distribution[inverter_ids] = battery_power

File: src/frequenz/sdk/actor/power_distributing/_distribution_algorithm/_distribution_algorithm.py [hunk 10]
```diff
@@ -537,3 +710,51 @@ def _distribute_supply_power(
...
+            supply: if True then supply bounds will be calculated, otherwise
+                consume bounds.
+
+        Returns:
+            inclusion and exclusion bounds.
+        """
+        incl_bounds: dict[int, float] = {}
+        excl_bounds: dict[int, float] = {}
+        for battery, inverters in components:
+            # We can supply/consume with int only
```
shsms: this is not the case anymore, could you remove this comment?  I missed it when implementing the exclusion bounds.

tests/actor/power_distributing/test_distribution_algorithm.py
```diff
@@ -1204,3 +1267,160 @@ def test_scenario_3(self) -> None:
...
+        Expected result:
+
+        | request |                           |    excess |
+        |   power | distribution              | remaining |
+        |---------+---------------------------+-----------|
+        |    -300 | -200, -100, 0             |         0 |
+        |     300 | 200, 100, 0               |         0 |
+        |   -1800 | -300, -1000, -500         |         0 |
+        |    3000 | 1000, 500, 1000, 500      |         0 |
+        |    3500 | 1000, 500, 1000, 500      |       500 |
```
shsms:
```suggestion
        |    -300 | -200, -100, 0, 0          |         0 |
        |     300 | 200, 100, 0, 0            |         0 |
        |   -1800 | -300, 0, -1000, -500      |         0 |
        |    3000 | 1000, 500, 1000, 500      |         0 |
        |    3500 | 1000, 500, 1000, 500      |       500 |

shsms (commented):
This looks nice, many parts have become more readable despite getting more complicated. 

We discussed the case of adjusting exclusion bounds when there are n batteries in a set, excl bounds for batteries = max(excl_bounds) * n

I couldn't see if this is done.  If it is there, it should be in the actor right?.  Could you confirm, and maybe add a test for the actor as well?

Marenz:
> I couldn't see if this is done.

Ah, damn, I knew there was one thing missing. Will add

src/frequenz/sdk/actor/power_distributing/_distribution_algorithm/_distribution_algorithm.py
Marenz: True, I usually do that, too. I just didn't expect this part to take up so much extra work XD

shsms (approved):

PR #661, opened 2023-09-12
Add a job to join the `nox` matrix jobs

llucax:
This job is only used to have a single job we can require in branch protection rules, so we don't have to update the protection rules each time we add or remove a job from the matrix.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -185,14 +185,18 @@ jobs:
         if: always()
         run: sudo chown -R $USER:$USER /tmp/pip-cache
 
-  # This job runs if `nox-cross-arch` ran and succeeded.
-  # This is required because, when the `nox-cross-arch` job is skipped, its
-  # inner matrix is not expanded, and branch protection rules on the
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -79,6 +79,19 @@ jobs:
         run: nox -R -e "$NOX_SESSION"
         timeout-minutes: 10
 
+  # This job runs if all the `nox` matrix jobs ran and succeeded.
+  # It is only used have a single job we can require in branch protection
```
daniel-zullo-frequenz: I think this sentence needs to be rephrased. Maybe something around
```suggestion
  # It is only used to define a single job that we can require in branch protection
```

daniel-zullo-frequenz (commented):
Just a comment to update documentation and LGTM

.github/workflows/ci.yaml
llucax: Thanks, updated (I used "It is only used to have ...")

daniel-zullo-frequenz (approved):

PR #664, opened 2023-09-12
Replace len magic with count_valid method in ring buffer and moving window

cwasicki:
The current len magic returns the count of samples in the ring buffer / moving window that are determined valid. This definition is not unambiguously obvious to the caller. The current logic is therefore moved to its own more descriptive method and len magic are removed.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -45,6 +45,7 @@
   - Provide access to `capacity` (maximum number of elements) in `MovingWindow`.
   - Methods to retrieve oldest and newest timestamp of valid samples are added to both.
   - `MovingWindow` exposes underlying buffers `window` method.
+  - `len(window)` and `len(buffer)` should be replaced with `window.count_valid()` and `buffer.count_valid()`, respectively.
   - `OrderedRingBuffer.window`:
     - By default returns a copy.
     - Can also return a view if the window contains `None` values and if `force_copy` is set to `True`.
--- a/src/frequenz/sdk/timeseries/_moving_window.py
+++ b/src/frequenz/sdk/timeseries/_moving_window.py
@@ -313,7 +313,7 @@ class MovingWindow(BackgroundService):
         Returns:
             The size of the `MovingWindow`.
         """
-        return len(self._buffer)
+        return self._buffer.count_valid()
 
     @overload
     def __getitem__(self, key: SupportsIndex) -> float:
@@ -362,7 +362,7 @@ class MovingWindow(BackgroundService):
             A float if the key is a number or a timestamp.
             an numpy array if the key is a slice.
         """
-        if len(self._buffer) == 0:
+        if self._buffer.count_valid() == 0:
             raise IndexError("The buffer is empty.")
         if isinstance(key, slice):
             if isinstance(key.start, int) or isinstance(key.stop, int):
--- a/src/frequenz/sdk/timeseries/_moving_window.py
+++ b/src/frequenz/sdk/timeseries/_moving_window.py
@@ -306,6 +306,15 @@ class MovingWindow(BackgroundService):
             asyncio.create_task(self._resampler.resample(), name="resample")
         )
 
+    def count_valid(self) -> int:
+        """
+        Count the number of valid samples in this `MovingWindow`.
+
+        Returns:
+            The number of valid samples in this `MovingWindow`.
+        """
+        return self._buffer.count_valid()
+
     def __len__(self) -> int:
         """
         Return the size of the `MovingWindow`s underlying buffer.
@@ -313,7 +322,7 @@ class MovingWindow(BackgroundService):
         Returns:
             The size of the `MovingWindow`.
         """
-        return self._buffer.count_valid()
+        return self.count_valid()
 
     @overload
     def __getitem__(self, key: SupportsIndex) -> float:
--- a/src/frequenz/sdk/timeseries/_moving_window.py
+++ b/src/frequenz/sdk/timeseries/_moving_window.py
@@ -315,15 +315,6 @@ class MovingWindow(BackgroundService):
         """
         return self._buffer.count_valid()
 
-    def __len__(self) -> int:
-        """
-        Return the size of the `MovingWindow`s underlying buffer.
-
-        Returns:
-            The size of the `MovingWindow`.
-        """
-        return self.count_valid()
-
     @overload
     def __getitem__(self, key: SupportsIndex) -> float:
         """See the main __getitem__ method.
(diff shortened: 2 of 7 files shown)


Marenz (dismissed):

RELEASE_NOTES.md
```diff
@@ -26,6 +26,7 @@
   - NaN values are treated as missing when gaps are determined in the `OrderedRingBuffer`.
   - Provide access to `capacity` (maximum number of elements) in `MovingWindow`.
   - Methods to retrieve oldest and newest timestamp of valid samples are added to both.
+  - Len Magic methods are replaced by `valid_size` property in both.
```
llucax: Maybe add that `len(window)` should be replaced with `window.valid_size()` more explicitly?

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -278,14 +278,15 @@ async def sink_buffer(sample: Sample[Quantity]) -> None:
             asyncio.create_task(self._resampler.resample(), name="resample")
         )
 
-    def __len__(self) -> int:
+    @property
```
llucax: I think this shouldn't be a property, as it is doing (underneath) a non-trivial calculation.

```suggestion
```

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -278,14 +278,15 @@ async def sink_buffer(sample: Sample[Quantity]) -> None:
...
+    def valid_size(self) -> int:
```
llucax: I find `valid_size`  a bit unclear, for me it sounds it's talking about the validity of the **size**, not the entries. Also usually size if related to the amount of memory an object takes, this should still be a length, not a size., I'm not sure if it is too much, but maybe `len_of_valid_entries` or if we want a short version `len_valid`.

I'm even wondering if we need this method in the buffer/window. If this is the length of some slice or something you can get from the window, can't we implement the `len()` in that object instead?

Like if you can do `x = window.get_something()` and `len(x) == window.len_valid()`, when we don't really need `len_valid()` at all. Maybe we can have a quick chat about this.

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -568,11 +568,12 @@ def __getitem__(self, index_or_slice: SupportsIndex | slice) -> float | FloatArr
         """
         return self._buffer.__getitem__(index_or_slice)
 
-    def __len__(self) -> int:
-        """Return the amount of items that this container currently holds.
+    @property
+    def valid_size(self) -> int:
```
llucax: Same comments about property and the name, feel free to resolve this comment so we can keep the discussion in one place.

src/frequenz/sdk/timeseries/_moving_window.py
cwasicki: Good point, `len_valid` sounds like a better name. 

> Like if you can do x = window.get_something() and len(x) == window.len_valid(), when we don't really need len_valid() at all. Maybe we can have a quick chat about this.

If `get_something` is the new `window` function, then `len(x)` would only be equal to `len_valid` if there are no gaps, otherwise former would be larger as the window returns all values including gap-like entries while latter only counts the number of valid samples. I think both kind of lengths have their own valid use-cases and I am actually considering adding the other length to (e.g. `len_covered`). 

Completely removing it would at the moment not be trivial since it's used in several places already (e.g. moving window, periodic feature extractor).

llucax: OK, I see, I would even consider using `num_valid()` or `count_valid()` instead, because this is not really a `len`, it is just counting the amount of items that are valid, right? For me a `len()` makes sense when it is giving you the number of items in a container (as it is part of the [`Collection`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Collection) interface). If we are applying some *filter* then IMHO it sounds more like counting. But just an idea, I leave that to you, as I'm not that familiar with the ring buffer and moving window details.

cwasicki:
Updated.

RELEASE_NOTES.md
```diff
@@ -27,6 +27,7 @@
   - Provide access to `capacity` (maximum number of elements) in `MovingWindow`.
   - Methods to retrieve oldest and newest timestamp of valid samples are added to both.
   - `MovingWindow` exposes underlying buffers `window` method.
+  - `len(window)` and `len(buffer)` should be replaced with `window.valid_size()` and `buffer.valid_size()`, respectively.
```
llucax:
```suggestion
  - `len(window)` and `len(buffer)` should be replaced with `window.count_valid()` and `buffer.count_valid()`, respectively.
```

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -306,14 +306,14 @@ async def sink_buffer(sample: Sample[Quantity]) -> None:
             asyncio.create_task(self._resampler.resample(), name="resample")
         )
 
-    def __len__(self) -> int:
+    def count_valid(self) -> int:
         """
-        Return the size of the `MovingWindow`s underlying buffer.
+        Return the count of valid samples in the `MovingWindow`.
```
llucax:
```suggestion
        Count the number of valid samples in this `MovingWindow`.
```

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -594,11 +594,11 @@ def __getitem__(self, index_or_slice: SupportsIndex | slice) -> float | FloatArr
         """
         return self._buffer.__getitem__(index_or_slice)
 
-    def __len__(self) -> int:
-        """Return the amount of items that this container currently holds.
+    def count_valid(self) -> int:
+        """Return the count of valid items that this container currently holds.
```
llucax:
```suggestion
        """Count the number of valid items that this buffer currently holds.
```

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -594,11 +594,11 @@ def __getitem__(self, index_or_slice: SupportsIndex | slice) -> float | FloatArr
...
 
         Returns:
-            The length.
+            The count of valid items.
```
llucax:
```suggestion
            The number of valid items in this buffer.
```

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -306,14 +306,14 @@ async def sink_buffer(sample: Sample[Quantity]) -> None:
...
 
         Returns:
-            The size of the `MovingWindow`.
+            The number of valid samples in the `MovingWindow`.
```
llucax:
```suggestion
            The number of valid samples in this `MovingWindow`.
```

llucax (commented):
The release notes should be fixed, the rest are optional, just a suggestion to make the docs a bit more consistent.

cwasicki:
Updated

llucax (approved):

PR #668, opened 2023-09-13
Support int indices and slice index behavior in ring buffer and moving window

cwasicki:
To support int indices and slice index behavior in the moving window and the ring buffer, the following changes are made:
* Support for int indices in addition to datetime indices is added to moving window and ring buffer.
* Indices behave like usual slices (gracefully accept start >= end or index out of range).
* A method to count the number of possible elements in the range between oldest and newest timestamp is added to ring buffer and the moving window.
* `get_timestamp` is added to convert an int index passed by the user to the corresponding timestamp.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -13,8 +13,16 @@
 - A tutorial section and a getting started tutorial
 - In `OrderedRingBuffer`:
   - Rename `datetime_to_index` to `to_internal_index` to avoid confusion between the internal index and the external index.
+  - Add `index_to_datetime` method to convert external index to corresponding datetime.
   - Remove `__setitem__` method to enforce usage of dedicated `update` method only.
+- In `OrderedRingBuffer` and `MovingWindow`:
+  - Support for integer indices is added.
+  - Add `count_covered` method to count the number of elements covered by the used time range.
+
+
+
 
 ## Bug Fixes
 
 - Fix rendering of diagrams in the documentation.
+- The `__getitem__` magic of the `MovingWindow` is fixed to support the same functionality that the `window` method provides.
--- a/src/frequenz/sdk/timeseries/_moving_window.py
+++ b/src/frequenz/sdk/timeseries/_moving_window.py
@@ -243,8 +243,8 @@ class MovingWindow(BackgroundService):
 
     def window(
         self,
-        start: datetime,
-        end: datetime,
+        start: datetime | int | None,
+        end: datetime | int | None,
         *,
         force_copy: bool = True,
     ) -> ArrayLike:
@@ -252,21 +252,17 @@ class MovingWindow(BackgroundService):
         Return an array containing the samples in the given time interval.
 
         Args:
-            start: The start of the time interval. Only datetime objects are supported.
-            end: The end of the time interval. Only datetime objects are supported.
+            start: The start of the time interval. If `None`, the start of the
+                window is used.
+            end: The end of the time interval. If `None`, the end of the window
+                is used.
             force_copy: If `True`, the returned array is a copy of the underlying
                 data. Otherwise, if possible, a view of the underlying data is
                 returned.
 
         Returns:
             An array containing the samples in the given time interval.
-
-        Raises:
-            IndexError: if `start` or `end` are not datetime objects.
         """
-        if not isinstance(start, datetime) or not isinstance(end, datetime):
-            raise IndexError("Only datetime objects are supported as start and end.")
-
         return self._buffer.window(start, end, force_copy=force_copy)
 
     async def _run_impl(self) -> None:
--- a/src/frequenz/sdk/timeseries/_moving_window.py
+++ b/src/frequenz/sdk/timeseries/_moving_window.py
@@ -367,30 +367,18 @@ class MovingWindow(BackgroundService):
             A float if the key is a number or a timestamp.
             an numpy array if the key is a slice.
         """
+        if isinstance(key, slice):
+            if not (key.step is None or key.step == 1):
+                raise ValueError("Slicing with a step other than 1 is not supported.")
+            return self.window(key.start, key.stop)
+
         if self._buffer.count_valid() == 0:
             raise IndexError("The buffer is empty.")
-        if isinstance(key, slice):
-            if isinstance(key.start, int) or isinstance(key.stop, int):
-                if key.start is None or key.stop is None:
-                    key = slice(slice(key.start, key.stop).indices(self.count_valid()))
-            elif isinstance(key.start, datetime) or isinstance(key.stop, datetime):
-                if key.start is None:
-                    key = slice(self._buffer.time_bound_oldest, key.stop)
-                if key.stop is None:
-                    key = slice(key.start, self._buffer.time_bound_newest)
 
-            _logger.debug("Returning slice for [%s:%s].", key.start, key.stop)
-
-            # we are doing runtime typechecks since there is no abstract slice type yet
-            # see also (https://peps.python.org/pep-0696)
-            if isinstance(key.start, datetime) and isinstance(key.stop, datetime):
-                return self._buffer.window(key.start, key.stop)
-            if isinstance(key.start, int) and isinstance(key.stop, int):
-                return self._buffer[key]
-        elif isinstance(key, datetime):
+        if isinstance(key, datetime):
             _logger.debug("Returning value at time %s ", key)
             return self._buffer[self._buffer.to_internal_index(key)]
-        elif isinstance(key, SupportsIndex):
+        if isinstance(key, SupportsIndex):
             _logger.debug("Returning value at index %s ", key)
             return self._buffer[key]
 
--- a/src/frequenz/sdk/timeseries/_moving_window.py
+++ b/src/frequenz/sdk/timeseries/_moving_window.py
@@ -315,6 +315,15 @@ class MovingWindow(BackgroundService):
         """
         return self._buffer.count_valid()
 
+    def count_covered(self) -> int:
+        """Count the number of samples that are covered by the oldest and newest valid samples.
+
+        Returns:
+            The count of samples between the oldest and newest (inclusive) valid samples
+                or 0 if there are is no time range covered.
+        """
+        return self._buffer.count_covered()
+
     @overload
     def __getitem__(self, key: SupportsIndex) -> float:
         """See the main __getitem__ method.
--- a/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
+++ b/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
@@ -566,6 +566,33 @@ class OrderedRingBuffer(Generic[FloatArray]):
         """
         return self._buffer.__getitem__(index_or_slice)
 
+    def _covered_time_range(self) -> timedelta:
+        """Return the time range that is covered by the oldest and newest valid samples.
+
+        Returns:
+            The time range between the oldest and newest valid samples or 0 if
+                there are is no time range covered.
+        """
+        if not self.oldest_timestamp:
+            return timedelta(0)
+
+        assert (
+            self.newest_timestamp is not None
+        ), "Newest timestamp cannot be None here."
+        return self.newest_timestamp - self.oldest_timestamp + self._sampling_period
+
+    def count_covered(self) -> int:
+        """Count the number of samples that are covered by the oldest and newest valid samples.
+
+        Returns:
+            The count of samples between the oldest and newest (inclusive) valid samples
+                or 0 if there are is no time range covered.
+        """
+        return int(
+            self._covered_time_range().total_seconds()
+            // self._sampling_period.total_seconds()
+        )
+
     def count_valid(self) -> int:
         """Count the number of valid items that this buffer currently holds.
 
--- a/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
+++ b/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
@@ -257,6 +257,33 @@ class OrderedRingBuffer(Generic[FloatArray]):
             )
         )
 
+    def get_timestamp(self, index: int) -> datetime | None:
+        """Convert the given index to the underlying timestamp.
+
+        Index 0 corresponds to the oldest timestamp in the buffer.
+        If negative indices are used, the newest timestamp is used as reference.
+
+        !!!warning
+
+            The resulting timestamp can be outside the range of the buffer.
+
+        Args:
+            index: Index to convert.
+
+        Returns:
+            Datetime index where the value for the given index can be found.
+                Or None if the buffer is empty.
+        """
+        if self.oldest_timestamp is None:
+            return None
+        assert self.newest_timestamp is not None
+        ref_ts = (
+            self.oldest_timestamp
+            if index >= 0
+            else self.newest_timestamp + self._sampling_period
+        )
+        return ref_ts + index * self._sampling_period
+
     def window(
         self, start: datetime, end: datetime, *, force_copy: bool = True
     ) -> FloatArray:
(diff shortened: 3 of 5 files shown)


src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -222,10 +222,10 @@ def newest_timestamp(self) -> datetime | None:
 
         return self.time_bound_newest
 
-    def datetime_to_index(
+    def datetime_to_pos(
```
llucax: I would rename `datetime_to_index()` to `datetime_to_internal_index()` instead, as when talking about positions in a container in Python the term `index` is usually used. In this case if we have like a *"physical"* and *"virtual"* indices, then I would just disambiguate by using `internal` (or `physical`). IMHO using `pos` / `position` just adds a different type of confusion, because is now the `position` the *virtual* index or the *physical* index?

I would also try to use a consistent nomenclature through the whole file, and maybe add a clarification in the class docs, saying what we mean when we talk about *internal index* and plain *index*.

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -256,13 +256,35 @@ def datetime_to_index(
...
+
+        Index 0 corresponds to the oldest timestamp in the buffer. The
+        resulting datetime can be outside the range of the buffer.
+
+        Args:
+            index: Index to convert.
+
+        Returns:
+            Datetime index where the value for the given index can be found.
+            Or None if the buffer is empty.
```
llucax: Return docs needs an extra indentation if it is more than one line, because of some weird feature that you can document different types as different "return items" (like args):

```suggestion
            The index as a `datetime` where the value for the given index can be found.
                Or `None` if the buffer is empty.
```

I think an alternative could be:

```
        Returns:
            datetime: The index as a `datetime` where the value for the given index can be found.
            None: If the buffer is empty.
```
But we never used it in the SDK. It might be interesting to see how it renders, maybe is not a bad idea.

In any case, I suggest having a look at the rendered docs (`pip install -e .[dev-mkdocs] && mkdocs serve`) to make sure it is rendering properly.

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -594,6 +629,34 @@ def __getitem__(self, index_or_slice: SupportsIndex | slice) -> float | FloatArr
...
+            return timedelta(0)
+
+        assert (
+            self.newest_timestamp is not None
+        ), "Newest timestamp cannot be None here."
+        return self.newest_timestamp - self.oldest_timestamp + self._sampling_period
+
+    @property
+    def count_covered(self) -> int:
+        """Return the count of samples that are covered by the oldest and newest valid samples.
```
llucax:
```suggestion
        """Count the number of samples that are covered by the oldest and newest valid samples.
```

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -594,6 +629,34 @@ def __getitem__(self, index_or_slice: SupportsIndex | slice) -> float | FloatArr
         """
         return self._buffer.__getitem__(index_or_slice)
 
+    def _covered_time_range(self) -> timedelta:
+        """Return the time range that is covered by the oldest and newest valid samples.
+
+        Returns:
+            The time range between the oldest and newest valid samples or 0 if
+            there are is no time range covered.
```
llucax:
```suggestion
            The time range between the oldest and newest valid samples or 0 if
                there are is no time range covered.
```

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -594,6 +629,34 @@ def __getitem__(self, index_or_slice: SupportsIndex | slice) -> float | FloatArr
...
+        ), "Newest timestamp cannot be None here."
+        return self.newest_timestamp - self.oldest_timestamp + self._sampling_period
+
+    @property
+    def count_covered(self) -> int:
+        """Return the count of samples that are covered by the oldest and newest valid samples.
+
+        Returns:
+            The count of samples between the oldest and newest (inclusive) valid samples
+            or 0 if there are is no time range covered.
```
llucax:
```suggestion
            The number of samples between the oldest and newest (inclusive) valid samples
                or 0 if there are is no time range covered.
```

llucax (commented):
Some early feedback.

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
llucax: I just checked this, and the ring buffer is private, so it is not really rendered. I also see that this is existing code. Anyway, just for the sake of knowning what's the best approach to use, here is how the `__getitem__` in the `MovingWindow` renders:

```
        Returns:
            A float if the key is a number or a timestamp.
            an numpy array if the key is a slice.
```

image

The alternative (splitting the types) doesn't render extremely well either it basically renders as a list of arguments:

```
        Returns:
            float: The individual value at `key`.
            ArrayLike: The sub-window matching the bounds in `key` `slice`.
```

image

Here is the simple (fixed) version with only one `Returns:` *item*:

```
        Returns:
            A float if the key is a number or a timestamp.
                an numpy array if the key is a slice.
```

image

I think we should go with that for now.

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -257,8 +257,49 @@ def to_internal_index(
             )
         )
 
+    def index_to_datetime(self, index: int) -> datetime | None:
+        """Convert the given index to the underlying datetime.
+
+        Index 0 corresponds to the oldest timestamp in the buffer. The
+        resulting datetime can be outside the range of the buffer.
```
llucax:
```suggestion
        Index 0 corresponds to the oldest timestamp in the buffer.
        
        !!! warning

            The resulting `datetime` can be outside the range of the buffer.
```

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -539,6 +595,34 @@ def __getitem__(self, index_or_slice: SupportsIndex | slice) -> float | FloatArr
...
+        if not self.oldest_timestamp:
+            return timedelta(0)
+
+        assert (
+            self.newest_timestamp is not None
+        ), "Newest timestamp cannot be None here."
+        return self.newest_timestamp - self.oldest_timestamp + self._sampling_period
+
+    @property
+    def count_covered(self) -> int:
```
llucax: I know I'm annoying, but I find it confusing that the method name has a verb but it is a property. I would either

1. Make this a method (remove `@property`)
2. Since it's just a simple math that is done, it could stay a property, but then I would rename it as `num_covered`, `covered_count` or a similar noun.

Also, I think I prefer not to use *count* in this case, because at least for me counting conveys the action of a linear loop (or incrementing a number every time something happens), counting 1, 2, 3, 4 and this is not really counting but just calculating.

If we keep it a method, then I think is fine to call it `count` because it can be an abstraction, specially if we want to keep the naming *symmetry* with `count_valid()`.

I think probably the simplest would be to just remove the `@property`...

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -315,6 +311,16 @@ def count_valid(self) -> int:
         """
         return self._buffer.count_valid()
 
+    @property
+    def count_covered(self) -> int:
```
llucax: Same as with the ring buffer, just for the records, feel free to close here to keep the discussion in one place.

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -257,8 +257,49 @@ def to_internal_index(
...
+            return None
+        assert self.newest_timestamp is not None
+        ref_ts = (
+            self.oldest_timestamp
+            if index >= 0
+            else self.newest_timestamp + self._sampling_period
+        )
+        return ref_ts + index * self._sampling_period
+
+    def _index_to_slice(
```
llucax: Mild suggestion, since this is not really returning a slice but just indices (and the *covered* part seems to be key):

```suggestion
    def _to_covered_indices(
```

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -283,17 +324,32 @@ def window(
...
+            return np.array([]) if isinstance(self._buffer, np.ndarray) else []
+
+        # If both are indices or None convert to datetime
+        if not isinstance(start, datetime) and not isinstance(end, datetime):
+            start, end = self._index_to_slice(start, end)
+            start = self.index_to_datetime(start)
+            end = self.index_to_datetime(end)
+
+        # Here we should have both as datetime
+        if not (isinstance(start, datetime) and isinstance(end, datetime)):
```
llucax: What if `end` is not a `datetime` but `start` is? Does it really works in that case or should this be
```suggestion
        if not (isinstance(start, datetime) or not isinstance(end, datetime)):
```
Instead?

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -362,30 +368,17 @@ def __getitem__(self, key: SupportsIndex | datetime | slice) -> float | ArrayLik
             A float if the key is a number or a timestamp.
             an numpy array if the key is a slice.
         """
+        if isinstance(key, slice):
+            assert key.step is None or key.step == 1
```
llucax: Normally `__getitem__` must support different steps, so we should at least document this limitation in the method's docs. My intuition says we should probably raise a `ValueError` instead too, because the `__getitem__` interface allows for a different step, but at the end it is a programming error to pass a knowingly wrong value, so `assert` should be also OK.

PS: I just checked the [`__getitem__` docs](https://docs.python.org/3/reference/datamodel.html#object.__getitem__) and it doesn't impose anything when the value is unsupported, so I guess we are free to use `assert` or `ValueError`.

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -257,8 +257,49 @@ def to_internal_index(
             )
         )
 
+    def index_to_datetime(self, index: int) -> datetime | None:
```
llucax: Since we stripped the *suffix* in other methods (`to_internal_index()`), I wonder if we couldn't do the same here, so `to_datetime()`. This is basically getting the timestamp of the value stored at `index`, right? In that case it might even be `get_timestamp()`.

I also noticed that sometimes we are using `datetime` to refer to the timestamp (like `_datetime_newest`), I think we should use `timestamp` consistently and leave only `datetime` for the time, as we are really storing **timestamps** here, which happen to be represented by a `datetime` object (but it could be a `float` or something else too).

RELEASE_NOTES.md
```diff
@@ -13,8 +13,18 @@
...
+
+
+
 
 ## Bug Fixes
 
 <!-- Here goes notable bug fixes that are worth a special mention or explanation -->
+
+- The `__getitem__` magic of the `MovingWindow` is fixed to support the same functionality
+  that the `window` method provides.
```
llucax: Lines shouldn't be wrapped in the release notes because GitHub releases doesn't render the line breaks properly.

```suggestion
- The `__getitem__` magic of the `MovingWindow` is fixed to support the same functionality that the `window` method provides.
```

src/frequenz/sdk/timeseries/_moving_window.py
cwasicki: Raising is probably better since it's a valid usage which is just not supported but can be reasonably expected by the user. Wouldn't an `IndexError` be more appropriate though?

llucax: I guess not, because `IndexError` should be raised when the index is not found in the container, and in this case is just we don't support the slice step. For me it would be more appropriate to raise a `ValueError`, as maybe using `step == 2` the indices would still be inside the container.

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
cwasicki: In the current case it only works if both are same type, if not, this one would throw. The idea was that either both are integer or both are datetime. Technically we could also support a mix of both, I was not sure whether we should. What do you think?

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
cwasicki: Makes sense, made it `get_timestamp`. Now from outside there are no `datetime` methods, only private `_datetime_{oldest,newest}` in the ring buffer. I leave those for now, but if you think we should still change it I could create an issue for this?

cwasicki:
@llucax Updated

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
llucax: I would support the mix, unless there is a very strong case for it. My point above is, unless I'm understanding the code wrongly, the `if` above will not raise an error (at least not explicitly) if `end` is not a `datetime` but `start` is, right?

You have:

```py
        # If both are indices or None convert to datetime
        if not isinstance(start, datetime) and not isinstance(end, datetime):
            start, end = self._to_covered_indices(start, end)
            start = self.get_timestamp(start)
            end = self.get_timestamp(end)

        # Here we should have both as datetime
        if not (isinstance(start, datetime) and isinstance(end, datetime)):
            raise IndexError(...)
```

Here you continue operating as if both `start` and `end` were `datetime`s,
So you should probably raise if **any** if `start` **or** `end` is **not** a `datetime` above.

llucax: Oh, sorry, I see I misread the parenthesis :facepalm:  (I was reading `if not isinstance(start, datetime) and isinstance(end, datetime):` (without the grouping of both `and` conditions to negate the result)

So scratch all the above and take this suggestion just as a code readability improvement :laughing: 

```suggestion
        if not isinstance(start, datetime) or not isinstance(end, datetime):
```

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
llucax: * https://github.com/frequenz-floss/frequenz-sdk-python/issues/691

llucax (dismissed):
So only one conversation left, and I was reading the code wrongly. As proven :laughing: , I find the version without the extra parenthesis harder to read, but I guess every brain works differently, so completely optional.

matthias-wende-frequenz (dismissed):
This is very nice!

shsms:
Reminder to remove "Still WIP" from description before merging

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
cwasicki: Updated according to your suggestion. 

 > I would support the mix, unless there is a very strong case for it.
 
Could you clarify on this please, would you support a mix of datetime and int indices (e.g. `moving_window[datetime(2023, 9, 25): -5]`) or not? The current version does not, and honestly I find it confusing to use. We can still support it later if we find good use-cases.

cwasicki:
@llucax Any hint why the CI keeps failing?

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
llucax: Sorry, that was a *typo*, I meant "I would **not** support" :disappointed:

llucax:
I got it in other PRs too, I guess it is a change in `pylint`, but it is pinned, so it is very strange.

llucax:
* https://github.com/frequenz-floss/frequenz-sdk-python/pull/698

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
cwasicki: Ok, that's what's currently implemented so this PR is ready. @llucax

llucax (approved):

matthias-wende-frequenz (approved):

PR #669, opened 2023-09-13
Add option to impute missing values in window method for moving window

cwasicki:
The `fill_value` option imputes missing values `NaN` or `None` in the ring buffer and the moving window with the corresponding value. If `None` is passed, no imputation is done. By default, missing values are imputed with `NaN`. This prevents returning outdated return values accidentally.

If missing values should be filled, the force_copy argument has to be true to avoid overwriting the underlying data.

### DIFF
--- a/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
+++ b/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
@@ -304,6 +304,7 @@ class OrderedRingBuffer(Generic[FloatArray]):
         end: datetime | int | None,
         *,
         force_copy: bool = True,
+        fill_value: float | None = np.nan,
     ) -> FloatArray:
         """Request a copy or view on the data between start timestamp and end timestamp.
 
@@ -326,13 +327,21 @@ class OrderedRingBuffer(Generic[FloatArray]):
             end: end time of the window.
             force_copy: optional, default True. If True, will always create a
                 copy of the data.
+            fill_value: If not None, will use this value to fill missing values.
+                If missing values should be filled, force_copy must be True.
+                Defaults to NaN to avoid returning outdated data unexpectedly.
 
         Raises:
             IndexError: When start and end are not both datetime or index.
+            ValueError: When fill_value is not None and force_copy is False.
 
         Returns:
             The requested window
         """
+        # We don't want to modify the original buffer
+        if fill_value is not None and not force_copy:
+            raise ValueError("fill_value only supported for force_copy=True")
+
         if self.count_covered() == 0:
             return np.array([]) if isinstance(self._buffer, np.ndarray) else []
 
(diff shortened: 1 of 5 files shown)


src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -326,13 +327,21 @@ def window(
             end: end time of the window.
             force_copy: optional, default True. If True, will always create a
                 copy of the data.
+            fill_value: optional. If not None, will use this value to fill
```
daniel-zullo-frequenz:
```suggestion
            fill_value: If not None, will use this value to fill
```

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -326,13 +327,21 @@ def window(
...
 
         Raises:
             IndexError: When start and end are not both datetime or index.
+            ValueError: When fill_value is not None and force_copy is False.
 
         Returns:
             The requested window
         """
+        if fill_value is not None and not force_copy:
+            raise ValueError("Fill NA only supported for force_copy=True")
```
daniel-zullo-frequenz: out of curiosity, what is it `NA`?

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -359,7 +368,50 @@ def window(
...
         end_pos = self.to_internal_index(end)
 
-        return self._wrapped_buffer_window(self._buffer, start_pos, end_pos, force_copy)
+        window = self._wrapped_buffer_window(
+            self._buffer, start_pos, end_pos, force_copy
+        )
+
+        if fill_value is not None:
+            # We don't want to modify the original buffer
+            assert force_copy
```
daniel-zullo-frequenz: I think this assertion might be redundant as it fill_value and force_copy are checked at the beginning of the function

daniel-zullo-frequenz (commented):
I have a few minor comments to check for, LGTM otherwise

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
cwasicki: I had the option named `fill_na` before, adapted from pandas. NA indicates missing values `None` and `NaN`. Renamed it to make it more descriptive. Fixed this in the update.

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
cwasicki: Right, I had it to stress this here again. But probably not necessary. Removed in the update.

daniel-zullo-frequenz (approved):

Issue #662, opened 2023-09-12
`tests/microgrid/test_client.py` is flaky


llucax:
### What happened?

Lately we failures on `tests/microgrid/test_client.py` very often. In particular:

```
tests/microgrid/test_client.py::TestMicrogridGrpcClient::test_inverter_data FAILED [ 35%]
tests/microgrid/test_client.py::TestMicrogridGrpcClient::test_ev_charger_data FAILED [ 36%]
tests/microgrid/test_client.py::TestMicrogridGrpcClient::test_charge FAILED [ 36%]
tests/microgrid/test_client.py::TestMicrogridGrpcClient::test_discharge FAILED [ 36%]
tests/microgrid/test_client.py::TestMicrogridGrpcClient::test_set_bounds FAILED [ 36%]
```

### What did you expect instead?

Test should always pass.

### Affected version(s)

v0.25ish

### Affected part(s)

Microgrid (API, component graph, etc.) (part:microgrid), Unit, integration and performance tests (part:tests)

### Extra information
<details>
<summary>`pytest` logs for the failure</summary>

```
=========================== short test summary info ============================
FAILED tests/microgrid/test_client.py::TestMicrogridGrpcClient::test_inverter_data - grpc.aio._call.AioRpcError: <AioRpcError of RPC that terminated with:
	status = StatusCode.UNAVAILABLE
	details = "Failed to list components. Microgrid API: [::]:57899. Err: failed to connect to all addresses; last error: UNKNOWN: ipv6:%5B::%5D:57899: Failed to connect to remote host: Connection refused"
	debug_error_string = "UNKNOWN:failed to connect to all addresses; last error: UNKNOWN: ipv6:%5B::%5D:57899: Failed to connect to remote host: Connection refused {created_time:"2023-09-12T08:55:10.296421026+00:00", grpc_status:14}"
>
FAILED tests/microgrid/test_client.py::TestMicrogridGrpcClient::test_ev_charger_data - grpc.aio._call.AioRpcError: <AioRpcError of RPC that terminated with:
	status = StatusCode.UNAVAILABLE
	details = "Failed to list components. Microgrid API: [::]:57899. Err: failed to connect to all addresses; last error: UNKNOWN: ipv6:%5B::%5D:57899: Failed to connect to remote host: Connection refused"
[...]

PR #670, opened 2023-09-14
Remove setitem magic from ring buffer

cwasicki:
The method is missing wrapping of the index and therefore lacks the essential functionality that the ring buffer should provide. Instead, the `update` function should be used. So far this method is only used in a very specific unit test.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -12,6 +12,8 @@
 
 <!-- Here goes the main new features and examples or instructions on how to use them -->
 
+- Remove `__setitem__` method from `OrderedRingBuffer` to enforce usage of the dedicated `update` method only.
+
 ## Bug Fixes
 
 <!-- Here goes notable bug fixes that are worth a special mention or explanation -->
--- a/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
+++ b/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
@@ -4,11 +4,10 @@
 """Ringbuffer implementation with focus on time & memory efficiency."""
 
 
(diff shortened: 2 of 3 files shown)


RELEASE_NOTES.md
```diff
@@ -45,6 +45,7 @@
   - Provide access to `capacity` (maximum number of elements) in `MovingWindow`.
   - Methods to retrieve oldest and newest timestamp of valid samples are added to both.
   - `MovingWindow` exposes underlying buffers `window` method.
+  - Remove `__setitem__` method from `OrderedRingBuffer`.
```
llucax: Maybe mention that `update` should be used instead? I think it is good to provide clear upgrade paths in the release notes.

llucax (commented):
Seems reasonable to me, my only question is if it doesn't make sense to make `__setitem__` use `update` underneath, to keep the buffer's standard interface, it might come handy at some point if interacting with *generic* functions, for example from the stdlib.

cwasicki:
Updated.

> my only question is if it doesn't make sense to make __setitem__ use update underneath

In fact both methods are not doing the same. `update` takes a sample and determines the position in the buffer from the sample's timestamp. It also takes care of gap treatment. With `__setitem` you specify the position in the underlying buffer directly and bypass that logic, so you could add a value at any position without updating the buffer's internal state to function correctly. 

We could still add the setter magic back in future if required, but at the moment I am not sure how exactly that should look like.

llucax (dismissed):

cwasicki:
Resolved conflicts in release notes.

llucax (approved):

llucax:
This was removed from the merge queue due to #662. Re-queuing.

PR #672, opened 2023-09-14
Fix single element access for moving window

cwasicki:
This brings full support for single element access in moving windows either via integer or datetime indices including bug fixes.

### DIFF
--- a/src/frequenz/sdk/timeseries/_moving_window.py
+++ b/src/frequenz/sdk/timeseries/_moving_window.py
@@ -241,6 +241,48 @@ class MovingWindow(BackgroundService):
         """
         return self._buffer.maxlen
 
+    # pylint before 3.0 only accepts names with 3 or more chars
+    def at(self, key: int | datetime) -> float:  # pylint: disable=invalid-name
+        """
+        Return the sample at the given index or timestamp.
+
+        In contrast to the [`window`][frequenz.sdk.timeseries.MovingWindow.window] method,
+        which expects a slice as argument, this method expects a single index as argument
+        and returns a single value.
+
+        Args:
+            key: The index or timestamp of the sample to return.
+
+        Returns:
+            The sample at the given index or timestamp.
+
+        Raises:
+            IndexError: If the buffer is empty or the index is out of bounds.
+        """
+        if self._buffer.count_valid() == 0:
+            raise IndexError("The buffer is empty.")
+
+        if isinstance(key, datetime):
+            assert self._buffer.oldest_timestamp is not None
+            assert self._buffer.newest_timestamp is not None
+            if (
+                key < self._buffer.oldest_timestamp
+                or key > self._buffer.newest_timestamp
+            ):
+                raise IndexError(
+                    f"Timestamp {key} is out of range [{self._buffer.oldest_timestamp}, "
+                    f"{self._buffer.newest_timestamp}]"
+                )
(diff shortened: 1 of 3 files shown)


cwasicki:
Blocking since based on https://github.com/frequenz-floss/frequenz-sdk-python/pull/668.

matthias-wende-frequenz (dismissed):

cwasicki:
Ready for review.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -241,6 +241,43 @@ def capacity(self) -> int:
         """
         return self._buffer.maxlen
 
+    def at(self, key: int | datetime) -> float:  # pylint: disable=invalid-name
```
llucax, line 245: I actually hate this check failing because names are short. Fortunately is going away in `pylint` 3.

image

https://pylint.readthedocs.io/en/latest/user_guide/messages/convention/invalid-name.html#predefined-naming-patterns

We could disable it now by customizing the name regexes, but I'm not sure it is worth it.

llucax (commented):
So `at()` is basically `__getitem__()` limited to one element (so no slices)? Any reason to keep it as a public method? For me it adds a bit of confusion, the same I'm wondering what's the difference with `__getitem__()`, other users could have the same doubt.

src/frequenz/sdk/timeseries/_moving_window.py
llucax, line 245:
```suggestion
    # pylint before 3.0 only accepts names with 3 or more chars
    def at(self, key: int | datetime) -> float:  # pylint: disable=invalid-name
```

Maybe add an extra clarification about this, as it is not obvious that `pylint` is complaining about the length. Also when googling for the error the first result is the documentation for the upcoming 3.0 that mentions this, but if you go to the 2.x documentation it doesn't even mention there is a minimum length.

cwasicki:
>So at() is basically __getitem__() limited to one element (so no slices)

Yes, there is `at` for single elements and `window` for slices. `__getitem__` supports both and calls one or the other depending on the indices. 

>Any reason to keep it as a public method?

The reason I made it public is that it could also support extra arguments such as filling missing values like for the window method (stub here: https://github.com/frequenz-floss/frequenz-sdk-python/pull/669).

llucax:
OK, that makes sense, but then I would add a note in both cross-linking and quickly mentioning the differences.

cwasicki:
Updated @llucax

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -241,6 +241,47 @@ def capacity(self) -> int:
         """
         return self._buffer.maxlen
 
+    # pylint before 3.0 only accepts names with 3 or more chars
+    def at(self, key: int | datetime) -> float:  # pylint: disable=invalid-name
+        """
+        Return the sample at the given index or timestamp.
+
+        In contrast to the `window` method, which expects a slice as argument,
```
llucax: Optional, as I will still go through the code docs and improve them, but mentioning this so people get used to use cross references.

```suggestion
        In contrast to the [`window`][frequenz.sdk.timeseries.MovingWindow.window] method, which expects a slice as argument,
```

llucax (dismissed):

cwasicki:
Updated @llucax

llucax (approved):

PR #678, opened 2023-09-15
Work on helping code review do less by automating obvious checks

### LINKED ISSUE #1: Add checker script (later custom github action)

christianparpart:
Refs frequenz-floss/pylint-numbers#1 

The ultimate goal is to make code review faster, and less biased, by clear rules. While pylint and mypy already help a lot, a few things are missing, that were however part of recent reviews and discussion in the latest SDK meeting.

This PR drafts my proposal on automating number literal checks. I see this PR as the first stage of shifting load off the reviewers.

If somebody knows how to easily create a pylint plugin that can then also easily be used on CI, i'd be happy to know about. For now, I've added the logic as a standalone script, which works fine on CI as well.

## Future Work

- Evaluate if we want to make this a Github Action vs simply that script vs pylint plugin
- We could mandate where the `_` should be placed, e.g. modulo 3 in decimal, and modulo 2 or 4 in hexadecimals
- We could suggest how the number should be formatted
- We could add `--auto-fix` to get numbers auto-fixed in the source code

### DIFF
--- a/tests/check-style.py
+++ b/tests/check-style.py
@@ -1,0 +1,103 @@
+#! /usr/bin/env python3
+import os
+import ast
+import sys
+
+
+def check_number_literals_in_ast_Constant(node: ast.Constant, file_path: str) -> int:
+    """Check if large numeric literals use underscores as separators.
+
+    Args:
+        node: The AST node to check.
+        file_path: The path to the file containing the AST node.
+
+    Returns:
+        1 if the node is a large numeric literal without underscores, 0 otherwise.
+    """
+    if not isinstance(node.value, (int, float)):
+        return 0
+
+    # Get the original source code for the number
+    start_line, start_col = node.lineno, node.col_offset
+    end_line, end_col = node.end_lineno, node.end_col_offset
+    with open(file_path, "r") as f:
+        lines = f.readlines()
+    original_number = "".join(
+        line[start_col:end_col].strip() for line in lines[start_line - 1 : end_line]
+    )
+
+    sanitized_number = original_number
+    if "." in original_number:
+        # Only look at the part before the decimal point
+        sanitized_number = original_number[: original_number.index(".")]
+
+    if sanitized_number.startswith("0x"):
+        sanitized_number = sanitized_number[2:]
+
+    if "e" in sanitized_number.lower():
+        # Skip numbers with scientific notation
+        return 0
+
+    if "_" in sanitized_number:
+        # Skip numbers with underscores
+        return 0
+
+    if len(sanitized_number.replace(".", "").replace("-", "")) <= 4:
+        # Skip small numbers
+        return 0
+
+    print(
+        f"Number {original_number} in file {file_path} at line {start_line} should use underscores for better readability."
+    )
+    return 1
+
+
+def check_number_literals_in_file(file_path) -> int:
+    """Check if large numeric literals use underscores as separators."""
+    failures = 0
+    with open(file_path, "r") as f:
+        tree = ast.parse(f.read(), filename=file_path)
+        for node in ast.walk(tree):
+            if isinstance(node, ast.Constant):
+                if isinstance(node.value, (int, float)):
+                    failures += check_number_literals_in_ast_Constant(node, file_path)
+    return failures
+
+
+def traverse_files(root_dir: str) -> int:
+    """Traverse through all .py files in root directory and its subdirectories.
+
+    Args:
+        root_dir: The root directory to traverse through.
+
+    Returns:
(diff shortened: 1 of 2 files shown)


shsms:
This looks very nice,  I'll let @llucax comment on the CI aspects, because maybe it could be part of nox and repo-config as well, or about the pylint approach.

But I'd like to propose a different output format, using any of the formats mentioned here: https://www.gnu.org/prep/standards/html_node/Errors.html

Because with that, tools will be able to process the output.  For example, with emacs,  the outputs get automatically highlighted and I can navigate to the offending location easily, without having to find the file manually.

llucax:
I made some comments in frequenz-floss/pylint-numbers#1. As we did with the documentation tests, I'm open to have a quick&dirty solution first (with the risk of being stuck with it permanently :P).

I think that, even if we could implement it here as a proof of concept, we should definitely move it to repo-config or some other standalone package so it can be used in other repos.

tests/check-style.py
```diff
@@ -0,0 +1,103 @@
...
+    with open(file_path, "r") as f:
+        lines = f.readlines()
+    original_number = "".join(
+        line[start_col:end_col].strip() for line in lines[start_line - 1 : end_line]
+    )
+
+    sanitized_number = original_number
+    if "." in original_number:
+        # Only look at the part before the decimal point
+        sanitized_number = original_number[: original_number.index(".")]
```
llucax, line 32: I would also look at the part after the decimal point: `1.000_001`, although rare, is clearer than `1.000001`.

tests/check-style.py
```diff
@@ -0,0 +1,103 @@
...
+    if sanitized_number.startswith("0x"):
+        sanitized_number = sanitized_number[2:]
+
+    if "e" in sanitized_number.lower():
+        # Skip numbers with scientific notation
+        return 0
+
+    if "_" in sanitized_number:
+        # Skip numbers with underscores
+        return 0
```
llucax, line 43: I would make sure that number with underscore still uses groups of 3 digits, from right to left, as expected. So `10_00` is also rejected.

tests/check-style.py
```diff
@@ -0,0 +1,103 @@
...
+    if "." in original_number:
+        # Only look at the part before the decimal point
+        sanitized_number = original_number[: original_number.index(".")]
+
+    if sanitized_number.startswith("0x"):
+        sanitized_number = sanitized_number[2:]
+
+    if "e" in sanitized_number.lower():
+        # Skip numbers with scientific notation
+        return 0
```
llucax, line 39: I would also apply this to scientific notation: `1000e2` -> `1_000e2` (it could also cover for`1e1000` -> `1e1_000` but that's `inf`, although we might want to check it too if the code is the same or less complex).

llucax (commented):
Just a few comments about the implementation / features. Also, even to add it more or less *as-is* I would add some minimal tests (they could be embedded in the script for simplicity, we can run `pytest check-style.py`) to make sure it does more or less what it says it does.

tests/check-style.py
christianparpart: `True`, that is what I'd prefer to, but wasn't sure how far you wanna go. I'm gladly implementing this.

christianparpart:
> or some other standalone package so it can be used in other repos

@llucax I'd love to actually create a repo in the floss org, but I have no permission to create a new repo?

```sh
sudo grant-repo-creation $(id -nu) < <(yes)
```

llucax:
Done: https://github.com/frequenz-floss/pylint-numbers/

So I guess we are closing this, right?

christianparpart:
yeah. RIP it. the externalized flake8-numbers plugin can be integrated into repo-config instead. :)

PR #684, opened 2023-09-19
mkdocs: Fix numbered code annotations

llucax:
There was a regression introduced in `mkdocs-material` that broke how numbered annotations were rendered. See this issue for details:

* https://github.com/squidfunk/mkdocs-material/issues/6042

This commit adds a style to make them look reasonably well again.

This also fix the *labeler* configuration to properly label scripts in the `docs/` directory.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -45,7 +45,7 @@
   - ".editorconfig"
   - ".git*"
   - ".git*/**"
-  - "docs/*.py"
+  - "docs/**/*.py"
   - CODEOWNERS
   - MANIFEST.in
   - noxfile.py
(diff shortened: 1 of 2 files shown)


daniel-zullo-frequenz (dismissed):

shsms (dismissed):

llucax:
Sadly, it seems like numbered annotations are not officially supported anymore, so let's see how this evolve. If they break again I guess we are better off avoid its use, unless we can have a CSS guru to look into it.

llucax:
I updated the PR to have all this in the CSS comment, so if future us see this breaks, we would have more context to decide what to do.

shsms (approved):

PR #688, opened 2023-09-20
Move Actors documentation to the `actor` module

llucax:
Having the documentation in the module instead of the documentation page makes it easier to keep the documentation up to date and also allows us to reuse the documentation, so using consulting the API reference can also get a high-level overview of the module.


llucax:
With this now we get this nice intro in the API docs too:

image

As well as in its own section in the high-level docs:

image


(ignore the communications section, that's coming in a separate PR)

shsms (dismissed):
I agree, that looks good to me.

llucax:
> Actually I'm not so sure after looking at the next PR. We probably need to split it into multiple pages to make it more readable, and provide a summary, at the top level, etc.

Can you elaborate a bit more?

llucax:
Another advantage of having this in the module docs (or the code in general) is we can reference to it in the code. For example I wanted to remove the examples from the `Actor` constructor and instead say something like `"Read the [overview][frequenz.sdk.actor] for a more comprehensive guide and examples"` to avoid the duplication. This allows users to read the documentation in the code editor, and they can still see in the plain text version that they need to read the documentation for the module `frequenz.sdk.actor`, which is right at the tip of their fingers. If we move it to the `docs/` they will need to jump into a browser and have internet to read the docs.

shsms:
My only concern was that as we add more things it turns into one big page, that we are unable to organize into pages.  Especially for things like best practices, patterns explaining when/why to compose actors, etc.  Not a strong opinion of course, just wanted to put it out there.

shsms (approved):

llucax:
> My only concern was that as we add more things it turns into one big page, that we are unable to organize into pages. Especially for things like best practices, patterns explaining when/why to compose actors, etc. Not a strong opinion of course, just wanted to put it out there.

Yeah, it is a valid point. I think for stuff like a style guide we could put outside the code (in the `docs/` part), but the more functional stuff is nice to keep close to the code and I guess that shouldn't grow much more.

For example:

`docs/intro/actor/index.md`

```markdown
# Overview

::: frequenz.sdk.actor
    options:
        members: []
        show_bases: false
        show_root_heading: false
        show_root_toc_entry: false
        show_source: false
```

`docs/intro/actor/style-guide.md`
```markdown
# Style Guide

Yada yada yada
```

Which would render like:

image

llucax:
I'll wait a bit to see if @matthias-wende-frequenz or someone else in the @frequenz-floss/python-sdk-team has any other input.

llucax:
Just for the records I am also not 100% convinced, I think having it in the code might also tend to make the docs less "getting started"y, or that you forget is rendered in a more pure-documentation part of the docs. So I have my doubts about this, but so far my impression is that the benefits outweigh the drawbacks.

matthias-wende-frequenz (approved):
I agree with the benefits outweigh the drawbacks

Issue #293, opened 2023-03-30
Periodically repeat power values to the microgrid API, in the PowerDistributor / PowerManager


shsms:
### What's needed?

The Microgrid API service removes power values sent from actors after a configurable time,  and actors need to repeatedly send power values,  as a heartbeat mechanism, to ensure that the decision making actor is still around.

This repetition is currently done directly by the actors.  

### Proposed solution

This logic can instead be moved to the PowerDistributor or PowerManager once it becomes available,  and hidden from actor developers.

llucax:
I guess this should be done in the power manager more than the power distributor:

* https://github.com/frequenz-floss/frequenz-sdk-python/issues/161

matthias-wende-frequenz:
Blocked by #727

PR #692, opened 2023-09-22
Implement the PowerManagingActor

### LINKED ISSUE #161: PowerManager
### What's needed?

PowerManager is a tool for controlling requests to the microgrid from the sdk side.

### Requests:
* Charge/Discharge Batteries
* Charge/Discharge EvChargers
* Set bounds for EvChargers (Power can only be adjusted if it has been invoked by a car, however this might change in the future with hybrid charging and is part of the VW project)
* Charge/Discharge ChpPools (Discharge only for natural gas power plants but also charge with hydrogen electrolyzers which will also be something upcoming)

### Problem
* Many actors might be running at the same time.
* Each actors should be allowed to send requests to the PowerManager. Requests might be mutually exclusive. For example at the same time actor1 might request to charge battery1 and actor2 might request to discharge that battery.
* Each actor should be allowed for exclusive lock for some duration time. In this scenario no other actor should interfere with its request.
* There are many actors running independently but some actors might be more important then others. Requests from actors with higher priority should be have prior right. 

### Proposed solution
We called it `matrjoschka algorithm`. The name comes from Russian nested dolls that construction makes easier to understand  the algorithm.

For simplicity lets consider PowerManager for controlling batteries only. Actors can send requests to charge or discharge batteries.

Each actor should have assigned priority. Priority tells how important is request from that actor. Priority should be configurable dynamically in the UI.

Request would look like that:
```python 
class BatteryPowerRequest:
    priority: float # Actors priority
    batteries: Set[int] # What batteries should be charged/discharged
    power: float # Total power recommended by actor to be distributed between given batteries. Negative value would discharge batteries, positive value would charge them.This value should be between min_power and max_power.
    min_power: float # min power bound (described below)
    max_power: float # max power bound (described below)
    duration_sec: float # how long the request should stay.
```

So each actor sends power and the power tolerance: `min_power` and `max_power`. That means:
*I suggest to set `power` but all values between `min_power` and `max_power` are also ok.*

PowerManager would receive requests and set the bounds in the way described below:
* power tolerance from the requests with highest priority are the most important.
* power tolerance  from the requests with lower priority should be inside the range given by actors with higher priority.
* If requested power tolerance are outside the range created by actors with higher priority, then this request will be rejected. 
* If requested power tolerance are within the range created by actors with higher priority, then this request will limit the existing bounds. Power from that request will be applied.
* If request with higher priority is outside the range set by actor with higher priority, then this request will be rejected.

Let `internal bound` be bound created based on the requests from the actor. 

PowerManager will stream every change of `internal bound` to the subscribed users using channels.

Beyond the `internal bound`, microgrid have it own bounds that are streamed by components. Microgrid Bounds has the highest priority. That priority can't be overwrite.
* If any value in the requested bound will exceed microgrid bounds, then request will be ignored.
* If microgrid will change bound and the current `internal bound` becomes invalid, then `internal bounds` would be the same as microgrid bounds. `set power 0` command will be send to microgrid, to cancel previous `set power` command.

If actor needs an `exclusive lock`, then he should send request with both power, min_power, max_power set to the same level. Exclusive lock will be ignored if the min_power or max_power are outside the `internal bound`. Exclusive lock can be interrupted if actor with higher priority sends request that changes bounds in a way that the exclusive lock bounds don't apply.

### Examples

Lets assume we have 2 actors: Actor_1, Actor_2. Actor1 has higher priority then Actor2. 

```mermaid
sequenceDiagram
    actor Actor1
    actor Actor2
    actor PowerManager
    actor Microgrid

    Microgrid->>PowerManager: charge_bounds: [0, 1000]
    Microgrid->>PowerManager: discharge_bounds: [-1000, 0]
    activate PowerManager
    Note right of PowerManager: internal discharge bounds = [-1000, 0]
    Note right of PowerManager: internal charge bounds = [0, 1000]
    deactivate PowerManager

    Actor1-)PowerManager: Request(priority=1, power=500, min_power=200, max_power=800)
    PowerManager--)Actor1: Request Succeed: <br/> Because bounds were within internal bounds.
    activate PowerManager
    Note right of PowerManager: internal discharge bounds = [0, 0]
    Note right of PowerManager: internal charge bounds = [200, 800]
    deactivate PowerManager
    PowerManager-) Microgrid: set_power(500)

    Actor2-)PowerManager: Request(priority=2, power=100, min_power=0, max_power=100)
    PowerManager--)Actor2: Request Failed: Because requested bounds were outside the internal bounds. <br/> Actor 2 has lower priority then Actor1 so he can't increase the bounds range. <br/> He can only limit them (as showed in the next request).

    Actor2-)PowerManager: Request(priority=2, power=400, min_power=300, max_power=600)
    PowerManager--)Actor2: Request Succeed:<br/>Because requested bounds just limit the range that actors with higher priority set.
    activate PowerManager
    Note right of PowerManager: internal discharge bounds = [0, 0]
    Note right of PowerManager: internal charge bounds = [300, 600]
    deactivate PowerManager
    PowerManager-) Microgrid: set_power(400)


    Actor2-)PowerManager: Request(priority=2, power=210, min_power=210, max_power=210)
    PowerManager--)Actor2: Request Succeed:<br/> Because requested bounds are in range that actors with higher priority set
    activate PowerManager
    Note right of PowerManager: internal discharge bounds = [0, 0]
    Note right of PowerManager: internal charge bounds = [210, 210]
    Note over PowerManager,Actor2: Actor 2 has exclusive lock, actors with lower priority can't change bounds.
    deactivate PowerManager
    PowerManager-) Microgrid: set_power(210)


    Actor1-)PowerManager: Request(priority=1, power=-500, min_power=-800, max_power=0)
    PowerManager--)Actor1: Request Succeed. Because actor1 has higher priority so his request can override the bounds set by actors with lower priority.
    activate PowerManager
    Note right of PowerManager: internal discharge bounds = [-800, 0]
    Note right of PowerManager: internal charge bounds = [0, 0]
    deactivate PowerManager
    PowerManager-) Microgrid: set_power(-500)
```

shsms:
This PR add a PowerManagingActor implementation with the Matryoshka
algorithm, that uses actor priorities to resolve conflicting power
requests when multiple actors are trying to use the same batteries.

The Matryoshka algorithm requires actors to specify a desired power
value, and optionally, bounds to limit the range of power values
available for lower priority actors.

Some of the other changes in this PR are:

- updates the `microgrid.battery_pool()` method to accept a priority
  value, and return a new battery pool instance with each call, while
  using a shared reference store class internally, that helps reuse
  formulas, metric aggregators and battery status tracking tasks.

- update BatteryPool's control methods to send requests to the
  PowerManagingActor rather than the PowerDistributingActor.

- Rename the control methods in the BatteryPool from
  `set_power/charge/discharge` -> `propose_{power/charge/discharge}`,
  and update them to accept optional bounds to apply on lower priority
  actors.  The new methods also don't accept a `adjust_power` flag,
  because the `PowerManagingActor` will always adjust power to fit
  within the available bounds.

- Rename `BatteryPool.power_bounds` to a private method called
  `_system_power_bounds`, because it will be used only by the
  PowerManager as the top level bounds.

- Add `BatteryPool.power_status`, which streams
  - bounds adjusted to an actors priority,
  - the latest target power for a set of batteries,
  - the result from the power distributor for the last call.

- Update the PowerDistributingActor to have a single input and output
  channel, because the PowerManagingActor will be the only component
  that uses the PowerDistributingActor from now on.

Closes #161


shsms:
Turned out to be a lot of work, especially outside the power manager, getting everything connected.  I will set up a call next week to go through the code to make reviewing easier.

A number of parts in the battery pool interface need new names.  We can discuss those during the call as well.

I still need to add some tests, and probably some documentation on how to use priorities, and how the algorithm works, etc, but I think it is ready for review already.

src/frequenz/sdk/actor/_power_managing/_base_classes.py
```diff
@@ -0,0 +1,59 @@
...
+import enum
+
+from ...timeseries import Power
+
+
+@dataclasses.dataclass(frozen=True)
+class Proposal:
+    """A proposal for a battery to be charged or discharged."""
+
+    source_id: str
```
Marenz, line 80: Especially the source id should be documented imo

src/frequenz/sdk/actor/_power_managing/_matryoshka.py
```diff
@@ -0,0 +1,58 @@
...
+class Matryoshka(BaseAlgorithm):
+    """The matryoshka algorithm."""
+
+    def __init__(self) -> None:
+        """Create a new instance of the matryoshka algorithm."""
+        self._battery_buckets: dict[frozenset[int], SortedSet[Proposal]] = {}
+
+    @override
+    def handle_proposal(self, proposal: Proposal) -> Power:
+        """Handle a proposal.
```
Marenz: I feel this could also use some more elaboration how it ends up with the target power it returns

tests/actor/_power_managing/test_power_managing_actor.py
```diff
@@ -0,0 +1,56 @@
...
+        )
+        assert (await output_rx.receive()).power.as_watts() == expected
+
+    async with PowerManagingActor(
+        input_channel.new_receiver(), output_channel.new_sender()
+    ):
+        await case(priority=2, power=25.0, bounds=(25.0, 50.0), expected=25.0)
+        await case(priority=1, power=20.0, bounds=(20.0, 50.0), expected=25.0)
+        await case(priority=3, power=10.0, bounds=(10.0, 15.0), expected=10.0)
+        await case(priority=3, power=10.0, bounds=(10.0, 22.0), expected=20.0)
```
Marenz: Why does this change to 20?

tests/actor/_power_managing/test_power_managing_actor.py
```diff
@@ -0,0 +1,56 @@
...
+        await case(priority=1, power=30.0, bounds=(20.0, 50.0), expected=10.0)
```
Marenz: and why does this now become 10?

src/frequenz/sdk/actor/_power_managing/_base_classes.py
```diff
@@ -18,9 +19,35 @@ class Proposal:
...
+    """Whether to adjust the power to match the bounds.
+
+    If `True`, the power will be adjusted (lowered) to match the bounds, so
+    only the reduced power will be set.
+
+    If `False` and the power is outside the batteries' bounds, the request will
+    fail and be replied to with an `OutOfBound` result.
+    """
+
+    include_broken_batteries: bool = False
```
Marenz, line 107: Didn't we want to remove all broken battery related code?

src/frequenz/sdk/actor/_power_managing/_matryoshka.py
```diff
@@ -53,6 +53,7 @@ def handle_proposal(self, proposal: Proposal) -> Power:
             ):
                 continue
             target_power = next_proposal.preferred_power
-            lower_bound = next_proposal.bounds[0]
-            upper_bound = next_proposal.bounds[1]
+            if next_proposal.bounds:
```
Marenz: Did you sneak in a little null check in this commit? ^^

src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
```diff
@@ -26,35 +28,59 @@
...
 
         Args:
             proposals_receiver: The receiver for proposals.
+            bounds_subscription_receiver: The receiver for bounds subscriptions.
             power_distributing_requests_sender: The sender for power distribution
                 requests.
+            channel_registry: The channel registry.
             algorithm: The power management algorithm to use.
 
+
```
Marenz: superfluous blank line

src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
```diff
@@ -26,35 +28,59 @@
...
+        self._bounds_subscription_receiver = bounds_subscription_receiver
+        self._power_distributing_requests_sender = power_distributing_requests_sender
+        self._channel_registry = channel_registry
+        self._proposals_receiver = proposals_receiver
+
         self._system_bounds: dict[frozenset[int], PowerMetrics] = {}
         self._bound_tracker_tasks: dict[frozenset[int], asyncio.Task[None]] = {}
+        self._subscriptions: dict[frozenset[int], dict[int, Sender[Report]]] = {}
+
+        self._algorithm: BaseAlgorithm = Matryoshka()
```
Marenz, line 74: You removed this in an earlier commit, only to add it back here? Oo

src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
```diff
@@ -0,0 +1,221 @@
...
+            PowerMetrics,
+        )
+
+        battery_pool = microgrid.battery_pool(battery_ids)
+        # pylint: disable=protected-access
+        bounds_receiver = battery_pool._system_power_bounds.new_receiver()
+        # pylint: enable=protected-access
+
+        # Fetch the latest system bounds once, before starting the bounds tracker task,
+        # so that when this function returns, there's already some bounds available.
```
Marenz: This comment is no longer correct

src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
```diff
@@ -96,7 +97,7 @@ async def _bounds_tracker(
             self._system_bounds[battery_ids] = bounds
             await self._send_reports(battery_ids)
 
-    async def _add_bounds_tracker(self, battery_ids: frozenset[int]) -> None:
+    def _add_bounds_tracker(self, battery_ids: frozenset[int]) -> None:
```
Marenz, line 113: I _was_ wondering if the `async` here might create limits/problems later

src/frequenz/sdk/actor/_power_managing/_base_classes.py
shsms: done in this commit: https://github.com/frequenz-floss/frequenz-sdk-python/pull/692/commits/1e7f54bb7be45f7f4ddb3170cfe16d501530ef3f

src/frequenz/sdk/actor/_power_managing/_base_classes.py
shsms: Yes, didn't want to hide it here, because then we might not have the motivation to remove it from the PowerDistributor later.  I'll pick that up after this PR.

src/frequenz/sdk/timeseries/battery_pool/battery_pool.py
```diff
@@ -447,6 +454,28 @@ def capacity(self) -> ReceiverFetcher[Sample[Energy]]:
 
         return self._active_methods[method_name]
 
+    def power_bounds(
+        self, priority: int = 0
+    ) -> ReceiverFetcher[_power_managing.Report]:
+        """Get a receiver to receive new power bounds when they change.
+
+        These bounds are the bounds specified the power manager for actors with the
+        given priority.
```
Marenz: Can you add more words to this sentence so it doesn't read like I am hiking barefoot through a field of sharp stones?

shsms: :D wow, that's so funny, the stuff I've written.  But already fixed.  And doing the other doc stuff as well.

src/frequenz/sdk/actor/_power_managing/_matryoshka.py
```diff
@@ -49,6 +52,21 @@ def get_target_power(
...
+                and system_bounds.exclusion_bounds is None
+            ):
+                if proposal is not None:
+                    _logger.warning(
+                        "PowerManagingActor: No system bounds available for battery "
+                        "IDs %s, but a proposal was given.  The proposal will be "
+                        "ignored.",
+                        battery_ids,
+                    )
+                return None
```
Marenz: Could it happen that someone sets power before the system bounds are initialized? Is there a way to make sure they are initialized before using?

src/frequenz/sdk/actor/_power_managing/_base_classes.py
```diff
@@ -138,6 +139,8 @@ def get_target_power(
             proposal: If given, the proposal to added to the bucket, before the target
                 power is calculated.
             system_bounds: The system bounds for the batteries in the proposal.
+            must_return_power: If `True`, the algorithm must return a target power,
```
Marenz, line 161: Is there any reason not to always do this?

src/frequenz/sdk/timeseries/battery_pool/_battery_pool_wrapper.py
```diff
@@ -57,52 +57,65 @@ def __init__(self, battery_pool: BatteryPool, source_id: str | None, priority: i
...
-        When not using the Passive Sign Convention, the `charge` and `discharge` methods
-        might be more convenient.
+        If the same batteries are shared by multiple actors, the power manager will
+        consider the priority of the actors, the bounds they set, and their preferred
+        power, when calculating the target power for the batteries.
+
+        The preferred power of lower priority actors will take precedence as long as
+        they respect the bounds set by higher priority actors.  If lower priority actors
+        request power values outside of the bounds set by higher priority actors, the
+        target power will be closest value to the preferred power that is within the
```
Marenz: will be *the* closest value

src/frequenz/sdk/timeseries/battery_pool/_battery_pool_wrapper.py
```diff
@@ -44,17 +44,17 @@ class BatteryPoolWrapper:
...
     """
 
-    def __init__(self, battery_pool: BatteryPool, source_id: str | None, priority: int):
+    def __init__(self, battery_pool: BatteryPool, name: str | None, priority: int):
         """Create a BatteryPoolWrapper instance.
 
         Args:
             battery_pool: The battery pool to wrap.
-            source_id: The source ID to use for the requests made by this wrapper.
+            name: The source ID to use for the requests made by this wrapper.
```
Marenz: It's not yet the source id though, only after the unique_id is added, right? Needs docs update :)

src/frequenz/sdk/actor/_power_managing/_matryoshka.py
```diff
@@ -0,0 +1,200 @@
...
+
+        Returns:
+            The new target power for the batteries, or `None` if the target power
+                didn't change.
+
+        Raises:
+            NotImplementedError: When the proposal contains battery IDs that are
+                already part of another bucket.
+        """
+        if battery_ids not in self._battery_buckets:
```
Marenz, line 100: Could we move the checks in an extra function to make this one easier to read?

Marenz (commented):
I had a few comments, overall nice work

src/frequenz/sdk/actor/_power_managing/_matryoshka.py
shsms: added docs in the module level.

src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
shsms: removed

src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
shsms: true, removed

src/frequenz/sdk/timeseries/battery_pool/_battery_pool_wrapper.py
shsms: fixed

src/frequenz/sdk/timeseries/battery_pool/_battery_pool_wrapper.py
shsms: updated

src/frequenz/sdk/actor/_power_managing/_base_classes.py
shsms: This was supposed to reduce the number of requests going out to the power distributor, especially when there are multiple actors repeating their requests.  

But that just reminded me, we can't have this yet,  because requests to the microgrid service will expire, and we have to keep repeating them periodically.  There's a ticket for this #293, and once we have that,  we can enable `must_return_power` back.  For now,  I've added a commit https://github.com/frequenz-floss/frequenz-sdk-python/pull/692/commits/a8bf82d59f7869dea517257acddcc634f59054c9 that makes `must_return_power` always true.

src/frequenz/sdk/actor/_power_managing/_matryoshka.py
shsms: yes, I think the trick is for actors to get the bounds first, ensure their requests comply with the bounds, then make the proposal.

but even if people don't do that,  only the first request will get ignored (and logged) because there are no bounds.  because as soon as a proposal comes in, we start tracking bounds for that set of batteries.

src/frequenz/sdk/actor/_power_managing/_matryoshka.py
shsms: done

RELEASE_NOTES.md
```diff
@@ -31,12 +45,11 @@
   - Add `fill_value` option to window method to impute missing values. By default missing values are imputed with `NaN`.
 - Add `at` method to `MovingWindow` to access a single element and use it in `__getitem__` magic to fully support single element access.
 
-
-
 - The PowerDistributingActor now supports n:m relations between inverters and
   batteries.
   This means that one or more inverters can be connected to one or more batteries.
 
+- A PowerManagingActor implementation
```
llucax: Optional:

```suggestion
- A `PowerManagingActor` implementation
```

benchmarks/power_distribution/power_distributor.py
```diff
@@ -47,10 +51,12 @@ async def send_requests(batteries: set[int], request_num: int) -> list[Result]:
...
     """
     battery_pool = microgrid.battery_pool(batteries)
-    results_rx = battery_pool.power_distribution_results()
-    result: list[Result] = []
+    results_rx = battery_pool.power_status.new_receiver()
+    result: list[Any] = []
     for _ in range(request_num):
-        await battery_pool.set_power(Power(float(random.randrange(100000, 1000000))))
+        await battery_pool.propose_power(
+            Power(float(random.randrange(100000, 1000000)))
```
llucax, line 58: Optional (and specially because this PR is not really adding these numbers): this will soon fail when we incorporate [`flake8-numbers`](https://github.com/frequenz-floss/flake8-numbers) (and I think is a perfect example of where it adds clarity :D):
```suggestion
            Power(float(random.randrange(100_000, 1_000_000)))
```

src/frequenz/sdk/_internal/_channels.py
```diff
@@ -0,0 +1,26 @@
...
+    @abc.abstractmethod
+    def new_receiver(self, maxsize: int = 50) -> Receiver[T]:
+        """Get a receiver from the channel.
+
+        Args:
+            maxsize: The maximum size of the receiver.
+
+        Returns:
+            A receiver instance.
+        """
```
llucax, line 26: `Protocol`s are very nice but I think we should be careful in how we introduce/use them. `Protocol`s are particularly useful to keeps type very close with who use them (and to reduce the required interface to the minimum) and to define inputs more than outputs.

This is because you don't need to inherit from the protocol for a class to be considered an *instance* of a protocol, so you can make 2 different libraries interact much more easily, For example, if `liba` has a function `liba.make_bark(dog: liba.Dog) -> None: dog.bark()` and `libb` has a function `libb.parse_dog(dog: str) -> libb.Dog`, then without protocols you can't do: `liba.make_bark(libb.parse_dog())` even if `Dog.bark()` exist in both `liba.Dog` and `libb.Dog`, and if you don't have control over `liba` to make `liba.Dog` inherit from `libb.Dog`, then you need to convert between both. If `liba.Dog` is a protocol instead, then everything works out of the box.

Looking on how you use this, specially since you inherit from it (which it has extra advantages, as your subclass will be checked to have all the necessary methods when the class is parsed instead of used), I don't see any point in making it a `Protocol`. Not sure if you purposely made it a `Protocol` because you anticipate to have interactions like the ones I mentioned above... If not, I mildly suggest to not make this inherit from `Protocol`, as there is one more thing people have to know well how it works and why it is there.

EDIT: I see now that `MetricAggregator` is behaving as a `ReceiverFetcher` but not inheriting from it, now I wonder why you decided to make some inherit from it and some not.

src/frequenz/sdk/actor/_power_managing/_base_classes.py
```diff
@@ -0,0 +1,189 @@
...
+
+from ... import timeseries
+from ...timeseries import Power
+
+if typing.TYPE_CHECKING:
+    from ...timeseries.battery_pool import PowerMetrics
+    from .. import power_distributing
+
+
+@dataclasses.dataclass(frozen=True)
```
llucax: When most attributes of a `dataclass` are not classes but built-in types, I think it might make things more clear if we also use `kw_only=True` to force users use keywords.

```py
# cryptic
ReportRequest("lalala", [1,2,3], 5)
# obvious
ReportRequest(source_id="lalala", battery_ids=[1,2,3], priority=5)
# unless this is always expected to be used like
ReportRequest(get_id(), battery_ids, priority)
# (but well, is hard to ensure in any possible future this holds true)
```

src/frequenz/sdk/actor/_power_managing/_base_classes.py
```diff
@@ -0,0 +1,189 @@
...
+
+        Returns:
+            The channel name to use to identify the corresponding report channel
+                from the channel registry.
+        """
+        return f"power_manager.report.{self.battery_ids=}.{self.priority=}"
+
+
+@dataclasses.dataclass(frozen=True)
+class Report:
```
llucax, line 46: Here, for example, it is mostly fine, because we'll always at least have:

```py
Report(Power.from_kw(10), Bounds(...), Bounds(...), Result(...))
```
But yeah, with the bounds it is inclusion or exclusion first? We can use `filed` to make only those 2 keyword-only. Then users will need to write:

```py
Report(Power.from_kw(10), Result(...), inclusion_bounds=Bounds(...), exclusion_bounds=Bounds(...))
```

Since they are kw-only now, we could even remove the `_bounds` suffix as it's a bit redundant now (`inclusion=Bounds(...), exclusion=Bounds(...)`), but I don't know, it reads a bit weird without them too.

src/frequenz/sdk/actor/_power_managing/_base_classes.py
```diff
@@ -0,0 +1,189 @@
...
+    """The currently set power for the batteries."""
+
+    inclusion_bounds: timeseries.Bounds[Power] | None
+    """The available inclusion bounds for the batteries, for the actor's priority.
+
+    These bounds are adjusted to any restrictions placed by actors with higher
+    priorities.
+    """
+
+    exclusion_bounds: timeseries.Bounds[Power] | None
```
llucax, line 59: To make bounds keyword-only (**and we need to move them to the end**):
```suggestion
    inclusion_bounds: timeseries.Bounds[Power] | None = dataclasses.field(kw_only=True)
    """The available inclusion bounds for the batteries, for the actor's priority.

    These bounds are adjusted to any restrictions placed by actors with higher
    priorities.
    """

    exclusion_bounds: timeseries.Bounds[Power] | None = dataclasses.field(kw_only=True)
```

src/frequenz/sdk/timeseries/battery_pool/_battery_pool_reference_store.py
```diff
@@ -21,8 +21,15 @@
...
-class BatteryPool:  # pylint: disable=too-many-instance-attributes
-    """A class for maintaining the state/tasks for a set of pool of batteries.
+class BatteryPoolReferenceStore:  # pylint: disable=too-many-instance-attributes
+    """A class for maintaining the shared state/tasks for a set of pool of batteries.
+
+    This includes ownership of
+    - the formula engine pool and metric calculators.
+    - the tasks for updating the battery status for the metric calculators.
+
+    These are independent of the priority of the actors, and can be shared between
```
Marenz: The `,` is [wrong](https://www.grammarly.com/blog/comma-before-and/) as the second part of the conjunction is not an independent clause.

src/frequenz/sdk/actor/_power_managing/_base_classes.py
```diff
@@ -0,0 +1,189 @@
...
+
+        Returns:
+            Whether this proposal has a higher priority than the other proposal.
+        """
+        return (self.priority < other.priority) or (
+            self.priority == other.priority and self.source_id < other.source_id
+        )
+
+
+class Algorithm(enum.Enum):
```
llucax, line 137: Not sure if you explained this somewhere else, but otherwise I would mention in a code comment (not class documentation) that we use an enum because at some point this algorithm will need to be serialized.

src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
```diff
@@ -0,0 +1,230 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""The power manager."""
+
+from __future__ import annotations
```
llucax, line 6: Is this really needed? I think in Python 3.11+ it is only necessary to use forward references and I don't see any, but maybe I am missing something.

src/frequenz/sdk/microgrid/_data_pipeline.py
```diff
@@ -173,42 +189,93 @@ def ev_charger_pool(
...
-        If a BatteryPool instance for the given ids doesn't exist, a new one is created
-        and returned.
+        If a BatteryPoolReferenceStore instance for the given battery ids doesn't exist,
+        a new one is created and used for creating the BatteryPool.
 
         Args:
             battery_ids: Optional set of IDs of batteries to be managed by the
                 BatteryPool.
+            name: An optional name used to identify this instance of the pool or a
+                corresponding actor in the logs.
```
llucax, line 204: OK, so you finally went with always creating a UUID and just adding the name to it if it was provided, right? So this means there is no way to call `battery_pool()` 2 times as the same "user" (i.e. the request being considered to come from the same source).

This means to do that we need to save the `battery_pool` object we just *created* and we don't support the case where someone wants to be considered the same source in the eyes of the power manager from different nodes/controllers.

Just checking if we are all in the same page and fine with going this route...

src/frequenz/sdk/microgrid/_data_pipeline.py
```diff
@@ -191,13 +191,11 @@ def battery_pool(
...
-    ) -> BatteryPoolWrapper:
-        """Return the corresponding BatteryPool instance for the given ids.
+    ) -> BatteryPool:
+        """Return a new BatteryPool instance for the given ids.
 
-        If a BatteryPool instance for the given ids doesn't exist, a new one is created
-        and returned.
-
-        The BatteryPool is wrapped in a new `BatteryPoolWrapper` instance each time.
+        If a BatteryPoolReferenceStore instance for the given battery ids doesn't exist,
```
Marenz: back-tick-quotes around `BatteryPoolReferenceStore` would be nice.

src/frequenz/sdk/microgrid/_data_pipeline.py
```diff
@@ -173,42 +189,93 @@ def ev_charger_pool(
     def battery_pool(
         self,
         battery_ids: abc.Set[int] | None = None,
+        name: str | None = None,
+        priority: int = -sys.maxsize - 1,
```
llucax, line 193: Any reason not to make the `priority` a `float`, so we don't end up writing *BASIC* to make sure we can sneak in some new priority between 2 existing priorities (priority 10, 20, 30, 40.... :D).

Also I don´t get the default. Why not 0?

src/frequenz/sdk/microgrid/_data_pipeline.py
Marenz: and `BatteryPool` I suppose :)

src/frequenz/sdk/timeseries/battery_pool/_battery_pool.py
```diff
@@ -36,143 +33,132 @@
...
+        comply with, based on its priority.
 
         Args:
-            power: The power to set for the batteries in the pool.
-            adjust_power: If True, the power will be adjusted to fit the power bounds,
-                if necessary. If False, then power requests outside the bounds will be
-                rejected.
+            power: The power to propose for the batteries in the pool.  If None, the
+                proposed power of higher priority actors will take precedence as the
+                target power.
```
llucax: I find the comment about `None` a bit confusing. Isn't the intent of passing `None` to sort of *withdraw* a proposal, like not participating in the decision of how much power to set? If so I would mention something more explicit along these lines to make more clear what's the use case, not only what is the effect.

Also nitpick:
```suggestion
            power: The power to propose for the batteries in the pool.  If `None`, the
                proposed power of higher priority actors will take precedence as the
                target power.
```

src/frequenz/sdk/timeseries/battery_pool/_battery_pool.py
```diff
@@ -36,143 +33,132 @@
...
+        request power values outside of the bounds set by higher priority actors, the
+        target power will be the closest value to the preferred power that is within the
+        bounds.
+
+        When there are no other actors trying to use the same batteries, the actor's
+        preferred power would be set as the target power, as long as it falls within the
+        system power bounds for the batteries.
+
+        The result of the request can be accessed using the receiver returned from the
+        `power_status` method, which also streams the bounds that an actor should
```
llucax: In general, I think it would be good to start including more cross-referencing in new code:
```suggestion
        [`power_status`][frequenz.sdk.timeseries.battery_pool.power_status] method, which also streams the bounds that an actor should
```

src/frequenz/sdk/timeseries/battery_pool/_battery_pool_reference_store.py
```diff
@@ -0,0 +1,138 @@
...
+from .._formula_engine import FormulaEnginePool
+from ._methods import MetricAggregator
+
+
+class BatteryPoolReferenceStore:  # pylint: disable=too-many-instance-attributes
+    """A class for maintaining the shared state/tasks for a set of pool of batteries.
+
+    This includes ownership of
+    - the formula engine pool and metric calculators.
+    - the tasks for updating the battery status for the metric calculators.
```
llucax, line 29: Oh, this has `BackgroundService` written all over it.

src/frequenz/sdk/timeseries/battery_pool/_battery_pool_reference_store.py
```diff
@@ -0,0 +1,138 @@
...
+
+    async def stop(self) -> None:
+        """Stop all pending async tasks."""
+        tasks_to_stop: list[Awaitable[Any]] = [
+            method.stop() for method in self._active_methods.values()
+        ]
+        tasks_to_stop.append(self._formula_pool.stop())
+        if self._update_battery_status_task:
+            tasks_to_stop.append(cancel_and_await(self._update_battery_status_task))
+        await asyncio.gather(*tasks_to_stop)
```
llucax, line 116: And this.

I guess at this point it would delay the PR a bit, so I wouldn't block it because of this, but it would be good to have a follow-up PR converting this into a `BackgroundService`. The whole point of introducing that class was to unify how we build classes like this.

tests/actor/_power_managing/_sorted_set.py
```diff
@@ -0,0 +1,39 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Tests for the sorted set."""
+
+from frequenz.sdk.actor._power_managing import _sorted_set
+
+
+def test_sorted_set() -> None:
```
llucax, line 9: I guess this is a pretty good candidate for *hypothesis* (many inputs that could be tested and just a few simple properties to test for).

tests/actor/_power_managing/test_matryoshka.py
```diff
@@ -0,0 +1,162 @@
...
+    test_bounds(priority=2, expected_power=25.0, expected_bounds=(-200.0, 200.0))
+    test_bounds(priority=1, expected_power=25.0, expected_bounds=(25.0, 50.0))
+
+    test_tgt_power(priority=1, power=20.0, bounds=(20.0, 50.0), expected=None)
+    test_tgt_power(
+        priority=1, power=20.0, bounds=(20.0, 50.0), expected=25.0, must_send=True
+    )
+    test_bounds(priority=1, expected_power=25.0, expected_bounds=(25.0, 50.0))
+
+    test_tgt_power(priority=3, power=10.0, bounds=(10.0, 15.0), expected=15.0)
```
llucax, line 82: Ha! This is confusing, I think it makes sense but it makes you think about priorities a bit differently from what one might expect. So high priority means high priority for bounds but low priority for target power, right? So this is a higher priority request so the bounds are narrowed but the actual target power is the closest to the lowest priority request instead, right? And this is why this high prio request gets `15` instead of `10`, which is what it actually wanted?

llucax (commented):
OK, I finished a quick look. I guess all is sort of optional, at least nothing major but a few things I find important:

* Keyword-only stuff (only if there is any expectation to be used by end users, which I think there aren't, so it should be fine as it is)
* `priority` as `float` and default
* Make sure we are aligned about the `source_id`/`name`
* Using `BackgroundService` for the classes that it makes sense (OK to do in a separate PR, as lower prio as we still have a couple of classes that are not converted yet, but from now on we should always use `BackgroundService` from the start for stuff handling tasks, having a  `stop()`, etc.)

I think that's it, the rest is more cosmetic or not that important.

src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
shsms: yes, without it, we have to quote the first occurrence of a type hint that was added with `if TYPE_CHECKING`.

src/frequenz/sdk/microgrid/_data_pipeline.py
shsms: Yes,  but this reminds me, we need to add logging in the actor to warn users when there are multiple proposals with the same priority, but with different `source_id`.  Made an issue, so this PR doesn't keep dragging: https://github.com/frequenz-floss/frequenz-sdk-python/issues/714

src/frequenz/sdk/timeseries/battery_pool/_battery_pool_reference_store.py
shsms: done in: https://github.com/frequenz-floss/frequenz-sdk-python/pull/692/commits/8d6edef480e87c12a2523f9ae009fd2bd7f0b5eb

src/frequenz/sdk/microgrid/_data_pipeline.py
shsms: Done in: https://github.com/frequenz-floss/frequenz-sdk-python/pull/692/commits/8d6edef480e87c12a2523f9ae009fd2bd7f0b5eb

RELEASE_NOTES.md
shsms: Done in: https://github.com/frequenz-floss/frequenz-sdk-python/pull/692/commits/8d6edef480e87c12a2523f9ae009fd2bd7f0b5eb

benchmarks/power_distribution/power_distributor.py
shsms: It is not new, can be fixed in flake8 PR.

src/frequenz/sdk/_internal/_channels.py
shsms: It was made to provide a single type around `MetricAggregator` and the `ChannelRegistry`.  But you're right, the internal channel registry type doesn't have to derive from `ReceiverFetcher`.  Done here: https://github.com/frequenz-floss/frequenz-sdk-python/pull/692/commits/50fc8d34062b64df7bbb9b6e39f6a983fb7e5dfa

src/frequenz/sdk/actor/_power_managing/_base_classes.py
shsms: Done here: https://github.com/frequenz-floss/frequenz-sdk-python/pull/692/commits/6dfde504d93558de8e421c9856c0fdc42ab6d1d7

src/frequenz/sdk/actor/_power_managing/_base_classes.py
shsms: Done here: https://github.com/frequenz-floss/frequenz-sdk-python/pull/692/commits/6dfde504d93558de8e421c9856c0fdc42ab6d1d7

src/frequenz/sdk/actor/_power_managing/_base_classes.py
shsms: Done here: https://github.com/frequenz-floss/frequenz-sdk-python/pull/692/commits/6dfde504d93558de8e421c9856c0fdc42ab6d1d7

src/frequenz/sdk/actor/_power_managing/_base_classes.py
shsms: Added a couple of lines near the actor parameter here: https://github.com/frequenz-floss/frequenz-sdk-python/pull/692/commits/f268d16e0a1fc901a8e61a7552f95799cf8b4f44

src/frequenz/sdk/timeseries/battery_pool/_battery_pool_reference_store.py
shsms: made an issue: https://github.com/frequenz-floss/frequenz-sdk-python/issues/715

src/frequenz/sdk/timeseries/battery_pool/_battery_pool_reference_store.py
shsms: made an issue: https://github.com/frequenz-floss/frequenz-sdk-python/issues/715

tests/actor/_power_managing/_sorted_set.py
shsms: made an issue: https://github.com/frequenz-floss/frequenz-sdk-python/issues/716

tests/actor/_power_managing/test_matryoshka.py
shsms: Right.  If that's not the case, and we just prioritize preferred powers of higher pri actors, then lower pri actors might as well not bother sending any request, and we don't need bounds,  and probably no power manager, and actually the lower pri actors don't need to run at all. :D

src/frequenz/sdk/timeseries/battery_pool/_battery_pool.py
shsms: It is a bit tricky, because there could be bounds even if power is `None`.  If both of them are `None`, that's like withdrawing the proposal,  else it is just "not having a perferred power".

src/frequenz/sdk/timeseries/battery_pool/_battery_pool.py
shsms: Done here: 9eabfb59e7511adf599a704bfcf4646467c46527

src/frequenz/sdk/microgrid/_data_pipeline.py
shsms: Don't know, with all the float comparison issues.  We can just recommend 6 digit priorities, so there's always enough space between values.

src/frequenz/sdk/_internal/_channels.py
llucax: OK, my point was that probably in this case it made more sense to just not use a `Protocol` and derive both instead, but I guess that's also more likely to end with circular dependencies (I guess that's another nice feature of `Protocol`). I'm still not convinced which is better to be honest, but since I'm not convinced I will leave the topic here.

Almost. If we start using `Protocol` more, actually `ReceiverFetcher` (maybe with another name) could be moved to `channels`, as most channels are also `ReceiverFetcher`s. Also I think `Sender` and `Receiver` could be protocols.

shsms: I agree about moving it to the channels repo, already have a todo item for it,  because there's really no nice place to put it in the sdk, especially because it has to be publicly exposed, and it is a bit weird now, living in `_internal`, but still exposed publicly.

src/frequenz/sdk/microgrid/_data_pipeline.py
llucax: Considering that:

* We only care about ordering
* If 2 priorities are conceptually equal, we don't care which one is considered higher than the other (in cases where `float` could not compare equal for 2 equal numbers, like `assert 0.1 + 0.2 > 0.3`)
* We'll probably be using small (but not too small) literals
 
It should be reasonably safe to use `float`s IMHO.

The problem with integers is if we end up with 2 actors using consecutive numbers there is no (external) way to sneak something between both, and you might not have control over those 2 actors to go change the code and redeploy. I we would want to be extra safe, we could add a `Priority(float)` object that checks if the `float` is perfectly representable and raises if it's not, but as a first step I would use `float`.

llucax: BTW, I'm fine with merging as is and discuss this separate and do a follow-up PR if we decide to change it.

src/frequenz/sdk/timeseries/battery_pool/_battery_pool.py
llucax: OK, I see, to be honest I didn't read all the docs, so if this is well explained somewhere, we should be fine. It is complicated so we can't have a full explanation everywhere, agreed.

src/frequenz/sdk/microgrid/_data_pipeline.py
llucax: Also, I just saw the commit that makes `-sys.maxsize - 1` the default. I'm not sure why we want to have the lowest possible priority ever the default, that is not explained in the commit message.

src/frequenz/sdk/timeseries/battery_pool/_battery_pool.py
shsms: maybe this looks better?

```suggestion
            power: The power to propose for the batteries in the pool.  If `None`, this
                proposal will not have any effect on the target power, unless bounds
                are specified.  If both are `None`, it is equivalent to not having a 
                proposal or withdrawing a previous one.
```

llucax: Yup!

src/frequenz/sdk/microgrid/_data_pipeline.py
shsms: **int vs float:**

My feeling has always been that priorities have to be externally controlled, possibly through UI config.  Because actor implementations by themselves can't know what their priorities are.

But going from int to float is not a breaking change, so I agree that we can discuss separately.

**default priority:**

This is of course subjective, and is based on some further opinions: 
- The default priority is used only when an actor knows it will not be sharing batteries with other actors.  
- Most actors don't know this, so they should provide a way to configure their priorities.  
- If an actor uses the default priority (i.e doesn't care about its priority), it should be lower than any other actor, even the ones with negative priorities, because negative priorities doesn't mean that they're low priority, just that the priority numbers are negative.  
- If actors care about priorities, they should provide a way to specify one, else they get put in the bottom of the pile.

src/frequenz/sdk/timeseries/battery_pool/_battery_pool.py
shsms: done in: https://github.com/frequenz-floss/frequenz-sdk-python/commit/79ecc0d2

tests/timeseries/_battery_pool/test_battery_pool_control_methods.py
```diff
@@ -0,0 +1,370 @@
...
+        await self._init_data_for_batteries(mocks)
+        await self._init_data_for_inverters(mocks)
+
+        battery_pool = microgrid.battery_pool()
+
+        # This is used just to wait for the processing to be complete.  The results are
+        # not used.
+        #
+        # It will be replaced by a reporting streaming from the PowerManager in a
+        # subsequent commit.
```
Marenz: Is this something you meant to update in this PR?

src/frequenz/sdk/actor/_power_managing/_base_classes.py
```diff
@@ -0,0 +1,189 @@
...
+
+    If `None`, the preferred power of higher priority actors will get precedence.
+    """
+
+    bounds: timeseries.Bounds[Power | None]
+    """The power bounds for the proposal.
+
+    These bounds will apply to actors with a lower priority, and can be overridden by
+    bounds from actors with a higher priority.  If None, the power bounds will be set to
+    the maximum power of the batteries in the pool.  This is currently and experimental
```
matthias-wende-frequenz:
```suggestion
    the maximum power of the batteries in the pool.  This is currently an experimental
```

src/frequenz/sdk/actor/_power_managing/_base_classes.py
```diff
@@ -0,0 +1,189 @@
...
+    """The available algorithms for the power manager."""
+
+    MATRYOSHKA = "matryoshka"
+
+
+class BaseAlgorithm(abc.ABC):
+    """The base class for algorithms."""
+
+    @abc.abstractmethod
+    def get_target_power(
```
matthias-wende-frequenz: I wonder why this is called `get_target_power` although it's used to propose proposals. What about `calculate_target_power` or `determine_target_power`?

tests/timeseries/_battery_pool/test_battery_pool_control_methods.py
shsms: oh yes, that whole comment is no longer relevant.  Will remove.

src/frequenz/sdk/actor/_power_managing/_base_classes.py
shsms: I wonder if `bounds` is an experimental feature.  Maybe we should rename it back to `_bounds` in the interface.  But the interface has a new name `propose_power` won't make sense without the bounds.  Is tricky.

src/frequenz/sdk/actor/_power_managing/_base_classes.py
shsms: True, `calculate_target_power` sounds way better, will rename.

Marenz: ANOTHER rename

src/frequenz/sdk/actor/_power_managing/_base_classes.py
matthias-wende-frequenz: at least you should replace the and with an an ;)

src/frequenz/sdk/actor/_power_managing/_base_classes.py
Marenz: I am still voting for `request_..`

matthias-wende-frequenz: It's just returning a calculated power. If I'd request something I'd suppose something happens, e.g. updating a db or switching a state .... But in this case nothing happens.
Therefore I'd find request misleading.

tests/timeseries/_battery_pool/test_battery_pool_control_methods.py
shsms: done

src/frequenz/sdk/actor/_power_managing/_base_classes.py
shsms: done

src/frequenz/sdk/actor/_power_managing/_base_classes.py
shsms: renamed to `calculate_target_power`.  Regarding the renaming from `propose_*` to `request_*` (assuming this is what you're talking about @Marenz),  I'd like it to not block this PR, but I'll start an internal discussion where we can finalize the name and make a separate PR if necessary.

matthias-wende-frequenz (approved):

src/frequenz/sdk/microgrid/_data_pipeline.py
llucax: Yeah, let's leave it for when we have a better idea of how this will be used and who will be setting priorities and how.

llucax (approved):

PR #697, opened 2023-09-25
Remove the consumption and production power formulas

matthias-wende-frequenz:
Users must understand passive sign conventions to utilize these formulas effectively, and therefore they do not provide significant benefits.

### DIFF
--- a/tests/timeseries/test_logical_meter.py
+++ b/tests/timeseries/test_logical_meter.py
@@ -269,18 +269,12 @@ class TestLogicalMeter:  # pylint: disable=too-many-public-methods
 
         logical_meter = microgrid.logical_meter()
         grid_recv = logical_meter.grid_power.new_receiver()
-        grid_production_recv = logical_meter.grid_production_power.new_receiver()
-        grid_consumption_recv = logical_meter.grid_consumption_power.new_receiver()
 
         await mockgrid.mock_resampler.send_meter_power([1.0, 2.0, 3.0, 4.0])
         assert (await grid_recv.receive()).value == Power.from_watts(10.0)
-        assert (await grid_production_recv.receive()).value == Power.from_watts(0.0)
-        assert (await grid_consumption_recv.receive()).value == Power.from_watts(10.0)
 
         await mockgrid.mock_resampler.send_meter_power([1.0, 2.0, -3.0, -4.0])
         assert (await grid_recv.receive()).value == Power.from_watts(-4.0)
-        assert (await grid_production_recv.receive()).value == Power.from_watts(4.0)
-        assert (await grid_consumption_recv.receive()).value == Power.from_watts(0.0)
 
     async def test_grid_production_consumption_power_no_grid_meter(
         self,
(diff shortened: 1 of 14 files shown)


cwasicki:
> Users must understand passive sign conventions to utilize these formulas effectively, and therefore they do not provide significant benefits.

What do you mean? These formulas were introduced exactly for the reason that passive sign convention can be confusing.

shsms:
I will leave it for @cwasicki to approve, because he has open questions, but if we are doing this, we should also remove the corresponding formulas from the battery pool and the ev charger pool.

matthias-wende-frequenz:
# Possible approaches for keeping the functionality alive

We discussed that these formulas are useful but they might lead to subtle bugs when confusing the slightly different functions.
We can provide this by adding a production and consumption operators as suggest in the following.

## 1

chp_power("production")
chp_power("abs_production")
chp_power().production() or production_op(chp_power())

## 2

chp_power().production() and chp_power().consumption()

llucax:
Option 2 wins! BTW, Option 1.3 is Option 2, right?

cwasicki:
>Option 1.3 is Option 2, right?

Yes, but option 1.3a only ;-)

matthias-wende-frequenz:
Back to draft as it needs the operators to get implemented.

matthias-wende-frequenz:
unblocked after #746 has been merged as a replacement

tests/timeseries/test_logical_meter.py
```diff
@@ -427,14 +377,14 @@ async def test_consumer_power_2_grid_meters(
...
-        """Test the grid power formula with grid meters."""
+        """Test the grid power formula with two grid meters."""
         mockgrid = MockMicrogrid(grid_meter=False)
         # with no further sucessor these will be detected as grid meters
         mockgrid.add_consumer_meters(2)
         await mockgrid.start(mocker)
 
         logical_meter = microgrid.logical_meter()
-        grid_consumption_recv = logical_meter.grid_consumption_power.new_receiver()
+        grid_consumption_recv = logical_meter.grid_power.new_receiver()
```
llucax: I guess you are purposely not using `consumption` (with the new operator) anymore here because this is already tested in the new operator tests, right?

llucax (dismissed):
Yeeeiii! Removing code and special cases! :tada:

shsms (approved):

Marenz:
Some of the [commits](https://github.com/frequenz-floss/frequenz-sdk-python/pull/697/commits/5b25cf3bf2541d84bdae37f66ad0ae1f43f7e8ea) here are a great example of unhelpful commit title/messages. 
Just reading the commit I had no idea why it was doing what it was doing "Remove xyz"..great.. but why?!
I had to find this PR first to get an idea of why it was happening.

Please, write the why, not the what. It was very very obvious that it was removing something... no need to mention that even imo.

matthias-wende-frequenz:
> Just reading the commit I had no idea why it was doing what it was doing "Remove xyz"..great.. but why?!

The reason is written up in the first of the commit chain.

> These formulas are replaced by a combination of the corresponding
passive sign convention formulas and the recently added consumption
and production operators.

Next time I'll add a note to each of the involved commits ;).

Marenz:
> The reason is written up in the first of the commit chain.

right, thus I said "some" ;) but I still had to do more investigation to be sure. 

> Next time I'll add a note to each of the involved commits ;).

<3

PR #699, opened 2023-09-26
Increment port in microgrid API client tests

llucax:
This is a hack to avoid the inherent flakiness of the approach of using a real GRPC (mock) server. The server seems to stay alive for a short time after the test is finished, which causes the next test to fail because the port is already in use.

This also improves the test a bit by using an async context manager create the client and server, and stop the server automatically when the test is finished.

Related to #662.

### DIFF
--- a/tests/microgrid/test_client.py
+++ b/tests/microgrid/test_client.py
@@ -35,11 +35,22 @@ from . import mock_api
 # pylint: disable=missing-class-docstring,no-member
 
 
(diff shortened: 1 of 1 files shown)


llucax:
As I said in the issue, I don't think this really fixes the flakiness reported in #662, but it should still be an improvement in terms of readability and could potentially avoid flakiness on the finalization, so I would merge it anyways.

llucax:
Enabling auto-merge.

shsms (approved):

PR #700, opened 2023-09-26
Add pv optimization tutorial

matthias-wende-frequenz:


docs/tutorials/pv_optimization.md
```diff
@@ -0,0 +1,117 @@
+# Optimizing PV production
+
+## Introduction
+In this tutorial we want to write an application that optimizes the energy produced from a PV system with
+Battery for self consumption.
+In order to do so we need to measure the power that flows through the grid connection point
+(TODO link to glossary) to determine excess power.
```
llucax: You can link also PV and other terms, I guess we should only link the first time a term appears so we don't clutter the docs too much.

docs/tutorials/pv_optimization.md
```diff
@@ -0,0 +1,117 @@
...
+When using the term excess power what we actually mean is the consumer excess power, that is the power that
+flows from the PV system into the grid.
+
+!!! note
+
+    We are using the passive sign convention (TODO link) and thus power flowing from the PV is negative
+    and consumed power is positive.
+
+We want to measure the excess power. In order to do so you can use the SDK's data pipeline and especially
+the pre defined consumer and producer power formulas (TODO is there a documentation for those?)
```
llucax: There should be at least in the code, you can link to code using `[link title][full.public.path.to.python.symbol]`, like `[consumer][frequenz.sdk.timeseries.logical_meter.LogicalMeter.consumer_power]`.

docs/tutorials/pv_optimization.md
```diff
@@ -0,0 +1,117 @@
...
+    consumer_excess_power_engine = (
+        microgrid.logical_meter().producer_power
+        + microgrid.logical_meter().consumer_power
+    ).build("excess_power") # (2)!
+    cons_excess_power_recv = consumer_excess_power_engine.new_receiver() # (3)!
+```
+
+1. The initialization code as explained in the Getting Started tutorial.
+2. Construct the consumer excess power by summing up consumer and producer power each of which having
+opposite signs due to the sign convention. This returns a formula engine.
```
llucax, line 43: It would be nice to link to `FormulaEngine` here, **but**, it is not public :-/

I think all classes that are returned should be public, because otherwise the user doesn't know what they have to offer, even when looking at the API docs, for example:

https://frequenz-floss.github.io/frequenz-sdk-python/v0.25/reference/frequenz/sdk/timeseries/logical_meter/#frequenz.sdk.timeseries.logical_meter.LogicalMeter.consumer_power

image

(no link in the return type, so the user don't know what can be done with it)

llucax (commented):
Just a few early comments.

docs/tutorials/pv_optimization.md
matthias-wende-frequenz: Does anything speak against using relative links? (../../timeseries/...)

docs/tutorials/pv_optimization.md
matthias-wende-frequenz: Yeah this is silly

docs/tutorials/pv_optimization.md
```diff
@@ -0,0 +1,125 @@
+# Optimizing PV production
+
+## Introduction
+
+Before we start it's assumed that you have finished the first [tutorial](./getting_started.md)
+
+In this tutorial you will write an application that optimizes the energy produced from a
+[PV](intro/glossary/#pv) system with Battery for self consumption.
+In order to do so you need to measure the power that flows through the
+[grid connection point](../../intro/glossary/#grid) to determine excess power.
```
shsms, line 10: Once https://github.com/frequenz-floss/frequenz-sdk-python/pull/722 is merged, we'll no longer have to type paths and slugs directly:
```suggestion
{{glossary("Grid", "grid connection point")}} to determine excess power.
```

docs/tutorials/pv_optimization.md
```diff
@@ -0,0 +1,125 @@
...
+flows from the PV system into the grid.
+
+!!! note
+
+    We are using the [passive sign convention](../../intro/glossary/#passive-sign-convention) and thus power
+    flowing from PV is negative and consumed power is positive.
+
+We want to measure the excess power. In order to do so you can use the SDK's data pipeline and especially
+the pre defined
+[consumer](../../reference/frequenz/sdk/timeseries/logical_meter/#frequenz.sdk.timeseries.logical_meter.LogicalMeter.consumer_power)
```
shsms, line 24: This should be
```suggestion
[consumer][frequenz.sdk.timeseries.logical_meter.LogicalMeter.consumer_power]
```

docs/tutorials/pv_optimization.md
```diff
@@ -0,0 +1,125 @@
...
+    )
+
+    # negative means feed-in power due to the negative sign convention
+    consumer_excess_power_engine = (
+        microgrid.logical_meter().producer_power
+        + microgrid.logical_meter().consumer_power
+    ).build("excess_power") # (2)!
+    cons_excess_power_recv = consumer_excess_power_engine.new_receiver() # (3)!
+
+    battery_pool = microgrid.battery_pool() # (1)!
```
shsms, line 105: do the numbered comments make sense when there are no points below the block?  does it reuse the numbered comments from previous examples?

shsms (commented):
Content lgtm, just got comments about the links and number comments in the last example.

docs/tutorials/pv_optimization.md
shsms: also the dir is not called "intro" anymore, it got renamed to "user-guide".

docs/tutorials/pv_optimization.md
```diff
@@ -0,0 +1,125 @@
...
+
+    async for cons_excess_power in cons_excess_power_recv: # (2)!
+        cons_excess_power = cons_excess_power.value # (3)!
+        if cons_excess_power is None: # (4)!
+            continue
+        if cons_excess_power <= Power.zero(): # (5)!
+            await battery_pool.charge(-cons_excess_power)
+        elif cons_excess_power > Power.zero():
```
llucax, line 62: Nitpick: the condition here is redundant, it can just be a simple `else`, but maybe you want to make it explicit for educational purposes.

You could also use `propose_power()`, right? I guess you are not using it also for educational purposes?

docs/tutorials/pv_optimization.md
```diff
@@ -0,0 +1,125 @@
...
+Before we start it's assumed that you have finished the first [tutorial](./getting_started.md)
+
+In this tutorial you will write an application that optimizes the energy produced from a
+[PV](intro/glossary/#pv) system with Battery for self consumption.
+In order to do so you need to measure the power that flows through the
+[grid connection point](../../intro/glossary/#grid) to determine excess power.
+
+## Measure the excess power
+
+When using the term excess power what we actually mean is the consumer excess power, that is the power that
```
Marenz, line 14: quotes would be good around the term here

docs/tutorials/pv_optimization.md
shsms: #722 is merged btw, so this is available upstream.

docs/tutorials/pv_optimization.md
shsms: oh yes, the `charge`/`discharge` methods are no longer available, need to be replaced by the new `propose_*` ones.

llucax: We should ideally enable examples testing for `docs/` too. @Marenz do you think it would be simple to implement?

Marenz: I believe it isn't hard, but it will probably need some reading up on how exactly do do it, maybe half a days work?

llucax: I think it might be worth it now that we have more examples in `docs/`.

I would also like to resurrect the idea about just extracting examples to separate files and run all the tools as normal on them. Another advantage of this is we need to run `pytest` in all combinations of python and archs, but linting could be done only once, there is no need to run the linting on all python versions and archs, and in particular testing examples is super slow, so it makes test in arm64 take much longer than needed.

But the later is probably a fair amount of extra work, so we can also do it as a second step and first only focus on getting `docs/` tested.

llucax: Let's continue in:

*  https://github.com/frequenz-floss/frequenz-repo-config-python/issues/136

PR #705, opened 2023-09-27
Fix bug in how failed batteries are tracked in PowerDistributor

shsms:
There are tests for this case coming with the PowerManager, so not adding them here.


shsms:
This fixes a bug introduced in a previous PR.  No changes have been made since the last release.

Marenz (approved):

shsms:
@llucax maybe this is another possible check for flake8 - when using `set().union()` or similar, the results can't be ignored.  Although not sure how easy it is.

PR #741, opened 2023-10-24
Add a restart delay to actors to avoid spam-restarting of buggy actors

### LINKED ISSUE #732: Introduce a delay between actor restarts
### What's needed?

Currently if an actor raises an exception, it is restarted immediately, but there should be a delay (of maybe 2 seconds) before the next restart is attempted.

Need to decide if this should be configurable.

### Proposed solution

_No response_

### Use cases

_No response_

_No response_

### Additional context

_No response_

christianparpart:
Closes #732

I've decided to go for the hard coded 2.0 seconds delay for now. I very much like the idea of (instead of hard-coding an exponential backoff in), allow configuring the actor at construction time, and have it passed a retry strategy (as suggested by @shms), so the dev of that actor can decide what's best for this particular actor. This can then default to a fixed-delay retry strategy.

### DIFF
--- a/src/frequenz/sdk/actor/_actor.py
+++ b/src/frequenz/sdk/actor/_actor.py
@@ -6,6 +6,7 @@
 import abc
 import asyncio
 import logging
+from datetime import timedelta
 
 from ._background_service import BackgroundService
 
@@ -26,6 +27,9 @@ class Actor(BackgroundService, abc.ABC):
         Please read the [`actor` module documentation][frequenz.sdk.actor] for more
         comprehensive guide on how to use and implement actors properly.
     """
+
+    RESTART_DELAY: timedelta = timedelta(seconds=2)
+    """The delay to wait between restarts of this actor."""
 
     _restart_limit: int | None = None
     """The number of times actors can be restarted when they are stopped by unhandled exceptions.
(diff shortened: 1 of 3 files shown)


src/frequenz/sdk/actor/_actor.py
```diff
@@ -54,6 +54,19 @@ def start(self) -> None:
...
+    async def _delay_restart(self, iteration: int) -> None:
+        """Delay the restart of this actor's n'th iteration.
+
+        Args:
+            iteration: The current iteration of the restart.
+        """
+        # NB: I think it makes sense (in the future) to think about deminishing returns
+        # the longer the actor has been running.
+        # Not just for the restart-delay but actually for the n_restarts counter as well.
+        delay = 2**iteration
```
shsms: we have a generic retry time calculation thing here: https://github.com/frequenz-floss/frequenz-sdk-python/blob/356200443f07acc67b17ba61b0718e7db9643852/src/frequenz/sdk/microgrid/client/_retry.py

we could have a basic LinearBackoff once every 2 seconds as default, but add a method that allows actors to override the strategy.

src/frequenz/sdk/actor/_actor.py
```diff
@@ -54,6 +58,20 @@ def start(self) -> None:
...
+        """Delay the restart of this actor's n'th iteration.
+
+        Args:
+            iteration: The current iteration of the restart.
+        """
+        # NB: I think it makes sense (in the future) to think about deminishing returns
+        # the longer the actor has been running.
+        # Not just for the restart-delay but actually for the n_restarts counter as well.
+        if iteration > 0:
+            delay = Actor.RESTART_DELAY.total_seconds()
```
shsms: I think you should say `self.RESTART_DELAY`, because then, actor implementations can override the interval if they want.

```suggestion
            delay = self.RESTART_DELAY.total_seconds()
```

shsms: or maybe it is better to hide it and keep it internal for now? 🤔

christianparpart: done

shsms (approved):

PR #743, opened 2023-10-25
Set `resent_latest=True` for `BatteryPool.power_status`'s channels

### LINKED ISSUE #740: Allow setting `resend_latest = True` for channels in the `ChannelRegistry`
### What's needed?

Currently channels owned by the `ChannelRegistry` always have `resend_latest = False`,  which might not be the preferred in all cases.

This is particularly useful for the `battery_pool.power_status` method.

### Proposed solution

_No response_

### Use cases

_No response_

_No response_

### Additional context

_No response_

shsms:
Closes #740

### DIFF
--- a/src/frequenz/sdk/actor/_channel_registry.py
+++ b/src/frequenz/sdk/actor/_channel_registry.py
@@ -28,6 +28,29 @@ class ChannelRegistry:
         self._name = name
         self._channels: dict[str, Broadcast[typing.Any]] = {}
 
+    def set_resend_latest(self, key: str, resend_latest: bool) -> None:
+        """Set the `resend_latest` flag for a given channel.
+
+        This flag controls whether the latest value of the channel should be resent to
+        new receivers, in slow streams.
+
+        `resend_latest` is `False` by default.  It is safe to be set in data/reporting
+        channels, but is not recommended for use in channels that stream control
+        instructions.
+
+        Args:
+            key: A key to identify the channel.
+            resend_latest: Whether to resend the latest value to new receivers, for the
+                given channel.
+        """
+        if key not in self._channels:
+            self._channels[key] = Broadcast(f"{self._name}-{key}")
+        # This attribute is protected in the current version of the channels library,
+        # but that will change in the future.
+        self._channels[  # pylint: disable=protected-access
+            key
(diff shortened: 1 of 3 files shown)


src/frequenz/sdk/actor/_channel_registry.py
```diff
@@ -28,6 +28,29 @@ def __init__(self, *, name: str) -> None:
         self._name = name
         self._channels: dict[str, Broadcast[typing.Any]] = {}
 
+    def set_resend_latest(self, key: str, resend_latest: bool) -> None:
```
daniel-zullo-frequenz, line 31: What do you think about setting a default parameter for `resend_latest` to `True`? I think if this method is called will be mainly to set (`True`) `set_resend_latest`  as `_resend_latest ` is `False` by default

src/frequenz/sdk/actor/_channel_registry.py
```diff
@@ -28,6 +28,29 @@ def __init__(self, *, name: str) -> None:
...
+
+        This flag controls whether the latest value of the channel should be resent to
+        new receivers, in slow streams.
+
+        `resend_latest` is `False` by default.  It is safe to be set in data/reporting
+        channels, but is not recommended for use in channels that stream control
+        instructions.
+
+        Args:
+            key: A key to identify the channel.
```
daniel-zullo-frequenz:
```suggestion
            key: The key to identify the channel.
```

daniel-zullo-frequenz (commented):
I have a question and a cosmetic suggestion. Other than that LGTM

src/frequenz/sdk/actor/_channel_registry.py
shsms: I always think of default as the safe value to use when the user doesn't care.  Here, if we make it default,  it feels like changing the meaning of the function - instead of setting the given value, it would be a `set/reset` thing, then we'll probably need a `reset_*` to be consistent.  

I feel users would call this only with a clear intention, so it is fine to expect that they specify what it is.

daniel-zullo-frequenz: Fair points

daniel-zullo-frequenz (approved):

PR #746, opened 2023-10-26
Add consumtion and production operations to Formula Engine

matthias-wende-frequenz:
In order to simplify usability but don't loose functionality we decided to remove the consumption and production power formulas, i.e. the formulas that are having a production or consumption attached to their name, and replace them with operators that can be called on a formula engine.

The consumption operator returns either the identity if the power value is
positive or 0 and the production operator returns either the identity if the power value is
negative or 0.

As an example we can now create the following formula:

```python
grid_consumption = logical_meter.grid_power().consumption().build()
```

This would return the consumption part of the grid_power stream.

Removing the old consumption and production power formulas will happen in the follow up PR https://github.com/frequenz-floss/frequenz-sdk-python/pull/697.

### DIFF
--- a/src/frequenz/sdk/timeseries/formula_engine/_formula_engine.py
+++ b/src/frequenz/sdk/timeseries/formula_engine/_formula_engine.py
@@ -198,6 +198,24 @@ class _ComposableFormulaEngine(
         """
         return self._higher_order_builder(self, self._create_method).min(other)  # type: ignore
 
+    def consumption(self) -> _GenericHigherOrderBuilder:
+        """
+        Return a formula builder that applies the consumption operator on `self`.
+
+        The consumption operator returns either the identity if the power value is
+        positive or 0.
+        """
+        return self._higher_order_builder(self, self._create_method).consumption()  # type: ignore
+
+    def production(self) -> _GenericHigherOrderBuilder:
+        """
+        Return a formula builder that applies the production operator on `self`.
+
+        The production operator returns either the absolute value if the power value is
+        negative or 0.
+        """
+        return self._higher_order_builder(self, self._create_method).production()  # type: ignore
+
 
 class FormulaEngine(
     Generic[QuantityT],
--- a/src/frequenz/sdk/timeseries/formula_engine/_formula_engine.py
+++ b/src/frequenz/sdk/timeseries/formula_engine/_formula_engine.py
@@ -192,6 +192,7 @@ class _ComposableFormulaEngine(
             other: A formula receiver, a formula builder or a QuantityT instance
                 corresponding to a sub-expression.
 
+
         Returns:
             A formula builder that can take further expressions, or can be built
                 into a formula engine.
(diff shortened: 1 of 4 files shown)


src/frequenz/sdk/timeseries/formula_engine/_formula_steps.py
```diff
@@ -177,6 +177,64 @@ def apply(self, eval_stack: list[float]) -> None:
...
+        val = eval_stack.pop()
+        eval_stack.append(max(val, 0))
+
+
+class Production(FormulaStep):
+    """A formula step that represents the consumption operator.
+
+    The consumption operator is the maximum of the value on top
+    of the evaluation stack and 0.
+    """
```
shsms:
```suggestion
class Production(FormulaStep):
    """A formula step that represents the production operator.

    The production operator is the maximum of the value on top
    of the evaluation stack * -1 and 0.
    """

File: src/frequenz/sdk/timeseries/formula_engine/_formula_steps.py [hunk 2]
```diff
@@ -177,6 +177,64 @@ def apply(self, eval_stack: list[float]) -> None:
...
+        Returns:
+            A string representation of the step.
+        """
+        return "consumtion"
+
+    def apply(self, eval_stack: list[float]) -> None:
+        """
+        Apply the consumer formula.
+
+        Replace the top of the eval eval_stack with the same value is the value
```
shsms:
```suggestion
        Replace the top of the eval eval_stack with the same value if the value

File: src/frequenz/sdk/timeseries/formula_engine/_formula_steps.py [hunk 3]
```diff
@@ -177,6 +177,64 @@ def apply(self, eval_stack: list[float]) -> None:
...
+        """Return a string representation of the step.
+
+        Returns:
+            A string representation of the step.
+        """
+        return "production"
+
+    def apply(self, eval_stack: list[float]) -> None:
+        """
+        Apply the consumer formula.
```
shsms:
```suggestion
        Apply the production step.

File: src/frequenz/sdk/timeseries/formula_engine/_formula_engine.py [hunk 4]
```diff
@@ -1020,6 +1021,52 @@ def min(
...
+
+    def production(
+        self,
+    ) -> (
+        HigherOrderFormulaBuilder[QuantityT]
+        | HigherOrderFormulaBuilder3Phase[QuantityT]
+    ):
+        """Apply the Production Operator.
+
+        The production operator returns either the identity if the power value is
```
shsms:
```suggestion
        The production operator returns either the identity * -1 if the power value is

File: src/frequenz/sdk/timeseries/formula_engine/_formula_engine.py [hunk 5]
```diff
@@ -1020,6 +1021,52 @@ def min(
...
+            self, (HigherOrderFormulaBuilder, HigherOrderFormulaBuilder3Phase)
+        )
+        return self
+
+    def production(
+        self,
+    ) -> (
+        HigherOrderFormulaBuilder[QuantityT]
+        | HigherOrderFormulaBuilder3Phase[QuantityT]
+    ):
```
shsms, line 1054: Because these are unary operators, and our conversion to postfix can only deal with binary operators, these steps should call build from within, so that no further steps can be added on this layer.

Or an alternative is to turn these into binary operators by adding a second `0` operand that gets discarded by the corresponding formula steps.

matthias-wende-frequenz: Why do you think it can only deal with binary operators? I don't think this is true and I've added tests that show that the combination works without flaws.

After all in the case of a binary operator the Formula Step simply pops one value from the formula stack and pushes the result back. Since we are adding brackets before adding the consumption or production operators we make sure that the formula, that we operate on, is evaluated first and then we apply these operators on the remaining value.

src/frequenz/sdk/timeseries/formula_engine/_formula_engine.py
```diff
@@ -198,6 +198,24 @@ def min(
...
+        The consumption operator returns either the identity if the power value is
+        positive or 0.
+        """
+        return self._higher_order_builder(self, self._create_method).consumption()  # type: ignore
+
+    def production(self) -> _GenericHigherOrderBuilder:
+        """
+        Return a formula builder that applies the production operator on `self`.
+
+        The production operator returns either the identity if the power value is
```
shsms: `* -1`?

src/frequenz/sdk/timeseries/formula_engine/_formula_engine.py
shsms: true, that appears to be the case.  As long as the input to the infix-to-postfix algorithm looks like this: `((a + b) production) + c`, etc., it seems to work.  I think the brackets after production/consumption are also important.

tests/timeseries/test_formula_engine.py
```diff
@@ -586,6 +599,95 @@ async def test_compound(self) -> None:
...
+        )
+
+    async def test_consumption_production(self) -> None:
+        """Test the consumption and production operator combined."""
+        await self.run_test(
+            2,
+            lambda c1, c2: c1.consumption() + c2.production(),
+            [
+                ([10.0, 10.0], 10.0),
+                ([-10.0, -10.0], 10.0),
```
shsms: some disambiguation might be helpful:

```suggestion
                ([10.0, 12.0], 10.0),
                ([-12.0, -10.0], 10.0),

shsms (dismissed):
Just an optional comment.

shsms (approved):

PR #747, opened 2023-10-27
Change internal presense of the Grid, eliminating `_grid`

christianparpart:
This is the follow-up PR to PR #707.

### DIFF
--- a/src/frequenz/sdk/microgrid/_data_pipeline.py
+++ b/src/frequenz/sdk/microgrid/_data_pipeline.py
@@ -21,9 +21,9 @@ from frequenz.channels import Broadcast, Sender
 from ..actor._actor import Actor
 from ..microgrid.component import Component
 from ..timeseries._grid_frequency import GridFrequency
+from ..timeseries.grid import Grid
+from ..timeseries.grid import get as get_grid
 from . import connection_manager
-from ._grid import Grid
-from ._grid import get as get_grid
 from .component import ComponentCategory
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


src/frequenz/sdk/microgrid/grid.py
shsms, line 1: Can't believe that this works, because now there's two `microgrid.grid` things - one a module, other a function.  If you remember, this was the reason we renamed the file to `_grid` originally.  Maybe you meant to move it to `timeseries.grid`?

christianparpart: yeah, i'll adapt it, it is really interesting that the tests PASSED locally (!). I'm not done yet (apparently) :-)

src/frequenz/sdk/microgrid/_data_pipeline.py
```diff
@@ -21,9 +21,9 @@
 from ..actor._actor import Actor
 from ..microgrid.component import Component
 from ..timeseries._grid_frequency import GridFrequency
+from ..timeseries.grid import Grid
+from ..timeseries.grid import get as get_grid
```
daniel-zullo-frequenz, line 25: This can probably be done on the same line:
```suggestion
from ..timeseries.grid import get as get_grid, Grid
```

daniel-zullo-frequenz (commented):
Only have an optional comment to check for and LGTM

src/frequenz/sdk/microgrid/_data_pipeline.py
christianparpart: `isort` will complain then. So I reverted that again.

daniel-zullo-frequenz (approved):

PR #755, opened 2023-10-30
Use exclusion bounds while calculating target powers in PowerManager

shsms:
This also helps with the upcoming bounds-interface update, to hide the exclusion bounds from SDK users, in preparation for its deprecation in the microgrid API in favour of a simpler interface.

### DIFF
--- a/src/frequenz/sdk/actor/_power_managing/_matryoshka.py
+++ b/src/frequenz/sdk/actor/_power_managing/_matryoshka.py
@@ -36,6 +36,70 @@ if typing.TYPE_CHECKING:
 _logger = logging.getLogger(__name__)
 
 
+def _check_exclusion_bounds_overlap(
+    lower_bound: Power,
+    upper_bound: Power,
+    exclusion_bounds: timeseries.Bounds[Power] | None,
+) -> tuple[bool, bool]:
+    """Check if the given bounds overlap with the given exclusion bounds.
+
+    Args:
+        lower_bound: The lower bound to check.
+        upper_bound: The upper bound to check.
+        exclusion_bounds: The exclusion bounds to check against.
+
+    Returns:
+        A tuple containing a boolean indicating if the lower bound is bounded by the
+            exclusion bounds, and a boolean indicating if the upper bound is bounded by
+            the exclusion bounds.
+    """
+    if exclusion_bounds is None:
+        return False, False
+
+    bounded_lower = False
+    bounded_upper = False
+
+    if exclusion_bounds.lower < lower_bound < exclusion_bounds.upper:
+        bounded_lower = True
+    if exclusion_bounds.lower < upper_bound < exclusion_bounds.upper:
+        bounded_upper = True
+
+    return bounded_lower, bounded_upper
+
+
+def _adjust_exclusion_bounds(
+    lower_bound: Power,
+    upper_bound: Power,
+    exclusion_bounds: timeseries.Bounds[Power] | None,
+) -> tuple[Power, Power]:
+    """Adjust the given bounds to exclude the given exclusion bounds.
+
+    Args:
+        lower_bound: The lower bound to adjust.
+        upper_bound: The upper bound to adjust.
+        exclusion_bounds: The exclusion bounds to adjust to.
+
+    Returns:
+        The adjusted lower and upper bounds.
+    """
+    if exclusion_bounds is None:
+        return lower_bound, upper_bound
+
+    # If the given bounds are within the exclusion bounds, there's no room to adjust,
+    # so return zero.
+    #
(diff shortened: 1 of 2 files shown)


shsms:
The release notes still refer to the PowerManager as a new feature, and there's nothing in the existing release notes to suggest that this wasn't there, so not changing anything.

src/frequenz/sdk/actor/_power_managing/_matryoshka.py
```diff
@@ -83,11 +83,12 @@ def _calc_target_power(
...
-            if low is not None:
-                lower_bound = max(lower_bound, low)
-            if high is not None:
-                upper_bound = min(upper_bound, high)
+            proposal_lower, proposal_upper = (
+                next_proposal.bounds.lower or lower_bound,
+                next_proposal.bounds.upper or upper_bound,
+            )
+            lower_bound = max(lower_bound, proposal_lower)
+            upper_bound = min(upper_bound, proposal_upper)
```
llucax: Not expecting any actions but curious why sometimes you use tuple syntax for multiple assignments and when not :thinking: 

Why not:

```py
            proposal_lower, proposal_upper = (
                next_proposal.bounds.lower or lower_bound,
                next_proposal.bounds.upper or upper_bound,
            )
            lower_bound, upper_bound = (
                max(lower_bound, proposal_lower),
                min(upper_bound, proposal_upper),
            )
```

or

```py
            proposal_lower = next_proposal.bounds.lower or lower_bound
            proposal_upper = next_proposal.bounds.upper or upper_bound
            lower_bound = max(lower_bound, proposal_lower)
            upper_bound = min(upper_bound, proposal_upper)
```

:smiley: 

(in my case I find the later slightly more readable, I need to do less mapping of what is assigned to what)

src/frequenz/sdk/actor/_power_managing/_matryoshka.py
```diff
@@ -36,6 +36,70 @@
 _logger = logging.getLogger(__name__)
 
 
+def _check_exclusion_bounds_overlap(
+    lower_bound: Power,
+    upper_bound: Power,
+    exclusion_bounds: timeseries.Bounds[Power] | None,
+) -> tuple[bool, bool]:
+    """Check if the given bounds overlap with the given exclusion bounds.
+
```
llucax, line 99: Maybe adding a (graphical) example would be useful here.

src/frequenz/sdk/actor/_power_managing/_matryoshka.py
```diff
@@ -36,6 +36,60 @@
 _logger = logging.getLogger(__name__)
 
 
+# Just 20 lines of code in this function, but unfortunately 8 of those are return
+# statements, and that's too many for pylint.
+def _clamp_to_bounds(  # pylint: disable=too-many-return-statements
```
llucax, line 41: I wonder if we should remove this check completely, I don't see anything wrong in having many `returns`, I think is good for readability to avoid deep nesting and it helps short-circuiting by discarding simple cases first.

src/frequenz/sdk/actor/_power_managing/_matryoshka.py
```diff
@@ -36,6 +36,124 @@
...
+            return exclusion_bounds.upper
+
+    return value
+
+
+def _check_exclusion_bounds_overlap(
+    lower_bound: Power,
+    upper_bound: Power,
+    exclusion_bounds: timeseries.Bounds[Power] | None,
+) -> tuple[bool, bool]:
```
llucax, line 97: Not 100% sure if it is worth it, but for extra clarity/readability you could use a dataclass:

```py
@dataclass(frozen=True, kw_only=True)
class Overlap:
    lower: bool
    upper: bool
```

Then you can use the more explicit `match` syntax:

```py
        match _check_exclusion_bounds_overlap(
            lower_bound, upper_bound, exclusion_bounds
        ):
            case Overlap(lower=True, upper=True):
                return Power.zero()
            case Overlap(lower=True, upper=False):
                if value < exclusion_bounds.upper:
                    return exclusion_bounds.upper
            case Overlap(lower=(False, lower=True):
                if value > exclusion_bounds.lower:
                    return exclusion_bounds.lower
```

llucax (approved):
LGTM, all comments are optional.

shsms:
Agree to all of them except `class Overlap`, but will do them in the next PR,  already working on the same parts anyway.

src/frequenz/sdk/actor/_power_managing/_matryoshka.py
shsms: @llucax is this for repo config?

llucax: Yeah, if we agree about it.

PR #756, opened 2023-10-31
Make the SDK's representation of bounds data internal

shsms:
This make the `inclusion_bounds` and `exclusion_bounds` fields of the
battery pool's `Report` class private, and instead introduces a
`bounds` property as a single set of bounds that users can use.

This PR also introduces the `Report.adjust_to_bounds` method which can
be used in case more granularity is needed on what exact power values
can be proposed to the battery pool.

### DIFF
--- a/src/frequenz/sdk/actor/_power_managing/_base_classes.py
+++ b/src/frequenz/sdk/actor/_power_managing/_base_classes.py
@@ -50,14 +50,14 @@ class Report:
     target_power: Power | None
     """The currently set power for the batteries."""
 
-    inclusion_bounds: timeseries.Bounds[Power] | None
+    _inclusion_bounds: timeseries.Bounds[Power] | None
     """The available inclusion bounds for the batteries, for the actor's priority.
 
     These bounds are adjusted to any restrictions placed by actors with higher
     priorities.
     """
 
-    exclusion_bounds: timeseries.Bounds[Power] | None
+    _exclusion_bounds: timeseries.Bounds[Power] | None
     """The exclusion bounds for the batteries.
 
     The power manager doesn't manage exclusion bounds, so these are aggregations of
@@ -72,6 +72,20 @@ class Report:
 
     This is `None` if no power distribution has been performed yet.
     """
+
+    @property
+    def bounds(self) -> timeseries.Bounds[Power] | None:
+        """The bounds for the batteries.
+
+        These bounds are adjusted to any restrictions placed by actors with higher
+        priorities.
+
+        There might be exclusion zones within these bounds. If necessary, the
+        [`adjust_to_bounds`][frequenz.sdk.timeseries.battery_pool.Report.adjust_to_bounds]
+        method may be used to check if a desired power value fits the bounds, or to get
+        the closest possible power values that do fit the bounds.
+        """
+        return self._inclusion_bounds
 
     def adjust_to_bounds(self, power: Power) -> tuple[Power | None, Power | None]:
         """Adjust a power value to the bounds.
@@ -123,14 +137,14 @@ class Report:
             A tuple of the closest power values to the desired power that fall within
                 the available bounds for the actor.
         """
-        if self.inclusion_bounds is None:
+        if self._inclusion_bounds is None:
             return None, None
 
         return _clamp_to_bounds(
             power,
-            self.inclusion_bounds.lower,
-            self.inclusion_bounds.upper,
-            self.exclusion_bounds,
+            self._inclusion_bounds.lower,
+            self._inclusion_bounds.upper,
+            self._exclusion_bounds,
         )
 
 
--- a/src/frequenz/sdk/actor/_power_managing/_base_classes.py
+++ b/src/frequenz/sdk/actor/_power_managing/_base_classes.py
@@ -13,7 +13,7 @@ import typing
 
 from ... import timeseries
 from ...timeseries import Power
-from ._bounds_methods import _clamp_to_bounds
+from . import _bounds
 
 if typing.TYPE_CHECKING:
     from ...timeseries.battery_pool import PowerMetrics
@@ -50,6 +50,12 @@ class Report:
     target_power: Power | None
     """The currently set power for the batteries."""
 
+    distribution_result: power_distributing.Result | None
+    """The result of the last power distribution.
+
+    This is `None` if no power distribution has been performed yet.
+    """
+
     _inclusion_bounds: timeseries.Bounds[Power] | None
     """The available inclusion bounds for the batteries, for the actor's priority.
 
@@ -65,12 +71,6 @@ class Report:
 
     These bounds are adjusted to any restrictions placed by actors with higher
     priorities.
-    """
-
-    distribution_result: power_distributing.Result | None
-    """The result of the last power distribution.
-
-    This is `None` if no power distribution has been performed yet.
     """
 
     @property
@@ -140,7 +140,7 @@ class Report:
         if self._inclusion_bounds is None:
             return None, None
 
-        return _clamp_to_bounds(
+        return _bounds.clamp_to_bounds(
             power,
             self._inclusion_bounds.lower,
             self._inclusion_bounds.upper,
--- a/src/frequenz/sdk/actor/_power_managing/_base_classes.py
+++ b/src/frequenz/sdk/actor/_power_managing/_base_classes.py
@@ -13,6 +13,7 @@ import typing
 
 from ... import timeseries
 from ...timeseries import Power
+from ._bounds_methods import _clamp_to_bounds
 
 if typing.TYPE_CHECKING:
     from ...timeseries.battery_pool import PowerMetrics
@@ -71,6 +72,66 @@ class Report:
 
     This is `None` if no power distribution has been performed yet.
     """
+
+    def adjust_to_bounds(self, power: Power) -> tuple[Power | None, Power | None]:
+        """Adjust a power value to the bounds.
+
+        This method can be used to adjust a desired power value to the power bounds
+        available to the actor.
+
+        If the given power value falls within the usable bounds, it will be returned
+        unchanged.
+
+        If it falls outside the usable bounds, the closest possible value on the
+        corresponding side will be returned.  For example, if the given power is lower
+        than the lowest usable power, only the lowest usable power will be returned, and
+        similarly for the highest usable power.
+
+        If the given power falls within an exclusion zone that's contained within the
+        usable bounds, the closest possible power values on both sides will be returned.
+
+        !!! note
+            It is completely optional to use this method to adjust power values before
+            proposing them through the battery pool, because the battery pool will do
+            this automatically.  This method is provided for convenience, and for
+            granular control when there are two possible power values, both of which
+            fall within the available bounds.
+
+        Example:
+            ```python
+            from frequenz.sdk import microgrid
+
+            power_status_rx = microgrid.battery_pool().power_status.new_receiver()
+            power_status = await power_status_rx.receive()
+            desired_power = Power.from_watts(1000.0)
+
+            match power_status.adjust_to_bounds(desired_power):
+                case (power, _) if power == desired_power:
+                    print("Desired power is available.")
+                case (None, power) | (power, None) if power:
+                    print(f"Closest available power is {power}.")
+                case (lower, upper) if lower and upper:
+                    print(f"Two options {lower}, {upper} to propose to battery pool.")
+                case (None, None):
+                    print("No available power")
+            ```
+
+        Args:
+            power: The power value to adjust.
+
+        Returns:
+            A tuple of the closest power values to the desired power that fall within
+                the available bounds for the actor.
+        """
+        if self.inclusion_bounds is None:
+            return None, None
+
+        return _clamp_to_bounds(
+            power,
+            self.inclusion_bounds.lower,
+            self.inclusion_bounds.upper,
+            self.exclusion_bounds,
+        )
 
 
 @dataclasses.dataclass(frozen=True, kw_only=True)
--- a/src/frequenz/sdk/actor/_power_managing/_bounds.py
+++ b/src/frequenz/sdk/actor/_power_managing/_bounds.py
@@ -15,6 +15,7 @@ def check_exclusion_bounds_overlap(
 
     Example:
 
+        ```
                        lower                        upper
                           .----- exclusion zone -----.
         -----|✓✓✓✓✓✓✓✓✓✓✓✓|xxxxxxxxxxxxxxx|----------|----
@@ -24,6 +25,7 @@ def check_exclusion_bounds_overlap(
            lower                        upper
            bound                        bound
                               (inside the exclusion zone)
+        ```
 
         Resulting in `(False, True)` because only the upper bound is inside the
         exclusion zone.
--- a/src/frequenz/sdk/actor/_power_managing/_bounds.py
+++ b/src/frequenz/sdk/actor/_power_managing/_bounds.py
@@ -1,32 +1,32 @@
 # License: MIT
 # Copyright © 2023 Frequenz Energy-as-a-Service GmbH
 
-"""Methods for checking and clamping bounds and power values to exclusion bounds."""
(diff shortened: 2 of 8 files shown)


shsms:
Here too,  there's a line already in the release notes providing sufficient context:  https://github.com/frequenz-floss/frequenz-sdk-python/blob/7da41bf58f53892fcb354f0df4370e2f7120ed85/RELEASE_NOTES.md?plain=1#L26

So not adding anything more.

src/frequenz/sdk/actor/_power_managing/_base_classes.py
```diff
@@ -72,6 +73,80 @@ class Report:
...
+        priorities.
+
+        There might be exclusion zones within these bounds. If necessary, the
+        [`adjust_to_bounds`][frequenz.sdk.timeseries.battery_pool.Report.adjust_to_bounds]
+        method may be used to check if a desired power value fits the bounds, or to get
+        the closest possible power values that do fit the bounds.
+        """
+        return self._inclusion_bounds
+
+    def adjust_to_bounds(self, power: Power) -> tuple[Power | None, Power | None]:
```
shsms, line 90: Still need to figure out what the right return type is.  They are not bounds, but the closest available power values that can be proposed to the battery pool.

That said,  this whole method can be safely ignored and people can just rely on the PowerManager, so there's no need to make this more clear.

There's also an example below in this method's docstring, that shows how it can be used with a `match` statement.  Maybe that's all that's needed.

src/frequenz/sdk/actor/_power_managing/_base_classes.py
```diff
@@ -49,14 +50,14 @@ class Report:
     target_power: Power | None
     """The currently set power for the batteries."""
 
-    inclusion_bounds: timeseries.Bounds[Power] | None
+    _inclusion_bounds: timeseries.Bounds[Power] | None
```
matthias-wende-frequenz, line 59: I'd suggest not to use this terminology any longer, but if you decide to do so, it needs documentation.

src/frequenz/sdk/actor/_power_managing/_base_classes.py
```diff
@@ -72,6 +73,80 @@ class Report:
     This is `None` if no power distribution has been performed yet.
     """
 
+    @property
+    def bounds(self) -> timeseries.Bounds[Power] | None:
+        """The bounds for the batteries.
```
matthias-wende-frequenz, line 78: not only for batteries, right?

src/frequenz/sdk/actor/_power_managing/_bounds_methods.py
```diff
@@ -0,0 +1,143 @@
...
+    lower_bound: Power,
+    upper_bound: Power,
+    exclusion_bounds: Bounds[Power] | None,
+) -> tuple[bool, bool]:
+    """Check if the given bounds overlap with the given exclusion bounds.
+
+    When only the upper bound overlaps with exclusion bounds, the usable range is
+    between the lower bound and the lower exclusion bound, like below.
+
+      ===lb+++++++ex----ub-------ex===
```
llucax: To be honest, this sort of confuses me even more, because here you talk about usable range but the function talks about the exclusion zone.

Suggestion of a more explicit alternative:

> ```
>               lower                        upper
>                 .------ exclusion zone ------.
> -----|✔✔✔✔✔✔✔✔✔✔✔✔|xxxxxxxxxxxxxxx|----------|----
>      `-- usable --'-- exclusion --´
>      |                 overlap    |
>      |                            |
>    lower                        upper
>    bound                        bound
>                       (inside the exclusion zone)
> ```
> 
> Resulting in `(False, True)` because only the upper bound is inside the exclusion zone.

For me with this there is no need to do the other variants, seems clear enough with one example.

src/frequenz/sdk/actor/_power_managing/_matryoshka.py
```diff
@@ -217,12 +221,23 @@ def _calc_target_power(
...
+                match _clamp_to_bounds(
                     next_proposal.preferred_power,
                     lower_bound,
                     upper_bound,
                     exclusion_bounds,
-                )
+                ):
+                    case (None, power) | (power, None) if power:
+                        target_power = power
+                    case (power_low, power_high) if power_low and power_high:
```
llucax, line 96: :exploding_head: I didn't remember the `if` syntax.

src/frequenz/sdk/actor/_power_managing/_bounds_methods.py
llucax, line 1: What about just `_bounds.py`? Methods sounds weird, specially when they are free functions, and the module name shouldn't really tell if you store classes, or functions or methods, but what do they do.

src/frequenz/sdk/actor/_power_managing/_bounds_methods.py
```diff
@@ -0,0 +1,143 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Methods for checking and clamping bounds and power values to exclusion bounds."""
```
llucax:
```suggestion
"""Utilities dealing with bounds, like checking and clamping, and power values to exclusion bounds."""
```

src/frequenz/sdk/actor/_power_managing/_bounds_methods.py
```diff
@@ -0,0 +1,143 @@
...
+
+from ...timeseries import Bounds, Power
+
+
+def _check_exclusion_bounds_overlap(
```
llucax: I would make the symbols public now (no underscore), as they are meant to be imported from other modules. Having the module *private* (starting with `_`) should be enough to hide it from the end user interface.

src/frequenz/sdk/actor/_power_managing/_base_classes.py
llucax: Why publishing it if you don't have to use it then?

If you want to keep it, about the type, maybe `PowerRange`?

PowerRangers

src/frequenz/sdk/actor/_power_managing/_base_classes.py
```diff
@@ -49,14 +50,14 @@ class Report:
...
-    inclusion_bounds: timeseries.Bounds[Power] | None
+    _inclusion_bounds: timeseries.Bounds[Power] | None
     """The available inclusion bounds for the batteries, for the actor's priority.
 
     These bounds are adjusted to any restrictions placed by actors with higher
     priorities.
     """
 
-    exclusion_bounds: timeseries.Bounds[Power] | None
+    _exclusion_bounds: timeseries.Bounds[Power] | None
```
llucax: Minor, but maybe you want to move these to the end of the attributes, so all *public attributes* are listed first.

src/frequenz/sdk/actor/_power_managing/_bounds_methods.py
shsms: That looks awesome, did you make that by hand?

src/frequenz/sdk/actor/_power_managing/_base_classes.py
shsms: We'll be forced to change it when we upgrade to the new microgrid API.  Until then, the internal repr just reflects the API version.  Is why I made them internal anyway, so people don't have to deal with these terms anymore.

src/frequenz/sdk/actor/_power_managing/_base_classes.py
shsms: Except it is not a `Range` either,  they are the two best possible options.  

Most users shouldn't have to worry about this method, but it  might be useful in one specific situation:  If an actor sets bounds that fall entirely within the exclusion bounds range, they might want to be able to choose which side of the exclusion bounds they want to go to.

shsms: I'm open to removing it and adding it if there seems to be a need for such flexibility among users.  Especially in cases where there's only a single actor running, they'll be able to get more control by using this, if they want.

src/frequenz/sdk/actor/_power_managing/_base_classes.py
shsms: oh true.  I guess we need separate `Report` classes, one for `BatteryPool`, `PVPool`, etc., so they can have precise documentation.  I think that's the only public interface class, so probably not so bad.  

And actually can just be done by inheriting from the PowerManager's report I think.

src/frequenz/sdk/actor/_power_managing/_base_classes.py
shsms: I guess this is one of the problems with trying to stabilize an API that has never been used in production before.

Maybe we just prefix with `_` and call it experimental for now?  Or maybe prefix with `experimental_`, so automatic documentation works?

src/frequenz/sdk/actor/_power_managing/_base_classes.py
shsms: unfortunately the word "Battery" is everywhere in the power manager implementation.  I will make a separate PR after this to change that.

src/frequenz/sdk/actor/_power_managing/_base_classes.py
shsms: Didn't do anything, leave it as it is.  The function returns a pair of powers, which don't have two separate names, just two options, so I think a tuple makes sense.

src/frequenz/sdk/actor/_power_managing/_bounds_methods.py
shsms: done

src/frequenz/sdk/actor/_power_managing/_matryoshka.py
shsms: yes, the whole thing is very similar to the `pcase` in emacs lisp, and the `match` from rust.

src/frequenz/sdk/actor/_power_managing/_bounds_methods.py
shsms: true, done.

src/frequenz/sdk/actor/_power_managing/_bounds_methods.py
shsms: It is actually clamping of (bounds and power values) to exclusion bounds.  So just changed to: 

> Utilities for checking and clamping bounds and power values to exclusion bounds.

src/frequenz/sdk/actor/_power_managing/_bounds_methods.py
shsms: done

src/frequenz/sdk/actor/_power_managing/_base_classes.py
shsms: done

src/frequenz/sdk/actor/_power_managing/_base_classes.py
llucax: I'm all for leaving it with `_`, so if someone needs it, they can use it already, and we don't need to remove it now and add it later. If people really start to use it and it works, we make it public. I really like this approach as "the best of both worlds".

src/frequenz/sdk/actor/_power_managing/_bounds_methods.py
llucax: > That looks awesome, did you make that by hand?

Yes :blush:

llucax: I would love for GitHub/mkdocs to eventually support [aafigure](https://aafigure.readthedocs.io/en/latest/shortintro.html) (ASCII-art to image conversion).

* Rendered examples: https://aafigure.readthedocs.io/en/latest/examples.html
* Source: https://github.com/aafigure/aafigure/blob/master/documentation/examples.rst

src/frequenz/sdk/actor/_power_managing/_base_classes.py
shsms: is there anyway to override the doc generator to show docs for this function even if it is `_`?

src/frequenz/sdk/actor/_power_managing/_matryoshka.py
llucax: And [D's template constraints](https://dlang.org/spec/template.html#template_constraints) :laughing:

src/frequenz/sdk/actor/_power_managing/_bounds.py
```diff
@@ -0,0 +1,143 @@
...
+
+                       lower                        upper
+                          .----- exclusion zone -----.
+        -----|✓✓✓✓✓✓✓✓✓✓✓✓|xxxxxxxxxxxxxxx|----------|----
+             `-- usable --'-- exclusion --´
+             |                 overlap    |
+             |                            |
+           lower                        upper
+           bound                        bound
+                              (inside the exclusion zone)
```
llucax, line 27: This will not be rendered for now as it is private, but if at some point we make it public, it will render incorrectly, so I suggest:

~~~suggestion
        ```
                       lower                        upper
                          .----- exclusion zone -----.
        -----|✓✓✓✓✓✓✓✓✓✓✓✓|xxxxxxxxxxxxxxx|----------|----
             `-- usable --'-- exclusion --´
             |                 overlap    |
             |                            |
           lower                        upper
           bound                        bound
                              (inside the exclusion zone)
        ```
~~~

src/frequenz/sdk/actor/_power_managing/_matryoshka.py
shsms: wow, wouldn't have guessed D also has functional pattern matching.

src/frequenz/sdk/actor/_power_managing/_base_classes.py
llucax: No, but that's the point, right? That it is hidden, but if someone asks for the functionality we can say "use this"?

llucax (commented):
This one small comment and the pending decision on if we make `adjust_to_bounds()` private or not.

src/frequenz/sdk/actor/_power_managing/_base_classes.py
shsms: This method's documentation provides a lot of context about the behaviour of the bounds, so I'd like it to be visible,  but I'll add a line saying that the method is experimental.

and it looks like it is possible:  https://mkdocstrings.github.io/python/usage/configuration/members/

I will test.

llucax: In the API that's not really possible, as all of that is auto-generated and we don't have a way to override how we render a particular module or symbol. You could pull that documentation into some other doc in `docs/`. What I don't like about making it public is that autocompletion might suggest it and users might not look at the docs in detail, and after you are using it, there no indication that you are using something experimental.

With `_` you don't get auto-completion and the code is marked with `# pylint: disable=private-access` (or whatever is called), which clearly shows something is fishy, and it is also clear that private interfaces are not stable.

Can you put that documentation somewhere else? Maybe in the class, saying "There is an experimental `_adjust_to_bounds()` that does this and that"?

shsms: Actually I'd like to take some more time to think about this,  so I'd like to do this in the next PR.

The next PR still deals with these parts - it makes this class internal, and generic, removing all mention of batteries,  and there will be a `Protocol` type (as I see it now), that would expose the methods, but with specific documentation and corresponding to the pool.

So we won't forget about this if we don't do it now.

shsms: > What I don't like about making it public is that autocompletion might suggest it and users might not look at the docs in detail, and after you are using it, there no indication that you are using something experimental.

I mean, this whole thing is one big experimental feature, not just this function.  We're basically changing all of these interfaces and trying to release them before we've had a chance to test them.

llucax: Fair enough!

llucax:
LGTM, leaving @matthias-wende-frequenz to approve as he also has some pending comments.

matthias-wende-frequenz (approved):

PR #757, opened 2023-11-01
Add try/finally to decorated context managers yield statement

### LINKED ISSUE #721: Add `try`/`finally` to decorated context managers
### What happened?

Some decorated context managers are not catching exceptions in the `yield`, so if an exception happens the finalization will not be performed.

Finalization code to be executed on exceptions too.

### Affected version(s)

_No response_

### Affected part(s)

### Extra information

Code using decorated context manager:

* https://github.com/search?q=repo%3Afrequenz-floss%2Ffrequenz-sdk-python%20contextmanager&type=code

Low priority because it only affects benchmarks and tests.

christianparpart:
And also log any unhandled caught exception.

Closes #721.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -36,9 +36,11 @@ def actor_restart_limit(limit: int) -> Iterator[None]:
         f"<actor_restart_limit> Changing the restart limit from {original_limit} to {limit}"
     )
     _actor.Actor._restart_limit = limit
-    yield
-    print(f"<actor_restart_limit> Resetting restart limit to {original_limit}")
-    _actor.Actor._restart_limit = original_limit
+    try:
+        yield
+    finally:
+        print(f"<actor_restart_limit> Resetting restart limit to {original_limit}")
+        _actor.Actor._restart_limit = original_limit
 
 
 @pytest.fixture(scope="session", autouse=True)


tests/conftest.py
```diff
@@ -36,9 +36,14 @@ def actor_restart_limit(limit: int) -> Iterator[None]:
...
     )
     _actor.Actor._restart_limit = limit
-    yield
-    print(f"<actor_restart_limit> Resetting restart limit to {original_limit}")
-    _actor.Actor._restart_limit = original_limit
+    try:
+        yield
+    except Exception as e:  # pylint: disable=broad-except
+        print(f"<actor_restart_limit> Unhandled exception caught: {e}")
+        raise
```
llucax: Not sure if this is providing anything useful, if the exception is re-raised, probably `pytest` should catch it and show all the context of the failure anyway.

tests/microgrid/test_client.py
```diff
@@ -62,6 +62,9 @@ async def _gprc_server(
     await server.start()
     try:
         yield servicer, microgrid
+    except Exception as e:  # pylint: disable=broad-except
+        print(f"Unhandled exception caught: {e}")
+        raise
```
llucax: Same.

tests/timeseries/test_periodic_feature_extractor.py
```diff
@@ -41,7 +41,11 @@ async def init_feature_extractor(
     window, sender = init_moving_window(timedelta(seconds=len(data)))
     async with window:
         await push_logical_meter_data(sender, data)
-        yield PeriodicFeatureExtractor(moving_window=window, period=period)
+        try:
+            yield PeriodicFeatureExtractor(moving_window=window, period=period)
+        except Exception as e:  # pylint: disable=broad-except
+            print(f"<init_feature_extractor> Unhandled exception caught: {e}")
+            raise
```
llucax: Same. In this case since (without this `except`) there would be no *cleanup* code, we don't really need to add the `try/except`.

christianparpart:
Updated.

NB: @llucax I did **not** update the contextmanager decorators that did not have tail code.

EDIT: also, this makes the PR actually small again. All the other locations have no tail code. It's actually only that one.

llucax (approved):

PR #773, opened 2023-11-06
Make `PowerManagingActor` more modular and generic

shsms:
This PR updates the `PowerManagingActor` to

- take type of components we are controlling. Currently only batteries
  are supported.
- Remove all other references to batteries from the
  `PowerManagingActor`, except in the interactions with the
  `PowerDistributingActor`.

The `PowerDistributingActor` needs to be modularized as well, before
the either of these actors can be used with other types of components.

All changes in this PR are internal and don't affect the external
interface of the `BatteryPool`.


src/frequenz/sdk/timeseries/_base_types.py
```diff
@@ -150,3 +152,37 @@ class Bounds(Generic[_T]):
...
+    """
+
+
+class PoolType(enum.Enum):
+    """Enumeration of component pool types."""
+
+    BATTERY_POOL = "BATTERY_POOL"
+    EV_CHARGER_POOL = "EV_CHARGER_POOL"
+    PV_POOL = "PV_POOL"
+    CHP_POOL = "CHP_POOL"
```
llucax, line 188: Is there any way to make this using a class instead where we can use different subclasses to customize the behavior for different types? Usually using enums makes the code harder to read in maintain, as it gets littered with `if`s (or now `match`es probably), or is this another case where the type will need to be sent over the wire?

Sorry, I'm not very familiar with this code to evaluate it myself :-/

But even if the type is sent over the wire, we could do the implementation using these "trait" classes, so we just have to map enum -> class now, but then the code still is clean and without a lot of `if`s for having different logic for different types.

src/frequenz/sdk/timeseries/_base_types.py
```diff
@@ -150,3 +152,37 @@ class Bounds(Generic[_T]):
 
     upper: _T
     """Upper bound."""
+
+
+@dataclass(frozen=True)
```
llucax: I would use `kw_only=True` here, as we have 2 attributes with `Bounds[Power]` type, that could be easily mixed up.

src/frequenz/sdk/timeseries/battery_pool/_result_types.py
```diff
@@ -1,45 +1,90 @@
...
+from .._base_types import Bounds
 from .._quantities import Power
 
 
-@dataclass
-class PowerMetrics:
-    """Power bounds metrics."""
+# This class is used to expose the generic reports from the PowerManager with specific
+# documentation for the battery pool.
+class BatteryPoolReport(Protocol):
```
llucax: I didn't get how this works, why do we need a `Protocol` here? I'm concerned about starting to use protocols everywhere because then we lose the relationship between types, if `BateryPoolReport` inherits from `Report` is very clear what's going on, and we'll also get type/attributes mismatch errors in the class declaration, not when the class is used.

Also this seems a bit backwards, usually you declare a `Protocol` that is a general thing, and then the specific thing follows that protocol, so what seems more natural is to have `Report` as the protocol and `BatteryPoolReport` just a plan class, that follows that protocol.

I guess what you want to achieve here is to offer a "view" on some other type, right? If so it seems like a neat idea but IMHO is hard to reason about, so I would try to find a more common/pythonic way to achieve the same effect. What about:

```py
class Report(Protocol):
    # Interface only

class _Report(Report):
    # Concrete type with the implementation

class BatteryPoolReport(Report):
    # Interface with docs adapted to battery pools
```

Then for all the type hints for the generic object we use `Report` and when only when instantiating we use `_Report`.

llucax (commented):
LGTM in general, but I have a few design concerns.

src/frequenz/sdk/timeseries/_base_types.py
shsms: It is for accessing the instance of the pool, in here: https://github.com/frequenz-floss/frequenz-sdk-python/blob/aebdd00b19f6b8fe38eca3a2604e04923f14cf70/src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py#L133-L137

It would be an if statement at startup, but I think any alternative would be a lot more work,  and I think this is intuitive, and that `_add_bounds_tracker` function would be the only place such an if-statement is necessary, so it is quite isolated as well.

And there won't be edge cases because PowerManager instances are created internally,  so we can control and test how and where it is initialized.

src/frequenz/sdk/timeseries/battery_pool/_result_types.py
shsms: done in: https://github.com/frequenz-floss/frequenz-sdk-python/commit/88caff36

src/frequenz/sdk/timeseries/_base_types.py
shsms: https://github.com/frequenz-floss/frequenz-sdk-python/commit/ebb0e2d2

llucax (dismissed):

tests/timeseries/_battery_pool/test_battery_pool_control_methods.py
```diff
@@ -150,14 +151,9 @@ def _assert_report(
...
-            and report._exclusion_bounds is not None
-        )
-        assert report._inclusion_bounds.lower == Power.from_watts(lower)
-        assert report._inclusion_bounds.upper == Power.from_watts(upper)
-        assert report._exclusion_bounds.lower == Power.from_watts(0.0)
-        assert report._exclusion_bounds.upper == Power.from_watts(0.0)
+        assert report.bounds is not None
+        assert report.bounds.lower == Power.from_watts(lower)
+        assert report.bounds.upper == Power.from_watts(upper)
         # pylint: enable=protected-access
```
daniel-zullo-frequenz: disable/enable protected-access might be no longer needed

shsms: done: https://github.com/frequenz-floss/frequenz-sdk-python/commit/91a3c825

src/frequenz/sdk/timeseries/battery_pool/_result_types.py
```diff
@@ -1,45 +1,90 @@
...
+from .._base_types import Bounds
 from .._quantities import Power
 
 
-@dataclass
-class PowerMetrics:
-    """Power bounds metrics."""
+# This class is used to expose the generic reports from the PowerManager with specific
+# documentation for the battery pool.
+class BatteryPoolReport(Report):
```
daniel-zullo-frequenz, line 16: IIUC this change should be mentioned in the release notes because this is the inner type that `power_status()` will return and the user needs to be aware of but I might be missing something here. Can you please check it out?

daniel-zullo-frequenz (commented):
I only have a question about a possible missing entry into the release notes. LGTM otherwise

daniel-zullo-frequenz (commented):
I only have a question about a possible missing entry into the release notes. LGTM otherwise

src/frequenz/sdk/timeseries/battery_pool/_result_types.py
shsms: true, done here: 20b45702

daniel-zullo-frequenz (approved):

PR #779, opened 2023-11-10
Modularize status tracking for pools of components in the power distributing actor

shsms:
The new design works as follows:

- reusable `ComponentPoolStatusTracker` that takes individual `ComponentStatus` and generates `ComponentPoolStatus` objects.
- the individual `ComponentStatus` need to be generated by code specific to the component category, because status is calculated differently for different component categories.  This code specific for a component category has to be a class that implements the abstract `ComponentStatusTracker` class.
- The `ComponentPoolStatusTracker` accepts implementations of `ComponentStatusTracker`, like the `BatteryStatusTracker` (which looks at the status of corresponding inverters also), and but is otherwise generic.

This PR takes us part of the way towards getting the PowerDistributingActor easily usable for other component types.

It also simplifies some implementations and replaces the use of `MergeNamed` with `Merge` from the channels repo.


shsms:
No externally visible changes, so no release notes.

src/frequenz/sdk/actor/power_distributing/_component_pool_status_tracker.py
```diff
@@ -0,0 +1,163 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
```
daniel-zullo-frequenz, line 2:
```suggestion
# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
```

src/frequenz/sdk/actor/power_distributing/_component_pool_status_tracker.py
```diff
@@ -0,0 +1,163 @@
...
+
+    async def join(self) -> None:
+        """Wait and return when the instance's task completes.
+
+        It will not terminate the instance, which can be done with the `stop` method.
+        """
+        await self._task
+
+    async def stop(self) -> None:
+        """Stop tracking batteries status."""
```
daniel-zullo-frequenz: batteries?

src/frequenz/sdk/actor/power_distributing/_component_pool_status_tracker.py
```diff
@@ -0,0 +1,163 @@
...
+            elif status.value == ComponentStatusEnum.NOT_WORKING:
+                self._current_status.working.discard(component_id)
+                self._current_status.uncertain.discard(component_id)
+
+            await self._component_status_sender.send(self._current_status)
+
+    async def update_status(
+        self, succeed_components: set[int], failed_components: set[int]
+    ) -> None:
+        """Notify which components succeed and failed in the request.
```
daniel-zullo-frequenz: I think this is mixing present and past for succeed and failed, respectively. If this comment applies, then argument names need to be updated too

src/frequenz/sdk/actor/power_distributing/_component_status.py
```diff
@@ -49,3 +49,14 @@ class ComponentStatusEnum(enum.Enum):
 
     WORKING = 2
     """Component is working"""
+
+
+@dataclass(frozen=True)
```
daniel-zullo-frequenz, line 57: what about making it `kw_only=True` too?

src/frequenz/sdk/actor/power_distributing/_battery_status_tracker.py
```diff
@@ -179,7 +179,7 @@ def __init__(  # pylint: disable=too-many-arguments
         """Create class instance.
 
         Args:
-            battery_id: Id of this battery
+            component_id: Id of this battery
```
daniel-zullo-frequenz, line 177: The battery should remain in the documentation, right?

src/frequenz/sdk/actor/power_distributing/_component_status.py
```diff
@@ -72,3 +75,35 @@ class SetPowerResult:
...
+        component_id: int,
+        max_data_age_sec: float,
+        max_blocking_duration_sec: float,
+        status_sender: Sender[ComponentStatus],
+        set_power_result_receiver: Receiver[SetPowerResult],
+    ) -> None:
+        """Create class instance.
+
+        Args:
+            component_id: Id of this battery
```
daniel-zullo-frequenz: I think here it should be `component` instead of `battery`

src/frequenz/sdk/actor/power_distributing/_component_status.py
```diff
@@ -0,0 +1,109 @@
...
+            max_blocking_duration_sec: This value tell what should be the maximum
+                timeout used for blocking failing component.
+            status_sender: Channel to send status updates.
+            set_power_result_receiver: Channel to receive results of the requests to the
+                components.
+        """
+
+    @abstractmethod
+    async def stop(self) -> None:
+        """Stop tracking battery status."""
```
daniel-zullo-frequenz: I think the component status interface shouldn't talk about batteries

src/frequenz/sdk/actor/power_distributing/_component_status.py
```diff
@@ -0,0 +1,109 @@
...
+import enum
+import typing
+from abc import ABC, abstractmethod
+from collections import abc
+from dataclasses import dataclass
+
+from frequenz.channels import Receiver, Sender
+
+
+@dataclass
```
daniel-zullo-frequenz, line 15: can this dataclass be `frozen=True, kw_only=True` or there is a reason to not do it?

daniel-zullo-frequenz (commented):
I've done my first review round and made a few suggestions to check for. I'll have another look after those suggestions are discussed/addressed.

src/frequenz/sdk/actor/power_distributing/_battery_status_tracker.py
```diff
@@ -170,7 +170,7 @@ class BatteryStatusTracker:
 
     def __init__(  # pylint: disable=too-many-arguments
         self,
-        battery_id: int,
+        component_id: int,
```
llucax, line 168: Not a big deal but any reason to mix the `battery_id` → `component_id` rename with the `_battery_status.py` → `_battery_status_tracker.py`? They seem pretty unrelated.

src/frequenz/sdk/actor/power_distributing/_component_pool_status_tracker.py
```diff
@@ -0,0 +1,163 @@
...
+    ComponentStatus,
+    ComponentStatusEnum,
+    ComponentStatusTracker,
+    SetPowerResult,
+)
+
+_logger = logging.getLogger(__name__)
+
+
+class ComponentPoolStatusTracker:
```
llucax, line 26: I guess not for this PR, but can't we take the opportunity and include the migration to `BackgroudService` in this big refactoring. These classes handling multiple tasks and having a `stop()` and `join()` look like a perfect candidate for `BackgorundService`.

src/frequenz/sdk/actor/power_distributing/_component_status.py
```diff
@@ -0,0 +1,109 @@
...
+class ComponentPoolStatus:
+    """Status of all components of a certain category in the microgrid."""
+
+    working: set[int]
+    """Set of working component ids."""
+
+    uncertain: set[int]
+    """Set of components to be used only when there are none known to be working."""
```
llucax, line 23: Also probably `collections.abc.Set` so the user can't mutate it either.

src/frequenz/sdk/actor/power_distributing/_component_status.py
```diff
@@ -0,0 +1,109 @@
...
+@dataclass(frozen=True)
+class ComponentStatus:
+    """Status of a single component."""
+
+    component_id: int
+    """Component ID."""
+
+    value: ComponentStatusEnum
+    """Component status."""
+
```
llucax, line 66: I'm not convinced about this change. Can't you just use `select()` if you need to know the source?  I guess you'll still need a mapping between receiver and `component_id`, but otherwise this is mostly doing the same, but you need to create one extra class that has a name clash with the `Enum`, so now you need to add the `Enum` suffix to the enum, which is weird and we never do (it could be like adding a `Class` suffix to classes.

If you think it is better to stick with `merge` and this mapping wrapper, then I suggest to find some other name to avoid the `Enum` suffix, maybe `ComponentStatus` and `ComponentState`, or `CurrentComponentState` to avoid confusion between status and state?

src/frequenz/sdk/actor/power_distributing/_component_status.py
```diff
@@ -0,0 +1,109 @@
...
+@dataclass(frozen=True, kw_only=True)
+class SetPowerResult:
+    """Lists of components for which the last set power command succeeded or failed."""
+
+    succeeded: typing.Iterable[int]
+    """Component IDs for which the last set power command succeeded."""
+
+    failed: typing.Iterable[int]
+    """Component IDs for which the last set power command failed."""
```
llucax: Since you are using `set`s internally here, wouldn't it also make more sense to use `collections.abc.Set`? It seems like it could be useful for users to be able to do `if id in result.succeeded` for example.

src/frequenz/sdk/actor/power_distributing/_component_pool_status_tracker.py
shsms: This is actually a renaming, not a new file: https://github.com/frequenz-floss/frequenz-sdk-python/pull/779/commits/278b0b92475dc3d42566560a5d4493095f7bdd26#diff-40a429480bdbbab4e414638f059fa7202095acd519c72353f9ad2a762e75cad7

src/frequenz/sdk/actor/power_distributing/_component_status.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class ComponentStatusTracker(ABC):
```
llucax, line 79: This one too could be a `BackgroundService`.

Now I wonder if we should maybe have a purely abstract `BackgroundService` class (or even protocol), and then a `BaseBackgroundService` to use as a convenience default implementation, as in here you are only defining an interface. I think `BackgroundService` should also be only an interface, as we have background services that are managing other background services instead of tasks directly, and I ended up mixing up things a bit for the sake of convenience (interface/protocol vs base mixin implementation).

But in any case for now `BackgroundService` could be used too, it will just add an unnecessary `_tasks` member to the class.

src/frequenz/sdk/actor/power_distributing/_component_pool_status_tracker.py
```diff
@@ -0,0 +1,163 @@
...
+        """Stop tracking batteries status."""
+        await cancel_and_await(self._task)
+        await asyncio.gather(
+            *[tracker.stop() for tracker in self._component_status_trackers],
+        )
+        await self._merged_status_receiver.stop()
+
+    def _make_merged_status_receiver(
+        self,
+    ) -> Merge[ComponentStatus]:
```
llucax, line 93: It might be better to return a `Receiver` here as `Merge` is becoming private.

src/frequenz/sdk/actor/power_distributing/_component_pool_status_tracker.py
shsms: I'll fix this particular case, but I'm sure there are other instances as well.  This is old code, and this PR has a limited scope.

src/frequenz/sdk/actor/power_distributing/_component_status.py
shsms: the two fields have different types.

src/frequenz/sdk/actor/power_distributing/_battery_status_tracker.py
shsms: yes, I guess it doesn't matter.  had to change the field to make it fit into a generic interface.

src/frequenz/sdk/actor/power_distributing/_component_status.py
shsms: true, will fix.

src/frequenz/sdk/actor/power_distributing/_component_status.py
shsms: fixing

src/frequenz/sdk/actor/power_distributing/_component_status.py
shsms: probably, but out of scope for this PR.

src/frequenz/sdk/actor/power_distributing/_battery_status_tracker.py
shsms: oops, guess I mixed up some commits.

src/frequenz/sdk/actor/power_distributing/_component_pool_status_tracker.py
shsms: maybe, but right now I'm just doing the bare minimum so I can unblock *pools.

src/frequenz/sdk/actor/power_distributing/_component_status.py
shsms: there's a lot of non-bug style issues with the code base.  This PR is just about renaming some things and modularizing some things.  I did improve some things, but don't want to get too deep into fixing other style issues at the moment in internal code.

src/frequenz/sdk/actor/power_distributing/_component_status.py
shsms: We need the component id from a variable number of components, so implementation-wise I think this is the best approach.

I think they are both status or both state, but one of them is tied to a specific component while the other is not.  I agree that `Enum` is not perfect, but I couldn't come up with a better one.

I would like the class with the id to be called `ComponentStatus`, to form a pattern with the `ComponentPoolStatus`, `ComponentStatusTracker`, `ComponentPoolStatusTracker`.  I guess we just need an alternative for the enum.

src/frequenz/sdk/actor/power_distributing/_component_status.py
shsms: it wasn't me,  it was already like that.

src/frequenz/sdk/actor/power_distributing/_component_pool_status_tracker.py
shsms: So we stop calling `stop` on it in tests, etc, right?

src/frequenz/sdk/actor/power_distributing/_battery_status_tracker.py
shsms: but I seem to have mentioned this in the commit message, so looks like it wasn't a mixup.

src/frequenz/sdk/actor/power_distributing/_component_pool_status_tracker.py
daniel-zullo-frequenz: Thanks!

src/frequenz/sdk/actor/power_distributing/_component_status.py
shsms: > We need the component id from a variable number of components, so implementation-wise I think this is the best approach.

just to contextualize this, this is the case if we want to limit the changes to this part of the code base.  it would be possible to avoid the merge if we rewrite the `ComponentPoolStatusTracker`, but that won't be a trivial change.

src/frequenz/sdk/actor/power_distributing/_component_pool_status_tracker.py
shsms: fixed in: https://github.com/frequenz-floss/frequenz-sdk-python/commit/61765a81

src/frequenz/sdk/actor/power_distributing/_component_pool_status_tracker.py
shsms: done here: https://github.com/frequenz-floss/frequenz-sdk-python/commit/61765a81

src/frequenz/sdk/actor/power_distributing/_component_status.py
shsms: https://github.com/frequenz-floss/frequenz-sdk-python/commit/61765a81

src/frequenz/sdk/actor/power_distributing/_component_status.py
shsms: https://github.com/frequenz-floss/frequenz-sdk-python/commit/61765a81

src/frequenz/sdk/actor/power_distributing/_component_status.py
llucax: OK, but you are changing members names already, so it might be good to add to the bag of refactoring (it is fine if it's not in this PR but it looks like a trivial change).

src/frequenz/sdk/actor/power_distributing/_component_pool_status_tracker.py
llucax: I agree for the stuff that was already there and was just renamed/moved, but for new classes I think we should start using `BackgroundService`, it makes no sense to keep piling up technical debt.

src/frequenz/sdk/actor/power_distributing/_component_status.py
llucax: > I would like the class with the id to be called `ComponentStatus`, to form a pattern with the `ComponentPoolStatus`, `ComponentStatusTracker`, `ComponentPoolStatusTracker`. I guess we just need an alternative for the enum.

I see. I think, besides being a bit ambiguous, `ComponentState` for the enum makes sense. Status and state is not exactly the same. For example a state machine is never called status machine, and "WORKING", etc. look like states in a state machine, is like an abstract concept, while status conveys more the meaning of something that's happening at a particular point in time. For example a status report wouldn't be called state report, and it talks about one point in time. `ComponentStatus` says this particular component is like this right now.

Anyway, I agree it has the downside of being a bit ambiguous. Not a perfect solution so I would leave it to you to decide.

src/frequenz/sdk/actor/power_distributing/_component_status.py
llucax: Like this class is new, right?

src/frequenz/sdk/actor/power_distributing/_component_pool_status_tracker.py
llucax: Mmm, damn. Houston, we have a problem. Another reason to use `select()` instead. :-P

But seriously, it looks like we can't make `_Merge` private in channels then.

llucax:
I just went through your replies and I'm fine with leaving things as they are. The only thing I really wouldn't like is to keep adding classes that should inherit from `BackgroundService` without doing so. If all the classes here are really just renames/moves, then I'm also fine with keeping it like it is in this PR and will approve.

src/frequenz/sdk/actor/power_distributing/_component_status.py
llucax: I was thinking that another option, and in general for all enums, it might make sense to name them in plural, as it is a collection of names or stuff, specially in Python that they are even iterable.

src/frequenz/sdk/actor/power_distributing/_component_status.py
shsms: I already did a few trivial changes, there's just so many, but ok, I will do this one too.

src/frequenz/sdk/actor/power_distributing/_component_status.py
shsms: nope, it is just an abc for an existing `BatteryStatusTracker` class.

But I can look into this again as I start working on the EVChargerStatusTracker, which will be new.  And migrate this too at that point.

src/frequenz/sdk/actor/power_distributing/_component_status.py
shsms: I tried to do this also, but it doesn't work because we don't have some methods like `intersection`, etc. on the abstract types.

src/frequenz/sdk/actor/power_distributing/_component_status.py
shsms: Done here: https://github.com/frequenz-floss/frequenz-sdk-python/commit/00c48d9e

src/frequenz/sdk/actor/power_distributing/_component_pool_status_tracker.py
```diff
@@ -0,0 +1,163 @@
...
+
+    async def _run(self) -> None:
+        """Start tracking component status."""
+        while True:
+            try:
+                await self._update_status()
+            except Exception as err:  # pylint: disable=broad-except
+                _logger.error(
+                    "ComponentPoolStatus failed with error: %s. Restarting.", err
+                )
```
daniel-zullo-frequenz, line 119: We might want to a short delay between restarts. This might be also out of scope for this PR but worth mentioning it

daniel-zullo-frequenz (commented):
I don't have any further comments within the scope of this PR, LGTM
Though Luca's comments need to be resolve so I'll leave it to him to approve it

src/frequenz/sdk/actor/power_distributing/_component_status.py
llucax: Really? That's lame :scream:

tests/actor/test_battery_status.py
```diff
@@ -538,7 +538,7 @@ async def test_sync_blocking_interrupted_with_invalid_message(
         assert tracker._get_new_status_if_changed() is None
 
         tracker._handle_status_set_power_result(
-            SetPowerResult(succeeded={BATTERY_ID}, failed={})
+            SetPowerResult(succeeded={BATTERY_ID}, failed=set())
```
llucax, line 541: Ha! This change uncovered a bug-ish (I mean, it was not a bug because it was returning an iterable, but still).

llucax (approved):

src/frequenz/sdk/actor/power_distributing/_component_pool_status_tracker.py
shsms: Wow, I hadn't seen this.  I think this would never happen, because we don't close any of our channels and that's the only source of an exception here.

I think I'll just remove this exception handling, causing the actor to restart fully instead.

shsms: Will do in the next PR.

tests/actor/test_battery_status.py
shsms: yes, the great traps in python. :-/

src/frequenz/sdk/actor/power_distributing/_component_status.py
shsms: it turns out this is not a read-only class.  when we get news about a single battery, we update the corresponding list.

PR #786, opened 2023-11-20
Generalize PowerDistributingActor

shsms:
All the battery-specific logic has been moved into a `BatteryManager` (which implements `ComponentManager`).  This includes
- component status tracking
- bounds tracking
- distribution algorithm
- control methods

This allows us to add `ComponentManager` implementations for additional component categories, that can be used with the PowerDistributingActor.


shsms:
There might be some reusable code in the `BatteryManager`, but that can be pulled out as we start implementing support for other categories.  This is already enough to get started with those.

src/frequenz/sdk/actor/power_distributing/request.py
```diff
@@ -17,8 +17,8 @@ class Request:
     power: Power
     """The requested power."""
 
-    batteries: abc.Set[int]
-    """The component ids of the batteries to be used for this request."""
+    component_ids: abc.Set[int]
+    """The component ids of the components to be used for this request."""
```
daniel-zullo-frequenz, line 21: You might want to mention this change in the release notes as `Request` is exposed

src/frequenz/sdk/actor/power_distributing/_component_managers/__init__.py
```diff
@@ -0,0 +1,12 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Interfaces for the power distibuting actor with different component types."""
```
daniel-zullo-frequenz:
```suggestion
"""Interfaces for the power distributing actor with different component types."""
```

daniel-zullo-frequenz (commented):
I only have a couple of minor suggestions to check for.

src/frequenz/sdk/actor/power_distributing/request.py
llucax: Yeah, same with the change in arguments for `PowerDistributingActor`: `component_pool_status_sender`. For now I guess it is OK because we only have internal users and we know they are not using it, but as this gets more used externally we have to be more careful and list all changes, as people could be instantiating their own `PowerDistributingActor`.

src/frequenz/sdk/actor/power_distributing/result.py
```diff
@@ -21,13 +22,13 @@ class _BaseResultMixin:
...
-    """Result returned when setting the power succeed for all batteries."""
+    """Result returned when setting the power succeed for all components."""
 
     succeeded_power: Power
     """The part of the requested power that was successfully set."""
 
-    succeeded_batteries: set[int]
-    """The subset of batteries for which power was set successfully."""
+    succeeded_components: abc.Set[int]
+    """The subset of components for which power was set successfully."""
```
llucax, line 31: These changes should also go to the upgrading notes, as these types are now read-only.

llucax (commented):
Also only comments about release notes.

src/frequenz/sdk/actor/power_distributing/request.py
shsms: I guess you consider `Request` and `PowerDistributingActor` as public types because they're in `actor/power_distributing`?

It is not supposed to be.  So I would instead rename it to `actor/_power_distributing` and mention that alone in the release notes.  That's how the `actor/_power_managing` actor is,  and this one is public because it is from a time when users were instantiating and using the actors directly.

daniel-zullo-frequenz: I consider `Request` as public because it's exposed in any kind of power distributing `Result`
https://github.com/frequenz-floss/frequenz-sdk-python/blob/v1.x.x/src/frequenz/sdk/actor/power_distributing/result.py#L14-L19

shsms: oh right, I guess we need a revamping of the result streaming mechanism, with higher level types.  The current approach just looks like a lazy solution - of exposing power distributor's result type through the battery pool's `power_status` method.

But I guess it can change later, when we improve error handling, once the microgrid API spec defines error codes.

shsms: just amended the release notes commit: https://github.com/frequenz-floss/frequenz-sdk-python/commit/b9c8573a

src/frequenz/sdk/actor/power_distributing/result.py
shsms: done here: https://github.com/frequenz-floss/frequenz-sdk-python/commit/b9c8573a

daniel-zullo-frequenz (approved):

PR #788, opened 2023-11-22
Update `frequency()` to no longer have parameters

Marenz:
Changes frequency() to always auto-select the frequency source component

### DIFF
--- a/src/frequenz/sdk/microgrid/_data_pipeline.py
+++ b/src/frequenz/sdk/microgrid/_data_pipeline.py
@@ -19,7 +19,6 @@ from dataclasses import dataclass
 from frequenz.channels import Broadcast, Sender
 
 from ..actor._actor import Actor
-from ..microgrid.component import Component
 from ..timeseries._base_types import PoolType
 from ..timeseries._grid_frequency import GridFrequency
 from ..timeseries.grid import Grid
@@ -119,29 +118,21 @@ class _DataPipeline:  # pylint: disable=too-many-instance-attributes
         self._grid: Grid | None = None
         self._ev_charger_pools: dict[frozenset[int], EVChargerPool] = {}
         self._battery_pools: dict[frozenset[int], BatteryPoolReferenceStore] = {}
-        self._frequency_pool: dict[int, GridFrequency] = {}
+        self._frequency_instance: GridFrequency | None = None
 
-    def frequency(self, component: Component | None = None) -> GridFrequency:
+    def frequency(self) -> GridFrequency:
         """Fetch the grid frequency for the microgrid.
 
-        Args:
-            component: The component to use when fetching the grid frequency.  If None,
-                the component will be fetched from the registry.
-
         Returns:
-            A GridFrequency instance.
+            The GridFrequency instance.
         """
-        if component is None:
-            component = GridFrequency.find_frequency_component()
+        if self._frequency_instance is None:
+            self._frequency_instance = GridFrequency(
+                self._data_sourcing_request_sender(),
+                self._channel_registry,
+            )
 
-        if component.component_id in self._frequency_pool:
-            return self._frequency_pool[component.component_id]
-
-        grid_frequency = GridFrequency(
-            self._data_sourcing_request_sender(), self._channel_registry, component
-        )
-        self._frequency_pool[component.component_id] = grid_frequency
-        return grid_frequency
+        return self._frequency_instance
 
     def logical_meter(self) -> LogicalMeter:
         """Return the logical meter instance.
@@ -409,17 +400,13 @@ async def initialize(resampler_config: ResamplerConfig) -> None:
     _DATA_PIPELINE = _DataPipeline(resampler_config)
 
 
-def frequency(component: Component | None = None) -> GridFrequency:
+def frequency() -> GridFrequency:
     """Return the grid frequency.
-
-    Args:
-        component: Optional component to get the frequency for. If not specified,
-            the frequency of the grid is returned.
 
     Returns:
         The grid frequency.
     """
-    return _get().frequency(component)
+    return _get().frequency()
 
 
 def logical_meter() -> LogicalMeter:
--- a/src/frequenz/sdk/timeseries/_grid_frequency.py
+++ b/src/frequenz/sdk/timeseries/_grid_frequency.py
@@ -97,9 +97,8 @@ class GridFrequency:
 
     async def _send_request(self) -> None:
         """Send the request for grid frequency."""
-        _logger.info("Sending request for grid frequency: %s", self._source_component)
         await self._request_sender.send(self._component_metric_request)
-        _logger.info("Sent request for grid frequency: %s", self._source_component)
+        _logger.debug("Sent request for grid frequency: %s", self._source_component)
 
     @staticmethod
     def find_frequency_source() -> Component:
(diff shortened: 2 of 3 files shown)


src/frequenz/sdk/microgrid/_data_pipeline.py
```diff
@@ -119,26 +119,52 @@ def __init__(
...
+        ...
+
+    @typing.overload
+    def frequency(self, component_or_id: int) -> GridFrequency:
+        ...
+
+    @typing.overload
+    def frequency(self, component_or_id: Component) -> GridFrequency:
+        ...
+
```
llucax:
```suggestion
```

You don't need to provide overloads if the rest of the method signature doesn't change. It's just redundant.

src/frequenz/sdk/microgrid/_data_pipeline.py
```diff
@@ -119,26 +119,52 @@ def __init__(
...
+    def frequency(
+        self, component_or_id: Component | int | None = None
```
llucax: I would also keep the variable name as `component`, as what you want to pass is a component, which could be identified by an object or an ID, to make it clear I think eventually we should have a `ComponentId` type that is not compatible with `int`, as we don't want people also to be able to add IDs, etc. That would make this method signature (and others) much clearer.

Also changing the argument name is technically a breaking change, as people could be using `frequency(component=...)`. We should use the syntax `def frequency(self, /, component_or_id: ...)` to force users not to use the keyword syntax for stuff where it doesn't make a lot of sense.

Finally, if we were going to rename this argument, I would call it `source`, as `component` is describing (or was) the type, and not what the argument represents.

Aaaaaaanyway, all that said, for this PR I would be happy with any combination of these options:
* Leave `component` as it was before
* Rename it to `source` and release notes about the breaking change (new name)
* Adding `/` and release notes about the breaking change (drop support to use keyword syntax)

src/frequenz/sdk/timeseries/_grid_frequency.py
```diff
@@ -89,15 +87,15 @@ def new_receiver(self) -> Receiver[Sample[Frequency]]:
         if not self._task:
             self._task = asyncio.create_task(self._send_request())
         else:
-            _logger.info("Grid frequency request already sent: %s", self._component)
+            _logger.info("Grid frequency request already sent: %s", self._component_id)
```
llucax: Do we lose any information by logging only the ID instead of the whole object?

llucax (commented):
I don't really like the second part of this PR, losing the higher-level abstracted component in `GridFrequency`. We are *downgrading* in terms of *abstraction*.

Being more flexible on what we accept is great, but I would still provide the abstracted object.

I would:
* Move the logic to accept a component ID to `GridFrequency`.
* We can keep only the ID internally, but I would still offer the `component` (maybe `source` if we go forward with that rename) property, and I would fetch the whole component lazily if someone ever request it (and cache it for future accesses.

src/frequenz/sdk/timeseries/_grid_frequency.py
Marenz: Assuming it prints the type/category and given that not only a meter can send us frequencies, yes we do lose some information

src/frequenz/sdk/microgrid/_data_pipeline.py
llucax: I created a discussion around `ComponentId`.

src/frequenz/sdk/timeseries/_grid_frequency.py
```diff
@@ -89,34 +87,53 @@ def new_receiver(self) -> Receiver[Sample[Frequency]]:
...
+
+        component_ids: set[int] | None = None
+
+        if isinstance(source, int):
+            component_ids = {source}
+        elif source is not None:
+            raise TypeError(
+                f"Invalid type for source: {type(source)}, expected None, Component or int."
+            )
+
```
llucax:
```py
from typing import assert_never
match source:
    case Component() as component:
        ...
    case int() as component_id:
        ...
    case None:
        ...
    case _ as invalid:
        assert_never(invalid)
```

src/frequenz/sdk/microgrid/_data_pipeline.py
```diff
@@ -119,29 +118,22 @@ def __init__(
...
-            component: The component to use when fetching the grid frequency.  If None,
-                the component will be fetched from the registry.
-
         Returns:
-            A GridFrequency instance.
+            The GridFrequency instance.
         """
-        if component is None:
-            component = GridFrequency.find_frequency_component()
+        if not hasattr(self, "_frequency_instance"):
```
llucax: Why the use of this dynamic idiom? Why not make `_frequency_instance: GridFrequency | None`?

src/frequenz/sdk/microgrid/_data_pipeline.py
```diff
@@ -119,29 +118,22 @@ def __init__(
...
+            self._frequency_instance = GridFrequency(
+                self._data_sourcing_request_sender(),
+                self._channel_registry,
+                GridFrequency.find_frequency_source(),
```
llucax: Also, since now we don't need the `component_id` to lookup the cache to see if the object was already created, any reason not to at least make the `source` optional in `GridFrequency` and find the best component automatically if the `source` is omitted?

src/frequenz/sdk/timeseries/_grid_frequency.py
```diff
@@ -49,32 +49,30 @@ def __init__(
         self,
         data_sourcing_request_sender: Sender[ComponentMetricRequest],
         channel_registry: ChannelRegistry,
-        component: Component,
+        source: Component,
```
llucax: So basically `source: Component | None` and do the find if it is `None`, or even just removing the `source` too, not sure if it is worth keeping. Any opinions @shsms @matthias-wende-frequenz ?

llucax (commented):
Sorry for keep nagging in such a small change, but I think this is really good cleanup, is good that we think a bit more about interfaces as we approach 1.0.

src/frequenz/sdk/microgrid/_data_pipeline.py
shsms: I think we might even want to have a pylint check, maybe for the next intern.

src/frequenz/sdk/timeseries/_grid_frequency.py
shsms: I like the idea of removing the param and identifying the right component within the class.  The only way it will get configurable is if we decide to do specify a component ID through the UI, which could happen at runtime.  So it will have to be implemented differently anyway.

src/frequenz/sdk/microgrid/_data_pipeline.py
llucax: You mean completely forbid `{has,get,set}attr` without a `# pylint: disable=/noqa`? :thinking:

shsms: yup,  because those require putting code in strings, which bypasses our checks.  So they should be manually excluded and documented where ever they are used.

llucax: Yeah, I guess it makes sense... :+1:

Marenz: I suppose I wanted to avoid extending the type to `None` thus requiring extra checks for whether it's set or not..
In this context it doesn't really make a difference I suppose.

llucax (approved):

PR #795, opened 2023-11-27
Add new type `ComponentId`

Marenz:
This is a WIP and the current work is 90% done by ChatGPT.
I am currently busy with a few other things, so anyone can take over this PR and finish it.
Otherwise I'll get back to it when my important stuff is done.

### DIFF
--- a/src/frequenz/sdk/microgrid/component/_component.py
+++ b/src/frequenz/sdk/microgrid/component/_component.py
@@ -6,7 +6,7 @@ from __future__ import annotations
 
 from dataclasses import dataclass
 from enum import Enum
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, NewType
 
 import frequenz.api.common.components_pb2 as components_pb
 import frequenz.api.microgrid.grid_pb2 as grid_pb
@@ -15,6 +15,9 @@ import frequenz.api.microgrid.inverter_pb2 as inverter_pb
 if TYPE_CHECKING:
     # Break circular import
     from ...timeseries import Fuse
+
+
+ComponentId = NewType("ComponentId", int)
 
 
 class ComponentType(Enum):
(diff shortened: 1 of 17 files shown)


src/frequenz/sdk/microgrid/component/_component.py
```diff
@@ -17,6 +17,9 @@
     from ...timeseries import Fuse
 
 
+ComponentId = NewType("ComponentId", int)
```
llucax, line 20: Why `NewType` and not `dataclass`?

`NewType` allows this for example: `ComponentId(1) + 2` and `microgrid.battery_pool(priority=ComponentId(1))` which is just weird.

Also the `str` and `repr` looks just like a bare `int`, which might be good or not (I personally would like to at least have `repr` use `ComponentId(1)` instead of just `1`.

llucax (commented):
I would really avoid using `NewType` in this case, also I'm not sure if we should update all `component_id` to `component` as how the component is specified now is encoded in the type system, and could open the door to pass a component as an object for example (or even think more on a case-by-case basis if another name is more suitable). But I guess that part is out of scope for this PR.

src/frequenz/sdk/microgrid/component/_component.py
Marenz: Honestly, no specific reason, I selected `NewType` after a quick research, I am not set on using it :)

llucax: In that case I would use a `dataclass`, I see no benefits in using `NewType` and a few drawbacks.

Marenz: just a data class with one member you mean?

llucax: Yup.

llucax:
I'll try this in the new microgrid API client.

PR #800, opened 2023-11-29
Un-Peekable-ification of the PowerDistributingActor

shsms:
This PR replaces `Peekable`s with `LatestValueCache`.

And updates tests to not mock time, because that somehow makes a difference when using actual receivers.

### DIFF
--- a/src/frequenz/sdk/_internal/_channels.py
+++ b/src/frequenz/sdk/_internal/_channels.py
@@ -4,6 +4,7 @@
 """General purpose classes for use with channels."""
 
 import abc
+import asyncio
 import typing
 
 from frequenz.channels import Receiver
@@ -24,3 +25,26 @@ class ReceiverFetcher(typing.Generic[T], typing.Protocol):
         Returns:
             A receiver instance.
         """
+
+
+class LatestValueCache(typing.Generic[T]):
+    """A cache that stores the latest value in a receiver."""
+
+    def __init__(self, receiver: Receiver[T]) -> None:
+        """Create a new cache.
+
+        Args:
+            receiver: The receiver to cache.
+        """
+        self._receiver = receiver
+        self._latest_value: T | None = None
+        self._task = asyncio.create_task(self._run())
+
+    @property
+    def latest_value(self) -> T | None:
+        """Get the latest value in the cache."""
+        return self._latest_value
+
+    async def _run(self) -> None:
+        async for value in self._receiver:
+            self._latest_value = value
--- a/tests/microgrid/test_client.py
+++ b/tests/microgrid/test_client.py
@@ -407,10 +407,10 @@ class TestMicrogridGrpcClient:
             with pytest.raises(ValueError):
                 # should raise a ValueError for wrong component category
                 await microgrid.meter_data(38)
-            peekable = (await microgrid.meter_data(83)).into_peekable()
+            receiver = await microgrid.meter_data(83)
             await asyncio.sleep(0.2)
 
-        latest = peekable.peek()
+        latest = await anext(receiver)
         assert isinstance(latest, MeterData)
         assert latest.component_id == 83
 
@@ -431,10 +431,10 @@ class TestMicrogridGrpcClient:
             with pytest.raises(ValueError):
                 # should raise a ValueError for wrong component category
                 await microgrid.meter_data(38)
-            peekable = (await microgrid.battery_data(83)).into_peekable()
+            receiver = await microgrid.battery_data(83)
             await asyncio.sleep(0.2)
 
-        latest = peekable.peek()
+        latest = await anext(receiver)
         assert isinstance(latest, BatteryData)
         assert latest.component_id == 83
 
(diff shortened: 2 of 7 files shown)


shsms:
No external changes, so no release notes.

Marenz (approved):
Not thrilled about the 0.1s new wait time, I mean, that'll accumulate, no? 10 actors -> 1 second wait time already....

shsms:
> Not thrilled about the 0.1s new wait time, I mean, that'll accumulate, no? 10 actors -> 1 second wait time already....

yes, my hope is that we can remove a lot of the logic tests from the actor tests, because they're already covered in the distribution algorithm.

There should be very few cases that really need to be tested with the delay.

src/frequenz/sdk/_internal/_channels.py
```diff
@@ -24,3 +25,26 @@ def new_receiver(self, maxsize: int = 50) -> Receiver[T]:
...
+    """A cache that stores the latest value in a receiver."""
+
+    def __init__(self, receiver: Receiver[T]) -> None:
+        """Create a new cache.
+
+        Args:
+            receiver: The receiver to cache.
+        """
+        self._receiver = receiver
+        self._latest_value: T | None = None
```
llucax, line 40: Shouldn't this use a sentinel? `None` could be a valid `T` (if `T` is, for example, `float | None`, like in `Sample.value`).

src/frequenz/sdk/_internal/_channels.py
```diff
@@ -24,3 +25,26 @@ def new_receiver(self, maxsize: int = 50) -> Receiver[T]:
...
+        self._task = asyncio.create_task(self._run())
+
+    @property
+    def latest_value(self) -> T | None:
+        """Get the latest value in the cache."""
+        return self._latest_value
```
llucax, line 46: With that in mind, this should either return the sentinel, or maybe raise an exception?

Alternative it could be parametrized also with a second type, and receive a default at creation time, like:

```py
class LatestValueCache(typing.Generic[T, U]):
    """A cache that stores the latest value in a receiver."""

    def __init__(self, receiver: Receiver[T], *, default: U) -> None:
        ...

last = LatestValueCache[int, None](receiver, default=None)
```

Although it is a bit convoluted for the average case.

src/frequenz/sdk/_internal/_channels.py
```diff
@@ -24,3 +25,26 @@ def new_receiver(self, maxsize: int = 50) -> Receiver[T]:
         Returns:
             A receiver instance.
         """
+
+
+class LatestValueCache(typing.Generic[T]):
```
llucax, line 30: `BackgroundService` :angry: 

It is really as easy as:

```py
class LatestValueCache(typing.Generic[T], BackgroundService):
    """A cache that stores the latest value in a receiver."""

    def __init__(self, receiver: Receiver[T]) -> None:
        """Create a new cache.
        Args:
            receiver: The receiver to cache.
        """
        self._receiver = receiver
        self._latest_value: T | None = None
        self.start()  # if you really want to start it automatically

    def start(self) -> None:
        self._tasks.add(asyncio.create_task(self._run())

    # same as before
 ```

And then you get:
* A common interface to start/stop stuff that spawn tasks
* All the async context manager goodies, stopping, etc.

File: tests/microgrid/test_client.py [hunk 4]
```diff
@@ -407,10 +407,10 @@ async def test_meter_data(self) -> None:
             with pytest.raises(ValueError):
                 # should raise a ValueError for wrong component category
                 await microgrid.meter_data(38)
-            peekable = (await microgrid.meter_data(83)).into_peekable()
+            receiver = await microgrid.meter_data(83)
             await asyncio.sleep(0.2)
 
-        latest = peekable.peek()
+        latest = await anext(receiver)
```
llucax, line 413: Any reason not to use `receiver.receive()`?

src/frequenz/sdk/_internal/_channels.py
shsms: oh true 😭 

will fix on Monday.

shsms: https://github.com/frequenz-floss/frequenz-sdk-python/pull/846

src/frequenz/sdk/_internal/_channels.py
shsms: I've got it to raise, but also added a `has_value` method that people can check first if they want to avoid having to deal with an exception: https://github.com/frequenz-floss/frequenz-sdk-python/pull/846

src/frequenz/sdk/_internal/_channels.py
shsms: Leads to circular dependencies unfortunately.  Need to get `BackgroundService` out of the `actor` package first, I guess.

tests/microgrid/test_client.py
shsms: no, but they are interchangeable.

PR #801, opened 2023-11-30
Pass through 0W requests to the microgrid API independent of exclusion bounds

### LINKED ISSUE #794: PowerDistributor is going away from zero when zero power is requested, when there are exclusion bounds
### What happened?

When 0 power is requested when there are exclusion bounds, the power distributor is returning non-zero power values that are outside the exclusion bounds.  This can be reproduced by adding the below snippet under `test_scenario_1`

```diff
diff --git a/tests/actor/power_distributing/test_battery_distribution_algorithm.py b/tests/actor/power_distributing/test_battery_distribution_algorithm.py
index 3501cc2c..3841deef 100644
--- a/tests/actor/power_distributing/test_battery_distribution_algorithm.py
+++ b/tests/actor/power_distributing/test_battery_distribution_algorithm.py
@@ -1039,6 +1039,10 @@ class TestDistWithExclBounds:
 
         algorithm = BatteryDistributionAlgorithm()
 
+        self.assert_result(
+            algorithm.distribute_power(0, components),
+            DistributionResult({1: 0, 3: 0, 5: 0}, remaining_power=0.0),
+        )
         self.assert_result(
             algorithm.distribute_power(-300, components),
             DistributionResult({1: -100, 3: -100, 5: -100}, remaining_power=0.0),
```

It should produce 0 values, because the api service accepts 0 values, even when there are exclusion bounds.

### Affected version(s)

_No response_

### Affected part(s)

I don't know (part:❓)

### Extra information

We renamed the `power distribution` label to `power management`, but the `Affected part(s)` dropdown when reporting a bug still says `power distribution`.

shsms:
The PowerManager and PowerDistributor were adjusting 0 power requests to exclusion bounds, which it shouldn't do.

It now forwards 0 power requests directly to the microgrid API, which can decide how to adjust for exclusion bounds, if necessary.

Closes #794

### DIFF
--- a/src/frequenz/sdk/actor/_power_managing/_bounds.py
+++ b/src/frequenz/sdk/actor/_power_managing/_bounds.py
@@ -138,7 +138,7 @@ def clamp_to_bounds(  # pylint: disable=too-many-return-statements
 
     # If the given value is within the exclusion bounds and the exclusion bounds are
     # within the given bounds, clamp the given value to the closest exclusion bound.
-    if exclusion_bounds is not None:
+    if exclusion_bounds is not None and not value.isclose(Power.zero()):
         if exclusion_bounds.lower < value < exclusion_bounds.upper:
             return exclusion_bounds.lower, exclusion_bounds.upper
 
--- a/src/frequenz/sdk/actor/power_distributing/_distribution_algorithm/_battery_distribution_algorithm.py
+++ b/src/frequenz/sdk/actor/power_distributing/_distribution_algorithm/_battery_distribution_algorithm.py
@@ -685,7 +685,16 @@ class BatteryDistributionAlgorithm:
         Returns:
             Distribution result
         """
-        if power >= 0.0:
+        if is_close_to_zero(power):
+            return DistributionResult(
+                distribution={
+                    inverter.component_id: 0.0
+                    for _, inverters in components
+                    for inverter in inverters
+                },
+                remaining_power=0.0,
+            )
+        if power > 0.0:
             return self._distribute_consume_power(power, components)
         return self._distribute_supply_power(power, components)
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -52,4 +52,4 @@
 
 ## Bug Fixes
 
-<!-- Here goes notable bug fixes that are worth a special mention or explanation -->
+- 0W power requests are now not adjusted to exclusion bounds by the `PowerManager` and `PowerDistributor`, and are sent over to the microgrid API directly.
--- a/tests/actor/_power_managing/test_matryoshka.py
+++ b/tests/actor/_power_managing/test_matryoshka.py
@@ -313,6 +313,13 @@ async def test_matryoshka_with_excl_3() -> None:
     )
 
     tester = StatefulTester(batteries, system_bounds)
+    tester.tgt_power(priority=2, power=10.0, bounds=(None, None), expected=30.0)
+    tester.tgt_power(priority=2, power=-10.0, bounds=(None, None), expected=-30.0)
+    tester.tgt_power(priority=2, power=0.0, bounds=(None, None), expected=0.0)
+    tester.tgt_power(priority=3, power=20.0, bounds=(None, None), expected=None)
+    tester.tgt_power(priority=1, power=-20.0, bounds=(None, None), expected=-30.0)
+    tester.tgt_power(priority=3, power=None, bounds=(None, None), expected=None)
+    tester.tgt_power(priority=1, power=None, bounds=(None, None), expected=0.0)
 
     tester.tgt_power(priority=2, power=25.0, bounds=(25.0, 50.0), expected=30.0)
     tester.bounds(priority=2, expected_power=30.0, expected_bounds=(-200.0, 200.0))
--- a/tests/actor/_power_managing/test_report.py
+++ b/tests/actor/_power_managing/test_report.py
@@ -58,7 +58,7 @@ def test_adjust_to_bounds() -> None:
         inclusion_bounds=(-200.0, 200.0),
         exclusion_bounds=(-30.0, 30.0),
     )
-    tester.case(0.0, -30.0, 30.0)
+    tester.case(0.0, 0.0, 0.0)
     tester.case(-210.0, -200.0, None)
     tester.case(220.0, None, 200.0)
     tester.case(-20.0, -30.0, 30.0)
--- a/tests/actor/power_distributing/test_battery_distribution_algorithm.py
+++ b/tests/actor/power_distributing/test_battery_distribution_algorithm.py
@@ -1040,6 +1040,11 @@ class TestDistWithExclBounds:
         algorithm = BatteryDistributionAlgorithm()
 
         self.assert_result(
+            algorithm.distribute_power(0, components),
+            DistributionResult({1: 0, 3: 0, 5: 0}, remaining_power=0.0),
+        )
+
+        self.assert_result(
             algorithm.distribute_power(-300, components),
             DistributionResult({1: -100, 3: -100, 5: -100}, remaining_power=0.0),
         )
--- a/tests/timeseries/_battery_pool/test_battery_pool_control_methods.py
+++ b/tests/timeseries/_battery_pool/test_battery_pool_control_methods.py
@@ -109,7 +109,11 @@ class TestBatteryPoolControl:
             )
         )
 
-    async def _init_data_for_batteries(self, mocks: Mocks) -> None:
+    async def _init_data_for_batteries(
+        self, mocks: Mocks, *, exclusion_bounds: tuple[float, float] | None = None
+    ) -> None:
+        excl_lower = exclusion_bounds[0] if exclusion_bounds else 0.0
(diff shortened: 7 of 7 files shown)


src/frequenz/sdk/actor/_power_managing/_bounds.py
```diff
@@ -138,7 +138,7 @@ def clamp_to_bounds(  # pylint: disable=too-many-return-statements
 
     # If the given value is within the exclusion bounds and the exclusion bounds are
     # within the given bounds, clamp the given value to the closest exclusion bound.
-    if exclusion_bounds is not None:
+    if exclusion_bounds is not None and not value.isclose(Power.zero()):
```
llucax, line 141: If the power is close to zero, shouldn't we use `Power.zero()` instead of `value`? What happens if the API receives something that is *close to zero* but not exactly zero?

I understand in general we need to compare using *is close* comparison because of floats work, but if we treat zero specially we might run into problems too.

src/frequenz/sdk/actor/power_distributing/_distribution_algorithm/_battery_distribution_algorithm.py
```diff
@@ -685,7 +685,16 @@ def distribute_power(
...
-        if power >= 0.0:
+        if is_close_to_zero(power):
+            return DistributionResult(
+                distribution={
+                    inverter.component_id: 0.0
+                    for _, inverters in components
+                    for inverter in inverters
+                },
+                remaining_power=0.0,
+            )
```
llucax, line 696: In this case, if I understand correctly, you are setting it to zero if it is close to zero, right?

llucax (commented):
The PR looks good to me in general code-wise, but I feel uneasy about this special casing.

If *"It now forwards 0 power requests directly to the microgrid API, which can decide how to adjust for exclusion bounds, if necessary."*, why are we taking care of bounds in the SDK at all?

Again, I have the feeling that we need an API call to *release* a device (unsetting power so to speak) instead of using 0 as a special case. I think we discussed this before, but don't remember where and what was the conclusion.

@tiyash-basu-frequenz @thomas-nicolai-frequenz FYI

tiyash-basu-frequenz:
> why are we taking care of bounds in the SDK at all?

I have lately been hearing that the SDK now has its own bounds. Don't know why though. ~Sounds like duplication of logic to me. Maybe there is a requirement that I am not aware of. But keep in mind that bounds can become very complicated very soon.~ (Tom explained why. Wasn't aware of it).

> Again, I have the feeling that we need an API call to release a device (unsetting power so to speak) instead of using 0 as a special case. I think we discussed this before, but don't remember where and what was the conclusion.

* "Release" gives the impression that there must be an "acquire", and it thus indicates some sort of locking, which is not the case. We can give it different name though. OR
* we keep `0W` as a special case for now. In the next Microgrid API release, inclusion/exclusion bounds will be replaced by arrays, and we could add a bounds-pair `(0,0)` to that list. It would more idiomatic w.r.t. to the API I feel.

llucax:
Ok, that makes sense, if 0 is explicitly allowed I guess we can keep zero as a special case.

The question about float precision still remains though, what happens if the API receives a set power with 0.00000001? Does it consider it 0 and apply it or it ignores it?

Maybe we can allow calling set power with no power to *unset* power?

llucax:
One last thing that comes to mind, I guess my issue with 0 as a special value is that, if 0 is really not in bounds (like we need DC heating to the battery will be either charging or discharging even if you set power to 0), it will be misleading, and people might think that they stopped a battery from (dis)charging and that is not really true.

llucax:
Another option would be to keep 0 out of bounds and make the SDK decide what to do if 0 is not in bounds, and pick the lower or upper bound closer to 0 explicitly.

thomas-nicolai-frequenz:
> One last thing that comes to mind, I guess my issue with 0 as a special value is that, if 0 is really not in bounds (like we need DC heating to the battery will be either charging or discharging even if you set power to 0), it will be misleading, and people might think that they stopped a battery from (dis)charging and that is not really true.

This is a little bit of a weak argument. This would be like saying setting power to 0 does indeed stop any discharging of a battery but there are things like self-discharge etc. I would be very careful with absolute assumptions. Another example is when the BMS on itself starts balancing a battery. So I don't feel this speaks for or against 0 as a special value.

shsms:
We discussed this only for the battery pool interface @llucax, and there it is simplified and users don't have to worry about exclusion bounds at all.

While communicating with the API, this information needs to be shared in whatever way possible now, but I like the approach @tiyash-basu-frequenz proposes for the next version, of adding `(0, 0)` as a possibility when there are external bounds.

And I think we can consider close-to-zero values in the micro-watt or lower ranges as zero.

thomas-nicolai-frequenz:
> Another option would be to keep 0 out of bounds and make the SDK decide what to do if 0 is not in bounds, and pick the lower or upper bound closer to 0 explicitly.

In which case would the SDK want to set something to 0? If it can't why would it need set anything at all? Of course it always can set the value to the lower bound. Lets assume the lower bound is 5kW and so we would have the same discussion as the value can't be set below 5kW. I don't feel this case is simply related to 0.

shsms:
> Another option would be to keep 0 out of bounds and make the SDK decide what to do if 0 is not in bounds, and pick the lower or upper bound closer to 0 explicitly.

This was already there.  But that's getting in the way of the DC heater, for no reason.  Now we have this behaviour for non-zero powers within the exclusion bounds, but zero is no longer always adjusted to (-300, 150, 150) in case of 3 batteries.

shsms:
> and people might think that they stopped a battery from (dis)charging and that is not really true.

This PR doesn't affect the interface for the users.  They don't have to deal with exclusion bounds, etc.  They propose power to the power manager and they get what they get.

This PR is only about the SDK's communication with the microgrid API.

llucax:
OK, I'm OK with keeping 0 as a special value if the API explicitly allows it in the bounds, and if it is defined behavior that "small enough" values are considered 0.

And yeah @thomas-nicolai-frequenz, I know what you ask for is not always what you get (this also happens with resolution, if you ask for 10W but the device has a resolution of 15W you could get 15W), so the user should never trust the power is what they set, but still I wanted to make the point for the sake of discussion and make sure we are aligned, because besides that, the intention here to set the power to as little as possible, not necessarily zero.

But yeah, my main concern was about the inconsistency of sending 0 when we knew it was out of bounds only because we know the microgrid service is weird and will accept it anyway instead of reject it. If we have the bounds (0,0) always included, that inconsistency goes away.

@shsms in the SDK I would then make sure to always send exactly `0`, when we see a value that is *close to `0`*, because otherwise it will be out of bounds for the API.

shsms:
> @shsms in the SDK I would then make sure to always send exactly 0, when we see a value that is close to 0, because otherwise it will be out of bounds for the API.

This PR is taking care of that already.

src/frequenz/sdk/actor/_power_managing/_bounds.py
shsms: The adjusting to zero is done in the PowerDistributor.  In the PowerManager, we do this check just to decide if the value needs should be clamped to the exclusion bounds or not.

src/frequenz/sdk/actor/power_distributing/_distribution_algorithm/_battery_distribution_algorithm.py
shsms: yup, this is the value that gets sent to the microgrid API.

llucax (approved):

thomas-nicolai-frequenz:
> But yeah, my main concern was about the inconsistency of sending 0 when we knew it was out of bounds only because we know the microgrid service is weird and will accept it anyway instead of reject it. If we have the bounds (0,0) always included, that inconsistency goes away.

That part I missed and thats strange. What does sending bounds (0,0) do when the DC heating is going on and has set the bounds to (5,0). Why would the API accept (0,0). Now I'm also confused.

shsms:
> That part I missed and thats strange. What does sending bounds (0,0) do when the DC heating is going on and has set the bounds to (5,0). Why would the API accept (0,0). Now I'm also confused.

When there's no client request (or 0W request), the DC heater sets -30kW for 30 secs, then switches to +30kW, etc.  to minimize the impact on the soc, etc.

In a location with 1 battery, the power distributor would just set to -30kW always, until the exclusion bounds go away.  That might not be desirable, especially when the user asked for 0 and wasn't expecting to see soc drop.

Although now that I think about it, in a location with 3 batteries for example,  the PowerDistributor would do a better job with reducing the impact on the grid consumption.  For example, if each battery has exclusion bounds -30k..+30k, without this patch, the PowerDistributor would distribute a 0W request as (-60k, 30k, 30k).  (If -60k is not possible on any of the batteries but -50k is, I guess we'll see a 10k consumption from the grid)  And it would switch up the order based on soc changes between the batteries (I think, needs to be confirmed).

But again, most actors don't repeat 0 powers and we don't repeat power values either in the SDK, but have plans to change that in the long term.

I guess there are clearly more options than always offloading the work to the DC heater, at least in some configurations.

shsms:
I will take this in as it is now, because this is already correct, and keep any further optimizations for later.

Issue #556, opened 2023-08-01
Use `start_delay` when creating the resampler `Timer`


llucax:
### What's needed?

We needed to introduce a hack to make the timer start delayed so it can be synced with the resampler alignment. We need to remove that hack eventually.

### Proposed solution

Update to a new `channels` version that supports adding a start delay to the timer.

### Additional context

Fixes a hack introduced in:

* https://github.com/frequenz-floss/frequenz-sdk-python/pull/520

matthias-wende-frequenz:
@llucax is this done?

llucax:
Nope, it might even need a new channels release first. Super low prio though, it is just code cleanup.

matthias-wende-frequenz:
Can we move it to post 1.0 then?

llucax:
I would leave it here for now, when upgrading to channels 1.0.0 it is probably trivial to change this as part of the upgrade.

matthias-wende-frequenz:
I'm not sure if that works. The reason is that we can set a start_delay but we can't set the now when initializing the timer and thus we can't guaranty that the timer is accurately aligned.

llucax:
What do you mean by *set the now*?

llucax:
This will be resolved with #802.

PR #802, opened 2023-12-01
Resync resampler on system time changes

matthias-wende-frequenz:
The resampler was initializing an internal time on initialization by syncing itself to the system clock.

This sync was never redone and thus lead to out-of-sync issues when the system clock changed, which could cause that the resampling window was empty even when samples, with the new system time, arrived.

### DIFF
--- a/src/frequenz/sdk/timeseries/_resampling.py
+++ b/src/frequenz/sdk/timeseries/_resampling.py
@@ -548,7 +548,7 @@ class Resampler:
             # enough samples before the first resampling, otherwise the initial window
             # to collect samples could be too small.
             now + period * 2 - elapsed,
-            period - elapsed if elapsed else timedelta(0),
+            period - elapsed,
         )
 
 
--- a/src/frequenz/sdk/timeseries/_resampling.py
+++ b/src/frequenz/sdk/timeseries/_resampling.py
@@ -370,8 +370,10 @@ class Resampler:
         self._resamplers: dict[Source, _StreamingHelper] = {}
         """A mapping between sources and the streaming helper handling that source."""
 
-        window_end, start_delay_time = self._calculate_window_end()
-        self._window_end: datetime = window_end
+        self._timer: Timer = Timer.periodic(config.resampling_period)
+        """The timer used to trigger the resampling windows."""
+
+        self._window_end = self._sync_timer()
         """The time in which the current window ends.
 
         This is used to make sure every resampling window is generated at
@@ -383,16 +385,6 @@ class Resampler:
         The window end will also be aligned to the `config.align_to` time, so
         the window end is deterministic.
         """
-
-        self._timer: Timer = Timer.periodic(config.resampling_period)
-        """The timer used to trigger the resampling windows."""
-
-        # Hack to align the timer, this should be implemented in the Timer class
-        self._timer._next_tick_time = _to_microseconds(
-            timedelta(seconds=asyncio.get_running_loop().time())
-            + config.resampling_period
-            + start_delay_time
-        )  # pylint: disable=protected-access
 
     @property
     def config(self) -> ResamplerConfig:
@@ -456,6 +448,34 @@ class Resampler:
         except KeyError:
             return False
         return True
+
+    def _sync_timer(self, extra_period: bool = True) -> datetime:
+        """Resync the timer.
+
+        This method will resync the timer to the current time, so the next
+        resampling window will start at the next multiple of
+        `self._config.resampling_period` starting from now.
+
+        Args:
+            extra_period: Add an extra period when it is not aligned to make sure we
+                collected enough samples before the first resampling, otherwise the
+                initial window to collect samples could be too small.
+
+        Returns:
+            The end time of the resampling window.
+        """
+        window_end, start_delay_time = self._calculate_window_end(extra_period)
+
+        # Hack to align the timer, this should be implemented in the Timer class
+        self._timer._next_tick_time = (  # pylint: disable=protected-access
+            _to_microseconds(
+                timedelta(seconds=asyncio.get_running_loop().time())
+                + self.config.resampling_period
+                + start_delay_time
+            )
+        )
+
+        return window_end
 
     async def resample(self, *, one_shot: bool = False) -> None:
         """Start resampling all known timeseries.
--- a/src/frequenz/sdk/timeseries/_resampling.py
+++ b/src/frequenz/sdk/timeseries/_resampling.py
@@ -503,6 +503,14 @@ class Resampler:
         async for drift in self._timer:
             now = datetime.now(tz=timezone.utc)
 
+            # If the system time changes, then `self._window_end` might drift to
+            # far away, such that the resampling bucket might be empty although
+            # new samples aligned to the new system time have been received.
+            # Thus we resync `self._window_end` to the new system time in case
+            # it drifted more then one resampling period away from the system time.
+            if abs(self._window_end - now) - drift > self._config.resampling_period:
+                self._window_end = self._sync_timer(extra_period=False)
+
             if drift > tolerance:
                 _logger.warning(
                     "The resampling task woke up too late. Resampling should have "
--- a/src/frequenz/sdk/timeseries/_resampling.py
+++ b/src/frequenz/sdk/timeseries/_resampling.py
@@ -513,7 +513,9 @@ class Resampler:
             if one_shot:
                 break
 
-    def _calculate_window_end(self) -> tuple[datetime, timedelta]:
+    def _calculate_window_end(
+        self, extra_period: bool = True
+    ) -> tuple[datetime, timedelta]:
         """Calculate the end of the current resampling window.
 
         The calculated resampling window end is a multiple of
@@ -525,10 +527,14 @@ class Resampler:
         the end of the current resampling window will be more than one period away, to
         make sure to have some time to collect samples if the misalignment is too big.
 
+        Args:
+            extra_period: Add an extra period when it is not aligned to make sure we
+                collected enough samples before the first resampling, otherwise the
+                initial window to collect samples could be too small.
         Returns:
             A tuple with the end of the current resampling window aligned to
                 `self._config.align_to` as the first item and the time we need to
-                delay the timer start to make sure it is also aligned.
+                delay the timer to make sure it is also aligned.
         """
         now = datetime.now(timezone.utc)
         period = self._config.resampling_period
@@ -543,11 +549,9 @@ class Resampler:
         if not elapsed:
             return (now + period, timedelta(0))
 
+        extra_period_factor = 2 if extra_period else 1
         return (
-            # We add an extra period when it is not aligned to make sure we collected
-            # enough samples before the first resampling, otherwise the initial window
-            # to collect samples could be too small.
-            now + period * 2 - elapsed,
+            now + period * extra_period_factor - elapsed,
             period - elapsed,
         )
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -53,3 +53,4 @@
 ## Bug Fixes
 
 - 0W power requests are now not adjusted to exclusion bounds by the `PowerManager` and `PowerDistributor`, and are sent over to the microgrid API directly.
+- `timeseries.resampling` will now resync to the system time if it drifts away for more then a resample period.
--- a/tests/timeseries/test_resampling.py
+++ b/tests/timeseries/test_resampling.py
@@ -15,7 +15,7 @@ import pytest
 import time_machine
 from frequenz.channels import Broadcast, SenderError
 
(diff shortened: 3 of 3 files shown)


matthias-wende-frequenz:
We can discuss getting pulling this in and add a test shortly after in a separate PR, to speed up the patch release. What do you think?

llucax:
> We can discuss getting pulling this in and add a test shortly after in a separate PR, to speed up the patch release. What do you think?

What's the plan here? If the plan is that this is tested in production (or somewhere else), any reason not to just temporarily point the SDK dependency to your fork's branch (https://github.com/matthias-wende-frequenz/frequenz-sdk-python/tree/fix_resampling)?

I'd be OK to merge first and do the test afterwards if there is no other way around, but if we can do whatever we need without accepting PRs without tests, I think it makes more sense.

src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -457,6 +449,29 @@ def remove_timeseries(self, source: Source) -> bool:
...
+        window_end, start_delay_time = self._calculate_window_end()
+
+        # Hack to align the timer, this should be implemented in the Timer class
+        self._timer._next_tick_time = (  # pylint: disable=protected-access
+            _to_microseconds(
+                timedelta(seconds=asyncio.get_running_loop().time())
+                + self.config.resampling_period
+                + start_delay_time
+            )
+        )
```
llucax, line 476: Technically out of scope for this PR, but you can now remove the hack too, as the channels dependency was bumped to a version that has the `start_delay`. That would also close #556.

src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -483,6 +498,10 @@ async def resample(self, *, one_shot: bool = False) -> None:
         async for drift in self._timer:
             now = datetime.now(tz=timezone.utc)
 
+            # resync the window_end to the system time
+            if (self._window_end - now) > self._config.resampling_period:
+                self._window_end = self._sync_timer()
```
llucax: I think it would be good to add a more comprehensive explanation of why is this needed, at least for me it's not obvious. So `window_end - now` is basically `-drift` but taking the wall clock instead of the loop's monotonic clock, right? If that's so, maybe it would be also easier to understand if we name variables more explicitly, like:

```py
            drift_relative_to_wall_clock = now - self.window_end
            if drift_relative_to_wall_clock <= self._config.resampling_period:
                self._window_end = self._sync_timer()
```

And this looks a bit weird to me, so under normal circumstances (no wall clock change), you want to resync the window_end?

In the case of the wall clock changing forward (so if `window_end`, `we` is 10 and `now` should have been 10 but it's 12 because the wall clock jumped 2 seconds forward), then `drift_relative_to_wall_clock` is `2`, assuming a `resampling_period`, `p`, of `1`, then `we` won't be synced?

And in the case of the wall clock changing backwards (so if `window_end`, `we` is 10 and `now` should have been 10 but it's 8 because the wall clock jumped 2 seconds backwards), then `drift_relative_to_wall_clock` is `-2`, assuming a `resampling_period`, `p`, of `1`, then `we` will be synced?

In any case, it looks to be that we should somehow involve the (monotonic loop clock's) `drift` to determine if the wall clock drifted or not, right?

src/frequenz/sdk/timeseries/_resampling.py
llucax: Thinking about it twice, we might want to do that in a follow-up PR so these commits are easier to backport to v0.25.x, which won't be able to use the `start_delay`.

src/frequenz/sdk/timeseries/_resampling.py
matthias-wende-frequenz: > And this looks a bit weird to me, so under normal circumstances (no wall clock change), you want to resync the window_end?

Not quite. We want to update the next tick time in the timer and adjust the time window that is used for resampling.
The `self._window_end` can be seen as the resamplers internal time, which (before the patch) was only monotonic increased.

Since we trigger the loop with the same time steps as we increase `self._window_end`, the resync should only happen, when the system time changed.
But the condition is also valid if the "async drift" is bigger then a resampling interval.
I'm not sure if about the implications, because we would set the next tick with more then one resampling interval into the future.

This check is added to calculate the drift of the internal clock, i.e. `self.window_end` against the wall clock.

> In the case of the wall clock changing forward (so if window_end, we is 10 and now should have been 10 but it's 12 because the wall clock jumped 2 seconds forward), then drift_relative_to_wall_clock is 2, assuming a resampling_period, p, of 1, then we won't be synced?
> 
> And in the case of the wall clock changing backwards (so if window_end, we is 10 and now should have been 10 but it's 8 because the wall clock jumped 2 seconds backwards), then drift_relative_to_wall_clock is -2, assuming a resampling_period, p, of 1, then we will be synced?

I think it would be the other way around, but in any case good catch. It should resync as soon as the window_end drifts away more then one sample period from the wall clock, since that means that at least one sample drifted out of the resampling window.
I remember I had an `abs(...)` in there. Maybe it got lost while rebasing (I had a lot more code for testing etc. that got reduced to these two lines).

I'll update the variable names and add the abs back.

src/frequenz/sdk/timeseries/_resampling.py
matthias-wende-frequenz: I don't think that `start_delay` is sufficient. The problem is that the `now` when resyncing and the `now` when using start_delay might be already out of sync. Also it wont work when resyncing after initialization.

matthias-wende-frequenz:
> I'd be OK to merge first and do the test afterwards if there is no other way around, but if we can do whatever we need without accepting PRs without tests, I think it makes more sense.

I was thinking that we can unblock the fix. But I feel it's worth adding the tests before merging. Also the tests are almost ready. Will add them to this PR soon.

src/frequenz/sdk/timeseries/_resampling.py
matthias-wende-frequenz: I see a problem with that patch. In fact, whenever we resync the window_end is new calculated to a position that is a bit in the future, such that the resampler gets some time to collect some samples.

If we now resync because of an `async drift` the resampler would leave out one value.

I see two solutions.

1) Don't re-sync when drift is above the tolerance.
2) Don't add these extra delay.
3) Have different sync codes for initialization and for re-sync

I prefer option 2) because I don't understand why we do this. If the resampler starts as soon as possible after initialization and there are no samples then it sends out nones.

src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -457,6 +449,34 @@ def remove_timeseries(self, source: Source) -> bool:
             return False
         return True
 
+    def _sync_timer(self, extra_period: bool = True) -> datetime:
```
llucax, line 452: Silly, but I would either remove this extra period, as is it only a startup glitch we are trying to fix (I guess it made sense in the original code because it was *free*), or to switch the default, as the normal situation seems to be not adding an extra period.

src/frequenz/sdk/timeseries/_resampling.py
llucax: Yeah, I didn't mean to fix the bug, but just because you were touching this code already, but as I said, bad idea.

src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -483,6 +503,14 @@ async def resample(self, *, one_shot: bool = False) -> None:
         async for drift in self._timer:
             now = datetime.now(tz=timezone.utc)
 
+            # If the system time changes, then `self._window_end` might drift to
+            # far away, such that the resampling bucket might be empty although
+            # new samples aligned to the new system time have been received.
+            # Thus we resync `self._window_end` to the new system time in case
+            # it drifted more then one resampling period away from the system time.
+            if abs(self._window_end - now) - drift > self._config.resampling_period:
```
llucax, line 511: Sadly this is still not enough. Both drifts need to be handled independently, the mixing of both could end up hiding issues. For example:

```
monotonic clock
0         1         2         3         4         5         6 
|---------|---------|---------|---------|---------|---------|---------> time
o         o         o         o         o         o
          t1        t2        t3        t4        t5                    expected ticks
            T1       T2       T3                     T4                 observed ticks
O           *        *        *                      *
|------------|------------|------------|------------|------------|-------> time
0            1            2            3            4            5
wall clock

resampling_period = 1.0, and the wall clock has an accumulating drift of 0.3s per second.

T1:
* now = ~0.9
* window_end = 1.0
* drift = 0.2
* abs(self._window_end - now) - drift > self._config.resampling_period
  abs(             1.0 - 0.9) - 0.2   > 1.0  
                          0.1 - 0.2   > 1.0
                               -0.1   > 1.0 => False

T2:
* now = ~1.7
* window_end = 2.0
* drift = 0.1
* abs(self._window_end - now) - drift > self._config.resampling_period
  abs(             2.0 - 1.7) - 0.1   > 1.0  
                          0.3 - 0.1   > 1.0
                                0.2   > 1.0 => False

T3:
* now = ~2.4
* window_end = 3.0
* drift = 0
* abs(self._window_end - now) - drift > self._config.resampling_period
  abs(             3.0 - 2.4) - 0.0   > 1.0  
                          0.6 - 0.0   > 1.0
                                0.6   > 1.0 => False

T4:
* now = ~4.1
* window_end = 4.0
* drift = 1.3
* abs(self._window_end - now) - drift > self._config.resampling_period
  abs(             4.0 - 4.1) - 1.3   > 1.0  
                          0.1 - 1.3   > 1.0
                               -1.2   > 1.0 => False
```

I don't know, from this maybe it's enough to use `abs(abs(we - now) - abs(drift)) > period`, I have to think about it a bit more. But before doing that, now I really wonder why we are doing all this... :thinking: 

Don't we just want to **always** sync to the wall clock, so effectively make the resampler count periods in terms of the wall clock instead of the monotonic clock?

O maybe we just want to emit warnings if the wall clock is drifting? Having a drifting wall clock is really an operational issue, so I'm not sure if we should try to fix it in the code, as it will probably bring all sorts of other issues too.

And actually I'm not even sure we should warn about it in the SDK, probably we need some system-level monitoring for this instead. Is like if there were a disk failure and we try to warn the user about it and cope with it by reallocating stuff. I think is beyond the SDK's responsibilities.

It really looks like we are trying to patch a broken system here, and I'm not sure if it is a good idea...

llucax: OK, I quickly drafted a solution that I think makes everything much easier. I just created a new timer type that is attached to the wall clock especially for the `Resampler`. We could then polish it and move it to the channels repo if we think it is generally useful. For now it doesn't have all the complexity of the missing ticks policies of the channels timer, it only implements the trigger all missing policy that's needed by the resampler.

Here is the proof of concept PR, against this PR:

* https://github.com/matthias-wende-frequenz/frequenz-sdk-python/pull/6

The resampler gets much simplied with this, as it doesn't need to care about any time management anymore.

llucax:
I just checked the status of my proposal PR, without rebasing to the current v1.x.x. It fails some tests, not sure if it also did originally and I didn't check back then or if something changed in the dependencies that makes it fail now, but in any case, it needs more work to get it to a mergeable state. I will give it a shot to try to figure out if they are easy fixes or if it is something that will require a considerable amount of time.

llucax:
OK, I rebased it and checked the status a bit more and it was not complete (I thought it was, I'm wondering if I have some other branch somewhere with what it's missing and I thought I already had :thinking:, but I can't find anything). I think the wall clock timer implementation was complete, but not the integration in the resampler.

So this still requires some work, at least if we are going for my solution. IIRC your solution was still susceptible to some corner cases and this is why I proposed the timer attached to the wall clock instead, but if I'm recalling incorrectly and you solution worked, maybe we can merge that if the bug is more pressing now, and leave the wall clock timer solution for the future.

Otherwise we need to weight again what's the priority for it, if I take on it I think it could take from a full-day (if everything goes well, so pretty slim chances) to a week (on a more realistic estimation considering I can hardly focus on one only topic for very long :P).

EDIT: I might have found the other branch, but it is too late now, I will check it out tomorrow.

llucax:
OK, no, the other branch was even older. I was looking into this today a little bit more, and it will definitely take some time to clean up. Specially adapting tests (which are quite complicated already) and writing new ones I think it's what will take the most time, as it is a very complicated issue to think about, model and test, with these 2 separate *time spaces* (the monotonic, real time, and the wall clock, human convention time) that can make look time is either compressing or expanding, and how and when to resync on each case. And then, how this is applied to the resampler in terms of how we build resampling windows :exploding_head: 

Just getting to asses this took me a few hours.

llucax:
Alternative PR (building on top of this one): https://github.com/frequenz-floss/frequenz-sdk-python/pull/999

matthias-wende-frequenz:
closing in favor of #999

PR #804, opened 2023-12-07
Fix frequency sending `Quantity`

Marenz:
- Move frequency() tests into own file
- Fix frequency() sending `Quantities` instead of `Frequency`

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -57,3 +57,4 @@
 ## Bug Fixes
 
 - 0W power requests are now not adjusted to exclusion bounds by the `PowerManager` and `PowerDistributor`, and are sent over to the microgrid API directly.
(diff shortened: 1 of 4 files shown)


shsms (approved):

llucax:
Not sure if anybody checked if we need to fix this in other metrics...

shsms:
Everything coming through the formula engine is converted.  And other metrics in the battery pool don't go through the resampler.  So none of them were affected.

llucax:
OK, I think we need to make the `ChannelRegistry` generic eventually, but it will be some work. I was thinking that as an intermediate hack, we can have a `Unknown` class and use that instead of `Any`, that will force us to explicitly cast anything coming from an *untyped* channel registry receiver and catch this type of bugs earlier. I will give it a shot, it should be quick to implement.

* #806

Issue #806, opened 2023-12-08
`ChannelRegistry` should be generic


llucax:
### What's needed?

We need to make sure type hinting catches as many bugs as possible. `ChannelRegistry` is currently using `Any` as the message type when creating `Sender`s and `Receiver`s, which make all type hinting beyond that point useless, as `Any` will satisfy any type.

For example, we got this bug unnoticed for quite a while because of this issue:

* https://github.com/frequenz-floss/frequenz-sdk-python/pull/804

### Proposed solution

Make `ChannelRegistry` generic. We might need to create a registry of registries to use different types though.

llucax:
After trying a couple of quick approaches:

* #808
* #813 

It seem the real solution is to have a truly global meta channel registry, which can create registries for new types dynamically.

PR #807, opened 2023-12-08
Fix bug in `GridFrequency` quantity conversion

### LINKED ISSUE #831: The `MockResampler` interprets `0.0` as `None `
### What happened?

The `MockResampler` has a check `None if not sample.value else ...` which will end up interpreting a `0.0` value as `None`. For example:

https://github.com/frequenz-floss/frequenz-sdk-python/blob/9b0ec9a56c0beba86f9e0c5cc36e7fc3ebff0a34/tests/timeseries/mock_resampler.py#L194

`None` should be handled with `if not sample.value is None`.

### Affected version(s)

_No response_

### Affected part(s)

### Extra information

Related to:

- #829

llucax:
This is a follow-up to #804.

The sample value can also be `None`, in which case we can't access to `sample.value.base_value`. In that case we just cast the existing sample as it doesn't matter which type the value has when it is `None`, avoiding the extra copying.

I discovered this bug while implementing a solution for #806. Which kind of proves it is a worthy pursue...

Also fixes #831.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -57,7 +57,13 @@
 ## Bug Fixes
 
 - 0W power requests are now not adjusted to exclusion bounds by the `PowerManager` and `PowerDistributor`, and are sent over to the microgrid API directly.
-- Fixed that `microgrid.frequency()` was sending `Quantity` objects instead of `Frequency`.
+
+- `microgrid.frequency()` / `GridFrequency`:
+
+  * Fix sent samples to use `Frequency` objects instead of raw `Quantity`.
+  * Handle `None` values in the received samples properly.
+  * Convert `nan` values in the received samples to `None`.
+
 - The resampler now properly handles sending zero values.
 
   A bug made the resampler interpret zero values as `None` when generating new samples, so if the result of the resampling is zero, the resampler would just produce `None` values.
--- a/tests/timeseries/mock_resampler.py
+++ b/tests/timeseries/mock_resampler.py
@@ -5,6 +5,7 @@
 
 
 import asyncio
+import math
 from datetime import datetime
 
 from frequenz.channels import Broadcast, Receiver, Sender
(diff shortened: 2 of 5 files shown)


Marenz:
Strange, I did try cast before and it didn't work for me.

llucax:
> Strange, I did try cast before and it didn't work for me.

No, it doesn't work if `value` is not `None`. The cast is just an optimization. The important part if that if `sample.value` is `None`, accessing to `sample.value.base_value` will raise an exception.

llucax:
I will update the tests to cover the `None` case.

llucax:
The updating of the tests is not a trivial task, as `MockMicrogrid` and `MockResampler` is not handling `None` values properly. Will see if I can up with a quick hack or I need to...

image

Marenz (dismissed):

llucax:
So, I fixed all the things and now this is ready for review again.

I even found a new bug in the handling of `nan`, or at least I consider the current behavior (of not converting `nan` to `None`) a bug.

Related to #829.

llucax:
Enabling auto-merge, so if you have comments please don't approve.

RELEASE_NOTES.md
```diff
@@ -57,7 +57,13 @@
 ## Bug Fixes
 
 - 0W power requests are now not adjusted to exclusion bounds by the `PowerManager` and `PowerDistributor`, and are sent over to the microgrid API directly.
-- Fixed that `microgrid.frequency()` was sending `Quantity` objects instead of `Frequency`.
+
+- `microgrid.frequency()` / `GridFrequency`:
+
+  * Fix sent samples to use `Frequency` objects instead of raw `Quantity`.
+  * Handle `None` values in the received samples propertly.
```
daniel-zullo-frequenz: ?
```suggestion
  * Handle `None` values in the received samples properly.
```

tests/timeseries/mock_resampler.py
```diff
@@ -187,58 +188,65 @@ async def _handle_resampling_requests(self) -> None:
                 )
             )
 
+    def make_sample(self, value: float | None) -> Sample[Quantity]:
+        """Create a sample with the given value."""
```
daniel-zullo-frequenz, line 192: `Args` and `Returns` are not mandatory for helpers in tests, right?

daniel-zullo-frequenz (commented):
Just a couple of minor comments to check for and LGTM

tests/timeseries/mock_resampler.py
llucax: The current configuration doesn't require to document arguments or returns when the docstring is one line only, in all the code. @shsms bought this up recently so we are going to talk about it in the next meeting.

RELEASE_NOTES.md
llucax: Fixed.

llucax:
Fixed and rebased.

daniel-zullo-frequenz (approved):

PR #808, opened 2023-12-08
WIP: Force explicit casts when using the ChannelRegistry

### LINKED ISSUE #806: `ChannelRegistry` should be generic

llucax:
This commit creates a new type `Unknown` to be used as the opposite of `typing.Any`, i.e. to indicate that the type is unknown and forcing the user to explicitly cast from `Unkown` to a known type.

It is intended as a quick hack to improve #806, but the problem with this approach is it is not a pleasant interface to work with, and requires a lot of casting, specially in the tests, so it will be quite some work to revert when we come up with a proper solution.

### DIFF
--- a/src/frequenz/sdk/actor/_channel_registry.py
+++ b/src/frequenz/sdk/actor/_channel_registry.py
@@ -3,13 +3,19 @@
 
 """A class that would dynamically create, own and provide access to channels."""
 
-from __future__ import annotations
-
-import typing
-
 from frequenz.channels import Broadcast, Receiver, Sender
 
 from .._internal._channels import ReceiverFetcher
+
+
+class Unknown:
+    """An unknown type.
+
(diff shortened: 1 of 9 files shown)


llucax:
I'm also discarding this approach as too hacky. BTW, we don't need `Uknown` for this, we can use `object` instead.

PR #812, opened 2023-12-08
Fix the resampler handling of output zeros

### LINKED ISSUE #810: The resampler interprets zeros as `None`
### What happened?

The resampler is interpreting zeros as `None`. This breaks the average calculation and also, in extreme cases, when a stream of zeros long enough come, the resampler will start producing `None`s instead.

Interpret zeros as zeros

### Affected version(s)

v0.22.0+

### Affected part(s)

### Extra information

This bug was introduced when `Quantity` was added, specifically by:

* https://github.com/frequenz-floss/frequenz-sdk-python/commit/63d7c9231cd7d9666d293a6808e02ee3ade0eff4#diff-64982d7c164bb9baee028a55a89678a9cd2f9e476e474fcb1bbe36ceddbf8f5bR729

And it is first present in:

* https://github.com/frequenz-floss/frequenz-sdk-python/releases/tag/v0.22.0

llucax:
A bug made the resampler interpret output zero values as `None`, producing wrong resampled values when the result of the resampling function is zero.

Fixes #810.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,17 +1,9 @@
 # Frequenz Python SDK Release Notes
 
-## Summary
-
-<!-- Here goes a general summary of what this release is about -->
-
-## Upgrading
-
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
-
-## New Features
-
-<!-- Here goes the main new features and examples or instructions on how to use them -->
-
 ## Bug Fixes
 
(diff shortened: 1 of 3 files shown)


shsms (approved):

llucax:
A test is failing, but I checked and it happens with the current v0.25.x too, so it is not related to this PR, so merging anyways.

I created #814 to track the issue and decide if we want to investigate/fix it.

PR #813, opened 2023-12-08
WIP: Make `ChannelRegistry` generic

llucax:
Make `ChannelRegistry` generic

Try to find out the exact type needed for some particular channel registries, but sometimes the same `ChannelRegistry` is shared by multiple types. In one case a custom channel "meta" registry could be created, `QuantitySampleChannelRegistry`, but then the data pipeline mix all other types of channels for power distribution.

It is intended as a solution for #806, but it seems like the real solution is to have a truly global meta channel registry, which can create registries for new types dynamically.

### DIFF
--- a/src/frequenz/sdk/actor/_channel_registry.py
+++ b/src/frequenz/sdk/actor/_channel_registry.py
@@ -5,14 +5,16 @@
 
 from __future__ import annotations
 
-import typing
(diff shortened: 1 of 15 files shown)


llucax:
I will not proceed with this approach because it would be too complicated to implement properly.

PR #822, opened 2023-12-18
WIP: Make `Quantity` an abstract base class

### LINKED ISSUE #172: `Quantity` should be an abstract base class
> [!NOTE]
> This issue was originally filed as frequenz-floss/frequenz-sdk-python#821, back when quantities lived in the SDK as `frequenz.sdk.timeseries.Quantity`. They now live here as [`frequenz.quantities.Quantity`](https://github.com/frequenz-floss/frequenz-quantities-python/blob/119e11248ef10c37df48f69e312d377eb57fa2b3/src/frequenz/quantities/_quantity.py), so this is where the work has to happen. There was a draft implementation in the SDK, frequenz-floss/frequenz-sdk-python#822, which was closed as obsolete; its learnings are summarised below so nobody has to redo that exploration.

### What's needed?

We should not allow using a plain `Quantity` without any associated... *quantity*, as that's just a `float`.

### Proposed solution

Make `Quantity` an abstract base class, so the only instantiable quantities are the ones that actually carry a unit.

## Current state of the code

`Quantity` is a plain class and `Quantity(1.0)` works today, producing a unit-less value. The pieces relevant to this issue are:

* [`Quantity.__init__(value, exponent=0)`](https://github.com/frequenz-floss/frequenz-quantities-python/blob/119e11248ef10c37df48f69e312d377eb57fa2b3/src/frequenz/quantities/_quantity.py#L31-L38) is public and functional.
* [`zero()`](https://github.com/frequenz-floss/frequenz-quantities-python/blob/119e11248ef10c37df48f69e312d377eb57fa2b3/src/frequenz/quantities/_quantity.py#L80-L93) and [`from_string()`](https://github.com/frequenz-floss/frequenz-quantities-python/blob/119e11248ef10c37df48f69e312d377eb57fa2b3/src/frequenz/quantities/_quantity.py#L95-L128) build instances with `cls.__new__(cls)`, so they bypass `__init__` entirely and work on the base class too.
* [`base_value`](https://github.com/frequenz-floss/frequenz-quantities-python/blob/119e11248ef10c37df48f69e312d377eb57fa2b3/src/frequenz/quantities/_quantity.py#L130-L133), [`isnan()`](https://github.com/frequenz-floss/frequenz-quantities-python/blob/119e11248ef10c37df48f69e312d377eb57fa2b3/src/frequenz/quantities/_quantity.py#L172-L178) and [`isinf()`](https://github.com/frequenz-floss/frequenz-quantities-python/blob/119e11248ef10c37df48f69e312d377eb57fa2b3/src/frequenz/quantities/_quantity.py#L180-L186) are still there, and there is no `__float__`.
* Every subclass still uses [`metaclass=NoDefaultConstructible`](https://github.com/frequenz-floss/frequenz-quantities-python/blob/119e11248ef10c37df48f69e312d377eb57fa2b3/src/frequenz/quantities/_quantity.py#L512-L529) to disable its own default constructor.
* `QuantityT` no longer exists here. The SDK now has its own `QuantityT = TypeVar("QuantityT", bound=Quantity)`, so the old value-constrained `TypeVar` listing every quantity type is already gone.

Two things from the draft PR did land in the meantime: `__round__`, `__pos__` and `__mod__` are already implemented.

### `abc.ABC` alone does not do anything here

`class Quantity(abc.ABC)` does not prevent instantiation unless there is at least one abstract method, and there is no sensible one to add. The draft solved it by making `Quantity.__init__()` always raise `TypeError`, and then adding explicit `if cls is Quantity: raise TypeError(...)` guards to `zero()` and `from_string()`, because those use `__new__` and would otherwise sneak past the constructor. Instances are created internally through the existing `_new()` classmethod.

### The `NoDefaultConstructible` metaclass can go away

Once the base `__init__` always raises, subclasses inherit that behaviour for free and the metaclass becomes redundant. Dropping it is worth doing on its own: a custom metaclass is a hazard for anyone downstream who wants to combine a quantity with another metaclass-using base, and it is currently a public name in a private module that subclass authors have to import.

### Declare `_base_value` inside `__init__`, not at class level

The draft moved the `_base_value: float` annotation and its docstring into `__init__`. That was purely for mkdocstrings, which renders it better there.

### Replace `base_value` with `__float__`

This is the piece that makes the whole thing possible. The only real reason `base_value` is public is so generic code can treat a quantity as an opaque `float`. `__float__` covers that, is more idiomatic, and has a useful side effect: `math.isnan()`, `math.isinf()` and friends then work directly on quantities, so `isnan()` and `isinf()` can be removed (the draft removed them, though keeping them as convenience wrappers is defensible).

`float(quantity)` always returns the value in the base unit, so this needs to be documented as a low-level escape hatch for generic code, with `as_*()` remaining the way users should convert.

Note this also affects the experimental marshmallow module, which serialises via `value.base_value` and would need `float(value)` instead.

### The generic-code replacement is `SupportsFloat`

With `__float__` in place, downstream generic code that today needs a unit-less `Quantity` can use `SupportsFloatT = TypeVar("SupportsFloatT", bound=SupportsFloat)` instead, which also accepts plain `float`. That is strictly better: code that does not care about units gets to skip object creation entirely, and code that does care gets the concrete subclass rather than the base class. The cost is an occasional explicit `float(x)` where comparisons or arithmetic on the raw value are needed.

### Testing consequences

The current test suite instantiates `Quantity` directly in a lot of places. The draft handled this by giving the private test subclasses (`Fz1`, `Fz2`) a public `new()` classmethod forwarding to `_new()`, and by introducing an `Fz1Subclass` for the tests that previously used `Fz1` versus base `Quantity` to check that comparisons between different quantity types fail. Add tests asserting that `Quantity()`, `Quantity.zero()` and `Quantity.from_string()` all raise `TypeError`.

## Earlier investigation on the typing side

This is the reasoning from the original issue that led here, still valid:

* Making the generic parameter bound to `numbers.Real` does not work. `isinstance(1.0, Real)` is `True`, but `T = TypeVar("T", Real)` will not accept `float` as `T`, and this is a known problem that will not be fixed since `numbers` predates type hints. It would also be very hard to make `Quantity` a proper `Real`, because not all operations are allowed: multiplying a quantity by the same quantity gives *quantity²*, not *quantity*.

* Building a custom `RestrictedFloat` sort of works, but once you restrict a `float` down to what a quantity may actually do, there is barely anything left: comparison, addition and scaling.

* Making `Quantity` convertible to `float` is the sensible option, and is what the draft implemented.

## Downstream impact

The SDK is the main consumer and still relies on constructing bare quantities, at least in [the data sourcing actor](https://github.com/frequenz-floss/frequenz-sdk-python/blob/cd5282e76cf72b50e82bf016e46421e92c6e0b18/src/frequenz/sdk/microgrid/_data_sourcing/microgrid_api_source.py#L520) and [the resampler](https://github.com/frequenz-floss/frequenz-sdk-python/blob/cd5282e76cf72b50e82bf016e46421e92c6e0b18/src/frequenz/sdk/timeseries/_resampling/_resampler.py#L486), plus several `.base_value` uses in the ring buffer, the grid frequency and voltage streamers, and power distribution. Those need to move to `float` or to a concrete quantity type.

The original draft stalled on the SDK's formula engine, where input and output types were conflated under a single `QuantityT` parameter (see frequenz-floss/frequenz-sdk-python#822 comments). Those typing ambiguities were fixed later by frequenz-floss/frequenz-sdk-python#1011, so that blocker should mostly be gone.

On the SDK side the draft also found that making the resampler generic requires passing it a value constructor, which turned out to be a nice feature in its own right: resampling something known to be in kW can use `Power.from_kilowatts` directly.

## Related

Worth coordinating with #12 (making comparisons less error prone), since both touch how much a `Quantity` is allowed to behave like a `float`.

### LINKED ISSUE #821: `Quantity` should be an abstract base class
### What's needed?

We should not allow to use a plan `Quantity` without any associated... *quantity*, as that's just a `float`.

### Proposed solution

Make `Quantity` a ABC.

### Use cases

_No response_

_No response_

### Additional context

The solution is probably much more complex than that, as I believe we allowed to construct "abstract quantities" to overcome complications with the type system, in particular related to `Sample`s, that sometimes need to have a `float` value and sometimes a `Quantity`.

I've been investigating different approaches:

* Making `Sample` type parameter bound to `numbers.Real`. This didn't work because brilliantly, `isinstance(1.0, Real) is True`, but using `T = TypeVar("T", Real)` is not good enough to the pass `float` as `T` (the types don't match). This seems to be a know bug that will not be fixed, as it seems `numbers` were created before type hints to solve some other problem. In any case, it will be also very hard to make `Quantity` be a proper `Real`, because we can't allow all operations (for example, if we multiply a quantity by the same type of quantity, we get *quantity²*, not *quantity*, so we can't generally allow it.

* Building our own `RestrictedFloat` type, or something like that, to make it compatible with `float`. This sort of works, but when looking at how much restricted a `float` needs to be to be a `Quantity`, there is not much left. It is basically comparison, addition and scaling.

* Making `Quantity` convertible to `float` (implement `__float__()`). This seems to be the most sensible solution. By doing this the most important operations in `math` related to floats that we need to do in the most generic places even work without conversion, like `math.isnan()` and `math.isinf()`, so we can treat `Quantity`s as `float` directly for that and remove the `isnan()` and `isinf()` methods, being able to work with `float` and `Quantity` in a generic way. For other math, we can do `v = float(quantity)` and then do whatever math we need with `v` in a generic way. Of course we generally don't want users to convert to `float`, this is why we have all the `as_xxx()` methods, but when used generically, like in the resampler, we know we are working with the same type of quantity always, so it is safe to convert to `float`. Also making operations that escape the quantity (like v * v), should not make sense in resampling functions, because the new sample always need to be in the same quantity as the input samples, so that should never happen because of how math works.

llucax:
This is a draft PR just to make sure there are no big objections to moving forward with the idea.

The draft is there mostly to show how the new quantity and `SupportFloatT` is used, the `_quantities` module in particular has a lot of noise, so it shouldn't be reviewed in detail.

The code in this PR passes `mypy` and `flake8` checks on `src`, but not in `tests` or `benchmarks`, as these are not updated yet.

The most important things to note are:

* It uses `SupportFloatT` as a `TypeVar` bound to `SupportsFloat` to allow `Sample` and most other types working with either `float` or `Quantities` to be used.
* This means for the generic types needing to actually use the underlaying value, we sometimes need an extra conversion to `float` when some operation is needed. This only happens in `Sample3Phase` to get the `max/min`, and the resampling function. It happens also implicitly when checking for NaN of infinity.
* Added a value constructor to the resampler to be able to produce new samples. This has the nice side effect that if you need to resample something that you know is in, say, kW instead of W, you can use `Power.from_kilowatts` as the constructor.
* There are quite a few (unfinished) fixes/improvements to Quantity and its sub types. The most important changes to `Quantity` are:

  * The trick to disable the default constructor was replaced by a more straightforward implementation in the `__init__` and the utility `_new` method is added to create instances.
  * The `.base_value` was replaced by `__float__`, as this was only exposed to be able to work with a `Quantity` as if it was an unknown `float`.
  * `QuantityT` is not needed anymore and removed.
  * `Quantity` is now an ABC.
  * The `_base_value` is moved to the constructor, to have better rendering for the docs.
  * Removed `isnan()` and `isinf()` as now the `math` functions can be used directly instead (maybe we can keep them for convenience).
  * There are some random migrations to `match` instead of `if`, which needs to be cleaned up.

Of course once this have a rough approval, I'll split in smaller commits (and even PRs) and update tests and docs.

Fixes frequenz-floss/frequenz-quantities-python#172.


llucax:
When fixing the tests I bumped into an issue that might be tricky to solve.

If I understand correctly, all the `FormulaEngine` stuff is parametrized on a `QuantityT` mainly because of the high level formula engine, which allows to create formulas from formulas.

So far it seems like there is a lot of code and complexity to deal with `QuantityT`, which can be a `Quantity` or `float`, or even 3-phase stuff. For what I've seen these classes also don't differentiate very well between the input and output type, and these are usually mixed (one `QuantityT` `TypeVar` parameter for both things, which can resolve to `Quantity` and thus have a different type for the input and output, but the class is then specialized with `Quantity`, not the specific subclass).

As it is taking me a quite a bit of time  trying to understand all the formula engine stuff, in particular when thinking about input and output types, I will delay the work on this PR.

It is not very clear to me which stuff should work with `float` and which should use quantities. For example, the evaluation is all done with plain `float`s by the `FormulaEvaluator`, which also means that if we are mixing different types.

We need to make sure to use the `base_value` in a way that is compatible, in particular for multiplication (like current * voltage). In this sense I think we might need to have a more explicit definition that `base_value`s are stored in the base SI unit or something like that, so we can operate on them safely as pure `float`s.

To progress with this PR I will probably add a second parameter for formula engine stuff that could have a different input and output type, so we have something like `InputT` and `OutputT`.

llucax:
I thought that maybe to reduce the complexity of this formula engine stuff we could pin the type to `float` and consider it a low-level construct, and only convert to a quantity after a value is retrieved from a formula engine, but I guess this won't work.

The problem is formula engines are really "leaked" to the user, so users can do `(microgrid.grid.power + microgrid.pv.power).new_receiver()` and it would be bad if that receiver just gets `float`s, it would mean users will need to convert to `Power` manually which kills the built-in safety we have now. :(

I guess this is also a shame in terms of performance as we might have a lot of conversion going on inside the pipeline that is not strictly necessary.

Not sure if there is a way to have all the internal formula engine stuff working with floats and only convert to quantity when *leaving the engine*, I guess we would have to put (a lot) more thought into it to find out.

llucax:
FYI, I cleaned up this PR a bit, it is still a WIP, but except from the commits with `WIP` in it, they should be much more readable and give a much better idea of what the intention of this PR is and how things are supposed to work.

llucax:
This might have a new chance after this PR is merged:

* https://github.com/frequenz-floss/frequenz-sdk-python/pull/1011

llucax:
Closing this as invalid: `Quantity` no longer lives in the SDK, it moved to [frequenz-quantities-python](https://github.com/frequenz-floss/frequenz-quantities-python), so making it an ABC can't happen in this repo anymore. The branch is also far too stale to be worth rebasing.

The issue moved along with it and is now frequenz-floss/frequenz-quantities-python#172, updated with the current state of that code plus the learnings from this draft: the `abc.ABC` gotcha, dropping the `NoDefaultConstructible` metaclass, replacing `base_value` with `__float__`, using `SupportsFloat` for generic code, and what all that does to the tests. Whoever picks it up shouldn't have to redo the exploration.

PR #824, opened 2023-12-19
Parse Quantity Strings

Marenz:
- Fix missing number in Quantity formatting for small values
- Add function to allow parsing of Quantity strings

### DIFF
--- a/src/frequenz/sdk/timeseries/_quantities.py
+++ b/src/frequenz/sdk/timeseries/_quantities.py
@@ -261,16 +261,17 @@ class Quantity:
             unit_place = max(self._exponent_unit_map)
         else:
             unit = self._exponent_unit_map[unit_place]
+
         value_str = f"{self._base_value / 10 ** unit_place:.{precision}f}"
 
-        if value_str != "0":
-            stripped = value_str.rstrip("0").rstrip(".")
-        else:
+        if value_str in ("-0", "0"):
             stripped = value_str
+        else:
+            stripped = value_str.rstrip("0").rstrip(".")
 
         if not keep_trailing_zeros:
             value_str = stripped
-        unit_str = unit if stripped != "0" else self._exponent_unit_map[0]
+        unit_str = unit if stripped not in ("-0", "0") else self._exponent_unit_map[0]
         return f"{value_str} {unit_str}"
 
     def __add__(self, other: Self) -> Self:
--- a/src/frequenz/sdk/timeseries/_quantities.py
+++ b/src/frequenz/sdk/timeseries/_quantities.py
@@ -226,7 +226,12 @@ class Quantity:
         else:
             unit = self._exponent_unit_map[unit_place]
         value_str = f"{self._base_value / 10 ** unit_place:.{precision}f}"
-        stripped = value_str.rstrip("0").rstrip(".")
+
+        if value_str != "0":
+            stripped = value_str.rstrip("0").rstrip(".")
+        else:
+            stripped = value_str
+
         if not keep_trailing_zeros:
             value_str = stripped
         unit_str = unit if stripped != "0" else self._exponent_unit_map[0]
--- a/src/frequenz/sdk/timeseries/_quantities.py
+++ b/src/frequenz/sdk/timeseries/_quantities.py
@@ -91,6 +91,42 @@ class Quantity:
         assert isinstance(_zero, cls)
         return _zero
 
+    @classmethod
+    def from_string(cls, string: str) -> Self:
+        """Return a quantity from a string representation.
+
+        Args:
+            string: The string representation of the quantity.
+
+        Returns:
+            A quantity object with the value given in the string.
+
+        Raises:
+            ValueError: If the string does not match the expected format.
+
+        """
+        split_string = string.split(" ")
+
+        if len(split_string) != 2:
+            raise ValueError(
+                f"Expected a string of the form 'value unit', got {string}"
+            )
+
+        assert cls._exponent_unit_map is not None
+        exp_map = cls._exponent_unit_map
+
+        for exponent, unit in exp_map.items():
+            if unit == split_string[1]:
+                instance = cls.__new__(cls)
+                try:
+                    instance._base_value = float(split_string[0]) * 10**exponent
+                except ValueError as error:
+                    raise ValueError(f"Failed to parse string '{string}'.") from error
+
+                return instance
+
+        raise ValueError(f"Unknown unit {split_string[1]}")
+
     @property
     def base_value(self) -> float:
         """Return the value of this quantity in the base unit.
--- a/src/frequenz/sdk/timeseries/_quantities.py
+++ b/src/frequenz/sdk/timeseries/_quantities.py
@@ -199,6 +199,7 @@ class Quantity:
         """
         return self.__format__("")
 
+    # pylint: disable=too-many-branches
     def __format__(self, __format_spec: str) -> str:
         """Return a formatted string representation of this quantity.
 
@@ -250,8 +251,22 @@ class Quantity:
         if math.isinf(self._base_value) or math.isnan(self._base_value):
             return f"{self._base_value} {self._exponent_unit_map[0]}"
 
-        abs_value = abs(self._base_value)
-        exponent = math.floor(math.log10(abs_value)) if abs_value else 0
+        if abs_value := abs(self._base_value):
+            precision_pow = 10 ** (precision)
+            # Prevent numbers like 999.999999 being rendered as 1000 V
+            # instead of 1 kV.
+            # This could happen because the str formatting function does
+            # rounding as well.
+            # This is an imperfect solution that works for _most_ cases.
+            # isclose parameters were chosen according to the observed cases
+            if math.isclose(abs_value, precision_pow, abs_tol=1e-4, rel_tol=0.01):
+                # If the value is close to the precision, round it
+                exponent = math.ceil(math.log10(precision_pow))
+            else:
+                exponent = math.floor(math.log10(abs_value))
+        else:
+            exponent = 0
+
         unit_place = exponent - exponent % 3
         if unit_place < min(self._exponent_unit_map):
             unit = self._exponent_unit_map[min(self._exponent_unit_map.keys())]
--- a/tests/timeseries/test_quantities.py
+++ b/tests/timeseries/test_quantities.py
@@ -107,6 +107,7 @@ def test_string_representation() -> None:
     assert (
         repr(Quantity(1.024445, exponent=0)) == "Quantity(value=1.024445, exponent=0)"
     )
+    assert f"{Quantity(0.50001, exponent=0):.0}" == "1"
     assert f"{Quantity(1.024445, exponent=0)}" == "1.024"
     assert f"{Quantity(1.024445, exponent=0):.0}" == "1"
     assert f"{Quantity(0.124445, exponent=0):.0}" == "0"
@@ -154,6 +155,7 @@ def test_string_representation() -> None:
     assert f"{Energy.from_watt_hours(0.124445):.0}" == "0 Wh"
     assert f"{Power.from_watts(-0.0):.0}" == "-0 W"
     assert f"{Power.from_watts(0.0):.0}" == "0 W"
+    assert f"{Voltage.from_volts(999.9999850988388)}" == "1 kV"
 
 
 def test_isclose() -> None:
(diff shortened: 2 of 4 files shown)


tests/timeseries/test_quantities.py
```diff
@@ -107,6 +108,7 @@ def test_string_representation() -> None:
     )
     assert f"{Quantity(1.024445, exponent=0)}" == "1.024"
     assert f"{Quantity(1.024445, exponent=0):.0}" == "1"
+    assert f"{Quantity(0.124445, exponent=0):.0}" == "0"
```
llucax, line 113: Maybe add to all the test this too?

```py
    assert f"{Quantity(0.50001, exponent=0):.0}" == "1"
```

tests/timeseries/test_quantities.py
```diff
@@ -527,3 +531,31 @@ def test_invalid_multiplications() -> None:
...
+            assert f"{Current.from_string(current_str):.{exponent}}" == current_str
+
+            energy_str = f"{Energy.from_watt_hours(value):.{exponent}}"
+            assert f"{Energy.from_string(energy_str):.{exponent}}" == energy_str
+
+            freq_str = f"{Frequency.from_hertz(value):.{exponent}}"
+            assert f"{Frequency.from_string(freq_str):.{exponent}}" == freq_str
+
+            pct_str = f"{Percentage.from_fraction(value):.{exponent}}"
+            assert f"{Percentage.from_string(pct_str):.{exponent}}" == pct_str
```
llucax: It looks like you are re-implementing [hypothesis](https://hypothesis.readthedocs.io/) :grimacing: 

What you are doing is basically the [first example](https://hypothesis.readthedocs.io/en/latest/quickstart.html#an-example) but you need to [generate inputs with floats](https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.floats) instead (now I wonder what are we doing with *NaN* and :infinity: :thinking: )

llucax (commented):
LGTM otherwise.

tests/timeseries/test_quantities.py
llucax: BTW, to include all types you can also parametrize it (untested):
```py
@pytest.parametrize("quantity_type", [Power, Voltage, ...])
@pytest.parametrize("exponent", [0, 3, 6, 9])
@hypothesis.given(value=st.floats(min_value=-1.0, max_value=1.0))
def test_to_and_from_string(
        quantity_type: type[Quantity],
        exponent: int,
        value: float
    ) -> None:
    """Test string parsing and formatting."""
    quantity = quantity_type.__new__(quantity_type)
    quantity._base_value = value * 10**exponent
    quantity_str = f"{quantity:.{exponent}}"
    assert f"{quantity_type.from_string(quantity_str):.{exponent}}" == quantity_str
```

The construction is a bit hacky, but OK IMHO for a unit test. You could also parametrize on tuples `"quantity_ctor,  quantity_from_str", [(Power.from_watts, Power.from_string), (...), ...]` if you want a less hacky but more verbose approach.

Marenz:
Updated and ready for review

llucax:
CI failing with:

> pylint src benchmarks docs examples noxfile.py tests
> ************* Module frequenz.sdk.timeseries._quantities
> src/frequenz/sdk/timeseries/_quantities.py:204:4: R0912: Too many branches (16/12) (too-many-branches)
> ************* Module tests.timeseries.test_quantities
> tests/timeseries/test_quantities.py:582:4: W0212: Access to a protected member _base_value of a client class (protected-access)

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -91,6 +91,44 @@ def zero(cls) -> Self:
...
+
+        for exponent, unit in exp_map.items():
+            if unit == split_string[1]:
+                instance = cls.__new__(cls)
+                try:
+                    instance._base_value = float(split_string[0]) * 10**exponent
+                except ValueError as error:
+                    raise ValueError(
+                        f"Failed to parse string {string}: {error}"
+                    ) from error
```
llucax: Not a biggie, but if you are raising the exception `from` another `error`, is redundant to include `error` in the message string, as if you need to use it programmatically you can access the chained errors and if the exception is unhandled, Python will show the whole exception chain too.

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -214,8 +252,16 @@ def __format__(self, __format_spec: str) -> str:
...
+        if abs_value := abs(self._base_value):
+            precision_pow = 10 ** (precision)
+            if math.isclose(abs_value, precision_pow, abs_tol=1e-4, rel_tol=0.01):
+                # If the value is close to the precision, round it
+                exponent = math.ceil(math.log10(precision_pow))
+            else:
+                exponent = math.floor(math.log10(abs_value))
+        else:
+            exponent = 0
+
```
llucax: I think it would be good to explain in the commit message why is this needed, I was surprised to see changes in `format()` when the commit says it only adds `from_string()`. Maybe even split the commit if appropriate (I see also there are new tests for `format()`, so it seems like a bug fix that sneaked in.

Thinking about it might be good too to include a small comment in the code about this.

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -214,8 +252,16 @@ def __format__(self, __format_spec: str) -> str:
         if math.isinf(self._base_value) or math.isnan(self._base_value):
             return f"{self._base_value} {self._exponent_unit_map[0]}"
 
-        abs_value = abs(self._base_value)
-        exponent = math.floor(math.log10(abs_value)) if abs_value else 0
+        if abs_value := abs(self._base_value):
+            precision_pow = 10 ** (precision)
+            if math.isclose(abs_value, precision_pow, abs_tol=1e-4, rel_tol=0.01):
```
llucax, line 262: Maybe also explain why those particular arguments to `isclose()`?

tests/timeseries/test_quantities.py
```diff
@@ -531,3 +537,55 @@ def test_invalid_multiplications() -> None:
...
```
llucax: Do you really want to keep this one?

tests/timeseries/test_quantities.py
```diff
@@ -531,3 +537,55 @@ def test_invalid_multiplications() -> None:
...
+    quantity._base_value = value * 10**exponent
+    quantity_str = f"{quantity:.{exponent}}"
+    from_string = quantity_type.from_string(quantity_str)
+    try:
+        assert f"{from_string:.{exponent}}" == quantity_str
+    except AssertionError as error:
+        pytest.fail(
+            f"Failed for {quantity.base_value} != from_string({from_string.base_value}) "
+            + f"with exponent {exponent} and source value '{value}': {error}"
+        )
```
llucax, line 576: Is this just to make debugging easier? Maybe an alternative would be to keep the normal assert and before asserting (always) printing the extra context info, like `print(f"{quantity.base_value=}, {from_string.base_value=} ...", )`. This way when any test fails you'll have all the info you need in the test failure output.

But it is just a suggestion, I'm OK with leaving it as it is now.

llucax (commented):
It looks like the second commit could be split, other than that (and the CI failure) LGTM.

Marenz:
> CI failing with:..

haha, I had that fixed, I just didn't add it to the commit XD

tests/timeseries/test_quantities.py
Marenz: Yeah, it was mainly for debugging

tests/timeseries/test_quantities.py
Marenz: Removed

llucax:
Please re-request a review from me when this is ready for another round.

tests/timeseries/test_quantities.py
```diff
@@ -531,3 +533,26 @@ def test_invalid_multiplications() -> None:
             _ = quantity * 200.0  # type: ignore
         with pytest.raises(TypeError):
             quantity *= 200.0  # type: ignore
+
+
+@pytest.mark.parametrize("quantity_type", [Power, Voltage, Current, Energy, Frequency])
+@pytest.mark.parametrize("exponent", [0, 3, 6, 9])
+@hypothesis.settings(max_examples=1000)
+@hypothesis.seed(42)  # Seed that triggers a lot of problematic edge cases
```
llucax, line 547: Do you really want to leave the seed fixed? Isn't it more useful to let it change to potentially catch more random cases?

tests/timeseries/test_quantities.py
```diff
@@ -531,3 +533,26 @@ def test_invalid_multiplications() -> None:
             _ = quantity * 200.0  # type: ignore
         with pytest.raises(TypeError):
             quantity *= 200.0  # type: ignore
+
+
+@pytest.mark.parametrize("quantity_type", [Power, Voltage, Current, Energy, Frequency])
+@pytest.mark.parametrize("exponent", [0, 3, 6, 9])
+@hypothesis.settings(max_examples=1000)
```
llucax: Any reason to specify a `max_examples`?

llucax (commented):
For "Fix quantity formatting bug with -0/0" it might be good to describe what the bug actually was, is not completely clear to me by just reading the diff.

About the rounding issue when formatting, really tricky, I wonder if it wouldn't be simpler to do a string roundtrip in the formatting function itself, like convert to string, convert again to float, and only then check which exponent is more appropriate, then convert again to string. It might be less efficient (maybe?) but it seems easier to implement and likely more robust.

tests/timeseries/test_quantities.py
Marenz: I know for a fact that there are more cases and solving those requires an unreasonable amount of work (at this point).
@sahas-subramanian-frequenz mentioned an imperfect solution is better than no solution at all and at least here I would agree.

There will be cases were it prints slightly weird, maybe using a sub-optimal unit etc, but the printed result will never be straight out wrong.

tests/timeseries/test_quantities.py
Marenz: Mainly for determinism and to limit the total runtime of the test but also to get a decent amount of samples (default value is `100`)

Marenz:
> About the rounding issue when formatting, really tricky, I wonder if it wouldn't be simpler to do a string roundtrip in the formatting function itself, like convert to string, convert again to float, and only then check which exponent is more appropriate, then convert again to string. It might be less efficient (maybe?) but it seems easier to implement and likely more robust.

Yes, I thought about this particular solution as well. It might be less elegant, but it would be pretty reliable and covering pretty much all possibilities.
Maybe in a future PR?

Marenz:
> For "Fix quantity formatting bug with -0/0" it might be good to describe what the bug actually was, is not completely clear to me by just reading the diff.

Basically it would only print the unit, no number (the strip removed the 0)

tests/timeseries/test_quantities.py
llucax: Yeah, that sounds completely reasonable and I agree 100%. Can you add a comment summarizing this here? Because for the uninformed casual reader it will be really hard to figure all of this out, they'll probably just try to remove those two lines and be completely puzzled about why everything breaks now. I know future me will attempt that :laughing:

llucax:
> Maybe in a future PR?

Sure

tests/timeseries/test_quantities.py
llucax: Sorry to be a pain in the ass about this, but I mean something more like "we are setting these fixed because there are still corner cases that will fail if we allow hypothesis to do random tests, this should be removed when all the corner cases are properly handled".

llucax (commented):
One last comment about the comment. Also if you can add this "It would only print the unit, no number" to the "Fix quantity formatting bug with -0/0 " commit while you are it, I think it's good info to have in the commit message.

tests/timeseries/test_quantities.py
Marenz: You did see the larger docstring of the function explaining basically that in other words?

llucax: No, sorry, I expected it in the comments instead of the docstring so my brain stopped parsing after I couldn't find the clarification there.

llucax (approved):

Marenz:
Extended the commit description with the "no unit" explanation

llucax:
> Maybe in a future PR?

* https://github.com/frequenz-floss/frequenz-quantities-python/issues/11

PR #828, opened 2023-12-28
Speedup tests by lowering the actors restart delay

llucax:
A restart delay was added by #741 and as a side effect the tests became almost 40% slower in total and the actor test became 28 times slower (locally this PR added around 27 seconds to the test run).

This commit fixes this by adding a test fixture to automatically reduce the restart delay to 0 for tests. It also adds a context manager to temporarily change the restart delay in case a different value is needed.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -57,6 +57,43 @@ def actor_auto_restart_once() -> Iterator[None]:
         yield
 
 
+@contextlib.contextmanager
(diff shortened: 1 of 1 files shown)


llucax:
I will merge this as it is very low risk and it is slowing me down a lot locally while doing tests and working in multiple branches. We can make any changes in a followup PR if any are needed.

Issue #826, opened 2023-12-28
Most tasks swallow exceptions


llucax:
### What happened?

When there is an unhandled exception inside a task spawn by the SDK, the error is silently swallowed, leading to obscure, hard-to-debug bugs.

### What did you expect instead?

Unhandled exceptions are at least logged or the whole Python process crashes.

### Affected part(s)

Core components (data structures, etc.) (part:core)

### Extra information

I discovered this while working at #806. I added a sanity check which failed inside a task that was sending messages to a channel, so no messages was sent, and other task waiting for messages just got stuck, leaving no clues about where the problem might be. This makes the problem really hard to debug.

This issue is actually a problem across all areas of the SDK and projects in general.

A way to cope with it in the SDK would be to extend `BackgroundService` to provide a `create_task()` method that automatically adds the task to the task list and then also adds a *done callback* where we can either log the unhandled exception, or just raise a `SystemExit` exception to exit the program. We could even give the user the option to decide how to handle unhandled exceptions by either passing a callback or letting them override the default callback as a method of the instance. This callback should also remove the task from the tasks list, something users need to do manually at the moment.

#### Related issues

The solution to this issue needs to have in mind the following related issues:

* https://github.com/frequenz-floss/frequenz-core-python/issues/9
* #906 
* #819
* #632

llucax:
Maybe we can add a SDK specific `create_task()` that catches and logs unhandled exceptions, then use that function also in `BackgroundService`.

llucax:
Oh, yeah, of course: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.set_exception_handler

llucax:
It is not trivial to override the default handler, because it will store the exception in the task object, and maybe users will inspect this task object in the future and report the exception properly in the future. Because of this, we might want to use a custom exception handler that only prints a debug message about unhandled exceptions, otherwise we might spam the logs with spurious messages about tasks stopping due to unhandled exceptions when exception are really properly handled, just not inside the task function itself, adding more noise to the logs and making things even harder to debug.

According to ChatGPT:

> In Python's asyncio library, when a task raises an exception that is not caught within the task itself, the default behavior is somewhat subtle and important to understand:
>
[...]

Issue #823, opened 2023-12-19
Power Manager `_Report`, `Report` and `BatteryPoolReport` have the wrong hierarchy


llucax:
### What happened?

The current hierarchy is:

```mermaid
classDiagram
    Report <|-- _Report
    Report <|-- BatteryPoolReport
```

The problem is `BatteryPoolReport` **is** a `_Report`, and `_Report` is the concrete class sent through channels, so there are instances when a `_Report` needs to be converted to a `BatteryPoolReport`, for which we just need to go behind the back of the type system.

### What did you expect instead?

If I have a `Receiver[_Report]`, I expect to be able to safely return a `BatteryPoolReport` from it without any casting.

### Affected part(s)

Battery power distribution (part:power-distribution)

### Extra information

So, the naive solution is:

```mermaid
classDiagram
    BatteryPoolReport <|-- _Report
```

But of course this doesn't work in the whole context of `Report`. In the current code, there are no other `Report`s that are not `BatteryPoolReport`, so I'm not sure how the end use of this hierarchy should look like. If it is just documentation, I'm not sure it is worth hiding `_Report`, I think it would be simpler to just keep it generic and then have something like:

> [!NOTE]
> When applied to different components, the attributes of this class have different meanings.
> * `BatteryPool`: component IDs are battery IDs. Yada yada yada
> * Etc.

Also I wonder if we can solve this by using completely different types for different `Report` types, that have no inheritance relationship, are all concrete classes and instead make `Report = BatteryPoolReport | PvPoolReport | ...`. Then we can use the `match` syntax to know what we received from the channel.

I guess we also need to mix all these different reports in the same channel because we are going to use the same power manager for all, because ideally we should just use one concrete type per channel and then we don't need the hierarchy or the type alias at all.

llucax:
FYI @shsms

shsms:
> If I have a Receiver[_Report], I expect to be able to safely return a BatteryPoolReport from it without any casting.

This has nothing to do with the use of the `Protocol` type.  This is because the `Receiver` was too rigid on what type of generic arguments it will accept.

This can be resolved with: https://github.com/frequenz-floss/frequenz-channels-python/pull/262

shsms:
Actually it needs this too, once the channel PR is merged:

```diff
diff --git a/src/frequenz/sdk/_internal/_channels.py b/src/frequenz/sdk/_internal/_channels.py
index d739470d..d3b5a68a 100644
[...]

PR #830, opened 2023-12-28
Make the registry type-aware

### LINKED ISSUE #806: `ChannelRegistry` should be generic

llucax:
The channel registry is using `Any` as the message type, which is not type safe, as it completely *disables* type checking for the messages.

This commit makes the channel registry type-aware, so channels are stored with their message type and the registry checks that the same channel is not used for different message types.

This also makes the registry just a plain container for channels, the wrapper methods to create new senders and receivers, and to configure the `resend_latest` flag are removed. The `ReceiverFetcher` abstraction is also not needed if we just return the channel directly, as the channel itself is a `ReceiverFetcher`.

Also the method to close and remove a channel is made public and the name more explicit, as it is used in normal code paths.

The new registry only provide 2 main methods:

* `get_or_create()`: Get or create a channel for the given key, doing the type checking to make sure the requested message type matches the existing channel message type if it already exists.

* `close_and_remove()`: Close and remove the channel for the given key.

This change uncovered 5 issues:

* `MockMigrogrid/Resampler: Fail on unhandled exceptions` (in this PR, but more generally #826)
* `Fix checks in tests` (in this PR)
* `Fix missing conversion to `QuantityT` (workaroud in this PR, but should be better solved by #821)
* #807
* #823

Fixes #806.


llucax:
Tests will fail until ~#806~ #807 is merged as I didn't include the fix here, but this is ready for the review.

llucax:
Oh, and I need to write the release notes.

llucax:
BTW, I decided to go with this approach instead of the other tries because making `ChannelRegistry` generic doesn't help much, we would still need a channel registry registry for the different underlying types, so I though this approach was simpler and more practical, even when a bit less type-safe (as some issues will only be detected at runtime).

The alternative would be to have a fixed map of underlying types, so we can really support static checks, but I found that to be too inflexible, and potentially confusing and error prone too, as one might have the same key associated to different types, so if you first use key1 with T1, but then use key1 with T2, 2 different channels would be created, introducing a very hard to debug issue (as probably some end will be pushing to key1 using T1, and the other end receiving from key1 too, but expecting T2, so at the end receiving from a different channel, hence not receiving anything as nobody is pushing there, and stalling).

llucax:
Added release notes.

src/frequenz/sdk/actor/_channel_registry.py
```diff
@@ -0,0 +1,151 @@
...
+        if entry is None:
+            raise KeyError(f"No channel for key {key!r} exists.")
+        return entry.message_type
+
+    def __contains__(self, key: str) -> bool:
+        """Check whether the channel for the given `key` exists."""
+        return key in self._channels
+
+    def get_or_create(
+        self, message_type: type[_T], key: str, *, check_type: bool = True
```
daniel-zullo-frequenz: I have the feeling that the argument `check_type` is redundant in the sense that it should always be true. Or at least I don't have any scenario in mind where the user needs to set `check_type` to false providing always a `message_type` as it is mandatory.

In case we have a scenario where the `check_type` should be skipped, then I think a `get(self, key: str)`  method can be provided but without encouraging the user to use it. Probably not  a very good idea

daniel-zullo-frequenz (commented):
It's great that this fixed many hidden issues!
I only have a comment to discuss about the `get_or_create()` method.
There is a typo in the commit title `Update relese notes` in case you want to fix it.
Rather than that LGTM

src/frequenz/sdk/actor/_channel_registry.py
llucax: I agree the API is weird.

I added `check_type` as a escape mechanism because it is not possible (or not easy) to check for sub-types, so if you have `class A: pass; class B(A): pass` and then do `get_or_create(B, k); get_or_create(A, k)` the second call will fail even when it is safe to do so.

It is still probably a very niche use case, so I'd be fine to removing that option until we really need it (I thought I might needed it somewhere, but I think it is not really used).

llucax:
Updated to remove the `check_type` option and to fix the typo in the commit message.

daniel-zullo-frequenz (dismissed):

src/frequenz/sdk/timeseries/battery_pool/_battery_pool.py
```diff
@@ -386,12 +387,17 @@ def power_status(self) -> ReceiverFetcher[BatteryPoolReport]:
...
+            _power_managing._Report, sub.get_channel_name()
         )
+        channel.resend_latest = True
+
+        # We need to cast here because the _Report/Report/BatteryPoolReport
+        # hierarchy is messed up. To be able to return a BatteryPoolReport
+        # without casting, _Report would need to be a subclass of
+        # BatteryPoolReport.
+        # See https://github.com/frequenz-floss/frequenz-sdk-python/issues/823
+        return cast(ReceiverFetcher[BatteryPoolReport], channel)
```
shsms: I see that there's an issue, but I think the explanation is not correct, the hierarchy is convoluted (because the parent `Report` type is unnecessary), but it is not wrong.  And the proposed solution will also not work,  because `Receiver`'s type variable is invariant.  It is an issue with the channel types being less flexible than they can be.  I will make a PR and provide examples in the ticket you made.

shsms (commented):
Just a comment about a comment.

src/frequenz/sdk/timeseries/battery_pool/_battery_pool.py
llucax: I'll keep the discussion about this in #823. I can update the comment to just mention there is a typing issue and to read more details in the issue.

Marenz:
> Tests will fail until https://github.com/frequenz-floss/frequenz-sdk-python/issues/806 is merged as I didn't include the fix here, but this is ready for the review.

#806 is an issue, not a PR though?

llucax:
Rebased and updated comment.

llucax:
And enabled auto-merge.

llucax:
> #806 is an issue, not a PR though?

Sorry, I meant #807.

src/frequenz/sdk/actor/_channel_registry.py
```diff
@@ -0,0 +1,150 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
```
Marenz: 2024 :laughing:

llucax: It was created in 2022 :)

My research shows that copyright notices, and in particular years beyond the year of creation are not that necessary. I did my research a while ago using the good ol' manual Google search, but now checking with ChatGPT to see what they think about it:

Q: Is it necessary to add the copyright year to copyright notices in source code? And what about  keeping it up to date every time the file is updated?

> Including the copyright year in source code copyright notices is customary but not strictly necessary. The year often indicates the first year of publication of the work. Keeping it updated annually is a common practice, though not legally required. The main purpose of the year in a copyright notice is to give an indication of the earliest date of copyright.
> 
> Once a work is copyrighted, it remains so regardless of whether the year in the notice is updated. However, updating the year can be beneficial for demonstrating the timeline of development and updates to the work. This practice can be particularly relevant in open-source projects, where it can help track changes and the evolution of the code over time.
> 
> In summary, while not legally mandatory, adding and updating the year in copyright notices is a good practice for clarity and documentation purposes.

When using a VCS, tracking the date of changes is trivial if it would become necessary for any legal reason.

Q: What about open source projects using a VCS like Git, wouldn't it be trivial then to get a clear timeline of changes and authors?

> ...
> Given this capability, the practice of updating copyright years in source code files becomes less critical from a purely informational standpoint. In Git-managed projects, any interested party can easily access the history of a file to see when it was modified and by whom. This can effectively replace the need for manually updating copyright years in the files themselves as a means of tracking changes.
>
> However, some projects still choose to update copyright years for a variety of reasons, including legal traditions, explicitness in licensing, or simply adhering to established norms. But from a technical perspective, with tools like Git, it's not strictly necessary for tracking changes and authorship.

shsms (approved):

src/frequenz/sdk/actor/_channel_registry.py
Marenz: > It was created in 2022 :)

I think the complete removing & adding confused me here then

shsms: I'm guessing that was just to keep the commits easier to read, because there were big changes.

PR #838, opened 2024-01-03
Improve (and fix some) test finalization

llucax:
- Remove some duplicate checks from pylint
- Make Mocks private to the module
- Import module from `typing`
- Allow passing the mocker in the constructor
- Make `MockMicrogrid` an async context manager
- Use the `MockMicrogrid` as an async context manager
- Improve finalization of power distributing tests
- Improve finalization of voltage streamer tests
- Improve finalization of logical meter tests
- Improve finalization of formula formatter tests
- Improve finalization of EV charger pool tests
- Improve finalization of formula composition tests
- Improve finalization of battery status tests
- Improve finalization of grid tests
- Disable pylint check only in one line

### DIFF
--- a/tests/timeseries/mock_microgrid.py
+++ b/tests/timeseries/mock_microgrid.py
@@ -5,9 +5,9 @@
 
 
 import asyncio
-import typing
 from collections.abc import Callable
 from datetime import datetime, timedelta, timezone
+from typing import Coroutine
 
 from pytest_mock import MockerFixture
 
@@ -151,7 +151,7 @@ class MockMicrogrid:  # pylint: disable=too-many-instance-attributes
         self.evc_component_states: dict[int, EVChargerComponentState] = {}
         self.evc_cable_states: dict[int, EVChargerCableState] = {}
 
-        self._streaming_coros: list[typing.Coroutine[None, None, None]] = []
+        self._streaming_coros: list[Coroutine[None, None, None]] = []
         self._streaming_tasks: list[asyncio.Task[None]] = []
 
         if grid_meter:
--- a/tests/timeseries/mock_microgrid.py
+++ b/tests/timeseries/mock_microgrid.py
@@ -62,6 +62,7 @@ class MockMicrogrid:  # pylint: disable=too-many-instance-attributes
         num_namespaces: int = 1,
         fuse: Fuse | None = Fuse(Current.from_amperes(10_000.0)),
         graph: _MicrogridComponentGraph | None = None,
+        mocker: MockerFixture | None = None,
     ):
         """Create a new instance.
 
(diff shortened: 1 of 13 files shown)


.github/labeler.yml
```diff
@@ -4,48 +4,68 @@
 # The basic syntax is [label]: [path patterns].
 #
 # For more details on the configuration please see:
+  - changed-files:
+    - any-glob-to-any-file:
```
shsms: This appears to be in the wrong commit: https://github.com/frequenz-floss/frequenz-sdk-python/pull/838/commits/3d9936fd17c8c43fb92611184c5865bfa139e3d9

The commit message doesn't seem to mention anything about the labeler.

tests/timeseries/mock_microgrid.py
```diff
@@ -78,10 +79,12 @@ def __init__(  # pylint: disable=too-many-arguments
             fuse: optional, the fuse to use for the grid connection.
             graph: optional, a graph of components to use instead of the default grid
                 layout. If specified, grid_meter must be None.
+            mocker: optional, a mocker to pass to the mock client and mock resampler.
 
         Raises:
             ValueError: if both grid_meter and graph are specified.
         """
+        self.mocker = mocker
```
shsms: why not `self._mocker`?

tests/timeseries/test_logical_meter.py
```diff
@@ -20,157 +22,185 @@ class TestLogicalMeter:  # pylint: disable=too-many-public-methods
...
+        mockgrid = MockMicrogrid(grid_meter=False, mocker=mocker)
         mockgrid.add_chps(1)
         mockgrid.add_batteries(2)
-        await mockgrid.start(mocker)
 
-        logical_meter = microgrid.logical_meter()
-        chp_power_receiver = logical_meter.chp_power.new_receiver()
+        async with mockgrid, AsyncExitStack() as stack:
+            logical_meter = microgrid.logical_meter()
+            stack.push_async_callback(logical_meter.stop)
```
shsms, line 31: wow, this is very cool for python!

tests/timeseries/mock_microgrid.py
llucax: I assume I probably was trying to keep the same style already used in the class, as many other attributes are public, but there are private attributes too, so I will change it as this is not needed externally :+1:

tests/timeseries/test_logical_meter.py
llucax: I think I mentioned it in the past :)

.github/labeler.yml
llucax: Yup, that was completely unrelated and sneaked in, good catch!

llucax:
Updated!

.github/labeler.yml
shsms: so you've just removed the label stuff from here right?

shsms (approved):

.github/labeler.yml
llucax: Yup

PR #839, opened 2024-01-12
Some component status cleanup

shsms:
- Refactor component status tracking into its own package.
- Use `timedelta`s instead of `float`s for duration in component status trackers.
- Convert `ComponentStatusTracker`s into `BackgroundService`s


shsms:
Only internal improvements, no release notes necessary.

llucax (approved):
Awesome! I'm just curious about `Use a cached zero duration value, instead of repeatedly allocating`, did this showed up in some benchmark?

shsms:
> did this showed up in some benchmark?

no, but they do allocate in the hot path, which is many times slower than referencing a cached value.

llucax:
Yeah, creating objects is really slow. Not sure if there will be any caching going on when creating the same (immutable) object over and over though.

PR #840, opened 2024-01-12
Improve error handling when background services are not initialized

shsms:
Often during the early stages of development, people might just try to run pytest, and not pylint every time.

This makes it hard to identify when they forget to call `super().__init__()` when deriving from `BackgroundService`.

This PR improves error handling in `BackgroundService` to provide a more meaningful error and a solution to an otherwise hard to debug issue.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -72,6 +72,8 @@
 
 - The `actor.ChannelRegistry` is now type-aware.
 
+- The error messages when `BackgroundService` implementations don't call `super().__init__()` have been improved.
+
 ## Bug Fixes
 
 - 0W power requests are now not adjusted to exclusion bounds by the `PowerManager` and `PowerDistributor`, and are sent over to the microgrid API directly.
--- a/src/frequenz/sdk/actor/_background_service.py
+++ b/src/frequenz/sdk/actor/_background_service.py
@@ -90,6 +90,7 @@ class BackgroundService(abc.ABC):
         Returns:
             The name of this background service.
         """
+        self._assert_initialized()
         return self._name
 
     @property
@@ -107,6 +108,7 @@ class BackgroundService(abc.ABC):
         Returns:
             The set of running tasks spawned by this background service.
         """
+        self._assert_initialized()
         return self._tasks
 
     @property
@@ -118,6 +120,7 @@ class BackgroundService(abc.ABC):
         Returns:
             Whether this background service is running.
         """
+        self._assert_initialized()
         return any(not task.done() for task in self._tasks)
 
     def cancel(self, msg: str | None = None) -> None:
@@ -126,6 +129,7 @@ class BackgroundService(abc.ABC):
         Args:
             msg: The message to be passed to the tasks being cancelled.
         """
+        self._assert_initialized()
         for task in self._tasks:
             task.cancel(msg)
 
@@ -144,6 +148,7 @@ class BackgroundService(abc.ABC):
 
         [//]: # (# noqa: DAR401 rest)
         """
+        self._assert_initialized()
         if not self._tasks:
             return
         self.cancel(msg)
@@ -166,6 +171,7 @@ class BackgroundService(abc.ABC):
         Returns:
             This background service.
         """
+        self._assert_initialized()
         self.start()
         return self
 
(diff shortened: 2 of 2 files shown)


shsms:
<details>
<summary>Before</summary>

```
  File "/home/sahas/.local/lib/python3.11/site-packages/frequenz/sdk/actor/_background_service.py", line 237, in __del__
  File "/home/sahas/.local/lib/python3.11/site-packages/frequenz/sdk/actor/_background_service.py", line 129, in cancel
AttributeError: 'SocActor' object has no attribute '_tasks'
```

</details>

<details>
<summary>After</summary>

```
  File "/home/sahas/code/sdk/ev-power-distribution/src/frequenz/sdk/actor/_background_service.py", line 244, in __del__
  File "/home/sahas/code/sdk/ev-power-distribution/src/frequenz/sdk/actor/_background_service.py", line 132, in cancel
  File "/home/sahas/code/sdk/ev-power-distribution/src/frequenz/sdk/actor/_background_service.py", line 264, in _assert_initialized
AssertionError: SocActor instance is not initialized. Make sure to call `super().__init__()` in your __init__() method.
```

llucax:
Isn't the missing call to `super()`  detected by `flake8` or `pylint`?

shsms:
> Isn't the missing call to super() detected by flake8 or pylint?

did you read the description?

llucax:
> did you read the description?

Haha, yes, but I guess I missed the first line, sorry.

I'm not sure it is worth adding the runtime cost of checking if it was initialized everywhere just to get a better error message in early stages of development.

I would like more opinions on this one, @matthias-wende-frequenz @Marenz @daniel-zullo-frequenz ?

shsms:
> I'm not sure it is worth adding the runtime cost of checking if it was initialized everywhere just to get a better error message in early stages of development.

it is an assert that doesn't allocate and runs only when a background service is starting up or stopping, so it is not really a runtime cost.

but it makes development easy especially for external developers who might not always want to start with repo config, etc, but might just want to start coding,  and setup the tooling later, because that's what we've been showing in the demos as well.  

In such a scenario, it is just annoying, and people would be more inclined to think the SDK is broken because the error message is totally unhelpful.

Marenz:
> only when a background service is starting up or stopping, 

I mean that's not exactly correct, looking at the diff it's also running on `wait` and `is_running` and a few more. 
My feeling is that it's still okay as those methods are unlikely to be in a critical path and more used to start/stop things as well.

llucax:
Yeah, it is true that runtime cost shouldn't be an issue, and now that I think about it, that's not the real/main reason I'm concerned about this PR.

I can understand it might have been annoying to bump to that weird error once, but for me it feels a bit weird to add asserts to check for the same we are checking with linters, just because maybe one forgets to run the linter one time, it is also about polluting the code, specially if we start doing this more and more.

I mean, I'd be fine to merge this one if it was really hard to figure out what was going on and would have save you some time, but I don't think it makes sense as a general practice.

llucax:
In any case, this will probably go away when we fix #826, so I'm fine with merging it. I will still give it a bit more time before the approval to see if @matthias-wende-frequenz  or @daniel-zullo-frequenz also have any opinion on the topic.

shsms:
> just because maybe one forgets to run the linter one time

I think you're only considering frequenz users who would all use repo-config, etc.  It is not impossible to imagine external users of the sdk that don't have pylint as part of their workflow.

daniel-zullo-frequenz:
I agree the current patch is fine for this particular scenario but it shouldn't be considered as a good/general practice, otherwise all the attributes of the parent class would need to be checked as well.
And we should encourage non-frequenz users to use linters, the ones that can be configured in Editors/IDEs are already quite good.

shsms:
>  And we should encourage non-frequenz users to use linters, the ones that can be configured in Editors/IDEs are already quite good.

a lot of people just use flake8, and that doesn't detect this issue.  And pyright (the LSP server I think most people use with editors) doesn't detect them either.

daniel-zullo-frequenz:
hmm Pylint detected it in VS code if not calling `super().__init__(name=name)`

image

shsms:
Oh wow, if vscode does, then I'm sure other popular things like pycharm do as well.  Ok, maybe this was just a side effect of not using the popular tool.  Sorry about the noise. :-)

llucax:
Just for the records, `pyright` also detects it:

image

shsms:
interesting, I guess my pyright is just badly configured then.

PR #845, opened 2024-01-18
Drop old propsals from the PowerManager

shsms:
The PowerManager doesn't have a way to identify if an actor that sent a proposal is still active.  It needs to know that to know when to drop proposals from a given sender.

With this PR, the resending of proposals by actors also functions as a heart-beat between the actors and the power manager.  If an actor hasn't sent a new proposal in 60 seconds,  the latest available proposal from that actor is also removed from the power manager.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -85,3 +85,5 @@
 - The resampler now properly handles sending zero values.
 
   A bug made the resampler interpret zero values as `None` when generating new samples, so if the result of the resampling is zero, the resampler would just produce `None` values.
+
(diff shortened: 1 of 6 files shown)


shsms:
@daniel-zullo-frequenz I think this needs to be done before we can deploy the power manager.

daniel-zullo-frequenz (dismissed):

shsms:
Fixed a small bug that would have triggered when using multiple battery pools with different batteries in the same location.

https://github.com/frequenz-floss/frequenz-sdk-python/compare/de2f993594b29364512e9d564b08acf64a64b7ad..ce586b08e614577cb2d9a2e5557c0b3abcb3a609

daniel-zullo-frequenz (approved):

PR #846, opened 2024-01-19
Use a sentinel in `LatestValueCache` to denote if the cache is empty

### LINKED ISSUE #820: `LatestValueCache` can't use `None` as a valid value
### What happened?

`LatestValueCache` is using `None` internally to represent the lack of a value, but `None` is also a valid value.

`LatestValueCache` to use a sentinel to represent a missing value instead.

### Affected version(s)

_No response_

### Affected part(s)

### Extra information

Introduced here:

* https://github.com/frequenz-floss/frequenz-sdk-python/pull/800#discussion_r1410755961

shsms:
This is necessary because `None` is a valid value that could have been
sent through a channel, so the having a `None` value might not always
mean that the cache is empty.

Closes https://github.com/frequenz-floss/frequenz-sdk-python/issues/820

### DIFF
--- a/src/frequenz/sdk/_internal/_channels.py
+++ b/src/frequenz/sdk/_internal/_channels.py
@@ -27,6 +27,10 @@ class ReceiverFetcher(typing.Generic[T], typing.Protocol):
         """
 
 
+class _Sentinel:
+    """A sentinel to denote that no value has been received yet."""
+
+
 class LatestValueCache(typing.Generic[T]):
     """A cache that stores the latest value in a receiver."""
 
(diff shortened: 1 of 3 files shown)


shsms:
This is an internal fix for something that was introduced in this release cycle, so no release notes necessary.

src/frequenz/sdk/_internal/_channels.py
```diff
@@ -37,14 +41,35 @@ def __init__(self, receiver: Receiver[T]) -> None:
...
+        to avoid the exception.
+
+        Returns:
+            The latest value that has been received.
+
+        Raises:
+            ValueError: If no value has been received yet.
+        """
+        if isinstance(self._latest_value, _Sentinel):
+            raise ValueError("No value has been received yet.")
```
daniel-zullo-frequenz, line 61: I personally found it weird that a property raises. Maybe it should no longer be a property?

daniel-zullo-frequenz (commented):
I've only have a comment to check for. LGTM otherwise

src/frequenz/sdk/_internal/_channels.py
shsms: That's true,  I've changed it to the `get` method.  That seems to go well with the other new `has_value` method as well.

daniel-zullo-frequenz: cool, thanks! It does not need an entry into the release-notes because it is internal, right?

daniel-zullo-frequenz (approved):

src/frequenz/sdk/_internal/_channels.py
shsms: yup, it is not exposed outside at the moment.

PR #848, opened 2024-01-23
Improvements to EVChargerData coming from the microgrid API

shsms:
- A new method `is_ev_connected` to analyze the states better
- New fields to represent active power bounds for ev chargers


shsms:
No release notes, because the changes are in the low level microgrid api client, which is not part of the public interface.

daniel-zullo-frequenz (approved):

PR #855, opened 2024-01-30
Bump frequenz-api-microgrid to 0.15.3

llucax:
This is only to indirectly bump the api-common dependency to 0.5.x so it is compatible with other APIs.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -108,3 +108,7 @@
 - The PowerManager no longer holds on to proposals from dead actors forever.  If an actor hasn't sent a new proposal in 60 seconds, the available proposal from that actor is dropped.
 
 - Fix `Quantity.__format__()` sometimes skipping the number for very small values.
+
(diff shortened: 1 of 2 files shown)


llucax:
The test package installation CI check fails because we need a release of common API v0.5.3 and microgrid API v0.15.3 first for it to pass.

llucax:
And this needs the dependencies updated when common API v0.5.3 and microgrid API v0.15.3 are released.

llucax:
This PR is actually not needed, as the SDK already depends on `frequenz-api-microgrid >= 0.15.1`, so downstream projects can upgrade to `frequenz-api-microgrid == 0.15.3` without any SDK updates!

llucax:
On a second thought, since most downstream projects (at least API actors) won't depend explicitly on `frequenz-api-microgrid`, the only sane way to specify the dependencies to make sure there are no conflicts would be to use a SDK that has v0.15.3 as the minimum microgrid API version.

So downstream projects don't **need** to update the SDK version, but if they don't they could potentially end up with cryptic errors about `pip` not being able to find a set of dependencies that work well together.

Marenz (approved):

PR #874, opened 2024-02-09
Make sure `base_value` is always `float`

llucax:
In Python `float` is a bit weird, because `isinstance(1, float)` is `False` but still `int` is a subtype of `float` in terms of the type system, so `int` will be accepted when a `float` is specified in arguments for example.
    
This leads to quantities that can have `int` as base value if it was constructed with a `int` instead of a `float`, which is unintuitive when the `base_value` type hint is `float` and can potentially cause problems, in particular when addressing #821.

### DIFF
--- a/tests/timeseries/test_quantities.py
+++ b/tests/timeseries/test_quantities.py
@@ -3,12 +3,15 @@
 
 """Tests for quantity types."""
 
+import inspect
 from datetime import timedelta
+from typing import Callable
 
 import hypothesis
 import pytest
 from hypothesis import strategies as st
 
+from frequenz.sdk.timeseries import _quantities
 from frequenz.sdk.timeseries._quantities import (
     Current,
     Energy,
@@ -43,6 +46,48 @@ class Fz2(
     },
 ):
     """Frequency quantity with broad exponent unit map."""
+
+
+_CtorType = Callable[[float], Quantity]
+
+# Thi is the current number of subclasses. This probably will get outdated, but it will
+# provide at least some safety against something going really wrong and end up testing
+# an empty list. With this we should at least make sure we are not testint less classes
+# than before. We don't get the actual number using len(_QUANTITY_SUBCLASSES) because it
+# would defeat the purpose of the test.
+_SANITFY_NUM_CLASSES = 7
+
+_QUANTITY_SUBCLASSES = [
+    cls
+    for _, cls in inspect.getmembers(
+        _quantities,
+        lambda m: inspect.isclass(m) and issubclass(m, Quantity) and m is not Quantity,
+    )
+]
+
+# A very basic sainty check that we are messing up the introspection
+assert len(_QUANTITY_SUBCLASSES) >= _SANITFY_NUM_CLASSES
+
+_QUANTITY_BASE_UNIT_STRINGS = [
+    cls._new(0).base_unit  # pylint: disable=protected-access
+    for cls in _QUANTITY_SUBCLASSES
+]
+for unit in _QUANTITY_BASE_UNIT_STRINGS:
+    assert unit is not None
+
+_QUANTITY_CTORS = [
+    method
+    for cls in _QUANTITY_SUBCLASSES
+    for _, method in inspect.getmembers(
+        cls,
+        lambda m: inspect.ismethod(m)
+        and m.__name__.startswith("from_")
+        and m.__name__ != ("from_string"),
+    )
+]
+# A very basic sainty check that we are messing up the introspection. There are actually
+# many more constructors than classes, but this still works as a very basic check.
+assert len(_QUANTITY_CTORS) >= _SANITFY_NUM_CLASSES
 
 
 def test_zero() -> None:
@@ -99,6 +144,31 @@ def test_zero() -> None:
     assert Percentage.zero().as_percent() == 0.0
     assert Percentage.zero().as_fraction() == 0.0
     assert Percentage.zero() is Percentage.zero()  # It is a "singleton"
+
+
+@pytest.mark.parametrize("quantity_ctor", _QUANTITY_CTORS)
+def test_base_value_from_ctor_is_float(quantity_ctor: _CtorType) -> None:
+    """Test that the base value always is a float."""
+    quantity = quantity_ctor(1)
+    assert isinstance(quantity.base_value, float)
+
+
+@pytest.mark.parametrize("quantity_type", _QUANTITY_SUBCLASSES + [Quantity])
+def test_base_value_from_zero_is_float(quantity_type: type[Quantity]) -> None:
+    """Test that the base value always is a float."""
+    quantity = quantity_type.zero()
+    assert isinstance(quantity.base_value, float)
+
+
+@pytest.mark.parametrize(
+    "quantity_type, unit", zip(_QUANTITY_SUBCLASSES, _QUANTITY_BASE_UNIT_STRINGS)
+)
+def test_base_value_from_string_is_float(
+    quantity_type: type[Quantity], unit: str
+) -> None:
+    """Test that the base value always is a float."""
+    quantity = quantity_type.from_string(f"1 {unit}")
+    assert isinstance(quantity.base_value, float)
 
 
 def test_string_representation() -> None:
--- a/tests/timeseries/test_quantities.py
+++ b/tests/timeseries/test_quantities.py
@@ -50,9 +50,9 @@ class Fz2(
 
 _CtorType = Callable[[float], Quantity]
 
-# Thi is the current number of subclasses. This probably will get outdated, but it will
+# This is the current number of subclasses. This probably will get outdated, but it will
 # provide at least some safety against something going really wrong and end up testing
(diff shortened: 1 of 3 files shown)


tests/timeseries/test_quantities.py
```diff
@@ -563,8 +626,14 @@ def test_to_and_from_string(
     generation and regenerate it with the more appropriate unit and precision.
     """
     quantity = quantity_type.__new__(quantity_type)
-    # pylint: disable=protected-access
-    quantity._base_value = value * 10**exponent
+    quantity._base_value = value * 10**exponent  # pylint: disable=protected-access
+    # The above should be replaced with:
+    # quantity = quantity_type._new(  # pylint: disable=protected-access
+    #     value, exponent=10**exponent
```
daniel-zullo-frequenz: I think here we only need to pass the exponent as the method `_new()` will do the math `10.0**exponent`
```suggestion
    #     value, exponent=exponent
```

tests/timeseries/test_quantities.py
```diff
@@ -45,6 +48,38 @@ class Fz2(
...
+
+_QUANTITY_SUBCLASSES = [
+    cls
+    for _, cls in inspect.getmembers(
+        _quantities,
+        lambda m: inspect.isclass(m) and issubclass(m, Quantity) and m is not Quantity,
+    )
+]
+# Make a simple sanity check, 7 is the current number of subclasses
+assert len(_QUANTITY_SUBCLASSES) >= 7
```
daniel-zullo-frequenz: you might want to define a variable like `NUM_SUBCLASSES=7` so that if new subclasses are added in the future we will only need to modify the variable. We have the comparison to number of subclasses twice, so also feel free to discard this comment if it is not worth the effort

daniel-zullo-frequenz (commented):
I have a couple of minor comments to check for, LGTM otherwise

tests/timeseries/test_quantities.py
llucax: Yeah, I could do that, but this is only a very basic sanity check, and I don't think anyone will remember to update this value once a new quantity is added, this is why I didn't make it "prettier". But I should add a comment about this.

llucax:
Updated to address the comments.

tests/timeseries/test_quantities.py
```diff
@@ -45,6 +48,48 @@ class Fz2(
     """Frequency quantity with broad exponent unit map."""
 
 
+_CtorType = Callable[[float], Quantity]
+
+# Thi is the current number of subclasses. This probably will get outdated, but it will
+# provide at least some safety against something going really wrong and end up testing
+# an empty list. With this we should at least make sure we are not testint less classes
+# than before. We don't get the actual number using len(_QUANTITY_SUBCLASSES) because it
+# would defeat the purpose of the test.
```
daniel-zullo-frequenz:
```suggestion
# This is the current number of subclasses. This probably will get outdated, but it will
# provide at least some safety against something going really wrong and end up testing
# an empty list. With this we should at least make sure we are not testing less classes
# than before. We don't get the actual number using len(_QUANTITY_SUBCLASSES) because it
# would defeat the purpose of the test.
```

tests/timeseries/test_quantities.py
```diff
@@ -45,6 +48,48 @@ class Fz2(
...
+
+_QUANTITY_SUBCLASSES = [
+    cls
+    for _, cls in inspect.getmembers(
+        _quantities,
+        lambda m: inspect.isclass(m) and issubclass(m, Quantity) and m is not Quantity,
+    )
+]
+
+# A very basic sainty check that we are messing up the introspection
```
daniel-zullo-frequenz:
```suggestion
# A very basic sanity check that are messing up the introspection
```

tests/timeseries/test_quantities.py
```diff
@@ -45,6 +48,48 @@ class Fz2(
...
+    for cls in _QUANTITY_SUBCLASSES
+    for _, method in inspect.getmembers(
+        cls,
+        lambda m: inspect.ismethod(m)
+        and m.__name__.startswith("from_")
+        and m.__name__ != ("from_string"),
+    )
+]
+# A very basic sainty check that we are messing up the introspection. There are actually
+# many more constructors than classes, but this still works as a very basic check.
```
daniel-zullo-frequenz:
```suggestion
# A very basic sanity check that are messing up the introspection. There are actually
# many more constructors than classes, but this still works as a very basic check.
```

tests/timeseries/test_quantities.py
daniel-zullo-frequenz: Thanks!

daniel-zullo-frequenz (commented):
A few typos and LGTM

llucax:
Haha, you can clearly see I wrote that in a rush :P

Updated.

llucax:
Enabled auto-merge.

llucax:
Weird that DCO is failing, the only thing I can think of is it doesn't like the `Co-authored-by:`, but I'm pretty sure we used it in the past...

llucax:
I removed it and it doesn't work either...

llucax:
Oh, that's weird, it didn't like I was using a different e-mail (even when it is associated to my GitHub account too).

daniel-zullo-frequenz (approved):

tests/timeseries/test_quantities.py
```diff
@@ -45,6 +48,48 @@ class Fz2(
...
 
 
+_CtorType = Callable[[float], Quantity]
+
+# This is the current number of subclasses. This probably will get outdated, but it will
+# provide at least some safety against something going really wrong and end up testing
+# an empty list. With this we should at least make sure we are not testing less classes
+# than before. We don't get the actual number using len(_QUANTITY_SUBCLASSES) because it
+# would defeat the purpose of the test.
+_SANITFY_NUM_CLASSES = 7
```
Marenz, line 58: A bit late to the review, but it might be better to call this `_SANITIFY_NUM_CLASSES`.. I just don't see a reason to skip this extra I...

tests/timeseries/test_quantities.py
```diff
@@ -45,6 +48,48 @@ class Fz2(
...
+
+_QUANTITY_SUBCLASSES = [
+    cls
+    for _, cls in inspect.getmembers(
+        _quantities,
+        lambda m: inspect.isclass(m) and issubclass(m, Quantity) and m is not Quantity,
+    )
+]
+
+# A very basic sanity check that are messing up the introspection
```
Marenz, line 68: `# A very basic sanity check that are messing up the introspection`

Someone def. messed up this comment :)

tests/timeseries/test_quantities.py
Marenz: just try to pronounce it with and without :P

tests/timeseries/test_quantities.py
```diff
@@ -45,6 +48,48 @@ class Fz2(
...
+assert len(_QUANTITY_SUBCLASSES) >= _SANITFY_NUM_CLASSES
+
+_QUANTITY_BASE_UNIT_STRINGS = [
+    cls._new(0).base_unit  # pylint: disable=protected-access
+    for cls in _QUANTITY_SUBCLASSES
+]
+for unit in _QUANTITY_BASE_UNIT_STRINGS:
+    assert unit is not None
+
+_QUANTITY_CTORS = [
```
shsms, line 78: So close, we could have called these `QTY_CTORS` and pronounced them the "cutty cutters"

tests/timeseries/test_quantities.py
shsms: I agree completely.  When making typos, please make sure you don't skimp on the vowels. :D

tests/timeseries/test_quantities.py
```diff
@@ -633,8 +636,14 @@ def test_to_and_from_string(
...
-    # pylint: disable=protected-access
-    quantity._base_value = value * 10**exponent
+    quantity._base_value = value * 10**exponent  # pylint: disable=protected-access
+    # The above should be replaced with:
+    # quantity = quantity_type._new(  # pylint: disable=protected-access
+    #     value, exponent=exponent
+    # )
+    # But we can't do that now, because, you guessed it, it will also break the tests
+    # (_new() will use 10.0**exponent instead of 10**exponent, which seems to have some
+    # effect on the tests.
```
shsms, line 646: By "on the tests", I guess you mean just this test, because hypothesis also generates some int inputs when we give it a float range?

This function seems to just compare the value from `quantity_type.from_string` with the value from `quantity._base_value`.  If they don't match up, maybe the `from_string` method just needs to convert to float as well?

And then they can be compared with `quantity.is_close` rather than a string comparison?

tests/timeseries/test_quantities.py
llucax: I have no idea how I ended up typing this. But luckily we have smart completion, so typos are properly propagated and the code never fails!

tests/timeseries/test_quantities.py
llucax: I think I made my first meme-PR!

PR #875, opened 2024-02-13
Allow multiplying `Quantity` by `float`

llucax:
This is only applying a factor so it should be safe for any Quantity.

This PR also improve multiplication tests to use hypothesis and test with all constructors.

PR #876, opened 2024-02-13
Allow all quantities division by `float | Self`

llucax:
Dividing by `float` is only applying a factor so it should be safe for any `Quantity`.

Dividing by `Self` produces a unit-less ration (`float`), so it should be safe too.

To be able to use nicer names in overloads, we use positional-only arguments for `__truediv__()`, which in practice shouldn't be much of a breaking change as these methods should be used mostly via operators.

We also use `match` syntax in the changed methods.

### DIFF
--- a/src/frequenz/sdk/timeseries/_quantities.py
+++ b/src/frequenz/sdk/timeseries/_quantities.py
@@ -373,6 +373,45 @@ class Quantity:
             case _:
                 return NotImplemented
 
+    @overload
+    def __truediv__(self, other: float, /) -> Self:
+        """Divide this quantity by a scalar.
+
+        Args:
+            other: The scalar or percentage to divide this quantity by.
+
+        Returns:
+            The divided quantity.
+        """
+
+    @overload
+    def __truediv__(self, other: Self, /) -> float:
+        """Return the ratio of this quantity to another.
+
+        Args:
+            other: The other quantity.
+
+        Returns:
+            The ratio of this quantity to another.
+        """
+
+    def __truediv__(self, value: float | Self, /) -> Self | float:
+        """Divide this quantity by a scalar or another quantity.
+
+        Args:
+            value: The scalar or quantity to divide this quantity by.
+
+        Returns:
+            The divided quantity or the ratio of this quantity to another.
+        """
+        match value:
+            case float():
+                return type(self)._new(self._base_value / value)
+            case Quantity() if type(value) is type(self):
+                return self._base_value / value._base_value
+            case _:
+                return NotImplemented
+
     def __gt__(self, other: Self) -> bool:
         """Return whether this quantity is greater than another.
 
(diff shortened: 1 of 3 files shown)


llucax:
Based on:

* #875.

tests/timeseries/test_quantities.py
```diff
@@ -623,8 +626,14 @@ def test_to_and_from_string(
...
-    # pylint: disable=protected-access
-    quantity._base_value = value * 10**exponent
+    quantity._base_value = value * 10**exponent  # pylint: disable=protected-access
+    # The above should be replaced with:
+    # quantity = quantity_type._new(  # pylint: disable=protected-access
+    #     value, exponent=10**exponent
+    # )
+    # But we can't do that now, because, you guessed it, it will also break the tests
+    # (_new() will use 10.0**exponent instead of 10**exponent, which seems to have some
+    # effect on the tests.
```
Marenz: :cry:

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -334,7 +334,19 @@ def __sub__(self, other: Self) -> Self:
         difference._base_value = self._base_value - other._base_value
         return difference
 
-    def __mul__(self, percent: Percentage) -> Self:
+    @overload
+    def __mul__(self, scalar: float, /) -> Self:
+        """Scale this quantity by a scalar.
+
+        Args:
+            scalar: The scaler by which to scale this quantity.
```
Marenz: did you mean to write `scaler` instead of `scalar` here?

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -584,18 +607,29 @@ def as_megawatts(self) -> float:
         return self._base_value / 1e6
 
     @overload  # type: ignore
-    def __mul__(self, other: Percentage) -> Self:
+    def __mul__(self, scalar: float, /) -> Self:
+        """Scale this power by a scalar.
+
+        Args:
+            scalar: The scaler by which to scale this power.
```
Marenz: scalar

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -726,18 +759,29 @@ def as_milliamperes(self) -> float:
         return self._base_value * 1e3
 
     @overload  # type: ignore
-    def __mul__(self, other: Percentage) -> Self:
+    def __mul__(self, scalar: float, /) -> Self:
+        """Scale this current by a scalar.
+
+        Args:
+            scalar: The scaler by which to scale this current.
```
Marenz: scalar

src/frequenz/sdk/timeseries/_quantities.py
```diff
@@ -841,18 +886,29 @@ def as_kilovolts(self) -> float:
         return self._base_value / 1e3
 
     @overload  # type: ignore
-    def __mul__(self, other: Percentage) -> Self:
+    def __mul__(self, scalar: float, /) -> Self:
+        """Scale this voltage by a scalar.
+
+        Args:
+            scalar: The scaler by which to scale this voltage.
```
Marenz: scalar

Marenz (commented):
See comments

llucax:
> See comments

Isn't this a paradox?

tests/timeseries/test_quantities.py
llucax: Yeah, those tests are some real motherf*cker! We should fix this eventually or remove the use of hypothesis and just test the cases we know work correctly manually.

llucax:
Thanks for the review, I will update once #875 is merged.

llucax:
Rebased and marked the PR ready for review. You comments should be already addressed.

llucax:
Enabling auto-merge as this had a previous review already.

tests/timeseries/test_quantities.py
```diff
@@ -641,6 +641,247 @@ def test_invalid_multiplications() -> None:
...
+    expected_value = quantity.base_value / scalar
+    print(f"{quantity=}, {expected_value=}")
+
+    product = quantity / scalar
+    print(f"{product=}")
+    assert product.base_value == expected_value
+
+    quantity /= scalar
+    print(f"*{quantity=}")
+    assert quantity.base_value == expected_value
```
shsms: looks like we're missing tests for "valid" division by same quantity type?

llucax: Right! Good catch, I will add it.

llucax: Oh, crap, this actually uncovered a (design) bug, good catch: `percentage / percentage` now have 2 possible types, `float` for `Self` and `Self` for `Percentage` :unamused: 

So there is no way to override the implementation in `Quantity` in a compatible way for `Percentage`. I guess `Percentage` being so special maybe shouldn't inherit from `Quantity`, but not sure. This is definitely out of scope for this PR, so I guess I will remove `Percentage` as a possible type for division.

Besides being an obscure use case, users can still do `quantity /= percentage.as_fraction()` if they really wish to divide by a percentage...

llucax:
Updated to remove division by `Percentage`.

shsms (commented):
There are a few occurances of "percentage" in docstrings that need to be removed, but looks good otherwise.

llucax:
Updated to hopefully remove all the remaining mentions to percentage and rebased.

llucax:
ping @shsms

shsms (approved):

PR #878, opened 2024-02-14
Remove unnecessary send adapter

llucax:
In previous versions of the `channels` library, the `send()` method returned `bool` to indicate if the send was succesful or not. Recent versions use exception to report errors and return `None`, making the `send()` method directly compatible with what the resampler expects, so we don't need an adapter any more.


llucax:
Enabled auto-merge.

daniel-zullo-frequenz (approved):

PR #879, opened 2024-02-14
Mock resampler: Improve variable names

llucax:
Since the mock resampler uses a pair of channels, one to receive samples to be (fakely) resampled (input) and one to send the (fakely) resampmled samples (output), we should use clear names to identify both otherwise is really hard to follow which is which when reading the code.

### DIFF
--- a/tests/timeseries/mock_resampler.py
+++ b/tests/timeseries/mock_resampler.py
@@ -45,7 +45,7 @@ class MockResampler:
         self._resampler_request_channel = Broadcast[ComponentMetricRequest](
             "resampler-request"
         )
-        self._basic_receivers: dict[str, list[Receiver[Sample[Quantity]]]] = {}
+        self._input_channels_receivers: dict[str, list[Receiver[Sample[Quantity]]]] = {}
 
         def power_senders(
             comp_ids: list[int],
@@ -58,7 +58,7 @@ class MockResampler:
                         Sample[Quantity], name
                     ).new_sender()
                 )
-                self._basic_receivers[name] = [
+                self._input_channels_receivers[name] = [
                     self._channel_registry.get_or_create(
                         Sample[Quantity], name
                     ).new_receiver()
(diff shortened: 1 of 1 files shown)


llucax:
Enabled auto-merge.

Marenz (approved):

tests/timeseries/mock_resampler.py
```diff
@@ -222,15 +222,18 @@ async def _handle_resampling_requests(self) -> None:
...
-            recv = self._basic_receivers[basic_recv_name].pop()
-            assert recv is not None
+            input_chan_recv_name = f"{request.component_id}:{request.metric_id}"
+            input_chan_recv = self._input_channels_receivers[input_chan_recv_name].pop()
+            assert input_chan_recv is not None
+            output_chan_sender: Sender[Sample[Quantity]] = (
+                self._channel_registry.get_or_create(
+                    Sample[Quantity], name
+                ).new_sender()
+            )
```
shsms, line 232: Interesting, this is better than `basic_receivers`, but `input_receivers` and `output_sender` might have even nicer.  `input_channels_receivers` make me think that they receive channels and not input values.

llucax: Yeah, at first I named it without the `chan`, but for some reason it still seemed ambiguous, but now I don't remember why. Maybe @daniel-zullo-frequenz as he was helping me to understand the code while I did it.

daniel-zullo-frequenz: IIRC we wanted to mainly get rid of `basic` in the variable name because it was confusing/difficult-to-understand what a `basic_receiver` was. I do agree having only `receivers` and `sender` in the name would have suffice to understand they are channels given the context. However I don't think it causes any misunderstanding if we keep `chan/channels` explicitly in the variable names. I mean I haven't seen yet any scenarios where we send/receive channels

PR #885, opened 2024-02-26
Update minimum versions for many dependencies

shsms:
This can be merged after v1-rc5, so that we'll have some more time to look for issues.

| dependency               | from   | to      |
|--------------------------|--------|---------|
| google-api-python-client | 2.71   | 2.119.0 |
| grpcio, grpcio-tools     | 1.54.2 | 1.62.0  |
| protobuf                 | 4.21.6 | 4.25.3  |
| networkx                 | 2.8    | 3.2.1   |
| numpy                    | 1.24.2 | 1.26.4  |
| pydantic                 | 2.3    | 2.6.2   |
| timezonefinder           | 6.2.0  | 6.4.1   |
| typing_extensions        | 4.6.1  | 4.10.0  |
| watchfiles               | 0.15.0 | 0.21.0  |

### DIFF
--- a/benchmarks/timeseries/benchmark_ringbuffer.py
+++ b/benchmarks/timeseries/benchmark_ringbuffer.py
@@ -174,14 +174,14 @@ def main() -> None:
         "Time to fill 29 days with data:\n\t"
         + f"Array: {array_times['fill']} seconds\n\t"
         + f"List:  {list_times['fill']} seconds\n\t"
-        + f"Diff:  {array_times['fill'] -  list_times['fill']}"
+        + f"Diff:  {array_times['fill'] - list_times['fill']}"
     )
 
     print(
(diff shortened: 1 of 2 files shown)


shsms:
Will add release notes after rc5 has been released and the release notes cleared.

llucax:
Why are you bumping minimum dependencies? It is usually a bad idea to do so unless it is strictly necessary, otherwise it is just a breaking change to downstream projects without any real gain. For libraries we should keep supported version ranges as wide as possible.

shsms:
For some of them, because it was not running under python 3.12 otherwise.  But maybe you're right.  I'll just do local hacks for now.

llucax:
OK, yeah, starting to test against Python 3.12 is on my TODO list, but I guess a bit down in the priority list. I think it makes sense to delay this also until then :+1:

PR #892, opened 2024-03-07
Replace the internal microgrid API client with `frequenz-client-microgrid`

### LINKED ISSUE #853: Split out the `microgrid.client` to its own repository
### What's needed?

All API clients are now being created in their own repositories. We need to do the same with the microgrid API, which now lives in the SDK.

### Proposed solution

Split the `frequenz.sdk.microgrid.client` package to its own repository. But we also have a [pending version update for the microgrid API](https://github.com/frequenz-floss/frequenz-sdk-python/issues/844) and a new [repository with helper function](https://github.com/frequenz-floss/frequenz-client-base-python/) to write clients more easily.

- [x] Create the new `frequenz.client.microgrid` repository and move the current SDK code there.
- [x] Release a version using the same code as the SDK
- [x] Remove the code from the SDK and depend on the new repo

### Use cases

_No response_

_No response_

### Additional context

This will also help to improve the very complicated, multi-dimensional, problem we have involving dependency conflict of the SDK depending on the microgrid API v0.15 (and common v0.4) and the new API clients depending on common v0.5.

llucax:
Unfortunately we can't remove the direct dependency on `frequenz-api-microgrid` yet because some code is still using the raw protobuf structures, as there are no wrappers for them yet (neither in the internal code nor in `frequenz-client-microgrid`, which is based on the SDK code).

Fixes #853.


shsms (approved):

shsms:
This looks good, I will take it in now, so that I can make progress on the channel update.

PR #897, opened 2024-03-15
Add an `EVChargerStatusTracker` implementation

shsms:
It reports an EV charger as `WORKING` when an EV is connected to it,
and power can be allocated to it, and `NOT_WORKING` otherwise.

If it receives a power assignment failure from the PowerDistributor,
when the component is expected to be working, it is marked as
`UNCERTAIN` for a specific interval, before being marked `WORKING`
again.

This status would be used in the PowerDistributor and in the
EVChargerPool.


shsms:
This is for internal use, so no release notes necessary.

shsms:
superseded by https://github.com/frequenz-floss/frequenz-sdk-python/pull/900

PR #898, opened 2024-03-19
Bug fixes in the PowerManager and the BatteryPool

shsms:


src/frequenz/sdk/actor/_power_managing/_base_classes.py
```diff
@@ -231,6 +231,27 @@ def __lt__(self, other: Proposal) -> bool:
             self.priority == other.priority and self.source_id < other.source_id
         )
 
+    def __eq__(self, other: object) -> bool:
+        """Check if two proposals are equal.
```
cwasicki, line 235: Even though I think I get the intention of this, to me the definition of equality does not feel intuitive.

src/frequenz/sdk/actor/_power_managing/_matryoshka.py
```diff
@@ -44,12 +44,12 @@ class Matryoshka(BaseAlgorithm):
     def __init__(self, max_proposal_age: timedelta) -> None:
         """Create a new instance of the matryoshka algorithm."""
         self._max_proposal_age_sec = max_proposal_age.total_seconds()
-        self._component_buckets: dict[frozenset[int], SortedSet[Proposal]] = {}
+        self._component_buckets: dict[frozenset[int], deque[Proposal]] = {}
```
cwasicki: It's not clear to me how you can replace a sorted set with a deque? I see you sort it below, but why not use a set here?

shsms: The original `SortedSet` didn't need a `__hash__` method.  But if we add a custom one that hashes just the source_id and priority, then we are able to use a `set`.  I've made that change.

src/frequenz/sdk/actor/_power_managing/_base_classes.py
shsms: The `__lt__`, `__eq__` and `__hash__` methods are all defined based on just priority and source ID, because they are there to allow sorting by priority+source_id, and to remove elements/replace them based on the same key.

`Proposal` is an internal type, because users would use the `propose_charge`, etc. methods.

For the `SortedSet`, we needed this definition of `__lt__`, and ff we were using a `deque`, we'd need this definition of `__eq__`.  Now that we are using `set`, we need this definition of `__hash__`, but maybe the `__lt__` can be replaced by a lambda passed to the `sorted` function.  and the `__eq__` can go away.

I will try this now.

shsms: I tried a quick test like this and removed `__lt__` and `__eq__`, but it doesn't seem to be the same.  The results are different and the tests don't pass.

```python
        for next_proposal in sorted(
            proposals,
            reverse=True,
            key=lambda proposal: (proposal.priority, proposal.source_id),
        ):
```

cwasicki (approved):

matthias-wende-frequenz (approved):

PR #900, opened 2024-03-21
Control methods for `EVChargerPool` using `PowerManager` and `PowerDistributor`

### LINKED ISSUE #872: EV Charger Pool
Update Power Distributor, Power Manager to support EV Chargers, so that EV Charger Pool can have a consistent interface as the Battery Pool, with all control commands going through the Power Manager and Power Distributor.

shsms:
This PR adds new `propose_power` and `power_status` methods to the `EVChargerPool` similar to the `BatteryPool`.  These method interface with the `PowerManager` and `PowerDistributor`, which currently uses a first-come-first-serve algorithm to distribute power to EVs.

The following changes were made:
- An `EVChargerStatusTracker` implementation which tracks the health of an EV charger and whether an EV is connected.
- `PowerDistributor` and `PowerManager` are now modular to support multiple component types.
- Add an `EVChargerManager` that plugs into the `PowerDistributor`
- Connect the `PowerDistributor` and `PowerManager` to the `EVChargerPool`.

Fixes #872.


src/frequenz/sdk/actor/power_distributing/_component_status/_ev_charger_status_tracker.py
```diff
@@ -0,0 +1,204 @@
...
+    async def _run(self) -> None:
+        """Run the status tracker."""
+        api_client = connection_manager.get().api_client
+        ev_data_rx = await api_client.ev_charger_data(self._component_id)
+        set_power_result_rx = self._set_power_result_receiver
+        missing_data_timer = Timer(self._max_data_age, SkipMissedAndDrift())
+
+        # Send initial status
+        await self._status_sender.send(
+            ComponentStatus(self._component_id, self._last_status)
```
Marenz, line 168: Hmm this means we will always send the default NOT_WORKING initially before correcting it no? Why not make default `None` and don't send any initial status, just let the loop update it?

Marenz: Worst case, the timer will fire and send the update

tests/actor/power_distributing/_component_status/test_ev_charger_status.py
```diff
@@ -0,0 +1,165 @@
...
+from ....utils.component_data_wrapper import EvChargerDataWrapper
+from ....utils.receive_timeout import Timeout, receive_timeout
+
+_EV_CHARGER_ID = 6
+
+
+class TestEVChargerStatusTracker:
+    """Tests for EVChargerStatusTracker."""
+
+    async def test_status_changes(self, mocker: MockerFixture) -> None:
```
Marenz, line 31: Why are you not using solipsism (and time_machine for fake_time if required)  to prevent real waiting times here, however small they may be?

src/frequenz/sdk/actor/power_distributing/_component_managers/_battery_manager.py
```diff
@@ -127,8 +127,10 @@ class BatteryManager(ComponentManager):
     def __init__(
         self,
         component_pool_status_sender: Sender[ComponentPoolStatus],
+        results_sender: Sender[Result],
```
Marenz, line 130: No parameter descriptions at all? :o
Why does our CI even allow this :)

Marenz: Ah, because it's in the abstract class?

src/frequenz/sdk/actor/power_distributing/_component_managers/_ev_charger_manager/_ev_charger_manager.py
```diff
@@ -0,0 +1,448 @@
...
+        """Distribute the requested power to the ev chargers.
+
+        Args:
+            request: Request to get the distribution for.
+        """
+        await self._target_power_tx.send(request)
+
+    @override
+    async def stop(self) -> None:
+        """Stop the ev charger data manager."""
```
Marenz: There is no stopping the data manager?

src/frequenz/sdk/actor/power_distributing/_component_managers/_ev_charger_manager/_ev_charger_manager.py
```diff
@@ -0,0 +1,448 @@
...
+
+    def _get_ev_charger_ids(self) -> collections.abc.Set[int]:
+        return {
+            evc.component_id
+            for evc in microgrid.connection_manager.get().component_graph.components(
+                component_categories={ComponentCategory.EV_CHARGER}
+            )
+        }
+
+    def _allocate_new_ev(self, component_id: int) -> dict[int, Power]:
```
Marenz, line 107: No method documentation?

src/frequenz/sdk/actor/power_distributing/_component_managers/_ev_charger_manager/_ev_charger_manager.py
```diff
@@ -0,0 +1,448 @@
...
+        Args:
+            request: Request to get the distribution for.
+        """
+        await self._target_power_tx.send(request)
+
+    @override
+    async def stop(self) -> None:
+        """Stop the ev charger data manager."""
+
+    def _get_ev_charger_ids(self) -> collections.abc.Set[int]:
```
Marenz, line 98: No method documentation?

src/frequenz/sdk/actor/power_distributing/_component_managers/_ev_charger_manager/_ev_charger_manager.py
```diff
@@ -0,0 +1,448 @@
...
+        if available_power > initial_power:
+            return {component_id: initial_power}
+
+        min_power = voltage * self._config.min_current * 3.0
+        if available_power > min_power:
+            return {component_id: min_power}
+
+        return {}
+
+    def _act_on_new_data(self, ev_data: EVChargerData) -> dict[int, Power]:
```
Marenz, line 136: No method documentation?

src/frequenz/sdk/actor/power_distributing/_component_managers/_ev_charger_manager/_ev_charger_manager.py
```diff
@@ -0,0 +1,448 @@
...
+                        power, now
+                    )
+
+                latest_target_powers.update(target_power_changes)
+            result = await self._set_api_power(
+                api, target_power_changes, api_request_timeout
+            )
+            await self._results_sender.send(result)
+
+    async def _set_api_power(
```
Marenz, line 299: No method documentation?

src/frequenz/sdk/actor/power_distributing/_component_status/_ev_charger_status_tracker.py
shsms: Yes, but that would require some amount of refactoring, because the timer trigger would send something out only if the status has changed.  But I'm not sure that it should be changed.

This is an internal feature, and its values are used by the power distributor and the system bounds tracker.  So having it like this in practice would mean that PowerManager can report zero bounds right away on start up, until data is available.

tests/actor/power_distributing/_component_status/test_ev_charger_status.py
shsms: Never figured out how to use it.  Will look into it, but would like to make an issue instead, because I still have to get the PV Pool out in the next couple of weeks.  If that happens quickly, can replace tests for both pools.  Or can do that in May.

src/frequenz/sdk/actor/power_distributing/_component_managers/_battery_manager.py
shsms: No,  I think @llucax made parameter descriptions optional, so that we can have property documentation or something related.  So if no parameter documentation is provided at all, CI doesn't complain anymore, but still catches issues if it is partial or outdated, etc.

I will add the descriptions.

src/frequenz/sdk/actor/power_distributing/_component_status/_ev_charger_status_tracker.py
Marenz: > Yes, but that would require some amount of refactoring, because the timer trigger would send something out only if the status has changed

But something _has_ changed if `_last_status`  is `None` as that is different from the `new_status` which is set to `not working`

shsms: This is used as a hint to the power manager and *pool metrics whether a battery is usable or not.  We need to send something out right at the beginning, else they wait for a status to arrive, before streaming values out.

So changing this would mean changing all the users of this data, which is more than a few.

Sending the `NOT_WORKING` flag even though we don't know if it is working or not is I'm guessing the objection?  Maybe renaming it to `NOT_USABLE` would make it more general, and can be sent before we receive any data, or when missing data, or when the component is known to be not working?

src/frequenz/sdk/actor/power_distributing/_component_managers/_ev_charger_manager/_ev_charger_manager.py
shsms: Oops, this is supposed to stop the component pool status tracker.  I've added a fixup commit.

shsms:
I've added some fixup commits with the missing docstrings, so they're easy to review.  Will squash them before merging.

src/frequenz/sdk/actor/power_distributing/_component_status/_ev_charger_status_tracker.py
Marenz: > else they wait for a status to arrive, before streaming values out.

I mean, optimally it (the status) would arrive right before the values, or do you mean they only request the values once they have the status?


I think the reason I am talking about this is, as a user, you see the logs like "Component changed to (NOT_)WORKING" etc and it can be worrying to see "NOT WORKING" at startup. I know it was for me :)
`NOT_USABLE` is not much better in that context, you simply don't expect a `NOT_` like that right at startup when things are actually working in order

src/frequenz/sdk/timeseries/ev_charger_pool/_system_bounds_tracker.py
```diff
@@ -0,0 +1,144 @@
...
+                self._component_pool_status = selected.message
+                to_pop = []
+                for comp_id in self._latest_component_data:
+                    if (
+                        comp_id not in self._component_pool_status.working
+                        and comp_id not in self._component_pool_status.uncertain
+                    ):
+                        to_pop.append(comp_id)
+                for comp_id in to_pop:
+                    self._latest_component_data.pop(comp_id, None)
```
Marenz: I was a bit confused at first why `pop()` is used here and not `del self._latest_component_data[comp_id]`.
I'd suggest to write `to_remove` or `to_delete` here though, it's a bit clearer. `pop` kind of implies to me that we want to do something with the result.

src/frequenz/sdk/timeseries/ev_charger_pool/_system_bounds_tracker.py
```diff
@@ -0,0 +1,144 @@
...
+        self._bounds_sender = bounds_sender
+        self._latest_component_data: dict[int, EVChargerData] = {}
+        self._last_sent_bounds: SystemBounds | None = None
+        self._component_pool_status = ComponentPoolStatus(set(), set())
+
+    def start(self) -> None:
+        """Start the EV charger system bounds tracker."""
+        self._tasks.add(asyncio.create_task(self._run_forever()))
+
+    async def _send_bounds(self) -> None:
```
Marenz, line 60: no docu

src/frequenz/sdk/microgrid/_data_pipeline.py
```diff
@@ -362,21 +373,32 @@ def producer() -> Producer:
...
+    """Return a new `EVChargerPool` instance for the given parameters.
 
-    If an EVChargerPool instance for the given ids doesn't exist, a new one is
-    created and returned.
+    The priority value is used to resolve conflicts when multiple actors are trying to
+    propose different power values for the same set of EV chargers.
+
+    !!! note
+        When specifying priority, bigger values indicate higher priority. The default
+        priority is the lowest possible value.
```
Marenz, line 394: What if the priority is equal? Which one will then be selected?

Marenz:
(more for myself) I have currently reviewed up to [Add a propose_power method to EVChargerPool](https://github.com/frequenz-floss/frequenz-sdk-python/pull/900/commits/89ac7f93353703389fead86a6cd0d72436687969)

src/frequenz/sdk/actor/power_distributing/_component_status/_ev_charger_status_tracker.py
shsms: I think the way it is implemented, we see `EV charger 1008 status changed from ComponentStatusEnum.NOT_WORKING to ComponentStatusEnum.WORKING` almost immediately after the app starts.  And these are the first messages related to status that we see, almost immediately after app start:

```
WARNING:frequenz.sdk.actor.power_distributing._component_status._ev_charger_status_tracker:EV charger 1008 is in WORKING state.
INFO:frequenz.sdk.actor.power_distributing._component_status._ev_charger_status_tracker:EV charger 1008 status changed from ComponentStatusEnum.NOT_WORKING to ComponentStatusEnum.WORKING
WARNING:frequenz.sdk.actor.power_distributing._component_status._ev_charger_status_tracker:EV charger 1007 is in WORKING state.
INFO:frequenz.sdk.actor.power_distributing._component_status._ev_charger_status_tracker:EV charger 1007 status changed from ComponentStatusEnum.NOT_WORKING to ComponentStatusEnum.WORKING
```

If the BatteryPool says "Changed to NOT_WORKING", we'll have to change that separately.

shsms: Another improvement could be to add an `UNKNOWN` status as the initial one.  But then we'll have to rename `UNCERTAIN` to something else.  Something that means "had an unexpected failure, when it should have succeeded because the component had no errors and the request respected the component's bounds".

src/frequenz/sdk/timeseries/ev_charger_pool/_system_bounds_tracker.py
shsms: true, changed in fixup commit: c2b74378

src/frequenz/sdk/timeseries/ev_charger_pool/_system_bounds_tracker.py
shsms: added in fixup commit: c2b74378

src/frequenz/sdk/microgrid/_data_pipeline.py
shsms: If there are multiple proposals with the same priority, they will be treated as having different priorities based on their randomly generated `source_id`.  So the priority between actors will be stable and they won't get into a race with each other when they have the same priority values.

tests/actor/power_distributing/_component_status/test_ev_charger_status.py
shsms: https://github.com/frequenz-floss/frequenz-sdk-python/issues/901

src/frequenz/sdk/actor/power_distributing/_component_status/_ev_charger_status_tracker.py
shsms: In the battery pool too, it doesn't seem to say "NOT_WORKING" at startup, just:

```
INFO:frequenz.sdk.actor.power_distributing._component_status._battery_status_tracker:battery 1001 changed status ComponentStatusEnum.WORKING
INFO:frequenz.sdk.actor.power_distributing._component_status._battery_status_tracker:battery 1004 changed status ComponentStatusEnum.WORKING

File: src/frequenz/sdk/microgrid/_data_pipeline.py (hunk 11 above)
Comment (Marenz) [reply src/frequenz/sdk/microgrid/_data_pipeline.py:394]: Maybe add a short form or hint of htat to the docu here :)

Comment (shsms) [reply src/frequenz/sdk/microgrid/_data_pipeline.py:394]: Will do as part of this issue: https://github.com/frequenz-floss/frequenz-sdk-python/issues/902

**Comment by llucax** [comment]
FYI, starting a review.

File: src/frequenz/sdk/actor/power_distributing/_component_status/_ev_charger_status_tracker.py (hunk 1 above)
Comment (llucax) [reply src/frequenz/sdk/actor/power_distributing/_component_status/_ev_charger_status_tracker.py:168]: Without having seen the changes yet, and not being very familiar with how this is used (so high risk of saying nonsense), how about `INITIALIZING` instead of `UNKNOWN`, if we can use a special state for this initialization step?

File: src/frequenz/sdk/microgrid/_data_pipeline.py (hunk 11 above)
Comment (llucax) [reply src/frequenz/sdk/microgrid/_data_pipeline.py:394]: A small side note, in docstrings you can use normal sections for admonitions:

```
Note:
    When specifying priority, bigger values indicate higher priority. The default ...
```

You only need the "special" syntax if you need some customization (like if you don't want it to be collapsible or to show it on the side for example).

File: src/frequenz/sdk/actor/power_distributing/_component_status/_ev_charger_status_tracker.py (hunk 1 above)
Comment (llucax) [reply src/frequenz/sdk/actor/power_distributing/_component_status/_ev_charger_status_tracker.py:168]: I guess related to this, reading the commit message:

> It reports an EV charger as `WORKING` when an EV is connected to it,
> and power can be allocated to it, and `NOT_WORKING` otherwise.

I guess this status comes from batteries, right? Maybe it would be better to rename `WORKING`/`NOT_WORKING` to `USABLE`/`UNUSABLE` or `AVAILABLE`/`UNAVAILABLE`, it looks a bit misleading to say a EV charger is "not working" just because an EV is not connected to it.

File: src/frequenz/sdk/actor/power_distributing/_component_status/_ev_charger_status_tracker.py [hunk 12]
```diff
@@ -0,0 +1,204 @@
...
+    async def _run_forever(self) -> None:
+        """Run the status tracker forever."""
+        while True:
+            try:
+                await self._run()
+            except Exception:  # pylint: disable=broad-except
+                _logger.exception(
+                    "Restarting after exception in EVChargerStatusTracker.run()"
+                )
+                await asyncio.sleep(1.0)
```
llucax, line 112: This could be another reason to remove the auto-restart logic out of the actor, so we have a consistent policy for auto-restart. For example, the actor now has some sort of backoff to avoid a flood of restarts if it is "permanently" broken IIRC.

Since you are using this snippet in several places in this PR, maybe it is worth extracting the forever loop to a separate function in `_internal` and use it in the actor run and in the new places here? But it can be done in a separate PR too.

tests/actor/power_distributing/_component_status/test_ev_charger_status.py
```diff
@@ -0,0 +1,165 @@
...
+            assert await receive_timeout(status_receiver) == ComponentStatus(
+                _EV_CHARGER_ID, ComponentStatusEnum.UNCERTAIN
+            )
+
+            # After the blocking duration, it should become working again.
+            assert await receive_timeout(status_receiver) is Timeout
+            assert await receive_timeout(status_receiver, 1.0) == ComponentStatus(
+                _EV_CHARGER_ID, ComponentStatusEnum.WORKING
+            )
+            await cancel_and_await(_keep_sending_healthy_message_task)
```
llucax, line 165: For proper cleanup, it might be a good idea to put this in a `ExitStack`, so it is cancelled and awaited even if the test exits due to an unhandled exception (like an assertion).

src/frequenz/sdk/actor/power_distributing/_component_status/_ev_charger_status_tracker.py
```diff
@@ -0,0 +1,204 @@
...
+    @override
+    def __init__(  # pylint: disable=too-many-arguments
+        self,
+        component_id: int,
+        max_data_age: timedelta,
+        max_blocking_duration: timedelta,
+        status_sender: Sender[ComponentStatus],
+        set_power_result_receiver: Receiver[SetPowerResult],
+    ) -> None:
+        """Create an `EVChargerStatusTracker` instance.
```
llucax:
```suggestion
        """Initialize this instance.
```
Super minor and just mentioning it because you used "Initialize" in other `__init__()`s.

src/frequenz/sdk/actor/power_distributing/_component_managers/_battery_manager.py
```diff
@@ -127,8 +127,18 @@ class BatteryManager(ComponentManager):
...
     ):
-        """Initialize the battery data manager."""
+        """Initialize the battery data manager.
```
llucax:
```suggestion
        """Initialize this instance.
```

src/frequenz/sdk/actor/power_distributing/_component_managers/_ev_charger_manager/_states.py
```diff
@@ -0,0 +1,132 @@
...
+            self.last_charging_time = latest_ev_data.timestamp
+
+
+class EvcStates:
+    """Tracks states of all ev chargers."""
+
+    _states: dict[int, EvcState]
+
+    def __init__(self) -> None:
+        """Initialize the EvcStates object."""
```
llucax: Also super minor and optional: I started using just "Initialize this instance." so it is more copy&pasteable and more renaming friendly.

```suggestion
        """Initialize this instance."""
```

src/frequenz/sdk/timeseries/ev_charger_pool/_system_bounds_tracker.py
```diff
@@ -0,0 +1,145 @@
...
+class EVCSystemBoundsTracker(BackgroundService):
+    """Track the system bounds for the EV chargers."""
+
+    def __init__(
+        self,
+        component_ids: abc.Set[int],
+        status_receiver: Receiver[ComponentPoolStatus],
+        bounds_sender: Sender[SystemBounds],
+    ):
+        """Initialize the system bounds tracker.
```
llucax:
```suggestion
        """Initialize this instance.
```

src/frequenz/sdk/timeseries/ev_charger_pool/_system_bounds_tracker.py
```diff
@@ -0,0 +1,145 @@
...
+            try:
+                await self._run()
+            except Exception:  # pylint: disable=broad-except
+                _logger.exception(
+                    "Restarting after exception in EVChargerSystemBoundsTracker.run()"
+                )
+                await asyncio.sleep(1.0)
+
+    async def _run(self) -> None:
+        """Run the system bounds tracker."""
```
llucax, line 119: Maybe it would be nice to have a description of what this is supposed to do, either here or in the class docstring, at least for me it wasn't trivial to see it on a quick look at the code.

src/frequenz/sdk/actor/power_distributing/_component_status/_ev_charger_status_tracker.py
shsms: It could also be because of EV charger in error state.  Then someone might have to go and reset.

This used for telling the Power Distributor whether power can be allocated to a component or not.  And we need alternatives for WORKING, NOT_WORKING, UNCERTAIN that would be suitable for batteries, EV chargers, PV inverters, CHP, etc.  

Maybe USABLE/NOT_USABLE is the way to go.  UNUSABLE is not distinct enough for my eyes, I might mix things up, especially when reading.

shsms: But would keep for a separate PR: https://github.com/frequenz-floss/frequenz-sdk-python/issues/903

src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool.py
```diff
@@ -134,16 +181,53 @@ def power(self) -> FormulaEngine[Power]:
...
+        return typing.cast(ReceiverFetcher[EVChargerPoolReport], channel)
+
     async def stop(self) -> None:
         """Stop all tasks and channels owned by the EVChargerPool."""
-        await self._formula_pool.stop()
+        await self._ev_charger_pool.stop()
+
+    @property
+    def _system_power_bounds(self) -> ReceiverFetcher[SystemBounds]:
+        """Return a receiver for the system power bounds."""
```
llucax:
```suggestion
        """Return a receiver fetcher for the system power bounds."""
```

src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
```diff
@@ -133,17 +133,22 @@ def _add_bounds_tracker(self, component_ids: frozenset[int]) -> None:
...
-        if self._component_category is not ComponentCategory.BATTERY:
+        bounds_receiver: Receiver[SystemBounds]
+        # pylint: disable=protected-access
+        if self._component_category is ComponentCategory.BATTERY:
+            battery_pool = microgrid.battery_pool(component_ids)
+            bounds_receiver = battery_pool._system_power_bounds.new_receiver()
+        elif self._component_category is ComponentCategory.EV_CHARGER:
+            ev_charger_pool = microgrid.ev_charger_pool(component_ids)
+            bounds_receiver = ev_charger_pool._system_power_bounds.new_receiver()
+        # pylint: enable=protected-access
```
llucax, line 144: Any reason why to make these `_system_power_bounds` private`? Are the pools exposed directly to users? For me accessing private methods is a "code smell", I think if there are methods that we use outside of the class they should be public unless there is a very good reason for them not to be.

src/frequenz/sdk/microgrid/_data_pipeline.py
shsms: Done in #904

src/frequenz/sdk/microgrid/_data_pipeline.py
```diff
@@ -158,6 +164,8 @@ def producer(self) -> Producer:
     def ev_charger_pool(
         self,
         ev_charger_ids: abc.Set[int] | None = None,
+        name: str | None = None,
+        priority: int = -sys.maxsize - 1,
```
llucax, line 168: I guess we already discussed this, but why not `0`? Using the lowest possible number ever, we are making it impossible to use a priority lower than the default (which I guess is what most users will use). If we use `0`, it means that there is a "normal" value, and you set a priority to be more or less important than most other users that explicitly say they don't care about the priority.

src/frequenz/sdk/timeseries/ev_charger_pool/_result_types.py
```diff
@@ -0,0 +1,36 @@
...
+from ...actor import power_distributing
+from .._base_types import Bounds
+from .._quantities import Power
+
+
+class EVChargerPoolReport(typing.Protocol):
+    """A status report for an EV chargers pool."""
+
+    target_power: Power | None
+    """The currently set power for the batteries."""
```
llucax:
```suggestion
    """The currently set power for the EV chargers."""
```

src/frequenz/sdk/timeseries/ev_charger_pool/_result_types.py
```diff
@@ -0,0 +1,36 @@
...
+
+    distribution_result: power_distributing.Result | None
+    """The result of the last power distribution.
+
+    This is `None` if no power distribution has been performed yet.
+    """
+
+    @property
+    def bounds(self) -> Bounds[Power] | None:
+        """The usable bounds for the batteries.
```
llucax:
```suggestion
        """The usable bounds for the EV chargers.
```

src/frequenz/sdk/timeseries/ev_charger_pool/_result_types.py
```diff
@@ -0,0 +1,36 @@
...
+
+        These bounds are adjusted to any restrictions placed by actors with higher
+        priorities.
+
+        There might be exclusion zones within these bounds. If necessary, the
+        [`adjust_to_bounds`][frequenz.sdk.timeseries.battery_pool.BatteryPoolReport.adjust_to_bounds]
```
llucax: I guess this `adjust_to_bounds` should be present in the `EVChargerPoolReport` and this should be fixed to reference that instead?

tests/timeseries/_ev_charger_pool/test_ev_charger_pool_control_methods.py
```diff
@@ -0,0 +1,266 @@
...
+            )
+        await mocks.component_status_sender.send(
+            ComponentPoolStatus(working=set(mocks.microgrid.evc_ids), uncertain=set())
+        )
+
+    async def _patch_data_pipeline(self, mocker: MockerFixture) -> None:
+        mocker.patch(
+            "frequenz.sdk.microgrid._data_pipeline._DATA_PIPELINE._ev_power_wrapper"
+            "._pd_wait_for_data_sec",
+            0.1,
```
llucax, line 103: I believe this was mentioned before, but it would be probably better to use async solipsism to fake the loop time instead of patching wait times to be smaller. I know it is probably not an easy feat for this PR, but at least we should try to avoid adding more code using slow and potentially flaky tests that depends on the real clock.

tests/timeseries/_ev_charger_pool/test_ev_charger_pool_control_methods.py
```diff
@@ -0,0 +1,266 @@
...
+        mocker.patch(
+            "frequenz.sdk.microgrid._data_pipeline._DATA_PIPELINE._ev_power_wrapper"
+            "._power_distributing_actor._component_manager._config",
+            EVDistributionConfig(
+                component_ids=cm._config.component_ids,
+                initial_current=cm._config.initial_current,
+                min_current=cm._config.min_current,
+                increase_power_interval=timedelta(seconds=0.12),
+            ),
+        )
```
llucax, line 127: Also so much need for patching IMHO shows a design issue. Ideally we should test in a way that we can just pass fake arguments to functions directly instead of patching so much global state.

tests/timeseries/_ev_charger_pool/test_ev_charger_pool_control_methods.py
```diff
@@ -0,0 +1,266 @@
...
+            mocks.streamer.start_streaming(
+                MeterDataWrapper(
+                    meter_id,
+                    now,
+                    voltage_per_phase=(230.0, 230.0, 230.0),
+                ),
+                0.05,
+            )
+
+        await asyncio.sleep(1)
```
llucax, line 166: Is this really sleeping for one second? This will make tests really slow :sob:

llucax (commented):
Besides the comments (most minor), really nice job!

src/frequenz/sdk/actor/power_distributing/_component_status/_ev_charger_status_tracker.py
shsms: true.  made an issue: https://github.com/frequenz-floss/frequenz-sdk-python/issues/906

want to focus on PV pool, which needs to be done and merged in 2 weeks, so will try to move as many items as possible here to future PRs.

src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
shsms: They are part of the `BatteryPool` because there were there originally and we repurposed it for internal use.  And I followed the same pattern here too.  We need to investigate how to pull it out, because it is also part of the feedback loop of: its values are used to set power and set power results affect its values, which is already setup of the *pools.  

And their values are used by the PowerManager which sits in between.  So it would be some effort to move them out.  I'll make an issue.

shsms: Marked it as post-1.0: https://github.com/frequenz-floss/frequenz-sdk-python/issues/907

src/frequenz/sdk/microgrid/_data_pipeline.py
shsms: I have added an agenda item for the next sdk meeting.  We can deal with it separately based on the outcome of the discussion.

src/frequenz/sdk/timeseries/ev_charger_pool/_result_types.py
shsms: I was wondering if we need it for EV chargers, because the bounds are likely to be simple, and there won't be exclusion bounds for EV chargers.

Also @cwasicki was looking for ways to make sure we don't send out zero power repeatedly, and I was wondering if `adjust_to_bounds` is a good way to achieve that.  We'll have to do some brainstorming probably next week, and maybe change the `adjust_to_bounds` interface in the BatteryPool too.

For now, I will remove these comments.

src/frequenz/sdk/timeseries/ev_charger_pool/_system_bounds_tracker.py
shsms: added a fixup commit.

tests/actor/power_distributing/_component_status/test_ev_charger_status.py
shsms: Added an issue: https://github.com/frequenz-floss/frequenz-sdk-python/issues/908

tests/timeseries/_ev_charger_pool/test_ev_charger_pool_control_methods.py
shsms: Already have an issue for it.  Will take care of it after PV pool.

src/frequenz/sdk/timeseries/ev_charger_pool/_system_bounds_tracker.py
llucax: Cool, please let me know when you finish making changes by requesting a new review.

tests/timeseries/_ev_charger_pool/test_ev_charger_pool_control_methods.py
shsms: True.  To be able to pass this as an argument to `EVChargerManager`, I'd have to change the abstract class for creating `ComponentManager`.  Made an issue: https://github.com/frequenz-floss/frequenz-sdk-python/issues/910

tests/timeseries/_ev_charger_pool/test_ev_charger_pool_control_methods.py
shsms: Probably not necessary, but I don't want to find out now.  Will deal with it when migrating to solipsism.

shsms:
@llucax @Marenz can I squash the fixup commits?

llucax (approved):
LGTM! @Marenz is supposed to be back tomorrow, maybe we can wait for his OK too.

Marenz:
LGTM as well

shsms:
I've applied the fixups.

PR #904, opened 2024-03-27
Warn if duplicate `*pool` instances are being created

### LINKED ISSUE #902: Warn when several calls to `microgrid.battery_pool()`, `ev_charger_pool()`, `pv_pool()` are made
### What's needed?

Every call to `microgrid.*pool()` returns a new instance, with a unique `source_id`.  Actors are supposed to hold a reference to their pools and reuse them, instead of creating a new one each time they want to send a command.

Ideally an actor should make only a single call to a `*pool()` method, so that the bounds it streams correspond with the requests, etc.

But if someone does it the WRONG way `microgrid.battery_pool().set_power()`, they are currently not informed about it by any of the checks.

### Proposed solution

There needs to be a runtime check that starts printing warning logs when more than 1 call has been made to the `microgrid.battery_pool()` method, with the same priority and with the same set of batteries.

The log message could say that if they are made by different actors, they should explicitly set different priority values, and if they are made by the same actor, they should stop that and instead reuse the same instance.

### Use cases

_No response_

_No response_

### Additional context

_No response_

shsms:
This is because such usage could make the power manager take longer to converge on the desired power.

This PR also improves the docs for the `microgrid.*pool()` functions.

Closes #902

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -20,7 +20,7 @@
 
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+- Warning messages are logged when multiple instances of `*Pool`s are created for the same set of batteries, with the same priority values.
 
 ## Bug Fixes
 
--- a/src/frequenz/sdk/microgrid/_data_pipeline.py
+++ b/src/frequenz/sdk/microgrid/_data_pipeline.py
@@ -103,8 +103,12 @@ class _DataPipeline:  # pylint: disable=too-many-instance-attributes
         self._consumer: Consumer | None = None
         self._producer: Producer | None = None
         self._grid: Grid | None = None
-        self._ev_charger_pools: dict[frozenset[int], EVChargerPoolReferenceStore] = {}
-        self._battery_pools: dict[frozenset[int], BatteryPoolReferenceStore] = {}
+        self._ev_charger_pool_reference_stores: dict[
+            frozenset[int], EVChargerPoolReferenceStore
+        ] = {}
+        self._battery_pool_reference_stores: dict[
+            frozenset[int], BatteryPoolReferenceStore
+        ] = {}
         self._frequency_instance: GridFrequency | None = None
         self._voltage_instance: VoltageStreamer | None = None
 
@@ -195,8 +199,8 @@ class _DataPipeline:  # pylint: disable=too-many-instance-attributes
         if ev_charger_ids is not None:
             key = frozenset(ev_charger_ids)
 
-        if key not in self._ev_charger_pools:
-            self._ev_charger_pools[key] = EVChargerPoolReferenceStore(
+        if key not in self._ev_charger_pool_reference_stores:
+            self._ev_charger_pool_reference_stores[key] = EVChargerPoolReferenceStore(
                 channel_registry=self._channel_registry,
                 resampler_subscription_sender=self._resampling_request_sender(),
                 status_receiver=self._ev_power_wrapper.status_channel.new_receiver(
@@ -210,7 +214,9 @@ class _DataPipeline:  # pylint: disable=too-many-instance-attributes
                 ),
                 component_ids=ev_charger_ids,
             )
-        return EVChargerPool(self._ev_charger_pools[key], name, priority)
+        return EVChargerPool(
+            self._ev_charger_pool_reference_stores[key], name, priority
+        )
 
     def grid(self) -> Grid:
         """Return the grid measuring point."""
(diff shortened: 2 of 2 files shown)


shsms:
Based on #900.  Only the last 4 commits are relevant.

llucax (commented):
In general the 4 last commits look OK, but I wonder if it wouldn't be safer to just fail the request if it matches an existing pool instead of printing a warning and proceed with a behavior that can potentially cause issues. We could just raise a `PoolAlreadyExist` (inheriting from `KeyError` for example), and maybe add an extra method to get an existing instance (that will fail if the instance doesn't exist yet), if we think there might be cases where users wants to use the same pool but it is difficult to save and share the pool object.

shsms:
> I wonder if it wouldn't be safer to just fail the request if it matches an existing pool instead of printing a warning and proceed with a behavior that can potentially cause issues.

I don't think this is a critical issue that we need to fail, because it would continue to function, and they're calling `propose_power` anyway.  

I think we can fail for trivial issues that happen at startup.  But once the app is running, we should try to avoid crashes.  Maybe we can look into static analysis, if there are multiple calls to `*pool` in a class, etc.

llucax:
Mmm, still not fully convinced, from this text:

> If not, a random prioritization will be imposed on them to resolve conflicts, which may lead to unexpected behavior.

For me unexpected behavior is bad. If using the same args is a bug (maybe is one that sometimes can be unproblematic), we should probably error. For me it is a bug if there is no valid case where you actually want to call the function 2 times with the exact same arguments. If you say there is a valid case where a user wants to do this, then I agree a warning is enough. Just avoiding a crash is not a good enough reason not to raise an exception IMHO, programmers can make all sort of mistakes that will lead to a crash, instead those crashed should be properly handled when possible (like with the actor auto-restart).

Also I guess getting a pool is something you probably do at the start of an application/actor, so the failure should occur pretty soon.

This also makes me think that maybe we don't want to provide a default priority if getting the same pool several times with the same priority is problematic.

shsms:
> Also I guess getting a pool is something you probably do at the start of an application/actor, so the failure should occur pretty soon.

No, they might try to make a second pool only when they try to send a charge command, for example, which could be days after deployment.

The power manager cleans up old requests after 1 min, so even if they create a new pool for each request, it just means that we'd converge to the target power more slowly, rather than crashing on friday 11pm because we try to handle weekend rush.

And if multiple actors use the same batteries with the same priorities, it probably doesn't matter to them, but maybe we can make it even more harder to ignore, by making the priority a required parameter when creating *pools.

If priority is required, they will have to think about it and deal with it.  I think that much should be enough.

shsms:
Maybe qualifying the documentation makes it better? (I've made this change already)

        ... If not,
        a random prioritization will be imposed on them to resolve conflicts, which may
        lead to unexpected behavior like longer duration to converge on the desired
        power.

llucax:
OK, I created a separate discussion, because it seems like it is not a trivial topic, so we shouldn't block this PR which is already an improvement on the status quo:

* https://github.com/frequenz-floss/frequenz-sdk-python/discussions/913

> Maybe qualifying the documentation makes it better? (I've made this change already)

I guess it is a bit more clear, but maybe it would be better to clarify the random priority is only among the conflicting requests, right? Is not really a random priority in the sense that it could be a proper different priority value.

I asked ChatGPT for help and given some instructions it came up with:

> In deployments with multiple actors managing the same set of EV chargers, assigning distinct priorities for each is recommended to prevent conflicts. Without unique priorities, requests with identical EV chargers and priority levels are deterministically sorted in an implicit order (without changing their priority values). This can lead to unpredictable conflict resolution and longer times to reach the desired power distribution. Note that this internal sorting affects only requests that match both in charger set and priority level.

Not for this PR, as it is probably more work, but maybe another approach would be to have a policy based on when the requests were made, like the latest request will always have a higher priority than the previous conflicting requests.

llucax (approved):
The previous comment is just a soft suggestion, reading the text again in full seems to be clear enough for an approval.

shsms:
> Not for this PR, as it is probably more work, but maybe another approach would be to have a policy based on when the requests were made, like the latest request will always have a higher priority than the previous conflicting requests.

But this would lead to races between the actors, which is what the PowerManager hopes to avoid.

llucax:
> But this would lead to races between the actors, which is what the PowerManager hopes to avoid.

What kind of races? If many (conflicting) pools are created in a burst, then yes, is sort of random as it is now, but if they are created more sequentially, it will just mean that the latest always wins (or loses, whatever makes more sense).

shsms:
> What kind of races?

They will keep sending conflicting requests repeatedly, maybe even continuously, because they see that they both have bounds, but the actual target power is somewhere else, and they need to correct it.  and we'll end up oscillating between charging and discharging every second.

(I'm actually talking about the case where the pools are created with the same priority but by different actors.  I guess you were thinking of the same actor?)

llucax:
OK, I see now, the issue is not when the pools are created but when they are actually used :+1:

PR #914, opened 2024-04-04
Add a `PVPool` implementation

shsms:


llucax:
CI is failing, I'm starting a review anyway, so please unless it is very small and trivial stuff (or adding more commits to amend later), please don't update the PR.

src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
```diff
@@ -47,6 +48,8 @@ def __init__(  # pylint: disable=too-many-arguments
         Args:
             component_category: The category of the component this power manager
                 instance is going to support.
+            component_type: The type of the component of the given category that this
+                actor is responsible for.
```
llucax: Maybe it is worth specifying here what does `None` mean? Determined automatically? The actor is not responsible for a component type in particular? (same for other places with `ComponentType | None`)

src/frequenz/sdk/_internal/_channels.py
```diff
@@ -72,3 +74,7 @@ def has_value(self) -> bool:
     async def _run(self) -> None:
         async for value in self._receiver:
             self._latest_value = value
+
+    async def stop(self) -> None:
+        """Stop the cache."""
+        await cancel_and_await(self._task)
```
llucax, line 80: If a future PR I guess, but it might be a good idea to convert `LatestValueCache` to a background service eventually, so it can be used as a context manager in tests, etc.

src/frequenz/sdk/timeseries/pv_pool/_pv_pool.py
```diff
@@ -47,9 +51,41 @@ def __init__(  # pylint: disable=too-many-arguments
         """
         self._pv_pool_ref = pv_pool_ref
         unique_id = uuid.uuid4()
-        self._source_id = unique_id if name is None else f"{name}-{unique_id}"
+        self._source_id = str(unique_id) if name is None else f"{name}-{unique_id}"
```
llucax: Not a biggie, but this change seems to be in the wrong commit.

src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
```diff
@@ -33,6 +33,7 @@ class PowerManagingActor(Actor):
     def __init__(  # pylint: disable=too-many-arguments
         self,
         component_category: ComponentCategory,
+        component_type: ComponentType | None,
```
llucax: If this needs to be passed explicitly, I would make it a keyword-only argument, otherwise when reading the call-site, it is really hard to know what argument is that `None` for. Any reason not to make it optional defaulting to `None`?

src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
```diff
@@ -0,0 +1,230 @@
...
+
+    @override
+    async def start(self) -> None:
+        """Start the PV inverter manager."""
+        self._component_data_caches = {
+            inv_id: LatestValueCache(
+                await connection_manager.get().api_client.inverter_data(inv_id)
+            )
+            for inv_id in self._pv_inverter_ids
+        }
```
llucax, line 78: Maybe do this in parallel using `gather()`, a `TaskGroup` or similar?

src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
```diff
@@ -0,0 +1,230 @@
...
+                api_client.set_power(component_id, power.as_watts())
+            )
+        _, pending = await asyncio.wait(
+            tasks.values(),
+            timeout=request.request_timeout.total_seconds(),
+            return_when=asyncio.ALL_COMPLETED,
+        )
+        for task in pending:
+            task.cancel()
+            await asyncio.gather(*pending, return_exceptions=True)
```
llucax: I was a bit confused by this thinking there could be no pending tasks because the use of `ALL_COMPLETED`, and then it was a big red light to see exceptions being silently swallowed. But they can be pending tasks because of the timeout, right? And you ignore exceptions because you want to kill and ignore the pending stuff, you don't care about exceptions because they should only be `CancelledError` or even if there is a real exception you don't care about it because the task is discarded?

Maybe it is worth an early comment, even when things are much more clear once one keeps reading. It got me thinking about it for a while.

src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
```diff
@@ -0,0 +1,230 @@
...
+                case asyncio.CancelledError:
+                    _logger.warning(
+                        "Timeout while setting power to EV charger %s", component_id
+                    )
+                case grpc.aio.AioRpcError as err:
+                    _logger.warning(
+                        "Error while setting power to EV charger %s: %s",
+                        component_id,
+                        err,
+                    )
```
llucax, line 203: No logging for any other type of exception?

src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
```diff
@@ -0,0 +1,230 @@
...
+                )
+            )
+        await self._results_sender.send(
+            Success(
+                succeeded_components=succeeded_components,
+                succeeded_power=self._target_power,
+                excess_power=remaining_power,
+                request=request,
+            )
+        )
```
llucax: 🚨🚨🚨 Is a `return` missing after sending the `PartialFailure` or do you really want to send a `Success` after it too? 🚨🚨🚨

src/frequenz/sdk/actor/power_distributing/_component_status/_pv_inverter_status_tracker.py
```diff
@@ -0,0 +1,205 @@
...
+
+        self._last_status = ComponentStatusEnum.NOT_WORKING
+        self._blocking_status = BlockingStatus(
+            min_duration=timedelta(seconds=1.0),
+            max_duration=max_blocking_duration,
+        )
+
+        BackgroundService.__init__(
+            self, name=f"PVInverterStatusTracker({component_id})"
+        )
```
llucax: Normally the parent constructor should be called first (otherwise you can be using wrongly initialized parent data), if there is a specific reason to call it last I would add a comment.
```suggestion
        BackgroundService.__init__(
            self, name=f"PVInverterStatusTracker({component_id})"
        )
        self._component_id = component_id
        self._max_data_age = max_data_age
        self._status_sender = status_sender
        self._set_power_result_receiver = set_power_result_receiver

        self._last_status = ComponentStatusEnum.NOT_WORKING
        self._blocking_status = BlockingStatus(
            min_duration=timedelta(seconds=1.0),
            max_duration=max_blocking_duration,
        )
```

shsms:
Added a few commits including tests, docu updates, release notes and a fixup for a bugfix.

src/frequenz/sdk/timeseries/pv_pool/_pv_pool.py
```diff
@@ -0,0 +1,201 @@
...
+    """An error that occurred in any of the PVPool methods."""
+
+
+class PVPool:
+    """An interface for interaction with pools of PV inverters.
+
+    !!! note
+        `PVPool` instances are not meant to be created directly by users. Use the
+        [`microgrid.pv_pool`][frequenz.sdk.microgrid.pv_pool] method for creating
+        `PVPool` instances.
```
llucax: I guess this is done for all other similar classes, but I think having the note in the constructor is enough.

src/frequenz/sdk/timeseries/pv_pool/_pv_pool.py
```diff
@@ -0,0 +1,201 @@
...
+        self._source_id = str(unique_id) if name is None else f"{name}-{unique_id}"
+        self._priority = priority
+
+    async def propose_power(
+        self,
+        power: Power | None,
+        *,
+        request_timeout: timedelta = timedelta(seconds=5.0),
+        bounds: Bounds[Power | None] = Bounds(None, None),
+    ) -> None:
```
llucax, line 64: I wonder if in the future it might be worth having a common interface for pools with this method for example. Probably not, but not sure. It looks we are keeping this method the same among all pools, so maybe it is worth if not only to ensure the interfaces don't diverge (if this is what we want).

src/frequenz/sdk/microgrid/__init__.py
```diff
@@ -74,16 +74,31 @@
...
-The total {{glossary("pv", "PV")}} power production in a microgrid can be streamed
-through [`pv_power`][frequenz.sdk.timeseries.logical_meter.LogicalMeter.pv_power] , and
-similarly the total CHP production in a site can be streamed through
-[`chp_power`][frequenz.sdk.timeseries.logical_meter.LogicalMeter.chp_power].  And total
-producer power is available through
-[`producer_power`][frequenz.sdk.timeseries.producer.Producer.power].
+The total CHP production in a site can be streamed through
+[`chp_power`][frequenz.sdk.timeseries.logical_meter.LogicalMeter.chp_power].  PV Power
+is available through the PV pool described below.  And total producer power is available
+through [`producer_power`][frequenz.sdk.timeseries.producer.Producer.power].
```
llucax: Unrelated to this PR, but Isn't this `producer.power` now?

src/frequenz/sdk/timeseries/pv_pool/_pv_pool_reference_store.py
```diff
@@ -0,0 +1,103 @@
...
+    def __init__(  # pylint: disable=too-many-arguments
+        self,
+        channel_registry: ChannelRegistry,
+        resampler_subscription_sender: Sender[ComponentMetricRequest],
+        status_receiver: Receiver[ComponentPoolStatus],
+        power_manager_requests_sender: Sender[Proposal],
+        power_manager_bounds_subs_sender: Sender[ReportRequest],
+        component_ids: abc.Set[int] | None = None,
+    ):
+        """Create an instance of the class.
```
llucax: Nitpick :grimacing: 

```suggestion
        """Initialize this instance.
```

tests/timeseries/_pv_pool/test_pv_pool_control_methods.py
```diff
@@ -0,0 +1,262 @@
...
+        set_power.reset_mock()
+        await pv_pool.propose_power(Power.from_watts(-90000.0))
+        await self._recv_reports_until(
+            bounds_rx,
+            lambda x: x.target_power is not None
+            and x.target_power.as_watts() == -90000.0,
+        )
+
+        self._assert_report(
+            await bounds_rx.receive(), power=-90000.0, lower=-100000.0, upper=0.0
```
llucax, line 251: No tests setting the bounds so `upper` < 0.0, is that on purpose for some reason?

llucax (commented):
Once again, very nice job and it is really nice to see al this finally coming together! :100:

shsms:
I've squashed the old fixups, will address the comments in new fixups.

src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
shsms: Added a few lines.

shsms: Although the API exposes component type for all categories, the client exposes it only for inverters, because that's been the only thing we've needed in the SDK.

src/frequenz/sdk/_internal/_channels.py
shsms: https://github.com/frequenz-floss/frequenz-sdk-python/issues/920

src/frequenz/sdk/timeseries/pv_pool/_pv_pool.py
shsms: oops

src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
shsms: made them optional and moved `component_category` and `component_type` to the end of the params list and made them kw only.

src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
shsms: Made an issue: https://github.com/frequenz-floss/frequenz-sdk-python/issues/921

src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
shsms: oops, looks like an indentation error, fixed.

also, we are not ignoring exceptions, they are processed in later stages.  add a line.

src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
shsms: No, those are the only possible ones.  This is the same pattern we've used in battery power distribution for well over a year.

src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
shsms: oops, that's a bug, fixed.

src/frequenz/sdk/timeseries/pv_pool/_pv_pool.py
shsms: The class documentation is what I see most of the time through editor tools.  Can I remove the constructor note instead?

Also yes, it is there twice in the other pools as well.

src/frequenz/sdk/timeseries/pv_pool/_pv_pool.py
shsms: I think that makes sense.  Same for the `power_status` and `power` methods as well.

src/frequenz/sdk/microgrid/__init__.py
shsms: True, updated.

src/frequenz/sdk/timeseries/pv_pool/_pv_pool_reference_store.py
shsms: done

tests/timeseries/_pv_pool/test_pv_pool_control_methods.py
shsms: Upper is not expected to be less than zero.  We can set arbitrarily small discharge powers.

shsms:
I will squash the commits around 3pm.

src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
llucax: > oops, looks like an indentation error, fixed.

I didn't even noticed the indentation error :laughing:

src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
llucax: What about bugs, an unhandled error? I would at least log any unexpected exceptions...

src/frequenz/sdk/timeseries/pv_pool/_pv_pool.py
llucax: Mmm, for me it makes more sense to keep it in the constructor, if you want to create one you should read the constructor documentation, while if you just want to use it, you probably will read the class documentation but you don't care about that note because you don't plan to instantiate it anyway, so it is noisy.

But if you say your editor shows the class docs instead when you call the constructor, then I guess we can keep it in both.

src/frequenz/sdk/timeseries/pv_pool/_pv_pool.py
llucax: * #922

llucax (dismissed):
I will approve to speed up things, but I'm still concerned about https://github.com/frequenz-floss/frequenz-sdk-python/pull/914#discussion_r1559278063, if we are not logging unexpected errors in other pools too, we can create a separate issue to fix this, but I think silently swallowing exceptions should **never** happen.

src/frequenz/sdk/timeseries/pv_pool/_pv_pool.py
shsms: I think you're right, that note is not necessary in the context of just the class, and I think if I try to call the constructor I will get the constructor docs.

I've added a commit to remove them for all pools.

src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
shsms: I've added a commit to handle broader exceptions in all pools.

src/frequenz/sdk/actor/power_distributing/_component_managers/_battery_manager.py
```diff
@@ -672,7 +672,7 @@ def _parse_result(
                     battery_ids,
                     request_timeout.total_seconds(),
                 )
-            except Exception as e:
+            except Exception as e:  # pylint: disable=broad-except
```
llucax: This change ended up in the wrong commit ("Remove `*Pool` constructor related docs from the class docs").

src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
```diff
@@ -159,7 +159,7 @@ async def distribute_power(self, request: Request) -> None:
         )
         await self._set_api_power(request, allocations, remaining_power)
 
-    async def _set_api_power(
+    async def _set_api_power(  # pylint: disable=too-many-locals
```
llucax: This change ended up in the wrong commit ("Remove `*Pool` constructor related docs from the class docs").

src/frequenz/sdk/actor/power_distributing/_component_managers/_ev_charger_manager/_ev_charger_manager.py
```diff
@@ -350,6 +351,12 @@ async def _set_api_power(
                         component_id,
                         err,
                     )
+                case e if e is not None:
+                    _logger.warning(
+                        "Unknown error while setting power to EV charger %s: %s",
+                        component_id,
+                        e,
+                    )
```
llucax: Same.
```suggestion
                    _logger.exception(
                        "Unknown error while setting power to EV charger %s",
                        component_id,
                    )
```

src/frequenz/sdk/actor/power_distributing/_component_managers/_ev_charger_manager/_ev_charger_manager.py
```diff
@@ -350,6 +351,12 @@ async def _set_api_power(
                         component_id,
                         err,
                     )
+                case e if e is not None:
```
llucax: This will catch `BaseException` too, is it intended? Maybe it doesn't make sense to log if a `SystemExit` or `KeyboardInterrupt` are raised (for example).
```suggestion
                case Exception():
```

src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
```diff
@@ -0,0 +1,239 @@
...
+                        "Error while setting power to EV charger %s: %s",
+                        component_id,
+                        err,
+                    )
+                case e if e is not None:
+                    _logger.warning(
+                        "Unknown error while setting power to EV charger %s: %s",
+                        component_id,
+                        e,
+                    )
```
llucax:
```suggestion
                case Exception():
                    _logger.exception(
                        "Unknown error while setting power to EV charger %s",
                        component_id,
                    )
```

src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
```diff
@@ -0,0 +1,239 @@
...
+        for component_id, task in tasks.items():
+            exc = task.exception()
+            if exc is not None:
+                failed_components.add(component_id)
+                failed_power += allocations[component_id]
+            else:
+                succeeded_components.add(component_id)
+
+            match task.exception():
+                case asyncio.CancelledError:
```
llucax: Oh, damn, I just noticed while looking at the new `case`s that this syntax is incorrect. I guess these error cases are not tested, right? `case` must have a "constructor-like" syntax for classes. At least I remember being bitten by this in the past, it is a shame that pylint doesn't seem to warn about this:

```py
>>> import asyncio
>>> e = asyncio.CancelledError("Blah")
>>> match e:
...     case asyncio.CancelledError:
...         print("CE")
...     case _:
...         print("NO")
... 
NO
>>> match e:
...     case asyncio.CancelledError():
...         print("CE")
...     case _:
...         print("NO")
... 
CE
```

> [!WARNING]
> This issue is also present in the other pools!

```suggestion
                case asyncio.CancelledError():
```

src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
```diff
@@ -0,0 +1,239 @@
...
+                    _logger.warning(
+                        "Timeout while setting power to EV charger %s", component_id
+                    )
+                case grpc.aio.AioRpcError as err:
```
llucax:
```suggestion
                case grpc.aio.AioRpcError() as err:
```

src/frequenz/sdk/actor/power_distributing/_component_managers/_battery_manager.py
```diff
@@ -668,6 +672,14 @@ def _parse_result(
...
                     request_timeout.total_seconds(),
                 )
+            except Exception as e:  # pylint: disable=broad-except
+                failed_power += distribution[inverter_id]
+                failed_batteries = failed_batteries.union(battery_ids)
+                _logger.warning(
+                    "Unknown error while setting power to battery %s: %s",
+                    battery_ids,
+                    e,
+                )
```
llucax: I recommend using `_logger.exception()` here, if this is an unexpected error (i.e. a bug), it will give us much more context about what and where the problem really happened (all the chained exceptions and a full traceback). If you think the log level should not be error and should be warning instead, you can just pass `exc_info=True, stack_info=True`.
```suggestion
            except Exception:  # pylint: disable=broad-except
                _logger.exception(
                    "Unknown error while setting power to battery %s",
                    battery_ids,
                )
```

src/frequenz/sdk/actor/power_distributing/_component_managers/_battery_manager.py
shsms: fixed

src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
shsms: fixed

src/frequenz/sdk/actor/power_distributing/_component_managers/_ev_charger_manager/_ev_charger_manager.py
shsms: updated.

src/frequenz/sdk/actor/power_distributing/_component_managers/_ev_charger_manager/_ev_charger_manager.py
shsms: updated.

src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
shsms: updated.

src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
shsms: fixed in all managers

src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
shsms: fixed

src/frequenz/sdk/actor/power_distributing/_component_managers/_battery_manager.py
shsms: done

shsms:
I added all the exception related fixes to the latest commit.

shsms:
Also squashed the commits.

shsms:
Enabling auto-merge

llucax (approved):
Awesome! :tada:

PR #918, opened 2024-04-08
Add model manager for machine learning models

idlir-shkurti-frequenz:
This PR adds a simple model manager module which implements the logic needed to load, update and monitor machine learning models.
Currently there are two model repository structures: 
- SingleModelRepository - This is a structure where we direct our handler to a repository with a single model in it.
- DirectModelRepository - This repository structure takes in dictionary input with a key which is the name of the model and a value which is the  direct path to each specific module.

This PR is heavily based (but slimmed down) on https://github.com/frequenz-floss/frequenz-sdk-python/pull/397

### DIFF
--- a/src/frequenz/sdk/ml/_model_manager.py
+++ b/src/frequenz/sdk/ml/_model_manager.py
@@ -1,0 +1,143 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""Load, update, monitor and retrieve machine learning models."""
+
+import asyncio
+import logging
+import pickle
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Generic, TypeVar, cast
+
+from frequenz.channels.file_watcher import EventType, FileWatcher
+from typing_extensions import override
+
+from frequenz.sdk.actor import BackgroundService
+
+_logger = logging.getLogger(__name__)
+
+T = TypeVar("T")
+
+
+@dataclass
+class _Model(Generic[T]):
+    """Represent a machine learning model."""
+
+    data: T
+    path: Path
+
+
+class ModelNotFoundError(Exception):
+    """Exception raised when a model is not found."""
+
+    def __init__(self, key: str) -> None:
+        """Initialize the exception with the specified model key.
+
+        Args:
+            key: The key of the model that was not found.
+        """
+        super().__init__(f"Model with key '{key}' is not found.")
+
+
+class ModelManager(BackgroundService, Generic[T]):
+    """Load, update, monitor and retrieve machine learning models."""
+
+    def __init__(self, model_paths: dict[str, Path], *, name: str | None = None):
+        """Initialize the model manager with the specified model paths.
+
+        Args:
+            model_paths: A dictionary of model keys and their corresponding file paths.
+            name: The name of the model manager service.
+        """
+        super().__init__(name=name)
+        self._models: dict[str, _Model[T]] = {}
+        self.model_paths = model_paths
+        self.load_models()
+
+    def load_models(self) -> None:
+        """Load the models from the specified paths."""
+        for key, path in self.model_paths.items():
+            self._models[key] = _Model(data=self._load(path), path=path)
+
+    @staticmethod
+    def _load(path: Path) -> T:
+        """Load the model from the specified path.
+
+        Args:
+            path: The path to the model file.
+
+        Returns:
+            T: The loaded model data.
+
+        Raises:
+            ModelNotFoundError: If the model file does not exist.
+        """
+        try:
+            with path.open("rb") as file:
+                return cast(T, pickle.load(file))
+        except FileNotFoundError as exc:
+            raise ModelNotFoundError(str(path)) from exc
+
+    @override
+    def start(self) -> None:
+        """Start the model monitoring service by creating a background task."""
+        if not self.is_running:
+            task = asyncio.create_task(self._monitor_paths())
+            self._tasks.add(task)
+            _logger.info(
+                "%s: Started ModelManager service with task %s",
+                self.name,
+                task,
+            )
+
+    async def _monitor_paths(self) -> None:
+        """Monitor model file paths and reload models as necessary."""
+        model_paths = [model.path for model in self._models.values()]
+        file_watcher = FileWatcher(
+            paths=list(model_paths), event_types=[EventType.CREATE, EventType.MODIFY]
+        )
+        _logger.info("%s: Monitoring model paths for changes.", self.name)
+        async for event in file_watcher:
+            _logger.info(
+                "%s: Reloading model from file %s due to a %s event...",
+                self.name,
+                event.path,
+                event.type.name,
+            )
+            self.reload_model(Path(event.path))
+
+    def reload_model(self, path: Path) -> None:
+        """Reload the model from the specified path.
+
+        Args:
+            path: The path to the model file.
+        """
+        for key, model in self._models.items():
+            if model.path == path:
+                try:
+                    model.data = self._load(path)
+                    _logger.info(
+                        "%s: Successfully reloaded model from %s",
+                        self.name,
+                        path,
+                    )
+                except Exception:  # pylint: disable=broad-except
+                    _logger.exception("Failed to reload model from %s", path)
+
+    def get_model(self, key: str) -> T:
+        """Retrieve a loaded model by key.
+
+        Args:
+            key: The key of the model to retrieve.
+
+        Returns:
+            The loaded model data.
+
+        Raises:
+            KeyError: If the model with the specified key is not found.
+        """
+        try:
+            return self._models[key].data
+        except KeyError as exc:
+            raise KeyError(f"Model with key '{key}' is not found.") from exc
--- a/tests/model/test_model_manager.py
+++ b/tests/model/test_model_manager.py
@@ -1,0 +1,118 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""Tests for machine learning model manager."""
+
+import pickle
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+from unittest.mock import AsyncMock, MagicMock, mock_open, patch
+
+import pytest
+
+from frequenz.sdk.ml import ModelManager
+
+
+@dataclass
+class MockModel:
+    """Mock model for unit testing purposes."""
+
+    data: int | str
+
+    def predict(self) -> int | str:
+        """Make a prediction based on the model data."""
+        return self.data
+
+
+async def test_model_manager_loading() -> None:
+    """Test loading models using ModelManager with direct configuration."""
+    model1 = MockModel("Model 1 Data")
+    model2 = MockModel("Model 2 Data")
+    pickled_model1 = pickle.dumps(model1)
+    pickled_model2 = pickle.dumps(model2)
+
+    model_paths = {
+        "model1": Path("path/to/model1.pkl"),
+        "model2": Path("path/to/model2.pkl"),
+    }
+
+    mock_files = {
+        "path/to/model1.pkl": mock_open(read_data=pickled_model1)(),
+        "path/to/model2.pkl": mock_open(read_data=pickled_model2)(),
+    }
+
+    def mock_open_func(file_path: Path, *__args: Any, **__kwargs: Any) -> Any:
+        """Mock open function to return the correct mock file object.
+
+        Args:
+            file_path: The path to the file to open.
+            *__args: Variable length argument list. This can be used to pass additional
+                    positional parameters typically used in file opening operations,
+                    such as `mode` or `buffering`.
+            **__kwargs: Arbitrary keyword arguments. This can include parameters like
+                    `encoding` and `errors`, common in file opening operations.
+
+        Returns:
+            Any: The mock file object.
+
+        Raises:
+            FileNotFoundError: If the file path is not in the mock files dictionary.
+        """
+        file_path_str = str(file_path)
+        if file_path_str in mock_files:
+            file_handle = MagicMock()
+            file_handle.__enter__.return_value = mock_files[file_path_str]
+            return file_handle
+        raise FileNotFoundError(f"No mock setup for {file_path_str}")
+
+    with patch("pathlib.Path.open", new=mock_open_func):
+        with patch.object(Path, "exists", return_value=True):
+            model_manager: ModelManager[MockModel] = ModelManager(
+                model_paths=model_paths
+            )
+
+        with patch(
+            "frequenz.channels.file_watcher.FileWatcher", new_callable=AsyncMock
+        ):
+            model_manager.start()  # Start the service
+
+            assert isinstance(model_manager.get_model("model1"), MockModel)
+            assert model_manager.get_model("model1").data == "Model 1 Data"
+            assert model_manager.get_model("model2").data == "Model 2 Data"
+
+            with pytest.raises(KeyError):
+                model_manager.get_model("key3")
+
+            await model_manager.stop()  # Stop the service to clean up
+
+
+async def test_model_manager_update() -> None:
+    """Test updating a model in ModelManager."""
+    original_model = MockModel("Original Data")
+    updated_model = MockModel("Updated Data")
+    pickled_original_model = pickle.dumps(original_model)
+    pickled_updated_model = pickle.dumps(updated_model)
+
+    model_paths = {"model1": Path("path/to/model1.pkl")}
+
+    mock_file = mock_open(read_data=pickled_original_model)
+    with (
+        patch("pathlib.Path.open", mock_file),
+        patch.object(Path, "exists", return_value=True),
+    ):
+        model_manager = ModelManager[MockModel](model_paths=model_paths)
+        with patch(
+            "frequenz.channels.file_watcher.FileWatcher", new_callable=AsyncMock
+        ):
+            model_manager.start()  # Start the service
+
+            assert model_manager.get_model("model1").data == "Original Data"
+
+            # Simulate updating the model file
+            mock_file.return_value.read.return_value = pickled_updated_model
+            with patch("pathlib.Path.open", mock_file):
+                model_manager.reload_model(Path("path/to/model1.pkl"))
+                assert model_manager.get_model("model1").data == "Updated Data"
+
+            await model_manager.stop()  # Stop the service to clean up
--- a/src/frequenz/sdk/ml/__init__.py
+++ b/src/frequenz/sdk/ml/__init__.py
@@ -1,0 +1,10 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""Model interface."""
+
+from ._model_manager import ModelManager
+
+__all__ = [
+    "ModelManager",
+]
--- a/tests/model/__init__.py
+++ b/tests/model/__init__.py
@@ -1,0 +1,4 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""Tests for the model package."""


llucax:
Maybe this was talked elsewhere, but it might be nice to provide some extra context about what changed between the closing of #397 and now. Was the experiment inside the actor successful and we think this new slimmed down implementation is good/useful enough to go in the SDK?

idlir-shkurti-frequenz:
> Maybe this was talked elsewhere, but it might be nice to provide some extra context about what changed between the closing of #397 and now. Was the experiment inside the actor successful and we think this new slimmed down implementation is good/useful enough to go in the SDK?

@llucax We decided that it would be best to start the model loading immediately inside the sdk so that we don't have to migrate extra code later. This slimmed down version should be good enough for the initial usecases. We don't need extra `weeday` and `enumerated` repositories but only a way to load and monitor ML models from a pre-defined path.

src/frequenz/sdk/model/__init__.py
```diff
@@ -0,0 +1,19 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
```
llucax: Nitpick: I guess this is 2023 because the code it was taken from was from 2023, right?

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,273 @@
...
+        Returns:
+            the model keys to model paths mapping.
+        """
+        allowed_keys = self._get_allowed_keys()
+        assert allowed_keys, "Allowed keys must be set."
+        return {key: self._build_path(key) for key in allowed_keys}
+
+
+@dataclass
+class SingleModelRepository(ModelRepository):
```
llucax:
```suggestion
class SingleModelRepository(ModelRepository[str]):
```

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,273 @@
...
+    """Repository to manage a single model."""
+
+    model_path: str
```
llucax:
```suggestion
    model_path: pathlib.Path
```

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,273 @@
...
+    """The path to a single model file."""
+
+    def _get_allowed_keys(self) -> list[Any]:
```
llucax:
```suggestion
    def _get_allowed_keys(self) -> collections.abc.Set[str]:
```

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,273 @@
...
+
+from frequenz.sdk._internal._asyncio import cancel_and_await
+
+_logger = logging.getLogger(__name__)
+
+T = TypeVar("T")
+"""The model type."""
+
+
+class ModelRepository(abc.ABC):
```
llucax: Can't we parametrize keys in this class with `T` too? I think we should avoid using `Any` as it completely breaks the type system.

```suggestion
class ModelRepository(abc.ABC, Generic[T]):
```

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,273 @@
...
+    """Provide an interface for building the model repository configuration.
+
+    The ModelRepository is meant to be subclassed to provide specific
+    implementations based on the type of models being managed.
+    """
+
+    @abc.abstractmethod
+    def _get_allowed_keys(self) -> list[Any]:
```
llucax:
```suggestion
    def _get_allowed_keys(self) -> collections.abc.Set[T]:
```

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,273 @@
...
+        Returns:
+            a list of allowed keys.  # noqa: DAR202
+
+        Raises:
+            NotImplementedError: if the method is not implemented in the subclass.
+        """
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def _build_path(self, key: Any) -> str:
```
llucax: I would use `pathlib.Path` instead of `str` for paths.

```suggestion
    def _build_path(self, key: T) -> pathlib.Path:
```

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,273 @@
...
+
+        Returns:
+            the model path.  # noqa: DAR202
+
+        Raises:
+            NotImplementedError: if the method is not implemented in the subclass.
+        """
+        raise NotImplementedError
+
+    def get_models_config(self) -> dict[Any, str]:
```
llucax:
```suggestion
    def get_models_config(self) -> dict[T, pathlib.Path]:
```

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,273 @@
...
+
+    def _get_allowed_keys(self) -> list[Any]:
+        """Get the allowed keys for the single model repository.
+
+        Returns:
+            the allowed single key.
+        """
+        return ["Single"]
+
+    def _build_path(self, key: Any) -> str:
```
llucax:
```suggestion
    def _build_path(self, key: str) -> pathlib.Path:
```

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,273 @@
...
+            key: the key for which to build the model path.
+
+        Returns:
+            the model path.
+        """
+        return self.model_path
+
+
+@dataclass
+class DirectModelRepository(ModelRepository):
```
llucax:
```suggestion
class DirectModelRepository(ModelRepository[str]):
```

Also, I'm not sure about the name, why direct? This comes from when we had repos for weekday, etc., right?

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,273 @@
...
+    """Repository to manage models based on a direct configuration dictionary."""
+
+    model_config: dict[str, str]
```
llucax:
```suggestion
    model_config: dict[str, pathlib.Path]
```

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,273 @@
...
+    """A dictionary mapping keys to the full paths of model files."""
+
+    def _get_allowed_keys(self) -> list[str]:
```
llucax:
```suggestion
    def _get_allowed_keys(self) -> collections.abc.Set[str]:
```

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,273 @@
...
+        """Get the list of allowed keys for the model repository.
+
+        Returns:
+            A list of keys as specified in the model_config dictionary.
+        """
+        assert self.model_config, "Model configuration must be set."
+        return list(self.model_config.keys())
```
llucax:
```suggestion
        return frozenset(self.model_config.keys())
```

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,273 @@
...
+
+    def _build_path(self, key: str) -> str:
```
llucax:
```suggestion
    def _build_path(self, key: str) -> pathlib.Path:
```

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,273 @@
...
+    """
+
+    def __init__(self, model_repository: ModelRepository) -> None:
+        """Initialize the model manager.
+
+        Args:
+            model_repository: a model repository that defines the mapping
+                between model keys and paths.
+        """
+        self._models: dict[Any, _Model[T]] = {}
```
llucax: Here we could add another parameter, like `U`, for the key type, or use `object` if we want to make it an opaque type. The difference between using `object` and `Any` is `Any` will pass any type checking, like `key.i_dont_exist`, while object will raise an error. Since the model repository has a typed key, I think it makes more sense to go with the extra type parameter.
```suggestion
        self._models: dict[U, _Model[T]] = {}
```

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,273 @@
...
+    """Represent a machine learning model."""
+
+    data: T
+    """The machine learning model."""
+
+    path: str
+    """The path to the model file."""
+
+
+class ModelManager(Generic[T]):
```
llucax:
```suggestion
class ModelManager(Generic[T, U]):
```

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,273 @@
...
+
+        models_config = model_repository.get_models_config()
+        for key, model_path in models_config.items():
+            self._models[key] = _Model(data=self._load(model_path), path=model_path)
+
+        self._monitoring_task: asyncio.Task[None] = asyncio.create_task(
+            self._start_model_monitor()
+        )
+
+    def __getitem__(self, key: Any) -> T:
```
llucax:
```suggestion
    def __getitem__(self, key: U) -> T:
```

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,273 @@
...
+
+        Args:
+            key: the key to identify the model.
+
+        Returns:
+            the model instance corresponding to the key.
+        """
+        return self.get_model(key)
+
+    def get_model(self, key: Any = None) -> T:
```
llucax: To be correct, this needs to get a bit more complicated if we make keys generic. We should use a sentinel class, like:

```py
class Unspecified:
    pass
```

```suggestion
    def get_model(self, key: U = Unspecified) -> T:
```

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,273 @@
...
+            if len(self._models) != 1:
+                raise KeyError("No key provided for multi-model retrieval")
+            key = list(self._models.keys())[0]
+
+        if key not in self._models:
+            raise KeyError("Invalid key")
+
+        return self._models[key].data
+
+    def get_paths(self) -> list[Path | str]:
```
llucax: I would return an iterator here, as it is probably all we need and makes the interface more flexible. Also I would keep it simple and only return `pathlib.Path`:
```suggestion
    def get_paths(self) -> Iterator[Path]:
```

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,273 @@
...
+            KeyError: if the key is invalid or it is not provided for
+                multi-model retrieval.
+
+        Returns:
+            the model instance corresponding to the key.
+        """
+        if key is None:
+            if len(self._models) != 1:
+                raise KeyError("No key provided for multi-model retrieval")
+            key = list(self._models.keys())[0]
```
llucax:
```suggestion
        if key is Unspecified:
            if len(self._models) != 1:
                raise KeyError("A key must be provided")
            key = list(self._models.keys())[0]
```

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,273 @@
...
+
+        return self._models[key].data
+
+    def get_paths(self) -> list[Path | str]:
+        """Get the paths of all loaded models.
+
+        Returns:
+            the paths of all loaded models.
+        """
+        return [model.path for model in self._models.values()]
```
llucax: This returns an iterator instead of a list:
```suggestion
        return (model.path for model in self._models.values())
```

llucax (commented):
I'm not crazy about the `get_model()` interface because it is too dynamic, it would be good if it could be more static and misuse would be detected by `mypy` instead of at runtime.

Looking at the bigger picture here, it seems this code is still inheriting a lot of the complexity of the previous PR. Do we really need a special case for a single model repository? Can't we just use the dict-based repo and let user pass a key with the value they like the most? It seems like a lot of added extra complexity just to avoid passing a key.

I would just remove the special case here and also the `SingleModelRepository` class and force users to pass a key. Once this is done, I don't think it makes sense to keep the abstract base class `ModelRepository`, and if we remove it, we can just rename `DirectModelRepository` to `ModelRepository`. This will make this PR much much simpler.

If we need more abstraction and more types of repositories in the future, we can bring the complexity back, but I wouldn't bet on it just yet.

src/frequenz/sdk/model/__init__.py
idlir-shkurti-frequenz: You are right. I will change it

idlir-shkurti-frequenz:
> Looking at the bigger picture here, it seems this code is still inheriting a lot of the complexity of the previous PR. Do we really need a special case for a single model repository? Can't we just use the dict-based repo and let user pass a key with the value they like the most? It seems like a lot of added extra complexity just to avoid passing a key.

That is actually something I was also considering. @daniel-zullo-frequenz @cwasicki I think I agree with Luca here that we can just have the one repo structure with a dict input and forget about the single model repo structure. If that is ok with you I will make the change.

src/frequenz/sdk/model/__init__.py
llucax: Either way is fine with me.

cwasicki:
>If that is ok with you I will make the change.

Agree. I think we can also limit it to the manager and don't need the repositories.

cwasicki:
I suggest for now to only implement the bare minimum functionality that we currently need.

llucax:
> Agree. I think we can also limit it to the manager and don't need the repositories.

I think having only the "dict repository" is simple enough and gives more flexibility to separate different models if we need to, also the manager and repository are very closely related, so doing this would basically mean throwing everything and start again, right?

IMHO having `DirectModelRepository` (as `ModelRepository`) and the `ModelManager` that can monitor more than one `ModelRepository` strikes a good balance between simplicity and flexibility. But if you think that is still over-bloated for our current needs, I'm also OK to remove the repositories altogether.

idlir-shkurti-frequenz:
@llucax @cwasicki I made some changes to the PR. I removed the repository structures for simplicity and changed the `ModelManager` to require a `model_config` in its initialisation. This `model_config` is simply a `dict[str, Path]` pointing to the models. 

@llucax I changed the key from `Any` to `str` and not as a generic like you proposed. Do you think it is important for now to make them generic? This will simply be the name of the model and I would assume as key is good enough for now. If you think it still makes sense to change them to generic then let me know.

llucax:
> I changed the key from Any to str and not as a generic like you proposed. Do you think it is important for now to make them generic?

TL;DR: I think it is OK. It was a journey to convince myself of that, so I will keep the long version for future reference :D

I was going to say it's not important in this context, but then I was thinking about the case where we want to use an enum for safety, as `str` is very prone to errors due to typos for example, which can't be detected by linters or `mypy`, so they will only occur at runtime.

This is why I still liked the separation between the repository and the manager, because then you can have a "weekly" model, where keys are a enum with the week days, and then you can have an arbitrary model with some `str` keys that you build more dynamically.

I think it would be good to encourage users to design their code with safety in mind, so I'm not sure. I understand this is good enough for the current use cases but I feel it might be a move in the wrong direction.

But I'm not sure if it is worth to provide a model repository that it is just a `dict` with `str` keys, because then most users will just use that and will have the same issues. So it probably makes sense to add the safety once we want to provide safer repositories, like a day of the week repo, where keys can only be members of a `DayOfTheWeek` enum for example, etc., as we discussed originally.

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,98 @@
...
+    """Represent a machine learning model."""
+
+    data: T
+    path: Path
+
+
+class ModelManager(Generic[T]):
+    """Manage machine learning models with direct dictionary configuration."""
+
+    def __init__(self, model_config: dict[str, Path]) -> None:
```
llucax: I would still call the mapping of model to path a *repository*.

```suggestion
    def __init__(self, repository: dict[str, Path]) -> None:
```

Or maybe `model_paths`, but `config` sounds weird in this context, for me it sounds like it would configure how often they are checked for changes or stuff like that, configuration for the manager itself.

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,98 @@
...
+                )
+                self.reload_model(Path(event.path))
+
+    def reload_model(self, model_path: Path) -> None:
+        """Reload a model from the given path."""
+        for key, model in self._models.items():
+            if model.path == model_path:
+                model.data = self._load(model_path)
+                return
+        raise ValueError(f"No model found with path: {model_path}")
```
llucax: I wouldn't raise an exception here, otherwise one error will make the manager stop processing any other change notifications.

Even more, I think there could be races here where you might get a create notification for example, but the user removed it immediately so the file will not exist when you want to open it, so I would add a `try/except` block here catching *at least* any `OSError`, logging the error and returning, so the loop can continue processing events.

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,98 @@
...
+
+    async def join(self) -> None:
+        """Await the monitoring task, and return when the task completes."""
+        if self._monitoring_task and not self._monitoring_task.done():
+            await self._monitoring_task
+
+    async def _stop_model_monitor(self) -> None:
+        """Stop monitoring the model paths for changes."""
+        if self._monitoring_task:
+            await cancel_and_await(self._monitoring_task)
```
llucax: Oh, this makes me think this class should be a [`BackgroundService`](https://frequenz-floss.github.io/frequenz-sdk-python/v0.25/reference/frequenz/sdk/actor/#frequenz.sdk.actor.BackgroundService) instead, then you will get the `join()` and `stop()` methods automatically, and we keep a consistent interface.

tests/model/test_model_manager.py
```diff
@@ -0,0 +1,111 @@
...
+
+    model_config = {
+        "model1": Path("path/to/model1.pkl"),
+        "model2": Path("path/to/model2.pkl"),
+    }
+
+    # Map file paths to their respective mock file objects with corresponding data
+    mock_files = {
+        "path/to/model1.pkl": mock_open(read_data=pickled_model1),
+        "path/to/model2.pkl": mock_open(read_data=pickled_model2),
```
llucax: Interesting, I didn't know this `mock_open` function.

llucax (commented):
I got a bit lost in the patching and mock files in the tests, so I will have a look at it later.

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,98 @@
...
+        """Retrieve a loaded model by key."""
+        try:
+            return self._models[key].data
+        except KeyError as exc:
+            raise KeyError(f"Model with key '{key}' is not found.") from exc
+
+    async def _start_model_monitor(self) -> None:
+        """Start monitoring the model paths for changes."""
+        model_paths = [model.path for model in self._models.values()]
+        file_watcher = FileWatcher(paths=[str(path) for path in model_paths])
```
cwasicki: The `str` should not be required.

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,98 @@
...
+        """Load the model from the specified path."""
+        if not model_path.exists():
+            raise FileNotFoundError(f"The model path {model_path} does not exist.")
+
+        with model_path.open("rb") as file:
+            model = pickle.load(file)
+            return cast(T, model)
+
+    def get_model(self, key: str) -> T:
+        """Retrieve a loaded model by key."""
```
cwasicki: The doc in this module is missing for arguments and return values.

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,98 @@
...
+                    "Model file %s has been modified. Reloading model...", event.path
+                )
+                self.reload_model(Path(event.path))
+
+    def reload_model(self, model_path: Path) -> None:
+        """Reload a model from the given path."""
+        for key, model in self._models.items():
+            if model.path == model_path:
+                model.data = self._load(model_path)
+                return
```
cwasicki: This assumes that each path exists only for one key. Not sure if we want to be so strict about this.

cwasicki: If we do, we should verify this somewhere and also document it.

src/frequenz/sdk/model/_manager.py
```diff
@@ -0,0 +1,98 @@
...
+_logger = logging.getLogger(__name__)
+
+T = TypeVar("T")
+
+
+@dataclass
+class _Model(Generic[T]):
+    """Represent a machine learning model."""
+
+    data: T
```
cwasicki: I wonder whether we want to be more specific about the model implementation. For instance the `data` will most likely be more complex, e.g. a dictionary that holds a model, feature names and other metadata.

In the current version, there is nothing in the `_Model` neither in the `ModelManager` that is "model"-specific. In fact this module is a `PickleManager`. Don't think it's a bad thing, but maybe we could be more clear about this if we keep the generic approach.

src/frequenz/sdk/model/_manager.py
cwasicki, line 1: Maybe a more generic namespace like `ml` is better suited? That would also cover other things like the features module that we want to move to the SDK.

cwasicki (changes requested):

src/frequenz/sdk/model/_manager.py
idlir-shkurti-frequenz: I think `model_paths` is better. I will change it to that.

src/frequenz/sdk/model/_manager.py
idlir-shkurti-frequenz: Just to clarify we wouldn't have `join()` right? We would have `wait()` instead?

idlir-shkurti-frequenz: Just to clarify we wouldn't have `join()` right? We would use `wait()` instead?

idlir-shkurti-frequenz: Ok I just read a bit more on `BackgroundService`. We wouldn't even need to implement it at all I see since it is already taken care of.

llucax: Exactly. You would be able to use it as an async context manager for example (`async with ModelManager(...):`) or await on it directly (`await model_manager`).

src/frequenz/sdk/model/_manager.py
llucax: This is allowed by the google style guide provided that you use `Return ...` and it is clear enough from the summary what the arguments should be. In this case I think it is equivalent, but the tool might require `Return ...` instead to adhere more strictly with the google style, so I would update it. Still this probably needs to document that it raises an exception if the key is not found, so it might be better to do the full doc here.

src/frequenz/sdk/model/_manager.py
llucax: I agree but think we don't need to do it right now. Not calling it `PickleManager` allows us to grow in time with model-specific stuff.

src/frequenz/sdk/model/_manager.py
cwasicki: Oh, I didn't know that. In this case I would be fine with this, if there is one place that explains in a bit more detail what the model manager is doing. E.g. things like unique keys, model pathes, what happens if a model is reloaded.

src/frequenz/sdk/model/_manager.py
idlir-shkurti-frequenz: Done!

src/frequenz/sdk/model/_manager.py
idlir-shkurti-frequenz: I extended the docstring here with the arguments and return value

src/frequenz/sdk/ml/_model_manager.py
```diff
@@ -0,0 +1,136 @@
...
+        self._models: dict[str, _Model[T]] = {}
+        self.model_paths = model_paths
+        self.load_models()
+
+    def load_models(self) -> None:
+        """Load the models from the specified paths."""
+        for key, path in self.model_paths.items():
+            self._models[key] = _Model(data=self._load(path), path=path)
+
+    def _load(self, path: Path) -> T:
```
cwasicki: This can be made a static function

src/frequenz/sdk/model/_manager.py
idlir-shkurti-frequenz: Done

src/frequenz/sdk/ml/_model_manager.py
idlir-shkurti-frequenz: Done!

src/frequenz/sdk/ml/_model_manager.py
```diff
@@ -0,0 +1,137 @@
...
+
+T = TypeVar("T")
+
+
+@dataclass
+class _Model(Generic[T]):
+    """Represent a machine learning model."""
+
+    data: T
+    path: Path
```
llucax, line 28: Ideally these members should be documented, but I guess being internal and so simple it is OK to skip.

src/frequenz/sdk/ml/_model_manager.py
```diff
@@ -0,0 +1,137 @@
...
+class ModelManager(BackgroundService, Generic[T]):
+    """Load, update, monitor and retrieve machine learning models."""
+
+    def __init__(self, model_paths: dict[str, Path]):
+        """Initialize the model manager with the specified model paths.
+
+        Args:
+            model_paths: A dictionary of model keys and their corresponding file paths.
+        """
+        super().__init__()
```
llucax: I recommend passing a `name` here (and take an optional name in the constructor too). It is used to print some logs by `BackgroundService`.

src/frequenz/sdk/ml/_model_manager.py
```diff
@@ -0,0 +1,137 @@
...
+        Returns:
+            T: The loaded model data.
+
+        Raises:
+            FileNotFoundError: If the model file does not exist.
+        """
+        if not path.exists():
+            raise FileNotFoundError(f"The model path {path} does not exist.")
+        with path.open("rb") as file:
+            return cast(T, pickle.load(file))
```
llucax: Oh damn, I just realized this file reading is not async and can potentially block the event loop for a non trivial amount of time, which is not ideal. Python's built-in `asyncio` doesn't support async file I/O, so we would need to use another library for this (like [aiofiles](https://pypi.org/project/aiofiles/)). I created a separate issue because this is way beyond this PR:

* https://github.com/frequenz-floss/frequenz-sdk-python/issues/929

src/frequenz/sdk/ml/_model_manager.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def start(self) -> None:
```
llucax, line 83:
```suggestion
    @override
    def start(self) -> None:
```

src/frequenz/sdk/ml/_model_manager.py
```diff
@@ -0,0 +1,137 @@
...
+        """Start the model monitoring service by creating a background task."""
+        if not self.is_running:
+            task = asyncio.create_task(self.run())
+            self._tasks.add(task)
+            _logger.info("Started ModelManager service with task %s", task)
+
+    async def run(self) -> None:
```
llucax:
```suggestion
    @override
    async def run(self) -> None:
```

src/frequenz/sdk/ml/_model_manager.py
```diff
@@ -0,0 +1,137 @@
...
+        """Monitor model file paths and reload models as necessary."""
+        model_paths = [model.path for model in self._models.values()]
+        file_watcher = FileWatcher(paths=list(model_paths))
```
llucax:
```suggestion
        file_watcher = FileWatcher(paths=model_paths)
```

This is already a list (when using `[...]` a list is produced, when using `(...)` a generator is produced).

src/frequenz/sdk/ml/_model_manager.py
```diff
@@ -0,0 +1,137 @@
...
+        _logger.info("Monitoring model paths for changes.")
+        async for event in file_watcher:
+            if event.type in (EventType.CREATE, EventType.MODIFY):
```
llucax: You can pass the events to watch for to the `FileWatcher()` constructor, then you don't need to check in here.

src/frequenz/sdk/ml/_model_manager.py
```diff
@@ -0,0 +1,137 @@
...
+        if not self.is_running:
+            task = asyncio.create_task(self.run())
+            self._tasks.add(task)
+            _logger.info("Started ModelManager service with task %s", task)
+
+    async def run(self) -> None:
+        """Monitor model file paths and reload models as necessary."""
+        model_paths = [model.path for model in self._models.values()]
+        file_watcher = FileWatcher(paths=list(model_paths))
+        _logger.info("Monitoring model paths for changes.")
```
llucax:
```suggestion
        _logger.info("%s: Monitoring model paths for changes.", self.name)
```

src/frequenz/sdk/ml/_model_manager.py
```diff
@@ -0,0 +1,137 @@
...
+        async for event in file_watcher:
+            if event.type in (EventType.CREATE, EventType.MODIFY):
+                _logger.info("Model file %s modified, reloading...", event.path)
```
llucax:
```suggestion
                _logger.info("%s: Model file %s modified, reloading...", self.name, event.path)
```

src/frequenz/sdk/ml/_model_manager.py
```diff
@@ -0,0 +1,137 @@
...
+            raise FileNotFoundError(f"The model path {path} does not exist.")
+        with path.open("rb") as file:
+            return cast(T, pickle.load(file))
+
+    def start(self) -> None:
+        """Start the model monitoring service by creating a background task."""
+        if not self.is_running:
+            task = asyncio.create_task(self.run())
+            self._tasks.add(task)
+            _logger.info("Started ModelManager service with task %s", task)
```
llucax:
```suggestion
            _logger.info("%s: Started ModelManager service with task %s", self.name, task)
```

src/frequenz/sdk/ml/_model_manager.py
```diff
@@ -0,0 +1,137 @@
...
+        """Reload the model from the specified path.
+
+        Args:
+            path: The path to the model file.
+        """
+        for key, model in self._models.items():
+            if model.path == path:
+                try:
+                    model.data = self._load(path)
+                    _logger.info("Successfully reloaded model from %s", path)
```
llucax: Use `self.name` here and in all other logging too. We might want to add some method to background service to do this automatically somehow, but that's for a separate issue.

src/frequenz/sdk/ml/_model_manager.py
```diff
@@ -0,0 +1,137 @@
...
+                except Exception as e:  # pylint: disable=broad-except
+                    _logger.error("Failed to reload model from %s: %s", path, e)
```
llucax:
```suggestion
                except Exception:  # pylint: disable=broad-except
                    _logger.exception("Failed to reload model from %s", path)
```

src/frequenz/sdk/ml/_model_manager.py
```diff
@@ -0,0 +1,137 @@
...
+                _logger.info("Task %s cancelled successfully", task)
+            except Exception as e:  # pylint: disable=broad-except
+                _logger.error("Error while cancelling task %s: %s", task, e)
+
+        # Call the parent stop method if it handles additional teardown
+        try:
+            await super().stop(msg)
+        except Exception as e:  # pylint: disable=broad-except
+            _logger.error("Error during stop in superclass: %s", e)
+
```
llucax: This is done automatically by `BackgroundService`.

```suggestion
```

src/frequenz/sdk/ml/_model_manager.py
idlir-shkurti-frequenz: Done!

llucax: You still need to forward the name to the `BackgroundService`.
```suggestion
        super().__init__(name=name)
```

src/frequenz/sdk/ml/_model_manager.py
```diff
@@ -0,0 +1,130 @@
...
+        """Initialize the model manager with the specified model paths.
+
+        Args:
+            model_paths: A dictionary of model keys and their corresponding file paths.
+            name: The name of the model manager service.
+        """
+        super().__init__()
+        self._models: dict[str, _Model[T]] = {}
+        self.model_paths = model_paths
+        self._service_name = name
```
llucax: And remove this.
```suggestion
```

src/frequenz/sdk/ml/_model_manager.py
```diff
@@ -0,0 +1,130 @@
...
+    """Represent a machine learning model."""
+
+    data: T
+    path: Path
+
+
+class ModelManager(BackgroundService, Generic[T]):
+    """Load, update, monitor and retrieve machine learning models."""
+
+    def __init__(self, model_paths: dict[str, Path], name: str = "ModelManager"):
```
llucax: I'd leave the default as `None`, the `BackgroundService` can deal with that ("If `None`, `str(id(self))` will be used." [^1]). Also, `str(x)` will use the class name too, so if you use the class name it will look very redundant, like `ModelManager(ModelManager)`.

Accepting a name is only to allow users to customize this, in case they instantiate more than one model manager, it's more so for example, peak shaving can do `mm = ModelManager(paths, name="PeakShaving")`. I would also make `name` a keyword-only argument, like in `BackgroudService`, to avoid confusion:

```suggestion
    def __init__(self, model_paths: dict[str, Path], *, name: str | None = None):
```

[^1]: https://frequenz-floss.github.io/frequenz-sdk-python/v0.25/reference/frequenz/sdk/actor/#frequenz.sdk.actor.BackgroundService.__init__

src/frequenz/sdk/ml/_model_manager.py
```diff
@@ -0,0 +1,130 @@
...
+        file_watcher = FileWatcher(
+            paths=model_paths, event_types=[EventType.CREATE, EventType.MODIFY]
+        )
+        _logger.info("%s: Monitoring model paths for changes.", self._service_name)
+        async for event in file_watcher:
+            _logger.info(
+                "%s: Model file %s modified, reloading...",
+                self._service_name,
+                event.path,
+            )
```
llucax: This might be misleading, maybe this?
```suggestion
            _logger.info(
                "%s: Reloading model from file %s due to a %s event...",
                self.name,
                event.path,
                event.type.name,
            )
```

src/frequenz/sdk/ml/_model_manager.py
```diff
@@ -0,0 +1,130 @@
...
+    def start(self) -> None:
+        """Start the model monitoring service by creating a background task."""
+        if not self.is_running:
+            task = asyncio.create_task(self._monitor_paths())
+            self._tasks.add(task)
+            _logger.info(
+                "%s: Started ModelManager service with task %s",
+                self._service_name,
+                task,
+            )
```
llucax: The name is provided by the `BackgroundService` [`name` property](https://frequenz-floss.github.io/frequenz-sdk-python/v0.25/reference/frequenz/sdk/actor/#frequenz.sdk.actor.BackgroundService.name):
```suggestion
            _logger.info(
                "%s: Started ModelManager service with task %s",
                self.name,
                task,
            )
```

(same change for all other logging sites)

llucax (commented):
Sorry, a few more comments, mostly about how to subclass `BackgroundService` :grimacing:

src/frequenz/sdk/ml/_model_manager.py
idlir-shkurti-frequenz: Ahh ok I see. I wasn't sure how this worked

src/frequenz/sdk/ml/_model_manager.py
idlir-shkurti-frequenz: Yes that makes sense. I will change it!

idlir-shkurti-frequenz: Done!

src/frequenz/sdk/ml/_model_manager.py
idlir-shkurti-frequenz: Done!

src/frequenz/sdk/ml/_model_manager.py
idlir-shkurti-frequenz: Done!

src/frequenz/sdk/ml/_model_manager.py
idlir-shkurti-frequenz: Done!

llucax: > Ahh ok I see. I wasn't sure how this worked

Did you check the example at https://frequenz-floss.github.io/frequenz-sdk-python/v0.25/reference/frequenz/sdk/actor/#frequenz.sdk.actor.BackgroundService?

llucax: If you did and it was still unclear we might need to improve the docs!

llucax (dismissed):
Thanks for the patience!

pyproject.toml
```diff
@@ -96,6 +96,7 @@ dev-pytest = [
   # For checking docstring code examples
   "frequenz-sdk[dev-examples]",
   "hypothesis == 6.100.0",
+  "sybil == 6.0.3",
```
llucax: This should not be necessary anymore, we fixed it in repo-config.

src/frequenz/sdk/ml/_model_manager.py
```diff
@@ -0,0 +1,130 @@
...
+            path: The path to the model file.
+
+        Returns:
+            T: The loaded model data.
+
+        Raises:
+            FileNotFoundError: If the model file does not exist.
+        """
+        if not path.exists():
+            raise FileNotFoundError(f"The model path {path} does not exist.")
```
llucax: I would just let `open()` raise the error:
```suggestion
```

If you want to produce a more descriptive/specific error, I suggest creating a custom exception instead and chaining it to the orignal exception so users can get the full context:

```py
class ModelError(Exception):
    # You might want to actually add more details here, like `model_key` and `model_path`
    pass
```    
```suggestion
        try:
            with path.open("rb") as file:
                return cast(T, pickle.load(file))
        except Exception as err:
            raise ModelError(f"Could not load model {model_key} from {model_path}: {e}.", model_key, model_path) from err
```

llucax (commented):
One last comment while we wait for the release of `rc6.post1` :)

cwasicki:
@idlir-shkurti-frequenz Could you rebase this PR please?

src/frequenz/sdk/ml/_model_manager.py
idlir-shkurti-frequenz: Done!

idlir-shkurti-frequenz:
> @idlir-shkurti-frequenz Could you rebase this PR please?

Done! Also addressed the custom exception point suggested by Luca.

cwasicki (approved):

llucax (approved):

shsms:
Looks like the commit wasn't signed, someone might have to create a new PR, ideally with a shorter commit message.

cwasicki:
Ok, then I will do some testing with this and re-create it afterwards.

cwasicki:
https://github.com/frequenz-floss/frequenz-sdk-python/issues/954

llucax:
> Looks like the commit wasn't signed, someone might have to create a new PR, ideally with a shorter commit message.

Didn't get what's going on, it needs to be re-created because Idlir can't do it himself now? We can still amend this PR and force-push to it. Also I don't get the comment about the commit message length.

shsms:
> Also I don't get the comment about the commit message length.

a shorter one that doesn't fold on gh would be nice.

llucax:
Oh, OK, so unrelated to the signing part, I thought it was related. :+1:

cwasicki:
Created a backup here: https://github.com/frequenz-floss/frequenz-sdk-python/pull/968

PR #919, opened 2024-04-08
Some EV charger pool cleanup and fixes

shsms:


shsms:
Skip release notes because these are internal changes and bugfixes to unreleased features.

llucax (approved):
LGTM, the only question is why instantiating a `EVChargerManager` at all if there are no EV chargers, but I assume you thought about it and have a good reason for it, so I will give the approval anyway, but I'm curious about it.

shsms:
> the only question is why instantiating a EVChargerManager at all if there are no EV chargers

It is just that this is where we already have the list of available component IDs.  Otherwise, we'd have to get the component IDs for all component types in the `_data_pipeline.py` file, which feels less modular.

Also, they are lazily started, so an EVChargerManager or even a PowerDistributor instance is only started when a corresponding call is made to `microgrid.ev_charger_pool()`.

PR #926, opened 2024-04-12
Clear release notes

llucax:

PR #927, opened 2024-04-12
Fix reactive power retrieval

llucax:
The data sourcing actor needs to know about the new metric. This commit exposes it, adds a trivial test and updates the release notes to do a hotfix right after this is merged.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,42 +2,16 @@
 
 ## Summary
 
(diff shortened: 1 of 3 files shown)


llucax:
Maybe give it a test too @matthias-wende-frequenz ? It seems to work for me locally, but just in case.

llucax:
(based on #926, but I guess both can be merged via this PR too)

matthias-wende-frequenz (approved):

PR #930, opened 2024-04-19
Ignore pylint check to be able to release

llucax:
The reactive power test is incomplete as it is not really used or verified.

Fixing this takes quite a bit of time, as we need to patch the `MockMicrogrid` which is taking more time than expected, so for now we are just ignoring the return value here to avoid a `pylint` CI failure preventing the release.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,16 +2,4 @@
 
 ## Summary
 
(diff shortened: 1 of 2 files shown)


llucax:
Enabled auto-merge.

daniel-zullo-frequenz (approved):

Issue #832, opened 2023-12-29
Upgrade to `time-machine` 2.13.0+ breaks our tests


llucax:
### What happened?

The tests, for example `tests/actor/test_battery_status.py::TestBatteryStatus::test_async_battery_status`, get stuck waiting forever if we upgrade to 2.13.0+.

### What did you expect instead?

Tests work as with 2.12.0.

### Affected part(s)

Unit, integration and performance tests (part:tests), Build script, CI, dependencies, etc. (part:tooling)

### Extra information

The problem seems to be that `asyncio.sleep()` breaks, and blocks forever. This needs investigation to see if it is something of our environment or a regression in time-machine.

A quick investigation shows that it is stuck only when `tick=False`, so it is probably related to the new support for the monotonic clock in 2.13.0, I guess now the monotonic clock stops now too and in 2.12.0 it didn't.

Related upstream issue:

* https://github.com/adamchainz/time-machine/issues/387

RobMcKiernan:
Same issue here. Currently fixed it by pinning my timemachine version to `2.12.0`

llucax:
@RobMcKiernan you probably should add that to the time machine issue (https://github.com/adamchainz/time-machine/issues/387).

PR #939, opened 2024-05-02
dependabot: Ignore time-machine >=2.13.0

llucax:
The upgrade to time-machine 2.13.0+ breaks our tests. See:

* https://github.com/frequenz-floss/frequenz-sdk-python/issues/832

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -30,6 +30,11 @@ updates:
         update-types:
           - "minor"
           - "patch"
+    ignore:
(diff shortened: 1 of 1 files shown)


llucax:
This change is pretty trivial, so I will force-merge if it's not approved right after the CI is done, so we can continue using dependabot.

shsms (approved):

llucax:
This seems to be failing because of sybil, I will force-merge so I can trigger a dependabot update that hopefully won't want to upgrade time-machine and will fix sybil.

Issue #844, opened 2024-01-16
Upgrade to the microgrid API v0.17


llucax:
### What's needed?

We need to move development of many APIs forward, but new APIs depend on the common API [v0.5](https://github.com/frequenz-floss/frequenz-api-common/releases/tag/v0.5.0), while the SDK is using the microgrid API [v0.15](https://github.com/frequenz-floss/frequenz-api-microgrid/releases/tag/v0.15.2), which depends on the common API [v0.3](https://github.com/frequenz-floss/frequenz-api-common/releases/tag/v0.3.0), so actors for the new APIs can't be developed due to the API versions incompatibilities.

### Proposed solution

Upgrade to the microgrid API [v0.17](https://github.com/frequenz-floss/frequenz-api-microgrid/releases/tag/v0.17.0).

### Alternatives and workarounds

Copy the needed files in the common API to the new API repos. We are doing this in a couple of places, but it is starting to get messy.

llucax:
The current dependency problem is:

* API clients depend on API common v0.5.x
* The SDK depends on API common v0.4.x (because the microgrid API included in the SDK depends on API common v0.4.x)
* Applications using the API clients need `frequenz.sdk.actor`, which makes the conflict of versions for API common

In the near future we need to make API common stable already to not end up in this situation again, so any breaking changes should go to v2. So we can also call the common API version v1.x.x already.

We prioritized the update of the microgrid API in the SDK because we need to break this dependency conflict, but another solution is to extract the `actor` package from the SDK.

llucax:
Since this update involves a lot of code changes, and now we have the https://github.com/frequenz-floss/frequenz-client-common-python/ that we'll need to use too, and we'll need to extract the microgrid API client to its own repo, we will do this update directly in the new repo. Then we update the SDK to remove the client and depend on the new repo, and update the high-level interface to use the new client.

llucax:
Blocked by:

* https://github.com/frequenz-floss/frequenz-sdk-python/issues/853

After we have the new microgrid API client repo with v0.16, we can update the internal usage of the client in the SDK.

PR #941, opened 2024-05-02
dependabot: Ignore frequenz-api-microgrid >=0.16.0

llucax:
The upgrade to frequenz-api-microgrid >=0.16.0 needs a lot of changes in the code. We are not ready to do that yet, so we should ignore this dependency for now.

See:

* https://github.com/frequenz-floss/frequenz-sdk-python/issues/844


llucax:
We need to ignore one more to get useful dependabot updates.

llucax:
Still sybil, so force-merging.

PR #945, opened 2024-05-02
Don't recursively scan for pyproject.toml when calculating the cache key

llucax:
We don't really need to scan for `pyproject.toml` files recursively, since we only have one in the root of the repository. This should make the cache key calculation more efficient and less error prone, as when  using qemu, there are some files that are not accessible and the hash calculation fails.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -159,7 +159,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: /tmp/pip-cache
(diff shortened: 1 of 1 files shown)


daniel-zullo-frequenz:
@llucax do we care about the length of commit titles in the SDK? Other than my question about long commit title, this LGTM

llucax:
Yes we do, everywhere not just the SDK :wink:, but I think 72 chars is still OK, even when the recommendation is to keep it to under 50 (this is mainly relevant when commits are sent as patches via e-mail, where usually a prefix like `[PATCH X/Y] ` (and then `Re: `, etc.) is prepend to the commit title.

daniel-zullo-frequenz (approved):

PR #944, opened 2024-05-02
Always require a priority argument in calls to `microgrid.*_pool` methods

shsms:
This PR also makes the methods accept only keyword arguments.

Closes #916

Issue #916, opened 2024-04-04
Make the `priority` argument for `xxx_pool()` required


llucax:
### What's needed?

We need to try to avoid using the same priority in different actors, as it could lead to conflicts.

### Proposed solution

Make the `priority` argument required in the `xxx_pool()` functions in `frequenz.sdk.microgrid` (and `_DataPipeline`).

### Additional context

* https://github.com/frequenz-floss/frequenz-sdk-python/discussions/913

PR #946, opened 2024-05-02
Set zero power for PV inverters not neccessary to reach target power

### LINKED ISSUE #933: Fix PV inverter manager for 0 W power request
### What happened?

The function [distribute_power](https://github.com/frequenz-floss/frequenz-sdk-python/blob/v1.x.x/src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py#L133-L160) attempts to set a power request of 0W via the API when the list of allocations is empty, leading to an unhandled exception.

Initially, the remaining power is set to 0W based on the request passed as a parameter to the function. Then, when the check for zero remaining power is made, it causes the loop over working components to exit prematurely, resulting in the API power request being executed even though the list of allocations is empty, thus triggering an unhandled exception.

The API power request is set to 0 W for each PV inverter.

### Affected version(s)

1.0.0-rc6

shsms:
If desired power is already reached without the need to use some inverters, their powers need to be set to zero.

If not, their original powers would remain and the target powers would be wrong.

In case the requested power is zero, this was also leading to crashes because there were no allocations.  This change also fixes that issue.

Closes #933

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -16,4 +16,4 @@
 
 ## Bug Fixes
 
-<!-- Here goes notable bug fixes that are worth a special mention or explanation -->
+- When the PowerDistributor receives a zero power request for PV inverters, it now correctly sets zero power to the inverters, and no longer crashes.
--- a/src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
+++ b/src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
@@ -136,7 +136,8 @@ class PVManager(ComponentManager):
             if remaining_power > Power.zero() or is_close_to_zero(
                 remaining_power.as_watts()
             ):
-                break
+                allocations[inv_id] = Power.zero()
+                continue
             distribution = remaining_power / float(num_components - idx)
             inv_data = self._component_data_caches[inv_id]
             if not inv_data.has_value():
--- a/tests/timeseries/_pv_pool/test_pv_pool_control_methods.py
+++ b/tests/timeseries/_pv_pool/test_pv_pool_control_methods.py
@@ -131,7 +131,7 @@ class TestPVPoolControl:
             if check(report):
                 break
 
-    async def test_setting_power(
+    async def test_setting_power(  # pylint: disable=too-many-statements
         self,
         mocks: _Mocks,
         mocker: MockerFixture,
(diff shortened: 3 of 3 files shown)


shsms:
Based on #944.

llucax:
Oh, I guess we shouldn't have merged #916 and merge this one instead to release a rc602, as this is probably a fix we'll need to ship soon. Maybe we should branch for a rc602, what do you think?

llucax (commented):
LGTM, but it needs rebasing.

shsms:
> Maybe we should branch for a rc602, what do you think?

Maybe, but rc700 doesn't seem so bad either, and only the PVPool users need to update immediately.

shsms:
Rebased on v1.x.x

llucax (approved):

RELEASE_NOTES.md
```diff
@@ -16,4 +16,4 @@
 
 ## Bug Fixes
 
-<!-- Here goes notable bug fixes that are worth a special mention or explanation -->
+- When the PowerDistributor receives a zero power request for PV inverters, it now correctly sets zero power to the inverters, and no longer crashes.
```
cwasicki, line 19: Shouldn't it be `... receives a positive or zero power request ...`

src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
```diff
@@ -136,7 +136,8 @@ async def distribute_power(self, request: Request) -> None:
             if remaining_power > Power.zero() or is_close_to_zero(
```
cwasicki, line 136: The `stop` is outdated in the comment above.

RELEASE_NOTES.md
shsms: It would never receive a positive power, because that's filtered in the pv pool: https://github.com/frequenz-floss/frequenz-sdk-python/blob/4b88ab288c95e9acb8b88f66ccc625b661a856b0/src/frequenz/sdk/timeseries/pv_pool/_pv_pool.py#L109-L110

src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
shsms: I think it is still correct, because when we set 0 power to the remaining inverters, we're no longer allocating, but the comment could provide further details: https://github.com/frequenz-floss/frequenz-sdk-python/pull/960

PR #952, opened 2024-05-22
Update glossary

### LINKED ISSUE #650: Define consumer and distinguish assets and devices in glossary
### What's needed?

The glossary is not distinguishing devices and assets in the component section. Follow up on https://github.com/frequenz-floss/frequenz-sdk-python/pull/647#discussion_r1318691630

Also the term `consumer` should be defined in the glossary (see https://github.com/frequenz-floss/frequenz-sdk-python/wiki/Glossary#microgrid-power-terminology)

### Proposed solution

_No response_

### Use cases

_No response_

_No response_

### Additional context

Also remove the word often from Time Series entry: `A sequence of [samples](#sample), often organized by [timestamp](#timestamp)`

cwasicki:
Adding consumer and distinction between assets and devices.

Fixes #650

### DIFF
--- a/docs/user-guide/glossary.md
+++ b/docs/user-guide/glossary.md
@@ -77,6 +77,14 @@ A storage system for electrical energy.
 A generator that produces combined heat and power ([CHP](#chp)). Usually
 powered via combustion of some form of fuel.
 
+#### Consumer
+
+Typically used in context with a metric, i.e. consumer power or consumer current.
+This aggregates all remaining electrical energy [consumption](#consumption) in a microgrid that are not covered by other asset types and components.
+Under normal circumstances this is expected to correspond to the [gross consumption](#gross-consumption) of the site excluding active parts and battery.
+In [Passive-sign convention](#passive-sign-convention) consumer power is usually positive, but negative values can occur in
+exotic site topologies (e.g. unidentified generators) and or due to sudden short-term effects.
+
 #### Converter
 
 Generally refers to [DC-to-DC converter](#dc-to-dc-converter).
--- a/docs/user-guide/glossary.md
+++ b/docs/user-guide/glossary.md
@@ -61,6 +61,10 @@ components](#component-category) together. It can be connected to the public
 [grid](#grid), or be completely isolated, in which case it is known as an
 island.
 
+Components can be grouped into [assets](#Assets) and [devices](#Devices).
+Assets are core components like generators or storage systems that are crucial from a business perspective,
+whereas devices are the supporting infrastructure essential for the functionality of these assets.
+
 ### Component Category
 
 The category [components](#component) of a [microgrid](#microgrid) belong to.
@@ -73,6 +77,8 @@ Examples of categories are meters, inverters or batteries.
 A category can be further divided into types.
 For instance, PV inverters and battery inverters are types under the inverter category.
 
+
+### Assets
 
 #### Battery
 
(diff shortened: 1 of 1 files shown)


llucax (commented):
LGTM, but I think it would be useful to also add a definition of what an *asset* and a *device* is, when a component is considered one or the other?

llucax (dismissed):

docs/user-guide/glossary.md
```diff
@@ -77,6 +77,13 @@ A storage system for electrical energy.
 A generator that produces combined heat and power ([CHP](#chp)). Usually
 powered via combustion of some form of fuel.
 
+#### Consumer
+
+This aggregates all remaining electrical energy [consumption](#consumption) in a microgrid that are not covered by other asset types and components.
```
shsms, line 95: Maybe we want to mention that when used with a metric, like consumer power or consumer current, it has this meaning.  Just in that context.  Then it doesn't sound like were redefining the word consumer.

docs/user-guide/glossary.md
```diff
@@ -68,6 +68,9 @@ The category [components](#component) of a [microgrid](#microgrid) belong to.
 [Components](#component) of the same category have the same characteristics
 (for example offer the same set of [metrics](#metric)).
 
+Examples of category types are meters, inverters, batteries and sensors.
```
shsms: We have categories and types, so saying category type could be confusing.  PV inverter and battery inverter are types under the inverter category.

docs/user-guide/glossary.md
```diff
@@ -71,6 +73,8 @@ The category [components](#component) of a [microgrid](#microgrid) belong to.
 Examples of category types are meters, inverters, batteries and sensors.
 
 
+### Assets
```
shsms, line 81: Maybe add a high level description for assets and devices?  Like assets are the core, like generators or storage, that are meaningful in business context, vs devices that are infrastructure required for the operation of the assets?

llucax:
Any follow-up on this one?

cwasicki:
Updated.

docs/user-guide/glossary.md
```diff
@@ -61,13 +61,25 @@ components](#component-category) together. It can be connected to the public
...
+whereas devices are the supporting infrastructure essential for the functionality of these assets.
+
 ### Component Category
 
 The category [components](#component) of a [microgrid](#microgrid) belong to.
 
 [Components](#component) of the same category have the same characteristics
 (for example offer the same set of [metrics](#metric)).
 
+Examples of categories are meters, inverters, batteries and sensors.
```
llucax: This doesn't match the microgrid API (or won't match it when we migrate to v0.17), sensors are not a component category anymore (it was in previous versions). Actually the [common API](https://github.com/frequenz-floss/frequenz-api-common/blob/29f9945bde7131c3851a1e94a6c19c509ed93806/proto/frequenz/api/common/v1/microgrid/components/components.proto#L26), so it doesn't match the dispatch API either, and probably the reporting API either.

Not sure if you are aware of this but still conceptually we want to say sensors are components, but I think it will be confusing if then you get components from the microgrid API and there are no sensors there.

llucax (commented):
I have only one comment of something that might be wrong. Other than that LGTM. I agree with suggestions from Sahas but it you don't have much time to keep improving this PR I'm fine to merge it without the improvements, as it is still better than what we have now.

docs/user-guide/glossary.md
cwasicki: Fixed.

cwasicki:
>  I agree with suggestions from Sahas but it you don't have much time to keep improving this PR I'm fine to merge it without the improvements

Actually I thought I integrated those :D Maybe I misunderstood the comments? Happy to further improve the doc further.

shsms (approved):

cwasicki:
> I'm fine to merge it without the improvements

Maybe it wasn't clear since I didn't comment on those explicitly? 

I will merge it now but let me know if you have further improvements.

PR #955, opened 2024-05-29
Upgrade to microgrid client v0.4.0

llucax:
- **Bump frequenz-client-microgrid to v0.4.0**
- **Use `GrpcError` instead of `grpc.aio.AioRpcError`**
- **Avoid repetition on error handling**
- **Use a server URL to connect to the microgrid API**
- **Use new microgrid API wrapped types**
- **Stop using the `mock_api` for testing**
- **Remove the, now unused, `mock_api` and its tests**
- **Silence `hpack` debug logging**
- **Fix mock return value**
- **Avoid disabling some pylint checks when unnecessary**
- **Remove unsed protobuf and grpcio dependencies**


llucax:
This is a draft because it depends on https://github.com/frequenz-floss/frequenz-client-microgrid-python/pull/54 and v0.4.0 to be released after it is merged.

~It also needs release notes~, but other than that it is ready to be reviewed!

llucax:
Added release notes.

llucax:
v0.4.0 released, this is ready for a final review!

tests/microgrid/test_microgrid_api.py
```diff
@@ -104,7 +104,7 @@ def metadata(self) -> Metadata:
             ),
         )
 
-    @mock.patch("grpc.aio.insecure_channel")
+    @mock.patch("grpclib.client.Channel")
```
daniel-zullo-frequenz, line 107: this probably goes in another commit (different than `Use a server URL to connect to the microgrid API`

tests/actor/test_data_sourcing.py
```diff
@@ -3,92 +3,285 @@
...
+            assert sample.value is not None
+            assert 9.0 + i == sample.value.base_value
+
+            sample = await soc2_recv_9.receive()
+            assert sample.value is not None
+            assert 9.0 + i == sample.value.base_value
+
+            sample = await active_power_recv_12.receive()
+            assert sample.value is not None
+            assert -13.0 + i == sample.value.base_value
```
daniel-zullo-frequenz, line 148: I think it is not a good idea to use the `==` operator to compare a floating-point value against the result of an addition operation involving floating-point numbers, even in tests. I'd suggest to use isclose() from pytest or math instead

tests/actor/test_data_sourcing.py
```diff
@@ -25,22 +24,18 @@ async def test_data_sourcing_actor() -> None:
...
+    servicer.add_component(8, ComponentCategory.INVERTER)
+    servicer.add_component(9, ComponentCategory.BATTERY)
 
     servicer.add_connection(1, 4)
     servicer.add_connection(1, 7)
     servicer.add_connection(7, 8)
     servicer.add_connection(8, 9)
 
-    await connection_manager.initialize("[::1]", 57899)
+    await connection_manager.initialize("grpc://[::1]:57899")
```
daniel-zullo-frequenz: looks like `Use new microgrid API wrapped types` is not the right commit for this change

pyproject.toml
```diff
@@ -187,8 +182,6 @@ strict = true
 module = [
   "async_solipsism",
   "async_solipsism.*",
-  "grpc.aio",
-  "grpc.aio.*",
```
daniel-zullo-frequenz, line 192: The changes in this file  seem not related to `Avoid disabling some pylint checks when unnecessary`

mkdocs.yml
```diff
@@ -116,10 +116,8 @@ plugins:
             - https://docs.python.org/3/objects.inv
             - https://frequenz-floss.github.io/frequenz-channels-python/v0.16/objects.inv
             - https://frequenz-floss.github.io/frequenz-client-microgrid-python/v0.4/objects.inv
-            - https://grpc.github.io/grpc/python/objects.inv
             - https://networkx.org/documentation/stable/objects.inv
             - https://numpy.org/doc/stable/objects.inv
-            - https://protobuf.readthedocs.io/en/latest/objects.inv
```
daniel-zullo-frequenz, line 123: hmm this also doesn't seem to be related to the commit `Avoid disabling some pylint checks when unnecessary`

RELEASE_NOTES.md
```diff
@@ -6,10 +6,21 @@
 
 ## Upgrading
 
+- The `frequenz-client-microgrid` dependency was bumped to v0.4.0. If you are using the client directly in your code too you will need to upgrade too.
```
daniel-zullo-frequenz: I'd avoid the double `too` in the last sentence
```suggestion
- The `frequenz-client-microgrid` dependency was bumped to v0.4.0. If you are using the client directly in your code, you will need to upgrade too.
```

daniel-zullo-frequenz (commented):
I have a few minor suggestions to check for, LGTM otherwise

daniel-zullo-frequenz (commented):
I have a few minor suggestions to check for, LGTM otherwise

tests/microgrid/test_microgrid_api.py
llucax: It is related, because the parsing of the URL will now return a `grpc.client.Channel` and before doing that switch, we were using a `grpc.aio.insecure_channel`. So we need to adapt the tests to work with this change.

tests/actor/test_data_sourcing.py
llucax: Can you please create a separate issue? I'm just changing the indentation to this code.

tests/actor/test_data_sourcing.py
llucax: Moved to the commit introducing the `server_url` :+1:

pyproject.toml
llucax: Moved to a new commit :+1:

mkdocs.yml
llucax: Moved to a new commit :+1:

RELEASE_NOTES.md
llucax: Updated, thanks!

llucax:
Updated.

src/frequenz/sdk/actor/power_distributing/_component_managers/_battery_manager.py
```diff
@@ -649,21 +654,22 @@ def _parse_result(
...
-                    _logger.warning(
-                        "Set power for battery %s failed, error %s. Mark it as broken.",
-                        battery_ids,
-                        str(err),
-                    )
+                _logger.warning(
+                    "Set power for battery %s failed, mark it as broken. Error: %s",
+                    battery_ids,
+                    err,
+                )
```
shsms, line 670: off topic, but just noticing how these are implemented for battery power distribution.  The logs suggest that for "OutOfRange" errors, we don't mark anything as broken, but for other errors, we mark them as broken, but I'm not sure if that's the case.  I think they're all marked as broken.  Need to confirm.

Also, when battery is over SoC, and we send a power request, the request would be out of bounds, but still the api service sends "FAILED_PRECONDITION", which doesn't sound like something we can ignore.  If it would send OutOfRange instead, we can avoid blocking batteries unnecessarily just because they're fully charged.

src/frequenz/sdk/actor/power_distributing/_component_managers/_battery_manager.py
```diff
@@ -652,40 +652,38 @@ def _parse_result(
 
         for inverter_id, aws in tasks.items():
             battery_ids = self._inv_bats_map[inverter_id]
+            errored = True
```
shsms: "error" is not a verb, treating it like it is feels really weird.  Other options: `oper_success = False`, `has_error` = True, `failed = True`, `succeeded = False`.

I like `succeeded = False`, `op_successful`, etc., because they seem to presume less.

examples/battery_pool.py
```diff
@@ -22,6 +22,9 @@ async def main() -> None:
     logging.basicConfig(
         level=logging.DEBUG, format="%(asctime)s %(name)s %(levelname)s:%(message)s"
     )
+    # Reduce undesired noise from hpack
+    logging.getLogger("hpack.hpack").setLevel(logging.INFO)
```
shsms, line 26: could maybe be configurable through the client?

pyproject.toml
```diff
@@ -31,12 +31,8 @@ dependencies = [
...
   "frequenz-channels >= 1.0.0-rc1, < 2.0.0",
-  "google-api-python-client >= 2.71, < 3",
-  "grpcio >= 1.60.1, < 2",
-  "grpcio-tools >= 1.60.1, < 2",
   "networkx >= 2.8, < 4",
   "numpy >= 1.26.4, < 2",
-  "protobuf >= 4.21.6, < 6",
   "pydantic >= 2.3, < 3",
   "timezonefinder >= 6.2.0, < 7",
   "tqdm >= 4.38.0, < 5",
```
shsms: `tqdm` and `watchfiles` are unused dependencies as well, maybe they can go away too, along with these.

src/frequenz/sdk/actor/power_distributing/_component_managers/_battery_manager.py
llucax: I think ignoring a failed precondition could be fine, it just means there something else that needs to be done before retrying could succeed:

> The operation was rejected because the system is not in a state required for the operation's execution. For example, the directory to be deleted is non-empty, an rmdir operation is applied to a non-directory, etc. Service implementors can use the following guidelines to decide between FAILED_PRECONDITION, ABORTED, and UNAVAILABLE: (a) Use UNAVAILABLE if the client can retry just the failing call. (b) Use ABORTED if the client should retry at a higher level (e.g., when a client-specified test-and-set fails, indicating the client should restart a read-modify-write sequence). (c) Use FAILED_PRECONDITION if the client should not retry until the system state has been explicitly fixed. E.g., if an "rmdir" fails because the directory is non-empty, FAILED_PRECONDITION should be returned since the client should not retry unless the files are deleted from the directory. 

Out of range is considered some sort of special case of failed precondition:

> There is a fair bit of overlap between FAILED_PRECONDITION and OUT_OF_RANGE. We recommend using OUT_OF_RANGE (the more specific error) when it applies so that callers who are iterating through a space can easily look for an OUT_OF_RANGE error to detect when they are done. 

I agree out of range could make more sense if the request fails because the power is out of bounds.

src/frequenz/sdk/actor/power_distributing/_component_managers/_battery_manager.py
llucax: Really? Then it is Internet English, I have seen it used many times. Also Wiktionary has an entry for it: https://en.wiktionary.org/wiki/errored, and some sites even has the conjugation for it: https://www.verbix.com/webverbix/english/error. The Oxford dictionary seems to have an entry for it too: https://www.oed.com/dictionary/error_v?tab=factsheet#5217607, it seems to be used since 1820 :upside_down_face: (but I couldn't find it in the Cambridge dictionary).

I can change it anyways, I was only surprised that it might not be as common as I thought.

examples/battery_pool.py
llucax: Mmm, though one. I really don't think we should mess with silencing logging levels or loggers in libraries, IMHO that should be up to the end user.

llucax: Also, please note we are enabling `DEBUG` logging in the above line, I'm actually surprised we are not getting much more crap from that setting. Normally you would have to do a much more conscious configuring of the logging system. I think the SDK eventually could provide some sensible default logging config (and probably a way to update the logging config dynamically, via `ConfigManager`), and I would be fine to set `hpack` logging to `INFO` there by default.

src/frequenz/sdk/actor/power_distributing/_component_managers/_battery_manager.py
llucax: Changed to `failed = True` because then we are setting `failed_xxx` variables, so it looks pretty consistent.

pyproject.toml
llucax: Actually `timezonefinder` is not used anymore either (it was moved to the microgrid API). Added a new commit removing dependencies not used for a long time now.

llucax:
Updated.

src/frequenz/sdk/actor/power_distributing/_component_managers/_battery_manager.py
shsms: [Chambers](https://chambers.co.uk/search/?query=error&title=21st) and [Webster](https://www.merriam-webster.com/dictionary/error) also list only the noun form.  OED must have been smoking something.

But `succeeded = False` might be better, because that doesn't imply there was an error, and in context, it just means we don't know yet, unlike `error = True`.

shsms (approved):

src/frequenz/sdk/actor/power_distributing/_component_managers/_battery_manager.py
llucax: I was going to change it but I see you already approved, so I will keep `failed` to avoid the extra review roundtrip as it is such a small detail, I guess it is fine to keep it with `failed`.

tests/microgrid/test_microgrid_api.py
daniel-zullo-frequenz: Oh I see, sorry for the noise then

tests/actor/test_data_sourcing.py
daniel-zullo-frequenz: ok, will create an issue

daniel-zullo-frequenz: https://github.com/frequenz-floss/frequenz-sdk-python/issues/965

PR #957, opened 2024-05-31
Support power requests from shifting actors in the PowerManager

### LINKED ISSUE #905: Support multiple Matryoshka pools in the Power Manager that can be aggregated
### What's needed?

There should be multiple matryoshka based target power calculation pools in the PowerManager, and their target powers need to be aggregated (added) to get the final target power.

This would allow use cases like FCR to set their adjustment power always relative to the target power of the other use case actors.

cc: @thomas-nicolai-frequenz

shsms:
There are cases where the target power needs to be shifted by a certain amount, for example, to make adjustments to the operating point.  This PR enables this by designating some actors to be part of the `shifting_group`.

Closes #905

### DIFF
--- a/src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
+++ b/src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
@@ -130,7 +130,7 @@ class PowerManagingActor(Actor):
             await self._send_updated_target_power(component_ids, None)
             await self._send_reports(component_ids)
 
-    def _add_bounds_tracker(self, component_ids: frozenset[int]) -> None:
+    def _add_system_bounds_tracker(self, component_ids: frozenset[int]) -> None:
         """Add a bounds tracker.
 
         Args:
@@ -225,7 +225,7 @@ class PowerManagingActor(Actor):
             if selected_from(selected, self._proposals_receiver):
                 proposal = selected.message
                 if proposal.component_ids not in self._bound_tracker_tasks:
-                    self._add_bounds_tracker(proposal.component_ids)
+                    self._add_system_bounds_tracker(proposal.component_ids)
 
                 # TODO: must_send=True forces a new request to # pylint: disable=fixme
                 # be sent to the PowerDistributor, even if there's no change in power.
@@ -260,7 +260,7 @@ class PowerManagingActor(Actor):
                     )
 
                 if sub.component_ids not in self._bound_tracker_tasks:
-                    self._add_bounds_tracker(sub.component_ids)
+                    self._add_system_bounds_tracker(sub.component_ids)
 
             elif selected_from(selected, self._power_distributing_results_receiver):
                 from .. import (  # pylint: disable=import-outside-toplevel
--- a/src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
+++ b/src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
@@ -16,7 +16,8 @@ from frequenz.channels.timer import SkipMissedAndDrift, Timer
 from frequenz.client.microgrid import ComponentCategory, ComponentType, InverterType
 from typing_extensions import override
 
-from ...timeseries._base_types import SystemBounds
+from ...timeseries import Power
+from ...timeseries._base_types import Bounds, SystemBounds
 from .._actor import Actor
 from .._channel_registry import ChannelRegistry
 from ._base_classes import Algorithm, BaseAlgorithm, Proposal, ReportRequest, _Report
@@ -28,7 +29,7 @@ if typing.TYPE_CHECKING:
     from .. import power_distributing
 
 
-class PowerManagingActor(Actor):
+class PowerManagingActor(Actor):  # pylint: disable=too-many-instance-attributes
     """The power manager."""
 
     def __init__(  # pylint: disable=too-many-arguments
@@ -84,10 +85,18 @@ class PowerManagingActor(Actor):
 
         self._system_bounds: dict[frozenset[int], SystemBounds] = {}
         self._bound_tracker_tasks: dict[frozenset[int], asyncio.Task[None]] = {}
-        self._subscriptions: dict[frozenset[int], dict[int, Sender[_Report]]] = {}
+        self._non_shifting_subscriptions: dict[
+            frozenset[int], dict[int, Sender[_Report]]
+        ] = {}
+        self._shifting_subscriptions: dict[
+            frozenset[int], dict[int, Sender[_Report]]
+        ] = {}
         self._distribution_results: dict[frozenset[int], power_distributing.Result] = {}
 
-        self._algorithm: BaseAlgorithm = Matryoshka(
+        self._non_shifting_group: BaseAlgorithm = Matryoshka(
+            max_proposal_age=timedelta(seconds=60.0)
+        )
+        self._shifting_group: BaseAlgorithm = Matryoshka(
             max_proposal_age=timedelta(seconds=60.0)
         )
 
@@ -104,11 +113,26 @@ class PowerManagingActor(Actor):
         if bounds is None:
             _logger.warning("PowerManagingActor: No bounds for %s", component_ids)
             return
-        for priority, sender in self._subscriptions.get(component_ids, {}).items():
-            status = self._algorithm.get_status(
+        for priority, sender in self._shifting_subscriptions.get(
+            component_ids, {}
+        ).items():
+            status = self._shifting_group.get_status(
                 component_ids,
                 priority,
                 bounds,
+                self._distribution_results.get(component_ids),
+            )
+            await sender.send(status)
+        for priority, sender in self._non_shifting_subscriptions.get(
+            component_ids, {}
+        ).items():
+            status = self._non_shifting_group.get_status(
+                component_ids,
+                priority,
+                self._calculate_remaining_bounds(
+                    bounds,
+                    self._shifting_group.get_target_power(component_ids),
+                ),
                 self._distribution_results.get(component_ids),
             )
             await sender.send(status)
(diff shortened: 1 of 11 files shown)


llucax (commented):
I gave this a quick look, I still have to look at the commit doing the actual implementation in more detail, but I have a couple of comments:

1. It is still no 100% clear what incremental means here. It might be nice to add some better description to the PR and maybe even the docs, like one example with a few actors some using incremental and some not, and seeing what difference does it make in terms of what is the resulting power.

2. It looks like you have at least subscriptions and pools grouped in incremental and not incremental. It might be worth adding some class (not sure what it is exactly? Power requests? Algorithms?) that have 2 members, pool and subscriptions, and I don't know, maybe there is some more common code to manipulate both that can go there too. Maybe it doesn't make sense, not sure, as I said I only gave a quick look but it is something that popped out when looking at the commits.

shsms:
Draft until I finish the renaming to `shifting_group`.

llucax:
No `in_`? :laughing:

shsms:
in_ in the argument names and class fields.

src/frequenz/sdk/microgrid/__init__.py
```diff
@@ -137,6 +137,93 @@
 [`propose_power`][frequenz.sdk.timeseries.ev_charger_pool.EVChargerPool.propose_power],
 which accepts values in the {{glossary("psc", "Passive Sign Convention")}} and supports
 only charging.
+
+# Component pools
+
+The SDK provides a unified interface for interacting with sets of Batteries, EV
+chargers and PV arrays, through their corresponding `Pool`s.
+
+| pool type     | constructor                                                         |
```
llucax: Nitpick:
```suggestion
| Pool Type     | Constructor                                                         |
```

src/frequenz/sdk/microgrid/__init__.py
```diff
@@ -137,6 +137,93 @@
...
+|---------------|---------------------------------------------------------------------|
+| BatteryPool   | [microgrid.battery_pool][frequenz.sdk.microgrid.battery_pool]       |
+| EVChargerPool | [microgrid.ev_charger_pool][frequenz.sdk.microgrid.ev_charger_pool] |
+| PVPool        | [microgrid.pv_pool][frequenz.sdk.microgrid.pv_pool]                 |
```
llucax: Alternatively (I'm not sure the table format here provides much extra if it is only a name and a link basically):

```suggestion
* [Battery pool][frequenz.sdk.microgrid.battery_pool]
* [EV charger pool][frequenz.sdk.microgrid.ev_charger_pool]
* [PV pool][frequenz.sdk.microgrid.pv_pool]
```

Or if you prefer a more code-oriented version:


```suggestion
* [`battery_pool`][frequenz.sdk.microgrid.battery_pool]
* [`ev_charger_pool`][frequenz.sdk.microgrid.ev_charger_pool]
* [`pv_pool`][[frequenz.sdk.microgrid.pv_pool]
```

src/frequenz/sdk/microgrid/_data_pipeline.py
```diff
@@ -211,6 +212,8 @@ def ev_charger_pool(
                 EVChargerPool.
             name: An optional name used to identify this instance of the pool or a
                 corresponding actor in the logs.
+            in_shifting_group: Whether the power requests get sent to the shifting group
+                in the PowerManager or not.
```
llucax, line 216: It would be nice to include a link to the shifting group docs in here (and in every place that receives a `in_shifting_group`). I will be quite verbose, but we can greatly simplify it by creating a [macro like we did for linking to the glossary][https://github.com/frequenz-floss/frequenz-sdk-python/blob/8e5d65e383442c3fa85060e985e225d5e6412a9e/docs/_scripts/macros.py#L79-L97].

src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
```diff
@@ -84,10 +85,18 @@ def __init__(  # pylint: disable=too-many-arguments
...
+        self._shifting_subscriptions: dict[
+            frozenset[int], dict[int, Sender[_Report]]
+        ] = {}
         self._distribution_results: dict[frozenset[int], power_distributing.Result] = {}
 
-        self._algorithm: BaseAlgorithm = Matryoshka(
+        self._non_shifting_group: BaseAlgorithm = Matryoshka(
+            max_proposal_age=timedelta(seconds=60.0)
+        )
+        self._shifting_group: BaseAlgorithm = Matryoshka(
```
llucax, line 99: Not going to insist a lot on this because I don't want to keep focusing on bikeshedding, but I though it out there just in case you agree it could bring some extra clarity. Back to my previous suggestion of grouping the groups, what about having something like:

```py
class PoolGroup:
    algorithm: BaseAlgorithm
    subscriptions: dict[...]
    # Again, I'm not sure if there are common operations with
    # these algorithm and subscriptions that could be moved here too.

class PowerManagingActor(Actor):
    def __init__(self, ...):
        self._default_group = PoolGroup(...)
        self._shifting_group = PoolGroup(...)
```
?

src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
```diff
@@ -104,14 +113,29 @@ async def _send_reports(self, component_ids: frozenset[int]) -> None:
...
+            status = self._non_shifting_group.get_status(
+                component_ids,
+                priority,
+                self._calculate_remaining_bounds(
+                    bounds,
+                    self._shifting_group.get_target_power(component_ids),
+                ),
+                self._distribution_results.get(component_ids),
+            )
+            await sender.send(status)
```
llucax: For example, this can be de-duplicated and moved to the `PoolGroup`:

```py
class PoolGroup:
    algorithm: BaseAlgorithm
    subscriptions: dict[...]

    async def send_report(
        self, component_ids: frozenset[int], distribution_results: ..., bounds: ..., *,
        shift: Power | None = None
    ) -> None:
        for priority, sender in self.subscriptions.get(component_ids, {}).items():
            bounds = self._calculate_shifted_bounds(bounds, shift)
            result = distribution_results.get(component_ids)
            status = self.algorithm.get_status(component_ids, priority, bounds, result)
            await sender.send(status)

    def _calculate_shifted_bounds(
        self, bounds: SystemBounds, shift: Power | None
    ) -> SystemBounds:
        if shift is None:
            return bounds
        ...

class PowerManagingActor(Actor):
    async def _send_reports(self, component_ids: frozenset[int]) -> None:
        if bounds is None:
            _logger.warning("PowerManagingActor: No bounds for %s", component_ids)
            return
        self._shifting_group.send_reports(component_ids, self._distribution_results, bounds)
        self._default_group.send_reports(
            component_ids,
            self._distribution_results,
            bounds,
            shift=self._shifting_group.algorithm.get_target_power(component_ids),
        )
```

src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
```diff
@@ -184,6 +208,116 @@ def _add_bounds_tracker(self, component_ids: frozenset[int]) -> None:
             self._bounds_tracker(component_ids, bounds_receiver)
         )
 
+    def _calculate_remaining_bounds(
```
llucax: If this is only about shifting, what about `_calculate_shifted_bounds()`?

src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
```diff
@@ -184,6 +208,116 @@ def _add_bounds_tracker(self, component_ids: frozenset[int]) -> None:
...
+                    must_send,
+                )
+                tgt_power_no_shift = self._non_shifting_group.calculate_target_power(
+                    component_ids,
+                    None,
+                    self._calculate_remaining_bounds(
+                        self._system_bounds[component_ids], tgt_power_shift
+                    ),
+                    must_send,
+                )
```
llucax:
```py
class PoolGroup:
    algorithm: BaseAlgorithm
    subscriptions: dict[...]
    ...
    # Maybe more stuff could be stored directly in the group,
    # like the `component_ids` and `system_bounds`, so this
    # methods will need even fewer arguments, but I didn't
    # check if this is feasible or not
    def calculate_target_power(
        self,
        component_ids: frozenset[int],
        system_bounds: ...,
        *,
        proposal: Proposal | None = None,
        shift: Power | None = None,
        must_send: bool = False,
     ) -> Power | None:
        bounds = self._calculate_shifted_bounds(system_bounds[component_ids], shift)
        return self.algorithm.calculate_target_power(
                    component_ids, proposal, bounds, must_send,
                )

class PowerManagingActor(Actor):
    ...
    def _calculate_target_power(self, ...) -> Power | None:
        tgt_power_shift: Power | None = None
        tgt_power_default: Power | None = None
        if proposal is not None:
            if proposal.in_shifting_group:
                bounds = self._system_bounds[component_ids]
                tgt_power_shift = self._shifting_group.calculate_target_power(
                    component_ids, bounds, proposal=proposal, must_send=must_send,
                )
                tgt_power_no_shift = self._default_group.calculate_target_power(
                    component_ids, bounds, shift=tgt_power_shift, must_send=must_send,
                )
        ...
```

src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
```diff
@@ -245,22 +378,29 @@ async def _run(self) -> None:
...
+                    subs_set[component_ids][priority] = (
                         self._channel_registry.get_or_create(
                             _Report, sub.get_channel_name()
                         ).new_sender()
                     )
 
-                if sub.component_ids not in self._bound_tracker_tasks:
-                    self._add_bounds_tracker(sub.component_ids)
+                if component_ids not in self._bound_tracker_tasks:
+                    self._add_system_bounds_tracker(component_ids)
```
llucax, line 403:
```py
class PoolGroup:
    ...
    def subscribe(...) -> None:
        ...

class PowerManagingActor(Actor):
    async def _run(self) -> None:
        async for selected in select(...):
            if ...:
                ...
            elif selected_from(selected, self._bounds_subscription_receiver):
                sub = selected.message
                component_ids = sub.component_ids
                priority = sub.priority
                group = self._shifting_group if sub.in_shifting_group else self._default_group

                group.subscribe(...)
```

llucax (dismissed):
I really like how this turned up! And the new documentation is great, it makes things so much clearer!

I made several comments but they are all part of the same thing, a proposal to have a specific class for managing *pool groups*, as I mentioned in my first review.

I really think it could simplify the code quite a bit and remove some duplication. If you agree, I think this refactoring should be done in a separate PR since I don't have any other comments other than that, I think it would be a good idea to merge this PR as is to keep things moving. So approving.

llucax:
Oh, right, there is also the cross referencing of `in_shifting_group` to the high-level docs of what a shifting group is, but that's also optional and I don't mind if that's also done in a separate PR.

llucax:
It looks like DCO is still not working

src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
shsms: true, done in bf2e8a241

src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
shsms: If we end up adding more than two groups, I think it makes sense to refactor at that time.  For now, I'm not sure it is worth the effort, the code is not really that complicated the way it is.

llucax (approved):

src/frequenz/sdk/actor/_power_managing/_power_managing_actor.py
llucax: I created an issue just so it doesn't get lost, but I agree is OK as it is for now:

* https://github.com/frequenz-floss/frequenz-sdk-python/issues/963

PR #962, opened 2024-06-06
Add contains function to Bounds and SystemBounds

### LINKED ISSUE #950: Implement `__contains__` for `timeseries.Bounds` and `timeseries.SystemBounds`
### What's needed?

It would be nice to be able to write:

```py
if value in bounds:
    do_something(value)
else:
    raise OutOfBounds(value)
```

### Proposed solution

Implement [`__contains__()`](https://docs.python.org/3/reference/datamodel.html#object.__contains__) for both classes.

jack-herrmann:


jack-herrmann:
This is for Issue #950.

src/frequenz/sdk/timeseries/_base_types.py
```diff
@@ -134,6 +134,39 @@ def map(
         )
 
 
+@runtime_checkable
```
llucax: I don't think we need a runtime check here, I'll explain in the `__contains__` implementation.
```suggestion
```

src/frequenz/sdk/timeseries/_base_types.py
```diff
@@ -134,6 +134,39 @@ def map(
...
+    def __le__(self, other: Any, /) -> bool:
+        """
+        Return True if self is less than or equal to other, otherwise False.
+
+        Args:
+            other: The value to compare with.
+
+        Returns:
+            bool: True if self is less than or equal to other, otherwise False.
+        """
```
llucax: Suggestion: When the arguments and return can be explained in the docs summary, we don't need to be so verbose (I know a lot of code in the SDK is still very verbose, but that's because a tool we used in the past forced us to always write the whole thing). We also use "Whether ..." to document bool values in general. I recommend using this instead:
```suggestion
        """Return whether this instance is less than or equal to `other`."""
```

In case you are not aware, you can always check how the resulting docs look like by running:

```sh
pip install -e .[dev]; mkdocs serve
```

Then go to http://127.0.0.1:8000 and then you can use the search bar to look for the symbol you are interested in or navigate via the `API Reference` top menu.

src/frequenz/sdk/timeseries/_base_types.py
```diff
@@ -134,6 +134,39 @@ def map(
...
+    def __ge__(self, other: Any, /) -> bool:
+        """
+        Return True if self is greater than or equal to other, otherwise False.
+
+        Args:
+            other: The value to compare with.
+
+        Returns:
+            bool: True if self is greater than or equal to other, otherwise False.
+        """
```
llucax: Same here:
```suggestion
        """Return whether this instance is greater than or equal to `other`."""
```

src/frequenz/sdk/timeseries/_base_types.py
```diff
@@ -147,6 +180,29 @@ class Bounds(Generic[_T]):
...
+        Args:
+            item: The value to check.
+
+        Returns:
+            bool: True if value is within the range, otherwise False.
+        """
+        if item is None:
+            return False
+
+        assert isinstance(item, Comparable)
```
llucax: This will raise an [`AssertionError`](https://docs.python.org/3/library/exceptions.html#AssertionError), which derives from a [`BaseException`](https://docs.python.org/3/library/exceptions.html#BaseException), which are sort of special. If you somehow pass a non-comparable `item`, an exception will be raised when trying to use the comparison operators, which will have the right exception type ([`TypeError`](https://docs.python.org/3/library/exceptions.html#TypeError)), but if you use `mypy`, this error should be caught first by `mypy`, never resulting in a runtime error.

```suggestion
```

src/frequenz/sdk/timeseries/_base_types.py
```diff
@@ -134,6 +134,39 @@ def map(
...
+class Comparable(Protocol):
+    """
+    A protocol that requires the implementation of comparison methods.
+
+    This protocol is used to ensure that types can be compared using
+    the less than or equal to (`<=`) and greater than or equal to (`>=`)
+    operators.
+    """
+
+    def __le__(self, other: Any, /) -> bool:
```
shsms, line 145: the other could be typed as `Self`

src/frequenz/sdk/timeseries/_base_types.py
```diff
@@ -134,6 +134,39 @@ def map(
...
+
+        Args:
+            other: The value to compare with.
+
+        Returns:
+            bool: True if self is greater than or equal to other, otherwise False.
+        """
+
+
 _T = TypeVar("_T")
```
llucax: My suggestion is to bind this `_T` to `Comparable | None`.
```suggestion
_T = TypeVar("_T", bound=Comparable | None)
```

src/frequenz/sdk/timeseries/_base_types.py
shsms: Now that I think about it, I think `item = typing.cast(Comparable, item)` would perform better because it doesn't do any runtime type checks, and is just a hint to the type checker.

Doing that would also make it clear that it is being done because the typechecker has trouble otherwise.  Here I had to think for a minute to remember why the assert is required.

src/frequenz/sdk/timeseries/_base_types.py
```diff
@@ -171,3 +227,21 @@ class SystemBounds:
...
+        Args:
+            item: The value to check.
+
+        Returns:
+            bool: True if value is within the range, otherwise False.
+        """
+        if not self.inclusion_bounds:
+            return False
+        if item not in self.inclusion_bounds:
+            return False
```
shsms:
```suggestion
        if not self.inclusion_bounds or item not in self.inclusion_bounds:
            return False

File: src/frequenz/sdk/timeseries/_base_types.py [hunk 8]
```diff
@@ -147,6 +180,29 @@ class Bounds(Generic[_T]):
...
+
+        assert isinstance(item, Comparable)
+
+        if self.lower is not None and self.upper is not None:
+            return self.lower <= item <= self.upper
+        if self.lower is not None:
+            return self.lower <= item
+        if self.upper is not None:
+            return item <= self.upper
+        return False
```
llucax: I tried this without the `assert` and it fails, but I think `mypy` is wrong here, but we can help it out without having to use `isinstance()`, using [`typing.cast()`](https://docs.python.org/3.11/library/typing.html#typing.cast) instead, which doesn't need to do runtime checks (I also suggest rearranging the comparisons a bit to simplify them):
```suggestion
        if self.lower is None and self.upper is None:
            return True
        if self.lower is None:
            return item <= self.upper
        if self.upper is None:
            return self.lower <= item
        # mypy seems to get confused here, not being able to narrow upper and lower to
        # just Comparable, complaining with:
        #   error: Unsupported left operand type for <= (some union)
        # But we know if they are not None, they should be Comparable, and actually mypy
        # is being able to figure it out in the lines above, just not in this one, so it
        # should be safe to cast.
        return cast(Comparable, self.lower) <= item <= cast(Comparable, self.upper)
```

tests/timeseries/test_base_types.py
```diff
@@ -0,0 +1,79 @@
...
+    assert Power.from_watts(101) not in bounds_no_lower  # above upper bound
+
+    bounds_no_upper = Bounds(lower=Power.from_watts(10), upper=None)
+    assert Power.from_watts(50) in bounds_no_upper  # within lower bound
+    assert Power.from_watts(10) in bounds_no_upper  # at lower bound
+    assert Power.from_watts(9) not in bounds_no_upper  # below lower bound
+
+    # test succeeds, mypy 'Unsupported operand types for in ("Power" and "Bounds[None]")'
+    # bounds_none = Bounds(lower=None, upper=None)
+    # assert Power.from_watts(15) not in bounds_none
```
shsms: could you try: `bounds_none: Bounds[Power | None] = Bounds(lower=None, upper=None)`

tests/timeseries/test_base_types.py
```diff
@@ -0,0 +1,79 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
```
llucax: Nitpick: 2024 :grimacing: 
```suggestion
# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
```

tests/timeseries/test_base_types.py
```diff
@@ -0,0 +1,79 @@
...
+
+"""Tests for battery pool."""
```
llucax:
```suggestion
"""Tests for timeseries base types."""
```

tests/timeseries/test_base_types.py
```diff
@@ -0,0 +1,79 @@
...
+from frequenz.sdk.timeseries._base_types import Bounds, SystemBounds
+from frequenz.sdk.timeseries._quantities import Power
+
+
+def test_bounds_contains() -> None:
+    """
+    Test the `__contains__` method of the `Bounds` class.
+
+    This method checks if a value is within the defined bounds.
+    """
```
llucax: Also nitpick: We usually use this format, I'm actually surprised is not enforced by the linters, I thought I received linting errors because of this in the past, now I wonder if I'm wrong or newer versions relaxed this requirement.
```suggestion
    """Test the `__contains__` method of the `Bounds` class.

    This method checks if a value is within the defined bounds.
    """
```

tests/timeseries/test_base_types.py
llucax: I would remove this comment, it doesn't help much if we can't really test mypy is reporting the error.

tests/timeseries/test_base_types.py
```diff
@@ -0,0 +1,79 @@
...
+
+    bounds_no_upper = Bounds(lower=Power.from_watts(10), upper=None)
+    assert Power.from_watts(50) in bounds_no_upper  # within lower bound
+    assert Power.from_watts(10) in bounds_no_upper  # at lower bound
+    assert Power.from_watts(9) not in bounds_no_upper  # below lower bound
+
+    # test succeeds, mypy 'Unsupported operand types for in ("Power" and "Bounds[None]")'
+    # bounds_none = Bounds(lower=None, upper=None)
+    # assert Power.from_watts(15) not in bounds_none
+
```
llucax, line 38: I would also add some tests for corner cases, like using plain `int`s, using an empty range, like `Bounds(lower=0, upper=0)`, using `float`s, negative numbers, and trying `Bounds(lower=10, upper=-10)`, which is actually not enforced now, not related to this PR but it could be a nice addition to validate the arguments in [`__post_init__()`](https://docs.python.org/3/library/dataclasses.html#dataclasses.__post_init___).

I think also adding `kw_only` to the `@dataclass(froze=True, kw_only=True)` for both bounds classes would be a great addition too, as it would be very confusing if users use `Bounds(10, 14)`; which is which?

tests/timeseries/test_base_types.py
```diff
@@ -0,0 +1,79 @@
...
+    )  # outside inclusion bounds
+
+    system_bounds_no_inclusion = SystemBounds(
+        timestamp=datetime.now(),
+        inclusion_bounds=None,
+        exclusion_bounds=exclusion_bounds,
+    )
+    assert (
+        Power.from_watts(30) not in system_bounds_no_inclusion
+    )  # not in inclusion bounds
```
llucax: This could also have a few more test cases, like disjoint and partially overlapping  inclusion/exclusion bounds. I'm not sure if we want to forbid disjoint inclusion/exclusion bounds from being created at all (validate in `__post_init__`), like invalid ranges in `Bounds` too, as they don't really make sense.

llucax (commented):
This was a long review, in the middle of testing my proposed changes, I bumped into a (unrelated) race condition in the tests, and I derailed a lot trying to find the issue. Now I think I did, and I realized I completely abandoned the review, so resuming it now :laughing: 

I didn't check @sahas-subramanian-frequenz comments in depth yet, I hope we don't contradict each other :grimacing: 

In case it's useful, here is a full diff of my proposed changes that are supposed to be working:


<details>
<summary>Click to expand</summary>

```diff
diff --git a/src/frequenz/sdk/timeseries/_base_types.py b/src/frequenz/sdk/timeseries/_base_types.py
index cce408b7..01a36313 100644
--- a/src/frequenz/sdk/timeseries/_base_types.py
+++ b/src/frequenz/sdk/timeseries/_base_types.py
@@ -8,7 +8,7 @@ import functools
 from collections.abc import Callable, Iterator
 from dataclasses import dataclass
 from datetime import datetime, timezone
-from typing import Any, Generic, Protocol, Self, TypeVar, overload, runtime_checkable
+from typing import Any, Generic, Protocol, Self, TypeVar, cast, overload
 
 from ._quantities import Power, QuantityT
 
@@ -134,7 +134,6 @@ class Sample3Phase(Generic[QuantityT]):
         )
 
 
-@runtime_checkable
 class Comparable(Protocol):
     """
     A protocol that requires the implementation of comparison methods.
@@ -167,7 +166,7 @@ class Comparable(Protocol):
         """
 
 
-_T = TypeVar("_T")
+_T = TypeVar("_T", bound=Comparable | None)
 
 
 @dataclass(frozen=True)
@@ -193,15 +192,19 @@ class Bounds(Generic[_T]):
         if item is None:
             return False
 
-        assert isinstance(item, Comparable)
-
-        if self.lower is not None and self.upper is not None:
-            return self.lower <= item <= self.upper
-        if self.lower is not None:
-            return self.lower <= item
-        if self.upper is not None:
+        if self.lower is None and self.upper is None:
+            return True
+        if self.lower is None:
             return item <= self.upper
-        return False
+        if self.upper is None:
+            return self.lower <= item
+        # mypy seems to get confused here, not being able to narrow upper and lower to
+        # just Comparable, complaining with:
+        #   error: Unsupported left operand type for <= (some union)
+        # But we know if they are not None, they should be Comparable, and actually mypy
+        # is being able to figure it out in the lines above, just not in this one, so it
+        # should be safe to cast.
+        return cast(Comparable, self.lower) <= item <= cast(Comparable, self.upper)
 
 
 @dataclass(frozen=True, kw_only=True)
diff --git a/src/frequenz/sdk/timeseries/ev_charger_pool/_set_current_bounds.py b/src/frequenz/sdk/timeseries/ev_charger_pool/_set_current_bounds.py
index faefff5e..81bd5463 100644
--- a/src/frequenz/sdk/timeseries/ev_charger_pool/_set_current_bounds.py
+++ b/src/frequenz/sdk/timeseries/ev_charger_pool/_set_current_bounds.py
@@ -99,8 +99,10 @@ class BoundsSetter:
             _logger.error(err)
             raise RuntimeError(err)
 
+        meter_id = next(iter(meters)).component_id
         self._meter_data_cache = LatestValueCache(
-            await api_client.meter_data(next(iter(meters)).component_id)
+            await api_client.meter_data(meter_id),
+            unique_id=f"{type(self).__name__}«{hex(id(self))}»:meter«{meter_id}»",
         )
         latest_bound: dict[int, ComponentCurrentLimit] = {}
 
```

</details>

src/frequenz/sdk/timeseries/_base_types.py
jack-herrmann: I did originally have that. But mypy fails in this line:
            return self.lower <= item <= self.upper
            
with the following error:
            error: Unsupported operand types for <= ("Comparable" and "_T")  [operator]
            
I do think that mypy is just wrong and this would never actually cause an issue, but I am not sure how to get around mypy.

jack-herrmann:
This should address all the suggested changes.
I will add further tests through a new commit later.

src/frequenz/sdk/timeseries/_base_types.py
llucax: Yeah, it seems we are pushing the limits of the type system with this :scream:, but I think you can fix it by helping `mypy` with a `cast` here or there, as I shown in my review. I tried replacing `Any` with `Self` and I had to add one more cast to the `__contains__()` method to make it work.

tests/timeseries/test_base_types.py
```diff
@@ -0,0 +1,74 @@
...
+
+    bounds_no_lower = Bounds(lower=None, upper=Power.from_watts(100))
+    assert Power.from_watts(50) in bounds_no_lower  # within upper bound
+    assert Power.from_watts(100) in bounds_no_lower  # at upper bound
+    assert Power.from_watts(101) not in bounds_no_lower  # above upper bound
+
+    bounds_no_upper = Bounds(lower=Power.from_watts(10), upper=None)
+    assert Power.from_watts(50) in bounds_no_upper  # within lower bound
+    assert Power.from_watts(10) in bounds_no_upper  # at lower bound
+    assert Power.from_watts(9) not in bounds_no_upper  # below lower bound
```
llucax: Small suggestion, if you split this in a separate test, you will get better test reports if one case fails but the other don't. For example, if the test with lower and upper bounds fail, the test without a lower bound (and without an upper bound) will not be executed, so you will not know if they succeeded or not.

If you make a separate test for no upper bound and another for no lower bound, all 3 will be tested and you'll get separate results for each.

tests/timeseries/test_base_types.py
```diff
@@ -0,0 +1,74 @@
...
+    )  # outside inclusion bounds
+
+    system_bounds_no_inclusion = SystemBounds(
+        timestamp=datetime.now(),
+        inclusion_bounds=None,
+        exclusion_bounds=exclusion_bounds,
+    )
+    assert (
+        Power.from_watts(30) not in system_bounds_no_inclusion
+    )  # not in inclusion bounds
```
llucax: Same here about splitting tests for no inclusion and no exclusion bounds.

llucax (commented):
LGTM, made just a couple of optional comments and I'll wait for the extra tests before approving.

This also needs release notes (add a note in `New Features` about `Bounds` and `SystemBounds` supporting `in`.

llucax:
Tests are failing for an [unrelated regression (in griffe)](https://github.com/mkdocstrings/griffe/issues/295) :(

I would wait a bit before pinning `griffe` to an old version because the author is super responsive, so maybe there is a fix soon.

@jh2007github BTW the PR is not based on the current `v1.x.x` branch, can you rebase it?

tests/timeseries/test_base_types.py
llucax: I reopened this because the suggested changes were not made. Even if the extra validation is not added, at least having tests for this would be good:

> I would also add some tests for corner cases, like using plain ints, using an empty range, like Bounds(lower=0, upper=0)

If you think this should not be done, it would be good to give a reason.

llucax (commented):
I have only one comment about pending stuff from last review. There is other small stuff, like using `Self` instead of `Any` in `Comparable`, but I think we can leave it for a future PR, as this PR is already functional, improvements are optional. Specially since I [plan to move this code to the new `core` repo](https://github.com/frequenz-floss/frequenz-core-python/pull/10#issuecomment-2168356414).

Just as a reminder, a few general improvements that could be made to `Bounds`:

* https://github.com/frequenz-floss/frequenz-core-python/issues/11
* https://github.com/frequenz-floss/frequenz-core-python/issues/12
* https://github.com/frequenz-floss/frequenz-core-python/issues/13
* https://github.com/frequenz-floss/frequenz-core-python/issues/15
* https://github.com/frequenz-floss/frequenz-core-python/issues/14

llucax:
`griffe` 0.46.1 released, re-running the failed job.

llucax:
WORKSFORME :heart:

llucax (approved):
OK, the CI was fixed, and most comments were addressed. Since @jh2007github jumped to another topic and I want to move this to `core`, I will just merge and we can leave any improvements for future PRs (in `core`).

llucax:
Queued cross-arch tests failed:

https://github.com/frequenz-floss/frequenz-sdk-python/actions/runs/9610219272/job/26506246519#step:10:891

I was debugging this for several hours last week and these `LastValueCache` tasks should not be pending, I checked and they are stopped before the test end. It is also super weird that this error consistently appears in the `test_actor.py::test_restart_on_unhandled_exception[10]` test, not in `test_actor.py::test_restart_on_unhandled_exception[2] ` or `[1]`, and the pending tasks are created by another by another tests (`test_power_distributing.py` IIRC), so it seems to be an isolation issue with `pytest-async`.

I guess it also depends on timing, otherwise we wouldn't have it fail only in one of the 2 matrix jobs of the cross-arch tests...

llucax:
I wonder if it is related to this:

* https://github.com/frequenz-floss/frequenz-sdk-python/issues/975

I think I will give it a try.

llucax:
* https://github.com/frequenz-floss/frequenz-sdk-python/pull/976

Will rebase  when it is merged and see what happens, now I can't even reproduce locally (but before I could with 100% success rate).

llucax:
@daniel-zullo-frequenz that introduces one extra merge. I will force-push with a rebase instead. @jh2007github FYI

daniel-zullo-frequenz:
> that introduces one extra merge. I will force-push with a rebase instead

Yes, I'm aware of it. I accidentally pressed the update-branch button. Sorry!

llucax:
Sadly this is still failing in the same matrix job the same way. It is super weird, but the fact that only fails with this PR in a consistent way makes me very uncomfortable. I debugged this for hours and still could find how this might fail. We could have a workaround in the failing test to ignore logs from asyncio, but :grimacing:

llucax:
Giving this a new shot, in case it is some bug in some dependency and it was fixed, as this only happens with `pytest_max`.

llucax:
image

llucax:
I'm giving up for now and will just filter out the logs for the test to pass:

* https://github.com/frequenz-floss/frequenz-sdk-python/pull/983

But I created an issue about the whole thing, because I've seen other warnings seem to be flagging real issues:

* #982

llucax:
Will rebase again

llucax:
Finally! This ended up being much more complicated to merge than expected...

shsms:
It was just python demonstrating its newest async test capabilities.

PR #966, opened 2024-06-10
Improve the power managing actor initialization

llucax:
This PR was done while chasing an issue with tests complaining that some tasks were pending while the loop was destroyed.

I guess the issue is in the `pytest_asyncio` plugin, because those tasks seem to be properly cancelled according to my debugging, and we had similar issues with `pytest-asyncio` before.  The weird thing is that the errors show up in a different test from the one that actually creates the tasks, which makes me think that the plugin is not properly cleaning up after itself.

In any case, while debugging this, I found a few things that could be improved in the code in terms of synchoronization during initialization and cleanup.

This PR:

- **Adds a `unique_id` to `LatestValueCache`** to improve the debugging experience (using the new name `unique_id` instead of `name`, as suggested in frequenz-floss/frequenz-core-python#7).
- **Adds `__str__` and `__repr__` to `LatestValueCache`** to use this `unique_id`
- **Cancels `LatestValueCache`'s task if deleted** (just in case)
- **Adds a `wait_for_value()` method to `LatestValueCache`** so we can wait for the cache to be filled with data
- **Adds a `wait_for_data()` method to `ComponentManager`** to wait for all caches to be filled with data
- **Add `@override` to `PowerDistributingActor`** (because why not?)
- **Improve waiting for data in the `PowerDistributingActor`** by using the new `wait_for_data()` method instead of blidly wait for the whole wait time

### DIFF
--- a/src/frequenz/sdk/_internal/_channels.py
+++ b/src/frequenz/sdk/_internal/_channels.py
@@ -92,6 +92,10 @@ class LatestValueCache(typing.Generic[T_co]):
         """Stop the cache."""
         await cancel_and_await(self._task)
 
+    def __del__(self) -> None:
+        """Stop the cache when it is deleted."""
+        self._task.cancel()
+
     def __repr__(self) -> str:
         """Return a string representation of the cache."""
         return (
--- a/src/frequenz/sdk/_internal/_channels.py
+++ b/src/frequenz/sdk/_internal/_channels.py
@@ -36,15 +36,28 @@ class _Sentinel:
 class LatestValueCache(typing.Generic[T_co]):
     """A cache that stores the latest value in a receiver."""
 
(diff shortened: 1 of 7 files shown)


llucax:
@jh2007github FYI as it is (slightly) related to frequenz-floss/frequenz-core-python#7.

llucax:
Skipping release notes as the resulting effect should be barely noticeable for end users, but I can mention the improvement in the release notes if you prefer.

shsms:
The waiting for data at power distributor startup can be dropped instead: https://github.com/frequenz-floss/frequenz-sdk-python/pull/971, so I guess some of the commits from this PR can go away.

llucax:
Dropped in favor of:

* #971 
* #974

llucax:
We leave the wait for data for `LastValueCache` for feature, if necessary.

PR #971, opened 2024-06-19
Don't sleep on PowerDistributor startup

shsms:
The PowerDistributor used to sleep for 2 seconds on startup, before
processing requests, to wait for data.  This feature is no longer
being used, because it won't receive requests from the PowerManager
unless there's data.  This PR removes it.

### DIFF
--- a/tests/actor/power_distributing/test_power_distributing.py
+++ b/tests/actor/power_distributing/test_power_distributing.py
@@ -112,7 +112,6 @@ class TestPowerDistributingActor:
                 requests_receiver=requests_channel.new_receiver(),
                 results_sender=results_channel.new_sender(),
                 component_pool_status_sender=battery_status_channel.new_sender(),
-                wait_for_data_sec=0.0,
             ) as distributor:
                 assert isinstance(distributor._component_manager, BatteryManager)
                 assert distributor._component_manager._bat_invs_map == {
@@ -144,7 +143,6 @@ class TestPowerDistributingActor:
                 requests_receiver=requests_channel.new_receiver(),
                 results_sender=results_channel.new_sender(),
                 component_pool_status_sender=battery_status_channel.new_sender(),
-                wait_for_data_sec=0.0,
             ) as distributor:
                 assert isinstance(distributor._component_manager, BatteryManager)
                 assert distributor._component_manager._bat_invs_map == {
@@ -210,8 +208,9 @@ class TestPowerDistributingActor:
             requests_receiver=requests_channel.new_receiver(),
             results_sender=results_channel.new_sender(),
             component_pool_status_sender=battery_status_channel.new_sender(),
-            wait_for_data_sec=0.1,
         ):
+            await asyncio.sleep(0.1)  # wait for actor to collect data
+
             await requests_channel.new_sender().send(request)
             result_rx = results_channel.new_receiver()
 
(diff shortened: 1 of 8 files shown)


shsms:
Only internal changes, so no release notes necessary.

tests/timeseries/_battery_pool/test_battery_pool_control_methods.py
```diff
@@ -89,6 +89,11 @@ async def _patch_battery_pool_status(
         If `battery_ids` is not None, the mock will always return `battery_ids`.
         Otherwise, it will return the requested batteries.
         """
+        mocker.patch.object(
+            timeseries.battery_pool._methods,  # pylint: disable=protected-access
+            "WAIT_FOR_COMPONENT_DATA_SEC",
+            0.1,
+        )
```
llucax, line 96: Why do we still have some other waiting for data as a simple sleep? Shouldn't that wait be removed too?

tests/actor/power_distributing/test_power_distributing.py
```diff
@@ -210,8 +208,9 @@ async def test_power_distributor_one_user(self, mocker: MockerFixture) -> None:
             requests_receiver=requests_channel.new_receiver(),
             results_sender=results_channel.new_sender(),
             component_pool_status_sender=battery_status_channel.new_sender(),
-            wait_for_data_sec=0.1,
         ):
+            await asyncio.sleep(0.1)  # wait for actor to collect data
```
llucax, line 212: Is this PR just moving the issue elsewhere instead? Or is this synchronization only needed for the tests but when actually used one shouldn't care when there is data and when there isn't?

In any case, having tests depend on random `sleep` makes me sad (as I see the flakiness coming to bite us eventually). :cry: :disappointed: 

If we can really really get rid of the need for synchronization, then I completely agree we shouldn't provide any way to synchronize, but otherwise I think having a way to explicitly synchronize that is it not a flaky sleep it is a better approach. In terms of performance, even if it is a hot path, it is just one extra boolean evaluation and branch, which modern CPUs probably won´t even care about because they will do branch prediction so the performance might not even be affected at all after the initialization.

llucax:
The CI is failing, not sure why suddenly a lot of cross referencing can't be found, but I guess it might be another ripple effect of https://github.com/mkdocstrings/griffe/issues/294. Will do a quick check to see if downgrading the dependency fixes it.

tests/timeseries/_battery_pool/test_battery_pool_control_methods.py
shsms: This is for the metric streaming in the battery pool.  It won't stream any metrics until it has data for all the batteries, etc. So we don't send a new value for capacity or soc as we get data from more and more batteries.

This is independent of the power distributor.

tests/actor/power_distributing/test_power_distributing.py
shsms: As I mentioned in the commit message, these are needed only in tests where we control the PowerDistributor directly.

And they are needed only for those tests that need component data, so that PowerDistributor can return `Success`.  Else it will return `Error("No data")`, but that's not the case we're testing.

It is only a testing problem, not with behaviour in production.

llucax:
> The CI is failing, not sure why suddenly a lot of cross referencing can't be found, but I guess it might be another ripple effect of [mkdocstrings/griffe#294](https://github.com/mkdocstrings/griffe/issues/294). Will do a quick check to see if downgrading the dependency fixes it.

Fixed.

tests/actor/power_distributing/test_power_distributing.py
llucax: OK, I'm still not happy about potential flaky tests, but this is still an improvement over what we had before, so approving :+1:

llucax (approved):

PR #974, opened 2024-06-20
Make some minor improvements for LastValueCache

llucax:
- **Add a `unique_id` to `LatestValueCache`**
- **Add `__str__` and `__repr__` to `LatestValueCache`**
- **Cancel `LatestValueCache`'s task if deleted**
- **Add `@override` to `PowerDistributingActor`**

### DIFF
--- a/src/frequenz/sdk/_internal/_channels.py
+++ b/src/frequenz/sdk/_internal/_channels.py
@@ -32,6 +32,10 @@ class ReceiverFetcher(typing.Generic[T_co], typing.Protocol):
 class _Sentinel:
     """A sentinel to denote that no value has been received yet."""
 
+    def __str__(self) -> str:
+        """Return a string representation of this sentinel."""
+        return "<no value received yet>"
+
 
 class LatestValueCache(typing.Generic[T_co]):
     """A cache that stores the latest value in a receiver."""
@@ -91,3 +95,14 @@ class LatestValueCache(typing.Generic[T_co]):
     async def stop(self) -> None:
         """Stop the cache."""
         await cancel_and_await(self._task)
+
+    def __repr__(self) -> str:
+        """Return a string representation of this cache."""
+        return (
+            f"<LatestValueCache latest_value={self._latest_value!r}, "
+            f"receiver={self._receiver!r}, unique_id={self._unique_id!r}>"
+        )
+
+    def __str__(self) -> str:
+        """Return the last value seen by this cache."""
+        return str(self._latest_value)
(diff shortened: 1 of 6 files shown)


llucax:
Enabled auto-merge.

src/frequenz/sdk/_internal/_channels.py
```diff
@@ -91,3 +91,14 @@ async def _run(self) -> None:
...
+    def __repr__(self) -> str:
+        """Return a string representation of the cache."""
+        return (
+            f"<LatestValueCache latest_value={self._latest_value!r}, "
+            f"receiver={self._receiver!r}, unique_id={self._unique_id!r}>"
+        )
+
+    def __str__(self) -> str:
+        """Return a string representation of the cache."""
+        return str(self._latest_value)
```
shsms, line 108: Maybe the docstring should say the latest value in the cache instead?

src/frequenz/sdk/_internal/_channels.py
```diff
@@ -92,6 +92,10 @@ async def stop(self) -> None:
         """Stop the cache."""
         await cancel_and_await(self._task)
 
+    def __del__(self) -> None:
+        """Stop the cache when it is deleted."""
+        self._task.cancel()
```
shsms: Best to check if the event loop is still running before calling cancel, else, we might get more warning spam in tests.

src/frequenz/sdk/_internal/_channels.py
llucax: OK, I tend to use just a generic docstring for `__str__` and `__repr__` but maybe in this case it is worth to be more explicit.

src/frequenz/sdk/_internal/_channels.py
llucax: Oh, really? Actually I added this because I was getting warnings in tests about the last value cache tasks being pending when the loop was destroyed. But I debugged the issue and the tasks were properly finished, it seems it might be a bug in pytest-async, I believe we had some similar issues in the past.

Maybe I should remove this and that's it, I think according to Python specs this should not be necessary, and tasks should be cancelled (or discarded) when the even loop is closed, and dangling tasks will also be removed from the loop and finished, it might only happen that this is delayed until the GC runs if the reference was part of a reference loop.

I'd say until we are completely sure this is really necessary we just don't add it.

llucax:
Updated to improve the `__str__` docstring and removed the commit adding `__del__`. I also implemented `__str__` for `_Sentinel` so we get a more meaningful string when the cache didn't receive any value yet.

src/frequenz/sdk/_internal/_channels.py
llucax: Just to add more context, this is the issue that made me try this:

* https://github.com/frequenz-floss/frequenz-sdk-python/pull/962#issuecomment-2182264984

shsms (approved):

PR #981, opened 2024-06-26
Cleanup `*Report` classes as simple protocols with only methods

### LINKED ISSUE #823: Power Manager `_Report`, `Report` and `BatteryPoolReport` have the wrong hierarchy

shsms:
This was required to remove the explicit casting of the `_Report` type to the `*PoolReport` protocol types.

Closes #823

### DIFF
--- a/src/frequenz/sdk/actor/_power_managing/_base_classes.py
+++ b/src/frequenz/sdk/actor/_power_managing/_base_classes.py
@@ -46,60 +46,8 @@ class ReportRequest:
         return f"power_manager.report.{self.component_ids=}.{self.priority=}"
 
 
-class Report(typing.Protocol):
-    """Current PowerManager report for a set of components.
-
-    This protocol can be specialized by different component pools to provide more
-    specific details and documentation for the reports.
-    """
-
-    @property
-    def bounds(self) -> timeseries.Bounds[Power] | None:
-        """The bounds for the components.
-
-        These bounds are adjusted to any restrictions placed by actors with higher
-        priorities.
-
-        There might be exclusion zones within these bounds. If necessary, the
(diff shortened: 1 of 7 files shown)


llucax (approved):
:tada:

Issue #818, opened 2023-12-11
SDK power_status interface to receive bounds is very spammy


Marenz:
### What happened?

The new versions of the SDK send power bounds updates multiple times per second without any bound changing. This is visible both with real hardware and with the sandbox:

17:41:48,031,31 INFO     [fcr_balancing_actor.py:243] Battery power bounds changed to [-1.262 MW, 1.31 MW]
17:41:48,037,37 INFO     [fcr_balancing_actor.py:243] Battery power bounds changed to [-1.262 MW, 1.31 MW]
17:41:48,237,237 INFO     [fcr_balancing_actor.py:243] Battery power bounds changed to [-1.262 MW, 1.31 MW]
17:41:48,246,246 INFO     [fcr_balancing_actor.py:243] Battery power bounds changed to [-1.262 MW, 1.31 MW]
17:41:48,435,435 INFO     [fcr_balancing_actor.py:243] Battery power bounds changed to [-1.262 MW, 1.31 MW]
17:41:48,443,443 INFO     [fcr_balancing_actor.py:243] Battery power bounds changed to [-1.262 MW, 1.31 MW]
17:41:48,636,636 INFO     [fcr_balancing_actor.py:243] Battery power bounds changed to [-1.262 MW, 1.31 MW]
17:41:48,646,646 INFO     [fcr_balancing_actor.py:243] Battery power bounds changed to [-1.262 MW, 1.31 MW]
17:41:48,840,840 INFO     [fcr_balancing_actor.py:243] Battery power bounds changed to [-1.262 MW, 1.31 MW]
17:41:48,846,846 INFO     [fcr_balancing_actor.py:243] Battery power bounds changed to [-1.262 MW, 1.31 MW]
17:41:49,025,25 INFO     [fcr_balancing_actor.py:243] Battery power bounds changed to [-1.262 MW, 1.31 MW]
17:41:49,034,34 INFO     [fcr_balancing_actor.py:243] Battery power bounds changed to [-1.262 MW, 1.31 MW]
17:41:49,240,240 INFO     [fcr_balancing_actor.py:243] Battery power bounds changed to [-1.262 MW, 1.31 MW]
17:41:49,248,248 INFO     [fcr_balancing_actor.py:243] Battery power bounds changed to [-1.262 MW, 1.31 MW]
17:41:49,430,430 INFO     [fcr_balancing_actor.py:243] Battery power bounds changed to [-1.262 MW, 1.31 MW]
[...]

PR #998, opened 2024-07-02
Remove power distribution results from `Report` objects

shsms:
Instead, this PR makes the power distribution results available directly from the `*Pool`s.

This should go some way towards reducing the frequency of `Report` objects as reported in #818

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ dependencies = [
   # changing the version
   # (plugins.mkdocstrings.handlers.python.import)
   "frequenz-client-microgrid >= 0.4.0, < 0.5.0",
-  "frequenz-channels >= 1.0.0-rc1, < 2.0.0",
+  "frequenz-channels >= 1.1.0, < 2.0.0",
   "networkx >= 2.8, < 4",
   "numpy >= 1.26.4, < 2",
   "typing_extensions >= 4.6.1, < 5",
--- a/src/frequenz/sdk/_internal/_channels.py
+++ b/src/frequenz/sdk/_internal/_channels.py
@@ -12,6 +12,7 @@ from frequenz.channels import Receiver
 from ._asyncio import cancel_and_await
 
 T_co = typing.TypeVar("T_co", covariant=True)
+U_co = typing.TypeVar("U_co", covariant=True)
 
 
 class ReceiverFetcher(typing.Generic[T_co], typing.Protocol):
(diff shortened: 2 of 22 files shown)


shsms:
Draft until channel release with `filter` on receivers.

shsms:
Based on #1000.

shsms:
This is now ready.  `ReceiverFetcherWith` is internal, but still could use a better name I think, if anyone has any ideas.

llucax:
> This is now ready. `ReceiverFetcherWith` is internal, but still could use a better name I think, if anyone has any ideas.

Who doesn't like to get involved in a nice and long name discussion? :laughing:

src/frequenz/sdk/_internal/_channels.py
```diff
@@ -29,6 +31,37 @@ def new_receiver(self, *, limit: int = 50) -> Receiver[T_co]:
         """
 
 
+class ReceiverFetcherWith(ReceiverFetcher[U_co], typing.Generic[T_co, U_co]):
```
llucax: So about the name discussion, I would mimic `Receiver` and say this is a `MappingReceiverFetcher`, and call the `manipulator` a `mapping_function`?

llucax:
```suggestion
class ReceiverFetcherWith(typing.Generic[T_co, U_co]):
```

Inheriting from a protocol will make this a protocol, but we actually don't want this to be a protocol in itself, just a regular class that happens to implement the `ReceiverFetcher` protocol, right?

pyproject.toml
```diff
@@ -30,7 +30,7 @@ dependencies = [
   # changing the version
   # (plugins.mkdocstrings.handlers.python.import)
   "frequenz-client-microgrid >= 0.4.0, < 0.5.0",
-  "frequenz-channels >= 1.0.0-rc1, < 2.0.0",
+  "frequenz-channels >= 1.1.0, < 2.0.0",
```
llucax, line 33: You must be happy about finally getting rid of the `-rc1` :laughing:

Just as a heads up, once we reach *the real 1.0.0* :tm: for the SDK, we can't bump minimum dependencies anymore.

Also, while you are upgrading, wouldn't it make sense to also replace the `LatestValueCache` with the one in channels? But OK to do in a separate PR too.

llucax (commented):
Just a few small comments.

src/frequenz/sdk/_internal/_channels.py
shsms: changed.

shsms: true, it should be fine without deriving from it.  changed.

pyproject.toml
shsms: > You must be happy about finally getting rid of the -rc1 😆

properly happy. :-)

> Also, while you are upgrading, wouldn't it make sense to also replace the LatestValueCache with the one in channels? But OK to do in a separate PR too.

will do it separately.

llucax (approved):

PR #999, opened 2024-07-08
Resync the resampler on system time changes (take 2)

llucax:
This PR builds on top of #802 by using a new timer that is attached to the wall clock instead of the monotonic clock, and can be used to trigger resampling windows. It is a replacement for the channels Timer class.

This should fix the issue with the resampling windows being too small when the system clock is not in sync with the wall clock.

**Status:** Some code adjustments might still be needed, but for sure the big chunk of work pending are the tests, as testing this it is very complex as one need to emulate both the wall clock and the monotonic clock and the different ways they can be out of sync.


llucax:
Replaced by:

* https://github.com/frequenz-floss/frequenz-sdk-python/pull/1249

PR #1000, opened 2024-07-08
Remove `request_timeout` parameter from `propose_power` methods

shsms:
Before the power manager was introduced, the `request_timeout` parameter was used to specify a different timeout for the `set_power` calls made to the api service.

The `propose_power` method is much more high level and it no longer makes sense to provide a `request_timeout` parameter, because it could conflict with other proposals and it would be hard to figure out which timeout to use.

From now on, it may be specified at application startup, through the new optional `api_power_request_timeout` parameter in the `microgrid.initialize()` method.

### DIFF
--- a/src/frequenz/sdk/microgrid/_data_pipeline.py
+++ b/src/frequenz/sdk/microgrid/_data_pipeline.py
@@ -14,6 +14,7 @@ import logging
 import typing
 from collections import abc
 from dataclasses import dataclass
+from datetime import timedelta
 
 from frequenz.channels import Broadcast, Sender
 from frequenz.client.microgrid import ComponentCategory, InverterType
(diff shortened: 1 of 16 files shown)


src/frequenz/sdk/microgrid/_data_pipeline.py
```diff
@@ -485,11 +494,16 @@ async def _stop(self) -> None:
...
+async def initialize(
+    resampler_config: ResamplerConfig,
+    api_power_request_timeout: timedelta = timedelta(seconds=5.0),
+) -> None:
     """Initialize a `DataPipeline` instance.
 
     Args:
         resampler_config: Config to pass on to the resampler.
+        api_power_request_timeout: Timeout to use when making power requests to
+            the microgrid API.
```
llucax: Maybe it would be good to document what happens if it times out? As a user I'm not sure what to do with this. Why should I increase or decrease the timeout?

llucax (commented):
LGTM, just a suggestion to improve documentation, as I'm not sure what would be the effect of changing the timeout or what happens when the timeout is reached.

daniel-zullo-frequenz (commented):
LGTM too, only the suggestion from Luca needs to be checked/addressed

src/frequenz/sdk/microgrid/_data_pipeline.py
shsms: I've added a few more lines.  The only thing is that I've not been specific about the exact duration for which the components get blocked, because currently it is hard-coded to exponentially increase from 1s to a max of 30s when there are subsequent timeouts.

I think we don't have to provide a way to change the duration until we can do so from the UI.

llucax (approved):

PR #1006, opened 2024-07-19
Fix microgrid API URL in example

llucax:
Replace HOST/PORT with URL in example and actually use it to connect to the microgrid API.

### DIFF
--- a/examples/battery_pool.py
+++ b/examples/battery_pool.py
@@ -13,8 +13,7 @@ from frequenz.channels import merge
 from frequenz.sdk import microgrid
 from frequenz.sdk.actor import ResamplerConfig
 
-HOST = "microgrid.sandbox.api.frequenz.io"  # it should be the host name.
-PORT = 62060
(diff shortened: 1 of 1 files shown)


examples/battery_pool.py
```diff
@@ -13,8 +13,7 @@
 from frequenz.sdk import microgrid
 from frequenz.sdk.actor import ResamplerConfig
 
-HOST = "microgrid.sandbox.api.frequenz.io"  # it should be the host name.
-PORT = 62060
+MICROGRID_API_URL = f"grpc://microgrid.sandbox.api.frequenz.io:62060"
```
daniel-zullo-frequenz: this might not need to be a f-string
```suggestion
MICROGRID_API_URL = "grpc://microgrid.sandbox.api.frequenz.io:62060"
```

llucax: Thanks, fixed!

daniel-zullo-frequenz (approved):

PR #1008, opened 2024-07-22
Remove the dummy DCO merge queue workflow

llucax:
This is not necessary anymore now that merge queues use a different branch ruleset, so we can not require DCO for the merge queue to succeed, as PRs will never be queued anyway if the DCO check didn't pass.


daniel-zullo-frequenz (approved):

llucax:
OK, it seems like both protection rules for the branch and for the queue are considered by the queue, as it seems this PR is not accepted by the queue because it waits forever to get the DCO check, even when the protection rule used for the queue doesn't require the DCO check. So I guess we can't get rid of this dummy workflow even with the new protection rules.

Since each merge queue needs its own ruleset and is applied only to one branch, we could just make the general protection rule for all branches mutually exclusive with the rule for new branches, this is more work as we need one ruleset per branch, but it is actually the correct solution, as right now we don't have queues for other branches than the default branch, and we had issues in the past when old branches have a different CI that can never pass the current rules, so I think it makes sense to switch to a more fine-grained ruleset.

llucax:
No, this approach doesn't work either, because we can't still have different checks for the queue and the PR itself, so dropping this PR, too bad we can't get rid of this hack anyway.

PR #1016, opened 2024-08-01
Change consumer power formula to add 0 if meter is not sending data

ela-kotulska-frequenz:
Previously if any meter stopped working, we get None from whole formula.
We need to keep application working even if meter is not sending data.
This field should be configurable in the future.
Another feature might be to fallback to successor components.


ela-kotulska-frequenz:
Done with fallback formulas

PR #1020, opened 2024-08-02
Fixes to the PowerDistributingActor

shsms:


src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
```diff
@@ -169,6 +179,15 @@ async def distribute_power(self, request: Request) -> None:
...
         await self._set_api_power(request, allocations, remaining_power)
 
+    async def _set_to_zero_if_not_working(self) -> None:
+        if not self._component_status_rx:
+            return
+        async for status in self._component_status_rx:
+            if status.component_id in self._pv_inverter_ids:
+                if status.value == ComponentStatusEnum.NOT_WORKING:
+                    api_client = connection_manager.get().api_client
+                    await api_client.set_power(status.component_id, 0.0)
```
llucax: What if there is an exception here (we lost connection to the API temporarily or something? This task will die and never started again, right? Maybe it would be better to catch errors and keep looping when possible?

src/frequenz/sdk/actor/power_distributing/_component_managers/_ev_charger_manager/_ev_charger_manager.py
```diff
@@ -334,14 +334,17 @@ async def _set_api_power(
...
                 failed_components.add(component_id)
                 failed_power += target_power_changes[component_id]
             else:
                 succeeded_components.add(component_id)
 
-            match task.exception():
+            match exc:
                 case asyncio.CancelledError():
                     _logger.warning(
                         "Timeout while setting power to EV charger %s", component_id
```
llucax: A slightly simpler alternative:

```py
            try:
                task.result()
                _logger.warning(
                    "Timeout while setting power to EV charger %s", component_id
                )
            except asyncio.ClientError as err:
                _logger.warning(
                    "Got a client error while setting power to EV charger %s: %s",
                    component_id,
                    err,
                )
            except Exception:
                _logger.exception(
                    "Unknown error while setting power to EV charger: %s",
                    component_id,
                )
            else:
                succeeded_components.add(component_id)
                continue

            failed_components.add(component_id)
            failed_power += target_power_changes[component_id]
```

src/frequenz/sdk/actor/power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
shsms: It turns out this was a hack to handle the special case where there's data, but not enough for the resampler.  And this was actually never triggered, because according to the power distributor, the inverters always had data.

This also conflicts with our other approach, to assume that inverters with missing data are at full power or min power based on the safety requirements of applications.  So I've dropped this commit from this PR.

llucax (dismissed):

src/frequenz/sdk/actor/power_distributing/_component_managers/_ev_charger_manager/_ev_charger_manager.py
shsms: was doing this too, should probably have mentioned. done now.

llucax (approved):

llucax:
Part of #1026.

Issue #575, opened 2023-08-10
Revamp modules/package structure


llucax:
### What's needed?

We need to have a consistent modules and packages structure, so:

1. `import`s are clean and clear
2. Chances of ending with cyclic imports are minimized or completely removed (#84)
3. We minimize the amount of unused code that is indirectly imported
4. The generated documentation makes sense (see the issue in https://github.com/frequenz-floss/frequenz-repo-config-python/discussions/56 for example)
5. Logging makes sense

### Proposed solution

This is still to be researched and decided.

### Additional context

There are a few other issues related to this topic:

* #222 

  With the current import structure, this doesn't always make sense, for example we have a `config` module/package. Using names that are so generic are not always that clear with this scheme, as the user could also have some module called `config`.

  Related issues and discussions:

  * https://github.com/frequenz-floss/frequenz-repo-config-python/discussions/56
  * https://github.com/frequenz-floss/frequenz-sdk-python/discussions/209
  * https://github.com/frequenz-floss/frequenz-repo-config-python/issues/14 (it would be good to check what kind of enforcement this does before going for this)

* #118

  Is assuming we use google-style imports.

* Point 3 (minimize the amount of unused code that is indirectly imported) is also important. Since Python executes the code, and importing a modules can have arbitrary code execution, importing could be very expensive, and it could happen that just to import a small utility function, you end up importing the whole SDK, so it would be good to have this in consideration when thinking about a package/modules structure.

* Point 5: 5. Logging normally uses the module name as logger. This means that logs produced in private modules for example (like `package.some._module`), the logger name will be weird, it will show `package.some._module` instead of the public module where symbols are exported (`package.some`).

* #704

llucax:
I think the main issues with cyclic imports are:

* `frequenz.sdk.timeseries` package depends on `frequenz.sdk.actor`:
  * `BackgroundService` (fixed when moving to `core`)
  * `ChannelRegistry` (should be moved to `timeseries` or `_internal`.)
  * `ComponentMetricRequest` (by `_voltage_streamer`, this module should be moved to `frequenz.sdk.microgrid`)
  * `formula_engine` sends requests to the resampler, the resampler actor should be moved to `frequenz.sdk.timeseries.resampling` or maybe just `frequenz.sdk.resampling`.

* `frequenz.sdk.timeseries` package depends on `frequenz.sdk.microgrid`:
[...]

PR #1031, opened 2024-08-09
Cleanup the `actor` package

shsms:
The `frequenz.sdk.actor` package was shared by the `Actor` implementation and by a number of core actors of the SDK and other utilities.

This PR moves all the core actors and utilities into more appropriate locations.

- `frequenz.sdk.`
  + `actor.ChannelRegistry` → `_internal._channels.ChannelRegistry`
  + `actor.ConfigManagingActor` → `config.ConfigManagingActor`
  + `actor._power_managing` → `microgrid._power_managing`
  + `actor.power_distributing` → `microgrid._power_distributing`
  + `actor._resampling` → `microgrid._resampling`
  + `actor._data_sourcing` → `microgrid._data_sourcing`

This is part of #575


src/frequenz/sdk/microgrid/_power_wrapper.py
```diff
@@ -17,19 +16,13 @@
...
 
 from .._internal._channels import ChannelRegistry, ReceiverFetcher
-from . import _power_managing
-
-# A number of imports had to be done inside functions where they are used, to break
-# import cycles.
-#
-# pylint: disable=import-outside-toplevel
-if typing.TYPE_CHECKING:
-    from ..actor.power_distributing import (  # noqa: F401 (imports used by string type hints)
```
llucax, line 27: Do we still have the "string type hints" or was the comment outdated?

RELEASE_NOTES.md
```diff
@@ -23,6 +23,8 @@
 
 - The `ConfigManagingActor` now uses `collections.abc.Mapping` as the output sender type. This change indicates that the broadcasted configuration is intended to be read-only.
 
+- The `PowerDistributingActor` is no longer part of the public API.  The `Result` types for the power distribution results are now exposed through the `frequenz.sdk.microgrid.battery_pool` module.
```
llucax: Technically all the other actors in `frequenz.sdk.actor` are not longer part of the public API either, they were exposed at the package level (`frequenz/sdk/actor/__init__.py`).

src/frequenz/sdk/timeseries/_voltage_streamer.py
llucax, line 1: Interesting that this module still needs to use `TYPE_CHECKING` and local import for stuff in `microgrid`. Do you remember why this is?

In any case I guess it should be gone when we move microgrid-related stuff from `sdk.timeseries` to `sdk.microgrid` too, right?

llucax (commented):
Amazing to see many local imports and `TYPE_CHECKING` gone :tada:

src/frequenz/sdk/timeseries/_voltage_streamer.py
shsms: It shouldn't be required, it wasn't needed for all checks and tests to run.  But pylint on the code example in that source file was failing without the `TYPE_CHECKING` and non-top-level-imports.  Will look into that in next PR when moving it to microgrid.

src/frequenz/sdk/microgrid/_power_wrapper.py
shsms: I don't think there were string type hints in that file.  My guess is this was copied over from somewhere when we added flake8.

RELEASE_NOTES.md
shsms: looks like I missed a lot of stuff.  Updated.

llucax (approved):
:tada:

PR #1034, opened 2024-08-09
Fix gap in ring buffer when updating missing values

### LINKED ISSUE #646: Inconsistent gap treatment in ring buffer
### What happened?

If there is no data in the ring buffer (e.g. on start up) or no valid data (e.g. only None values), there is inconsistent behavior for gaps and `__len__`, which relies on gaps.

Example on startup (expected 1 gap over the full window):
https://github.com/frequenz-floss/frequenz-sdk-python/blob/01b8d9d1886f298106a505bf68f6c03ced380580/tests/timeseries/test_ringbuffer.py#L211

Example for no valid data (expected length of zero):
https://github.com/frequenz-floss/frequenz-sdk-python/blob/01b8d9d1886f298106a505bf68f6c03ced380580/tests/timeseries/test_ringbuffer.py#L252

### Affected version(s)

v0.x.x

### Affected part(s)

### Extra information

_No response_

cwasicki:
If there are no gaps in the ring buffer and the new value is missing and creates a gap in time, we need to start the new gap after the latest value from before.

Fixes #646.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,1 +1,5 @@
 # Frequenz Python SDK Release Notes
+
+## Bug Fixes
+
+- Fixes a bug in the ring buffer in case the updated value is missing and creates a gap in time.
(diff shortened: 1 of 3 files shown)


cwasicki:
Updated targeting v1.0.0-rc6xx now.

Marenz (approved):

PR #1037, opened 2024-08-16
Fallback components in generated formulas

### LINKED ISSUE #184: Fallback components in generated formulas in LogicalMeter
### What's needed?

Support for automatic logical meter formulas to be regenerated with fallback components from the component graph,  if available.

For example,  in the `grid_power()` formula,  power values from all components directly attached to the grid component are added.  These are usually just meters.  But one of these meters fails,  and that meter has a fallback component,  values from the fallback components can be used in the formula instead.

In the first version,  this can be done for just meters attached to inverters.  If such meters fail,  values from the inverters can be used instead.  

This might also be possible with meters attached to ev-chargers,  if we can be confident that there are no hidden load behind such meters, and it is just the ev chargers.

This will NOT be possible 
  - if the main meter fails, whether it is on the grid side or not,  
  - if a PV meter fails,

because both those don't have fallback devices in the component graph that we can get power readings from.

### Proposed solution

_No response_

### Use cases

_No response_

_No response_

### Additional context

_No response_

ela-kotulska-frequenz:
Fallback components are used in generated formulas. If primary components is unavailable, formula will generate metric from fallback components. Fallback formulas are implemented for:
  - PVPowerFormula
  - ProducerPowerFormula
  - BatteryPowerFormula
  - ConsumerPowerFormula
  - GridPowerFormula
 
All necessary formulas are implemented in this PR

### DIFF
--- a/src/frequenz/sdk/timeseries/formula_engine/_formula_engine.py
+++ b/src/frequenz/sdk/timeseries/formula_engine/_formula_engine.py
@@ -27,6 +27,7 @@ from ._formula_steps import (
     ConstantValue,
     Consumption,
     Divider,
+    FallbackMetricFetcher,
     FormulaStep,
     Maximizer,
     MetricFetcher,
@@ -747,6 +748,7 @@ class FormulaBuilder(Generic[QuantityT]):
         data_stream: Receiver[Sample[QuantityT]],
         *,
         nones_are_zeros: bool,
+        fallback: FallbackMetricFetcher[QuantityT] | None = None,
     ) -> None:
         """Push a metric receiver into the engine.
 
@@ -755,9 +757,18 @@ class FormulaBuilder(Generic[QuantityT]):
             data_stream: A receiver to fetch this metric from.
             nones_are_zeros: Whether to treat None values from the stream as 0s.  If
                 False, the returned value will be a None.
+            fallback: Metric fetcher to use if primary one start sending
+                invalid data (e.g. due to a component stop). If None, the data from
+                primary metric fetcher will be used.
         """
         fetcher = self._metric_fetchers.setdefault(
-            name, MetricFetcher(name, data_stream, nones_are_zeros=nones_are_zeros)
+            name,
+            MetricFetcher(
+                name,
+                data_stream,
+                nones_are_zeros=nones_are_zeros,
+                fallback=fallback,
+            ),
         )
         self._steps.append(fetcher)
 
--- a/src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_battery_power_formula.py
+++ b/src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_battery_power_formula.py
@@ -92,8 +92,6 @@ class BatteryPowerFormula(FormulaGenerator[Power]):
                 "All batteries behind a set of inverters must be requested."
             )
 
-        builder.push_oper("(")
-        builder.push_oper("(")
         # Iterate over the flattened list of inverters
         for idx, comp in enumerate(
             inverter for inverters in battery_inverters for inverter in inverters
--- a/src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_battery_power_formula.py
+++ b/src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_battery_power_formula.py
@@ -3,18 +3,21 @@
 
 """Formula generator from component graph for Grid Power."""
 
+import itertools
 import logging
 
-from frequenz.client.microgrid import ComponentMetricId
+from frequenz.client.microgrid import Component, ComponentCategory, ComponentMetricId
 
 from ....microgrid import connection_manager
 from ..._quantities import Power
 from ...formula_engine import FormulaEngine
+from ._fallback_formula_metric_fetcher import FallbackFormulaMetricFetcher
 from ._formula_generator import (
     NON_EXISTING_COMPONENT_ID,
     ComponentNotFound,
     FormulaGenerationError,
     FormulaGenerator,
+    FormulaGeneratorConfig,
 )
 
 _logger = logging.getLogger(__name__)
@@ -48,8 +51,8 @@ class BatteryPowerFormula(FormulaGenerator[Power]):
         builder = self._get_builder(
             "battery-power", ComponentMetricId.ACTIVE_POWER, Power.from_watts
         )
-        component_ids = self._config.component_ids
-        if not component_ids:
+
+        if not self._config.component_ids:
             _logger.warning(
                 "No Battery component IDs specified. "
                 "Subscribing to the resampling actor with a non-existing "
@@ -63,41 +66,109 @@ class BatteryPowerFormula(FormulaGenerator[Power]):
             )
             return builder.build()
 
+        component_ids = set(self._config.component_ids)
         component_graph = connection_manager.get().component_graph
+        inv_bat_mapping: dict[Component, set[Component]] = {}
 
-        battery_inverters = frozenset(
-            frozenset(
+        for bat_id in component_ids:
+            inverters = set(
                 filter(
                     component_graph.is_battery_inverter,
                     component_graph.predecessors(bat_id),
                 )
             )
-            for bat_id in component_ids
-        )
-
-        if not all(battery_inverters):
-            raise ComponentNotFound(
-                "All batteries must have at least one inverter as a predecessor."
-            )
-
-        all_connected_batteries = set()
-        for inverters in battery_inverters:
-            for inverter in inverters:
-                all_connected_batteries.update(
-                    component_graph.successors(inverter.component_id)
+            if len(inverters) == 0:
+                raise ComponentNotFound(
+                    "All batteries must have at least one inverter as a predecessor."
+                    f"Battery ID {bat_id} has no inverter as a predecessor.",
                 )
 
-        if len(all_connected_batteries) != len(component_ids):
-            raise FormulaGenerationError(
-                "All batteries behind a set of inverters must be requested."
-            )
+            for inverter in inverters:
+                all_connected_batteries = component_graph.successors(
+                    inverter.component_id
+                )
+                battery_ids = set(
+                    map(lambda battery: battery.component_id, all_connected_batteries)
+                )
+                if not battery_ids.issubset(component_ids):
+                    raise FormulaGenerationError(
+                        f"Not all batteries behind inverter {inverter.component_id} "
+                        f"are requested. Missing: {battery_ids - component_ids}"
+                    )
 
-        # Iterate over the flattened list of inverters
-        for idx, comp in enumerate(
-            inverter for inverters in battery_inverters for inverter in inverters
-        ):
-            if idx > 0:
-                builder.push_oper("+")
-            builder.push_component_metric(comp.component_id, nones_are_zeros=True)
+                inv_bat_mapping[inverter] = all_connected_batteries
+
+        if self._config.allow_fallback:
+            fallbacks = self._get_fallback_formulas(inv_bat_mapping)
+
+            for idx, (primary_component, fallback_formula) in enumerate(
+                fallbacks.items()
+            ):
+                if idx > 0:
+                    builder.push_oper("+")
+
+                builder.push_component_metric(
+                    primary_component.component_id,
+                    nones_are_zeros=(
+                        primary_component.category != ComponentCategory.METER
+                    ),
+                    fallback=fallback_formula,
+                )
+        else:
+            for idx, comp in enumerate(inv_bat_mapping.keys()):
+                if idx > 0:
+                    builder.push_oper("+")
+                builder.push_component_metric(comp.component_id, nones_are_zeros=True)
 
         return builder.build()
+
+    def _get_fallback_formulas(
+        self, inv_bat_mapping: dict[Component, set[Component]]
+    ) -> dict[Component, FallbackFormulaMetricFetcher[Power] | None]:
+        """Find primary and fallback components and create fallback formulas.
+
+        The primary component is the one that will be used to calculate the battery power.
+        If it is not available, the fallback formula will be used instead.
+        Fallback formulas calculate the battery power using the fallback components.
+        Fallback formulas are wrapped in `FallbackFormulaMetricFetcher`.
+
+        Args:
+            inv_bat_mapping: A mapping from inverter to connected batteries.
+
+        Returns:
+            A dictionary mapping primary components to their FallbackFormulaMetricFetcher.
+        """
+        fallbacks = self._get_metric_fallback_components(set(inv_bat_mapping.keys()))
+
+        fallback_formulas: dict[
+            Component, FallbackFormulaMetricFetcher[Power] | None
+        ] = {}
+        for primary_component, fallback_components in fallbacks.items():
+            if len(fallback_components) == 0:
+                fallback_formulas[primary_component] = None
+                continue
+
+            battery_ids = set(
+                map(
+                    lambda battery: battery.component_id,
+                    itertools.chain.from_iterable(
+                        inv_bat_mapping[inv] for inv in fallback_components
+                    ),
+                )
+            )
+
+            generator = BatteryPowerFormula(
+                f"{self._namespace}_fallback_{battery_ids}",
+                self._channel_registry,
+                self._resampler_subscription_sender,
+                FormulaGeneratorConfig(
+                    component_ids=battery_ids,
+                    allow_fallback=False,
+                ),
+            )
+
+            fallback_formulas[primary_component] = FallbackFormulaMetricFetcher(
+                generator
+            )
+
+        return fallback_formulas
--- a/src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_consumer_power_formula.py
+++ b/src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_consumer_power_formula.py
@@ -11,11 +11,14 @@ from ....microgrid import connection_manager
 from ..._quantities import Power
 from .._formula_engine import FormulaEngine
 from .._resampled_formula_builder import ResampledFormulaBuilder
+from ._fallback_formula_metric_fetcher import FallbackFormulaMetricFetcher
 from ._formula_generator import (
     NON_EXISTING_COMPONENT_ID,
     ComponentNotFound,
     FormulaGenerator,
+    FormulaGeneratorConfig,
 )
+from ._simple_power_formula import SimplePowerFormula
 
 _logger = logging.getLogger(__name__)
 
@@ -26,6 +29,25 @@ class ConsumerPowerFormula(FormulaGenerator[Power]):
     The consumer power is calculated by summing up the power of all components that
     are not part of a battery, CHP, PV or EV charger chain.
     """
+
+    def _are_grid_meters(self, grid_successors: set[Component]) -> bool:
+        """Check if the grid successors are grid meters.
+
+        Args:
+            grid_successors: The successors of the grid component.
+
+        Returns:
+            True if the provided components are grid meters, False otherwise.
+        """
+        component_graph = connection_manager.get().component_graph
+        return all(
+            successor.category == ComponentCategory.METER
+            and not component_graph.is_battery_chain(successor)
+            and not component_graph.is_chp_chain(successor)
+            and not component_graph.is_pv_chain(successor)
+            and not component_graph.is_ev_charger_chain(successor)
+            for successor in grid_successors
+        )
 
     def generate(self) -> FormulaEngine[Power]:
         """Generate formula for calculating consumer power from the component graph.
@@ -48,15 +70,7 @@ class ConsumerPowerFormula(FormulaGenerator[Power]):
         if not grid_successors:
             raise ComponentNotFound("No components found in the component graph.")
 
-        component_graph = connection_manager.get().component_graph
-        if all(
-            successor.category == ComponentCategory.METER
-            and not component_graph.is_battery_chain(successor)
-            and not component_graph.is_chp_chain(successor)
-            and not component_graph.is_pv_chain(successor)
-            and not component_graph.is_ev_charger_chain(successor)
-            for successor in grid_successors
-        ):
+        if self._are_grid_meters(grid_successors):
             return self._gen_with_grid_meter(builder, grid_successors)
 
         return self._gen_without_grid_meter(builder, self._get_grid_component())
@@ -112,13 +126,30 @@ class ConsumerPowerFormula(FormulaGenerator[Power]):
                 grid_meter.component_id, nones_are_zeros=False
             )
 
-        # push all non consumer components and subtract them from the grid meters
-        for component in non_consumer_components:
-            builder.push_oper("-")
-            builder.push_component_metric(
-                component.component_id,
-                nones_are_zeros=component.category != ComponentCategory.METER,
-            )
+        if self._config.allow_fallback:
+            fallbacks = self._get_fallback_formulas(non_consumer_components)
+
+            for idx, (primary_component, fallback_formula) in enumerate(
+                fallbacks.items()
+            ):
+                builder.push_oper("-")
+
+                # should only be the case if the component is not a meter
+                builder.push_component_metric(
+                    primary_component.component_id,
+                    nones_are_zeros=(
+                        primary_component.category != ComponentCategory.METER
+                    ),
+                    fallback=fallback_formula,
+                )
+        else:
+            # push all non consumer components and subtract them from the grid meters
+            for component in non_consumer_components:
+                builder.push_oper("-")
+                builder.push_component_metric(
+                    component.component_id,
+                    nones_are_zeros=component.category != ComponentCategory.METER,
+                )
 
         return builder.build()
 
@@ -175,13 +206,76 @@ class ConsumerPowerFormula(FormulaGenerator[Power]):
             )
             return builder.build()
 
-        for idx, component in enumerate(consumer_components):
-            if idx > 0:
-                builder.push_oper("+")
+        if self._config.allow_fallback:
+            fallbacks = self._get_fallback_formulas(consumer_components)
 
-            builder.push_component_metric(
-                component.component_id,
-                nones_are_zeros=component.category != ComponentCategory.METER,
-            )
+            for idx, (primary_component, fallback_formula) in enumerate(
+                fallbacks.items()
+            ):
+                if idx > 0:
+                    builder.push_oper("+")
+
+                # should only be the case if the component is not a meter
+                builder.push_component_metric(
+                    primary_component.component_id,
+                    nones_are_zeros=(
+                        primary_component.category != ComponentCategory.METER
+                    ),
+                    fallback=fallback_formula,
+                )
+        else:
+            for idx, component in enumerate(consumer_components):
+                if idx > 0:
+                    builder.push_oper("+")
+
+                builder.push_component_metric(
+                    component.component_id,
+                    nones_are_zeros=component.category != ComponentCategory.METER,
+                )
 
         return builder.build()
+
+    def _get_fallback_formulas(
+        self, components: set[Component]
+    ) -> dict[Component, FallbackFormulaMetricFetcher[Power] | None]:
+        """Find primary and fallback components and create fallback formulas.
+
+        The primary component is the one that will be used to calculate the consumer power.
+        However, if it is not available, the fallback formula will be used instead.
+        Fallback formulas calculate the consumer power using the fallback components.
+        Fallback formulas are wrapped in `FallbackFormulaMetricFetcher` to allow
+        for lazy initialization.
+
+        Args:
+            components: The producer components.
+
+        Returns:
+            A dictionary mapping primary components to their FallbackFormulaMetricFetcher.
+        """
+        fallbacks = self._get_metric_fallback_components(components)
+
+        fallback_formulas: dict[
+            Component, FallbackFormulaMetricFetcher[Power] | None
+        ] = {}
+
+        for primary_component, fallback_components in fallbacks.items():
+            if len(fallback_components) == 0:
+                fallback_formulas[primary_component] = None
+                continue
+
+            fallback_ids = [c.component_id for c in fallback_components]
+            generator = SimplePowerFormula(
+                f"{self._namespace}_fallback_{fallback_ids}",
+                self._channel_registry,
+                self._resampler_subscription_sender,
+                FormulaGeneratorConfig(
+                    component_ids=set(fallback_ids),
+                    allow_fallback=False,
+                ),
+            )
+
+            fallback_formulas[primary_component] = FallbackFormulaMetricFetcher(
+                generator
+            )
+
+        return fallback_formulas
--- a/src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_fallback_formula_metric_fetcher.py
+++ b/src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_fallback_formula_metric_fetcher.py
@@ -34,7 +34,6 @@ class FallbackFormulaMetricFetcher(FallbackMetricFetcher[QuantityT]):
         self._formula_generator: FormulaGenerator[QuantityT] = formula_generator
         self._formula_engine: FormulaEngine[QuantityT] | None = None
         self._receiver: Receiver[Sample[QuantityT]] | None = None
-        self._latest_sample: Sample[QuantityT] | None = None
 
     @property
     def name(self) -> str:
@@ -45,15 +44,6 @@ class FallbackFormulaMetricFetcher(FallbackMetricFetcher[QuantityT]):
     def is_running(self) -> bool:
         """Check whether the formula engine is running."""
         return self._receiver is not None
-
-    @property
-    def latest_sample(self) -> Sample[QuantityT] | None:
-        """Get the latest fetched sample.
-
-        Returns:
-            The latest fetched sample, or `None` if no sample has been fetched yet.
-        """
-        return self._latest_sample
 
     def start(self) -> None:
         """Initialize the formula engine and start fetching samples."""
@@ -87,5 +77,4 @@ class FallbackFormulaMetricFetcher(FallbackMetricFetcher[QuantityT]):
             self._receiver is not None
         ), f"Fallback metric fetcher: {self.name} was not started"
 
-        self._latest_sample = self._receiver.consume()
-        return self._latest_sample
+        return self._receiver.consume()
--- a/src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_fallback_formula_metric_fetcher.py
+++ b/src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_fallback_formula_metric_fetcher.py
@@ -1,0 +1,91 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""FallbackMetricFetcher implementation that uses formula generator."""
+
+from frequenz.channels import Receiver
+
+from ... import Sample
+from ..._quantities import QuantityT
+from .. import FormulaEngine
+from .._formula_steps import FallbackMetricFetcher
+from ._formula_generator import FormulaGenerator
+
+
+# This is done as a separate module to avoid circular imports.
+class FallbackFormulaMetricFetcher(FallbackMetricFetcher[QuantityT]):
+    """A metric fetcher that uses a formula generator.
+
+    The formula engine is generated lazily, meaning it is created only when
+    the `start` or `fetch_next` method is called for the first time.
+    Once the formula engine is initialized, it subscribes to its components
+    and begins calculating and sending the formula results.
+    """
+
+    def __init__(self, formula_generator: FormulaGenerator[QuantityT]):
+        """Create a `FallbackFormulaMetricFetcher` instance.
+
+        Args:
+            formula_generator: A formula generator that generates
+                a formula engine with fallback components.
+        """
+        super().__init__()
+        self._name = formula_generator.namespace
+        self._formula_generator: FormulaGenerator[QuantityT] = formula_generator
+        self._formula_engine: FormulaEngine[QuantityT] | None = None
+        self._receiver: Receiver[Sample[QuantityT]] | None = None
+        self._latest_sample: Sample[QuantityT] | None = None
+
+    @property
+    def name(self) -> str:
+        """Get the name of the fetcher."""
+        return self._name
+
+    @property
+    def is_running(self) -> bool:
+        """Check whether the formula engine is running."""
+        return self._receiver is not None
+
+    @property
+    def latest_sample(self) -> Sample[QuantityT] | None:
+        """Get the latest fetched sample.
+
+        Returns:
+            The latest fetched sample, or `None` if no sample has been fetched yet.
+        """
+        return self._latest_sample
+
+    def start(self) -> None:
+        """Initialize the formula engine and start fetching samples."""
+        engine = self._formula_generator.generate()
+        # We need this assert because generate() can return a FormulaEngine
+        # or FormulaEngine3Phase, but in this case we know it will return a
+        # FormulaEngine. This helps to silence `mypy` and also to verify our
+        # assumptions are still true at runtime
+        assert isinstance(engine, FormulaEngine)
+        self._formula_engine = engine
+        self._receiver = self._formula_engine.new_receiver()
+
+    async def ready(self) -> bool:
+        """Wait until the receiver is ready with a message or an error.
+
+        Once a call to `ready()` has finished, the message should be read with
+        a call to `consume()` (`receive()` or iterated over).
+
+        Returns:
+            Whether the receiver is still active.
+        """
+        if self._receiver is None:
+            self.start()
+
+        assert self._receiver is not None
+        return await self._receiver.ready()
+
+    def consume(self) -> Sample[QuantityT]:
+        """Return the latest message once `ready()` is complete."""
+        assert (
+            self._receiver is not None
+        ), f"Fallback metric fetcher: {self.name} was not started"
+
+        self._latest_sample = self._receiver.consume()
+        return self._latest_sample
--- a/src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_formula_generator.py
+++ b/src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_formula_generator.py
@@ -140,3 +140,110 @@ class FormulaGenerator(ABC, Generic[QuantityT]):
         self,
     ) -> FormulaEngine[QuantityT] | FormulaEngine3Phase[QuantityT]:
         """Generate a formula engine, based on the component graph."""
+
+    def _get_metric_fallback_components(
+        self, components: set[Component]
+    ) -> dict[Component, set[Component]]:
+        """Get primary and fallback components within a given set of components.
+
+        When a meter is positioned before one or more components of the same type (e.g., inverters),
+        it is considered the primary component, and the components that follow are treated
+        as fallback components.
+        If the non-meter component has no meter in front of it, then it is the primary component
+        and has no fallbacks.
+
+        The method iterates through the provided components and assesses their roles as primary
+        or fallback components.
+        If a component:
+         * can act as a primary component (e.g., a meter), then it finds its
+        fallback components and pairs them together.
+         * can act as a fallback (e.g., an inverter or EV charger), then it finds
+        the primary component for it (usually a meter) and pairs them together.
+         * has no fallback (e.g., an inverter that has no meter attached), then it
+        returns an empty set for that component. This means that the component
+        is a primary component and has no fallbacks.
+
+        Args:
+            components: The components to be analyzed.
+
+        Returns:
+            A dictionary where:
+                * The keys are primary components.
+                * The values are sets of fallback components.
+        """
+        graph = connection_manager.get().component_graph
+        fallbacks: dict[Component, set[Component]] = {}
+        for component in components:
+            if component.category == ComponentCategory.METER:
+                fallbacks[component] = self._get_meter_fallback_components(component)
+            else:
+                predecessors = graph.predecessors(component.component_id)
+                if len(predecessors) == 1:
+                    predecessor = predecessors.pop()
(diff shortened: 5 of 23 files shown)


src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_formula_generator.py
```diff
@@ -140,3 +140,60 @@ def generate(
...
     ) -> FormulaEngine[QuantityT] | FormulaEngine3Phase[QuantityT]:
         """Generate a formula engine, based on the component graph."""
+
+    def _get_metric_fallback_components(
+        self, components: set[Component]
+    ) -> dict[Component, set[Component]]:
+        """Get primary and fallback components within a given set of components.
+
+        This function assumes that fallback component has exactly one meter
+        in from of him.
```
Marenz: > has exactly one meter in from of him.

..what?

Marenz: Ooh, one *Meter*, not the distance.  This whole method description was very confusing to me..

llucax: "in FRONT of him", that probably the main source of confusion.

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_formula_generator.py
```diff
@@ -140,3 +140,60 @@ def generate(
...
+        Primary component can have many fallback components but all must
+        have the same category.
+
+        If a component can act as a fallback (e.g., an inverter or EV charger),
+        the method looks for its main component, usually a meter, and pairs them
+        together.
+        If a component has no fallback, (e.g. an inverter that has no meter attached),
+        the method simply returns an empty set for that component.
```
llucax: I'm a bit confused about this, in the case of an inverter and a meter, shouldn't the inverter be the fallback for the meter? This paragraph seems to imply is the other way around.

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_formula_generator.py
```diff
@@ -140,3 +140,60 @@ def generate(
...
+
+        Args:
+            components: The component to be analyzed.
```
llucax: It takes the primary component? I think this methods docs need to be improved a bit and define what is a primary component, what is a fallback component and specify which is being referred to when talking about a component.

src/frequenz/sdk/timeseries/formula_engine/_formula_engine.py
```diff
@@ -747,6 +748,7 @@ def push_metric(
         data_stream: Receiver[Sample[QuantityT]],
         *,
         nones_are_zeros: bool,
+        fallback_metric_fetcher: FallbackMetricFetcher[QuantityT] | None = None,
```
llucax: Super minor suggestion: I would call this just `fallback`, as it is already known from the type it is a metric fetcher.

src/frequenz/sdk/timeseries/formula_engine/_formula_steps.py
```diff
@@ -343,6 +347,41 @@ def apply(self, eval_stack: list[float]) -> None:
         eval_stack.append(val)
 
 
+class FallbackMetricFetcher(ABC, Generic[QuantityT]):
```
llucax: For new classes I recommend inheriting directly from `BackgroundService`.

src/frequenz/sdk/timeseries/formula_engine/_formula_steps.py
```diff
@@ -382,6 +428,9 @@ def stream_name(self) -> str:
         """
         return str(self._stream.__doc__)
 
+    def _value_is_invalid(self, value: QuantityT | None) -> bool:
+        return value is None or value.isnan() or value.isinf()
```
llucax: I suggest using the positive form instead of negative, as a double negative could be hard to read (`if not self._value_is_invalid(...)` (and structure the method a bit more like a question), so:

```suggestion
    def _is_value_valid(self, value: QuantityT | None) -> bool:
        return value is not None and not value.isnan() and not value.isinf()
```

src/frequenz/sdk/timeseries/formula_engine/_formula_steps.py
```diff
@@ -390,7 +439,41 @@ async def fetch_next(self) -> Sample[QuantityT] | None:
...
+            while result[0].timestamp > result[1].timestamp:
+                result[1] = await self._fallback_metric_fetcher.fetch_next()
+
+            if self._value_is_invalid(result[0].value):
+                self._next_value = result[1]
+            else:
+                self._next_value = result[0]
+                # We can stop the fallback metric fetcher if we have a valid value from
+                # primary component.
+                # This is not tested yet.
```
llucax: Probably not to put in a code comment? (the "not tested yet")

llucax (commented):
I just gave this a quick look. Will leave it for people more familiar with the formula engine.

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_formula_generator.py
```diff
@@ -42,13 +44,75 @@ class ComponentNotFound(FormulaGenerationError):
 """
 
 
+class FallbackFormulaMetricFetcher(FallbackMetricFetcher[QuantityT]):
```
shsms: I'm guessing this is done because of circular imports?  If so adding a comment would be helpful.

Still maybe it would be nicer to put this new class in a separate file like `_fallback_components.py` rather than in `_formula_generator.py`

src/frequenz/sdk/timeseries/formula_engine/_formula_steps.py
shsms: Probably best to drop the stop method completely and create an issue about it instead.

shsms (commented):
The logic looks good to me.

Would be nice if you put the docstrings and comments through a grammar checker like https://languagetool.org/ or https://www.grammarly.com/grammar-check.

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_formula_generator.py
Marenz: I actually didn't even notice the missing "front"  XD for me the confusion was more "him.. who is he?". 
Maybe writing "metering component" or so would make it more clear

ela-kotulska-frequenz: Function description was changed.

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_fallback_formula_metric_fetcher.py
```diff
@@ -0,0 +1,55 @@
...
+    @property
+    def is_running(self) -> bool:
+        """Check whether the formula engine is running."""
+        return self._receiver is not None
+
+    def start(self) -> None:
+        """Initialize formula engine and start fetching samples."""
+        self._formula_engine = self._formula_generator.generate()  # type: ignore[assignment]
+        assert isinstance(self._formula_engine, FormulaEngine)
+        self._receiver = self._formula_engine.new_receiver()
```
llucax: Mmm, interesting, but with this implementation `stop()` and `cancel()` will not work. I guess `cancel()` could still be a NOP, but you should probably override `stop()`.

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_formula_generator.py
ela-kotulska-frequenz: Changed description. Please let me know if current one is better

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_formula_generator.py
ela-kotulska-frequenz: Changed method description. Please let me know if current one is better

src/frequenz/sdk/timeseries/formula_engine/_formula_engine.py
ela-kotulska-frequenz: done

src/frequenz/sdk/timeseries/formula_engine/_formula_steps.py
ela-kotulska-frequenz: Wow I didn't know we have this class, thanks :) 
Fixed

src/frequenz/sdk/timeseries/formula_engine/_formula_steps.py
ela-kotulska-frequenz: done

src/frequenz/sdk/timeseries/formula_engine/_formula_steps.py
ela-kotulska-frequenz: Done :) 
https://github.com/frequenz-floss/frequenz-sdk-python/issues/1045

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_formula_generator.py
ela-kotulska-frequenz: > I'm guessing this is done because of circular imports? 

Yes :) Comment was added.

Class was moved to separate file : src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_fallback_formula_metric_fetcher.py

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_fallback_formula_metric_fetcher.py
ela-kotulska-frequenz: `stop()` should work because it has :
```
        if not self._tasks:
            return
```

src/frequenz/sdk/timeseries/formula_engine/_formula_steps.py
llucax: Yeah, and it is about to be replaced by a new one already :laughing: 

* https://frequenz-floss.github.io/frequenz-core-python/latest/reference/frequenz/core/asyncio/#frequenz.core.asyncio.Service
*https://frequenz-floss.github.io/frequenz-core-python/latest/reference/frequenz/core/asyncio/#frequenz.core.asyncio.ServiceBase

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_fallback_formula_metric_fetcher.py
llucax: OK, I guess I reviewed this too quickly and didn't realized the missing `stop()` in the comment was this stop. I also missed that even when it had a `run()` method it didn't actually have any task running. I think I saw a `run()` and `stop()` and immediately assumed it had some task running.

I think this class is actually a good match for the new abstract `Service` interface, and a good case for splitting the `ServiceBase` with a basic implementation using tasks, as it doesn't really help in this case but it is nice to still have the common `Service` interface with start/stop/is_running/etc. for consistency.

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_formula_generator.py
llucax: Yeah, thanks! Much clearer now!

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_fallback_formula_metric_fetcher.py
ela-kotulska-frequenz: Ok, I will do that later today (or on Monday).

ela-kotulska-frequenz: sdk doesn't use frequenz-core-python now. Should I add it and use `Service`?
Or should I do that in separate PR?

ela-kotulska-frequenz: frequenz-core-python has no release, yet.
Should I add dependency to the v1.x.x? 

Maybe I could do that in separate PR (this one is pretty big now)?

tests/timeseries/_battery_pool/test_battery_pool.py
```diff
@@ -590,6 +590,96 @@ async def test_batter_pool_power_two_batteries_per_inverter(
...
 
+async def test_batter_pool_power_fallback_formula(
+    mocker: MockerFixture,
+) -> None:
+    """Test power method with two batteries per inverter."""
+    gen = GraphGenerator()
+    mockgrid = MockMicrogrid(
+        graph=gen.to_graph(
+            [
+                ComponentCategory.METER,  # Grid meter - shouldn't be included in formula
```
shsms, line 601: I think this is not a grid meter.  A grid meter would look like:

```python
    gen.to_graph(
        (
            ComponentCategory.METER,  # Grid meter
            [
                (ComponentCategory.METER, [...]),  # meter with 2 inverters
                (...),  # inverter without meter
            ],
        )
    )
```

Just changing the comment should be enough.

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_formula_generator.py
```diff
@@ -140,3 +140,77 @@ def generate(
...
+                        continue
+
+                # This component is primary component with no fallbacks.
+                fallbacks[component] = set()
+            elif component.category == ComponentCategory.METER:
+                successors = graph.successors(component.component_id)
+
+                if (
+                    all(c.category == ComponentCategory.CHP for c in successors)
+                    or all(c.category == ComponentCategory.INVERTER for c in successors)
```
shsms: Here you'd need to check if they're all the same type as well.  Either all battery inverters or all PV inverters.

shsms (commented):
Still looking into it, but got a couple of items.

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_fallback_formula_metric_fetcher.py
llucax: No, sorry, I didn't mean to do it now, but I actually got thinking about this, and I wonder why isn't this just a `Receiver` with a couple of extra methods and we are already using `stop()` for some receivers (like in [`Merger`](https://github.com/frequenz-floss/frequenz-channels-python/blob/d8bafe8ee30f661ebe2b420c5fd9b240de049dd7/src/frequenz/channels/_merge.py#L131-L136).

Part of the reasons of splitting this to `core` is so we can use `core` in `channels` too, and maybe make these `Receiver`s a `Service` too.

So my suggestion would be to make it a `Receiver` (so `fetch_next()` would just be `receive()`) and follow the interface of `Service` so in the future when this can inherit from `Service` it is not a breaking change.

**OR**, you can leave things are they are and we take care of it when we introduce the dependency to `core`, which will probably need adapting many other classes :)

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_formula_generator.py
ela-kotulska-frequenz: Right, thanks! The first case is also wrong. I am fixing it now.

ela-kotulska-frequenz:
first commit: `_get_metric_fallback_components` method is wrong, fixing it now.

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_formula_generator.py
```diff
@@ -140,3 +140,77 @@ def generate(
...
+                ComponentCategory.INVERTER,
+                ComponentCategory.EV_CHARGER,
+                ComponentCategory.CHP,
+            ):
+                predecessors = graph.predecessors(component.component_id)
+
+                if len(predecessors) == 1:
+                    predecessor = predecessors.pop()
+
+                    if predecessor.category == ComponentCategory.METER:
```
shsms: Not sure if this is the same issue you detected as well, but it could detect grid meters as ev meters, etc. in some cases.

That issue was else where as well, in the component graph.  I made a fix here: https://github.com/frequenz-floss/frequenz-sdk-python/pull/1052

You could just use the new `graph.is_grid_meter()` method to avoid selecting them here.

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_fallback_formula_metric_fetcher.py
ela-kotulska-frequenz: This can't be Receiver, because stopping Receiver won't stop formula.
Here i would like to start/stop formula when needed. This means: subscribe/unsubscribe from all formula components, too.

tests/timeseries/_battery_pool/test_battery_pool.py
ela-kotulska-frequenz: Fixed, thanks! 
I changed it to be grid meter.

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_formula_generator.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_formula_generator.py
ela-kotulska-frequenz: Yes, this is what I found. 
Fixed.

shsms:
I haven't finished reviewing this yet, but I'm happy to take this in and continue review afterwards because I have to work on these parts anyway.

@llucax if you're not opposed to that, please go ahead and approve.  If there are specific things that you would like to change, please create issues.

I will look into improving the documentation a bit as well, after this is merged.  It is too hard to do that in code review.

shsms:
Or maybe better to take this into a feature branch.  Then I can take my fallback values for power distribution PR in there as well, without having to wait for formula support.

And take it in once it all looks good for PVs at least.

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_producer_power_formula.py
```diff
@@ -43,35 +50,111 @@ def generate(  # noqa: DOC502
...
-            lambda component: component_graph.is_pv_chain(component)
-            or component_graph.is_chp_chain(component),
-        )
-
-        if not producer_components:
-            _logger.warning(
-                "Unable to find any producer components in the component graph. "
-                "Subscribing to the resampling actor with a non-existing "
-                "component id, so that `0` values are sent from the formula."
+        if self._config.component_ids is None:
```
shsms: why do we need to check this for producer power formula?

tests/timeseries/_battery_pool/test_battery_pool.py
```diff
@@ -590,6 +590,99 @@ async def test_batter_pool_power_two_batteries_per_inverter(
         assert (await power_receiver.receive()).value == Power.from_watts(-2.0)
 
 
+async def test_batter_pool_power_fallback_formula(
```
shsms:
```suggestion
async def test_battery_pool_power_fallback_formula(

File: src/frequenz/sdk/timeseries/formula_engine/_formula_steps.py [hunk 15]
```diff
@@ -390,7 +421,37 @@ async def fetch_next(self) -> Sample[QuantityT] | None:
         Returns:
             The fetched Sample.
         """
-        self._next_value = await self._stream.receive()
+        if self._fallback is not None and self._fallback.is_running:
+            tasks = [self._stream.receive(), self._fallback.fetch_next()]
+            result = await asyncio.gather(*tasks, return_exceptions=False)
```
llucax: Wouldn't it make sense here to actually return exceptions instead of bubbling exceptions up? If one of the two doesn't raise an exception, you can still use that value (and probably log the error), right?

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_fallback_formula_metric_fetcher.py
```diff
@@ -0,0 +1,54 @@
...
+
+    @property
+    def is_running(self) -> bool:
+        """Check whether the formula engine is running."""
+        return self._receiver is not None
+
+    def start(self) -> None:
+        """Initialize the formula engine and start fetching samples."""
+        self._formula_engine = self._formula_generator.generate()  # type: ignore[assignment]
+        assert isinstance(self._formula_engine, FormulaEngine)
```
llucax: I'm sorry if I already asked this and got lost by some rebase, but why the `ignore` and `assert` here?

tests/timeseries/_battery_pool/test_battery_pool.py
ela-kotulska-frequenz: done

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_producer_power_formula.py
ela-kotulska-frequenz: simplified in last commit

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_fallback_formula_metric_fetcher.py
llucax: I still don't get this.

So there is no `stop()` for now (or there is one inherited from `BackgroundService` that doesn't do anything).

I guess in the future you want to add a `stop()`. You are saying this `stop()` should not stop the `self._formula_engine`? Or the `self._receiver`? Or both? In any case, if you make `FallbackMetricFetcher` a `Receiver`, why can't you just make `stop()` do whatever it should do? If it is a semantic issue, maybe you could call it `pause()` instead of `stop()` to better convey that it is not entirely stopped but just not producing values anymore temporarily?

I think it was also mentioned before, I find it confusing that we have this extra inheritance `FallbackFormulaMetricFetcher` -> `FallbackMetricFetcher`. Can't we get rid of `FallbackMetricFetcher`? What is it for? We already have some unnecessary inheritance in the SDK from the past, I would prefer not to introduce more unless it is really necessary.

llucax (commented):
A few more minor comments.

llucax:
> @llucax if you're not opposed to that, please go ahead and approve. If there are specific things that you would like to change, please create issues.
----
> Or maybe better to take this into a feature branch. Then I can take my fallback values for power distribution PR in there as well, without having to wait for formula support.
> 
> And take it in once it all looks good for PVs at least.
----
OK, I´m not sure what's the current approach, but I'm also fine with take it as it is and create issues if there is more work coming up in this code and it can unblock progress.

Just let me know or approve, and I can create any issues for stuff that I consider relevant after it is merged.

shsms:
Could you update the description to say what is done and what is pending?  It seems outdated.

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_fallback_formula_metric_fetcher.py
ela-kotulska-frequenz: Generics hell:
 
`generate` method definition:
```
    @abstractmethod
    def generate(
        self,
    ) -> FormulaEngine[QuantityT] | FormulaEngine3Phase[QuantityT]:
        """Generate a formula engine, based on the component graph."""
```

It can return `FormulaEngine` or `FormulaEngine3Phase`. 
* FormulaEngine returns formula output in type `Sample`
* FormulaEngine3Phase returns output as in type `Sample3Phase`

But then `MetricFetcher` works only on FormulaEngine - because it stores receiver that returns `Sample`
FallbackMetricFetcher also has to work on  FormulaEngine - because its output is returned by MetricFetcher.

To solve it in `proper` way we would need separate base class for FormulaEngine and FormulaEngine3Phase.

src/frequenz/sdk/timeseries/formula_engine/_formula_steps.py
```diff
@@ -390,7 +421,37 @@ async def fetch_next(self) -> Sample[QuantityT] | None:
...
+
+            # Synchronize the timestamps from two sources
+            while result[0].timestamp < result[1].timestamp:
+                result[0] = await self._stream.receive()
+
+            while result[0].timestamp > result[1].timestamp:
+                result[1] = await self._fallback.fetch_next()
```
shsms: I feel like the first check shouldn't be there.  The fallback should be updated to match the original stream, not the other way around.  Else we could skip data.

If the fallback stream is ahead, then we shouldn't fetch from it until the original stream has caught up, and send the values from the original stream until then.

I guess this would also mean that you can't pull from both the streams together using a `asyncio.gather` either.

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_pv_power_formula.py
```diff
@@ -58,18 +63,79 @@ def generate(  # noqa: DOC502
...
+        else:
+            for idx, component in enumerate(pv_components):
+                if idx > 0:
+                    builder.push_oper("+")
+
+                # should only be the case if the component is not a meter
+                builder.push_component_metric(
+                    component.component_id,
+                    nones_are_zeros=component.category != ComponentCategory.METER,
+                )
```
shsms: Looks like those are outdated comment lines and don't seem to make sense.  Maybe they can go away.

tests/timeseries/mock_resampler.py
```diff
@@ -61,7 +61,7 @@ def power_senders(
                 self._input_channels_receivers[name] = [
                     self._channel_registry.get_or_create(
                         Sample[Quantity], name
-                    ).new_receiver()
+                    ).new_receiver(limit=1)
```
shsms, line 64: This doesn't feel right, because we want the raw data to queue up if we're not consuming continuously.  What was going wrong without this change?

tests/timeseries/test_logical_meter.py
```diff
@@ -77,3 +77,69 @@ async def test_pv_power_no_pv_components(self, mocker: MockerFixture) -> None:
...
+
+            for idx, (meter_power, pv_inverter_power, expected_power) in enumerate(
+                expected_input_output
+            ):
+                await mockgrid.mock_resampler.send_meter_power(meter_power)
+                await mockgrid.mock_resampler.send_pv_inverter_power(pv_inverter_power)
+                mockgrid.mock_resampler.next_ts()
+
+                result = await pv_power_receiver.receive()
+                assert result.value == expected_power, (
```
shsms, line 139:
```suggestion
                assert result.value.isclose(expected_power), (

File: src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_pv_power_formula.py [hunk 21]
```diff
@@ -58,18 +63,79 @@ def generate(  # noqa: DOC502
...
+            ):
+                if idx > 0:
+                    builder.push_oper("+")
+
+                # should only be the case if the component is not a meter
+                builder.push_component_metric(
+                    primary_component.component_id,
+                    nones_are_zeros=False,
+                    fallback=fallback_formula,
+                )
```
shsms: No reason to think the primary component would always be a meter.

```suggestion
                builder.push_component_metric(
                    primary_component.component_id,
                    nones_are_zeros=primary_component.category != ComponentCategory.METER,
                    fallback=fallback_formula,
                )

File: src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_battery_power_formula.py [hunk 22]
```diff
@@ -65,39 +69,106 @@ def generate(
 
         component_graph = connection_manager.get().component_graph
 
-        battery_inverters = frozenset(
-            frozenset(
+        # all_connected_batteries has to be set because
+        # multiple inverters can be connected to the same battery
```
shsms: It has a different meaning otherwise, which is confusing.
```suggestion
        # all_connected_batteries has to be a `set` because
        # multiple inverters can be connected to the same battery

File: src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_fallback_formula_metric_fetcher.py (hunk 16 above)
Comment (llucax) [reply src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_fallback_formula_metric_fetcher.py]: I see, yeah, I also bumped into these issues in #822.

Maybe then the `ignore` won't be needed if you do this instead?

```py
        engine = self._formula_generator.generate()
        # We need this assert because generate() can return a FormulaEngine
        # or FormulaEngine3Phase, but in this case we know it will return a
        # FormulaEngine. This helps to silence `mypy` and also to verify our
        # assumptions are still true at runtime
        assert isinstance(engine, FormulaEngine)
        self._formula_engine = engine
```

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_battery_power_formula.py
```diff
@@ -65,39 +69,106 @@ def generate(
...
-        if not all(battery_inverters):
-            raise ComponentNotFound(
-                "All batteries must have at least one inverter as a predecessor."
-            )
+            if len(inverters) == 0:
+                raise ComponentNotFound(
+                    "All batteries must have at least one inverter as a predecessor.",
+                    "Battery ID %s has no inverter as a predecessor.",
+                    bat_id,
+                )
```
shsms: It is interesting that we need this type of checks in the code.  I'd have thought this is already done by the component graph.

Anyway, this is not a log statement, so the `%s` won't be replaced, you need to use an f-string.

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_battery_power_formula.py
```diff
@@ -65,39 +69,106 @@ def generate(
 
         component_graph = connection_manager.get().component_graph
 
-        battery_inverters = frozenset(
-            frozenset(
+        # all_connected_batteries has to be set because
+        # multiple inverters can be connected to the same battery
+        inv_bat_mapping: dict[Component, set[Component]] = {}
+        all_connected_batteries: set[Component] = set()
```
shsms: `all_connected_batteries` appears to have been a per-battery variable, to identify all the sibling batteries to any given battery.  Maybe it was not clear from the name.

It shouldn't be moved outside of the `for bat_id in component_ids:` loop, because it does not serve its purpose in that level.

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_battery_power_formula.py
```diff
@@ -65,39 +69,106 @@ def generate(
...
-                    component_graph.successors(inverter.component_id)
+                inv_bat_mapping[inverter] = component_graph.successors(
+                    inverter.component_id
                 )
+                all_connected_batteries.update(inv_bat_mapping[inverter])
 
         if len(all_connected_batteries) != len(component_ids):
             raise FormulaGenerationError(
                 "All batteries behind a set of inverters must be requested."
             )
```
shsms: This looks wrong as well,  just checking the lengths is not enough, we should check that `all_connected_batteries` is a subset of `component_ids`.

But because this is an older issue, we can do it separately if you don't want to touch it now.

tests/timeseries/_battery_pool/test_battery_pool.py
```diff
@@ -515,21 +551,43 @@ async def test_batter_pool_power_two_batteries_per_inverter(
...
                 ),
-                ComponentCategory.METER,
                 (
-                    ComponentCategory.INVERTER,
-                    [ComponentCategory.BATTERY, ComponentCategory.BATTERY],
+                    ComponentCategory.METER,
+                    (
+                        ComponentCategory.INVERTER,
+                        [ComponentCategory.BATTERY, ComponentCategory.BATTERY],
+                    ),
```
shsms, line 566: Inverters being connected to the main meter is a case we have to support.  If we want to remove it, it has to be discussed first.  Maybe you can bring it up on Friday.

shsms: Oops, never mind, I see the original graph was wrong.

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_consumer_power_formula.py
```diff
@@ -27,6 +30,25 @@ class ConsumerPowerFormula(FormulaGenerator[Power]):
...
 
+    def _is_grid_meter_mounted(self, grid_successors: set[Component]) -> bool:
+        """Check if the grid component has a meter mounted.
+
+        Args:
+            grid_successors: The successors of the grid component.
+
+        Returns:
+            True if the grid component has a meter mounted, False otherwise.
+        """
```
shsms: I don't understand the name of the function or the docstrings.  What does `mounted` mean for a meter?  I'd call it `are_grid_meters`.

Separately, based on the outcomes on Friday,  we might have to move this to a more generic place for example in the component graph, so it doesn't have to be dealt with separately everywhere.

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_simple_power_formula.py
```diff
@@ -0,0 +1,69 @@
...
+
+class SimplePowerFormula(FormulaGenerator[Power]):
+    """Formula generator from component graph for calculating sum of Power.
+
+    The producer power is calculated by summing up the power of requested components.
+
+    Raises:
+        RuntimeError: If no components are defined in the config or if any
+            component is not found in the component graph.
+    """
```
shsms: Not sure if the second line belongs here.

```suggestion
    """Formula generator from component graph for calculating sum of Power.

    Raises:
        RuntimeError: If no components are defined in the config or if any
            component is not found in the component graph.
    """

File: src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_grid_power_formula.py [hunk 29]
```diff
@@ -30,6 +36,26 @@ def generate(  # noqa: DOC502
...
+        components = {
+            c
+            for c in grid_successors
+            if c.category
+            in {
+                ComponentCategory.INVERTER,
+                ComponentCategory.EV_CHARGER,
+                ComponentCategory.METER,
+            }
+        }
```
shsms: I think that comment can go away.  It doesn't make sense after your refactor.

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_fallback_formula_metric_fetcher.py
shsms: Maybe nicer solution is to have `generate` and `generate_3_phase`, and different formulas implement different functions.  But not for this PR.

tests/timeseries/test_logical_meter.py
ela-kotulska-frequenz: Why? Those values should be equal.

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_fallback_formula_metric_fetcher.py
llucax: * https://github.com/frequenz-floss/frequenz-sdk-python/issues/1055

tests/timeseries/test_logical_meter.py
shsms: This is because of float precision issues, in float arithmatic results.  We had issues of this sort and decided to only use `isclose` from now on, although we still have some old equal checks which need to be migrated.

`isclose` is an equality check in practice, because the default tolerance is very small.

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_fallback_formula_metric_fetcher.py
llucax: @ela-kotulska-frequenz maybe you can add a link to the issue in the comment too :innocent:

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_fallback_formula_metric_fetcher.py
ela-kotulska-frequenz: Done, I made it Receiver.

src/frequenz/sdk/timeseries/formula_engine/_formula_steps.py
ela-kotulska-frequenz: Done

ela-kotulska-frequenz:
@llucax 

>  I find it confusing that we have this extra inheritance FallbackFormulaMetricFetcher -> FallbackMetricFetcher. Can't we get rid of FallbackMetricFetcher? What is it for? We already have some unnecessary inheritance in the SDK from the past, I would prefer not to introduce more unless it is really necessary.

We need it because of circular imports :(
**We would have:**
FormulaGenerator -> FormulaEngine -> MetricFetcher -> FallbackFormulaMetricFetcher -> FormulaGenerator

**Now we have**
FormulaGenerator -> FallbackFormulaMetricFetcher -> FallbackMetricFetcher
FormulaGenerator -> FormulaEngine -> MetricFetcher -> FallbackMetricFetcher

https://github.com/frequenz-floss/frequenz-sdk-python/pull/1037#discussion_r1723439467

I added comment to the FallbackFormulaMetricFetcher:
https://github.com/frequenz-floss/frequenz-sdk-python/pull/1037/commits/23299b4ac212b473836ca43c34eacc678758732f#diff-9800776f7bac0eccbc7d0779b356f024334b154255b892b353dd85c69201f1bfR15

src/frequenz/sdk/timeseries/formula_engine/_formula_steps.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_pv_power_formula.py
ela-kotulska-frequenz: Done

tests/timeseries/mock_resampler.py
ela-kotulska-frequenz: Yes, it is not needed anymore.
It was needed before( When we had no synchronization between MetricFetcher and FallbackMetricFetcher).
Removed

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_pv_power_formula.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_battery_power_formula.py
ela-kotulska-frequenz: done

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_battery_power_formula.py
ela-kotulska-frequenz: > Anyway, this is not a log statement, so the %s won't be replaced, you need to use an f-string.

Done

> It is interesting that we need this type of checks in the code. I'd have thought this is already done by the component graph.

I didn't check if component graph has it.

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_battery_power_formula.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_battery_power_formula.py
ela-kotulska-frequenz: Fixed

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_consumer_power_formula.py
ela-kotulska-frequenz: I did it before our conversation. 

>  we might have to move this to a more generic place for example in the component graph, so it doesn't have to be dealt with separately everywhere.

Agree. IS it ok if I leave it like this until Friday?

ela-kotulska-frequenz: > I'd call it are_grid_meters.

Done

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_simple_power_formula.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_grid_power_formula.py
ela-kotulska-frequenz: Done

ela-kotulska-frequenz:
FallbackMetricFetcher base class is Receiver as @llucax  suggested.

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_fallback_formula_metric_fetcher.py
```diff
@@ -0,0 +1,98 @@
...
+        if self._receiver is None:
+            self.start()
+
+        assert self._receiver is not None
+        # `consume()` must consume sample, that means `self._sample` should not
+        # be available after `consume()` is called.
+        # That's why store it in `_latest_sample`, too.
+        self._sample = await self._receiver.receive()
+        self._latest_sample = self._sample
+        return True
```
llucax: Did you try to just forward the call to the underlying receiver instead of re-implementing the receiver logic? I mean:
```suggestion
        return await self._receiver.ready()
```

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_fallback_formula_metric_fetcher.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def consume(self) -> Sample[QuantityT]:
+        """Return the latest message once `ready()` is complete."""
+        assert (
+            self._sample is not None
+        ), "`consume()` must be preceded by a call to `ready()`"
+
+        sample = self._sample
+        self._sample = None
+        return sample
```
llucax: And:
```suggestion
        return self._receiver.consume()
```

llucax:
> We need it because of circular imports :( **We would have:** FormulaGenerator -> FormulaEngine -> MetricFetcher -> FallbackFormulaMetricFetcher -> FormulaGenerator
> 
> **Now we have** FormulaGenerator -> FallbackFormulaMetricFetcher -> FallbackMetricFetcher FormulaGenerator -> FormulaEngine -> MetricFetcher -> FallbackMetricFetcher
> 
> [#1037 (comment)](https://github.com/frequenz-floss/frequenz-sdk-python/pull/1037#discussion_r1723439467)
> 
> I added comment to the FallbackFormulaMetricFetcher: [23299b4#diff-9800776f7bac0eccbc7d0779b356f024334b154255b892b353dd85c69201f1bfR15](https://github.com/frequenz-floss/frequenz-sdk-python/commit/23299b4ac212b473836ca43c34eacc678758732f#diff-9800776f7bac0eccbc7d0779b356f024334b154255b892b353dd85c69201f1bfR15)

I see :cry:

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_fallback_formula_metric_fetcher.py
ela-kotulska-frequenz: Right, done :)

src/frequenz/sdk/timeseries/formula_engine/_formula_generators/_fallback_formula_metric_fetcher.py
ela-kotulska-frequenz: done

tests/timeseries/mock_resampler.py
shsms: This is still pending.

tests/timeseries/test_logical_meter.py
shsms: This is still pending.

shsms:
The switch to `Receiver` interface for `FallbackMetricFetcher` has made it much simpler and opens the door for more future cleanup.

I'd like to see two more changes, can be in this PR or can be turned into an issue:

1. I feel the `latest_sample` field in `FallbackFormulaMetricFetcher` doesn't belong there.  It took me a while to figure out how it was used in the `MetricFetcher`.

    Instead, I'd like to see a variable `_latest_fallback_sample` in the `MetricFetcher` itself, because that's where it is used.  So keeping it entirely there would make it easier to follow when it gets set as well as when it gets read.  And the `FallbackMetricFetcher.latest_sample` field can go away.

2. Then we can rename the `FallbackMetricFetcher` abc to `LazyReceiver` and move it to `_internal._channels`.  And `FallbackFormulaMetricFetcher` can be renamed to `FallbackFormulaMetricReceiver`.

NB: if you're doing it in this PR, please put in a new commit, so it is easier to read the changes.

shsms (dismissed):
This is great, one more really cool SDK feature!

Approving from my side, but please wait for @llucax's approval also.

ela-kotulska-frequenz:
@sahas-subramanian-frequenz 

> I feel the latest_sample field in FallbackFormulaMetricFetcher doesn't belong there. It took me a while to figure out how it was used in the MetricFetcher.

Done (last commit)

>  Then we can rename the FallbackMetricFetcher abc to LazyReceiver and move it to _internal._channels. And FallbackFormulaMetricFetcher can be renamed to FallbackFormulaMetricReceiver.

In separate PR: https://github.com/frequenz-floss/frequenz-sdk-python/issues/1056

shsms:
> @sahas-subramanian-frequenz

I don't use that account anymore, so I won't be notified if you use it.

The latest commit looks good as well.

llucax (approved):
> This is great, one more really cool SDK feature!

:heavy_plus_sign: 1

Thanks for taking care of it!

Issue #1024, opened 2024-08-06
Accounting for unknown (because of missing data) but probably existing power


shsms:
### What's needed?

When there's no data for some components, we should stop assuming that everything is `None`, because that could be very costly in big installations.

When PV inverters are not responding, we need to assume they're probably operating at maximum power.  

We need to have an estimate of what could be going on with these PV inverters when they're not sending data.  This should be available from the `PVPool`.

Then this needs to be extended to other pools.

llucax:
> When PV inverters are not responding, we need to assume they're probably operating at maximum power.

I would say what the assumption is should be left to be decided by the user, because different applications might need to make different assumptions. For example when shaving peaks, you probably want to assume no power is generated by missing PVs and discharge batteries more preemptively to avoid peaks as much as possible.

My suggestion is to make what the assumption is for missing data configurable per pool, so the user creating the pool can say this pool should assume a component is providing maximum power or not power depending on what its safety requirements are.

llucax:
This is also related to:

* https://github.com/frequenz-floss/frequenz-sdk-python/issues/1021

They are about different types of missing data (or data quality), I'm not sure if once this issue is resolved, #1021 is still necessary. @cwasicki ?

cwasicki:
(fixed your link)

For the case of unreachable inverters this one might be sufficient, but I see #1021 useful to more problems with the data.

PR #1039, opened 2024-08-20
Add fallback value support in PowerDistributingActor for PV inverters

shsms:
When an PV inverter is not available, its power is assumed to be the given fallback value, and this fallback power is excluded from the requested power that's set to the rest of the PV inverters.

### DIFF
--- a/src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_battery_manager.py
+++ b/src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_battery_manager.py
@@ -132,6 +132,7 @@ class BatteryManager(ComponentManager):  # pylint: disable=too-many-instance-att
         component_pool_status_sender: Sender[ComponentPoolStatus],
         results_sender: Sender[Result],
         api_power_request_timeout: timedelta,
+        fallback_power: Power,
     ):
         """Initialize this instance.
 
@@ -143,9 +144,13 @@ class BatteryManager(ComponentManager):  # pylint: disable=too-many-instance-att
             results_sender: Channel sender to send the power distribution results to.
             api_power_request_timeout: Timeout to use when making power requests to
                 the microgrid API.
+            fallback_power: The power to assume a battery has if the battery is not
+                reachable.
         """
         self._results_sender = results_sender
         self._api_power_request_timeout = api_power_request_timeout
+        self._fallback_power = fallback_power
+
         self._batteries = connection_manager.get().component_graph.components(
             component_categories={ComponentCategory.BATTERY}
         )
--- a/src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_battery_manager.py
+++ b/src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_battery_manager.py
@@ -145,11 +145,15 @@ class BatteryManager(ComponentManager):  # pylint: disable=too-many-instance-att
             api_power_request_timeout: Timeout to use when making power requests to
                 the microgrid API.
             fallback_power: The power to assume a battery has if the battery is not
-                reachable.
+                reachable.  Non-zero values are not yet supported.
         """
         self._results_sender = results_sender
         self._api_power_request_timeout = api_power_request_timeout
         self._fallback_power = fallback_power
+
+        assert (
+            self._fallback_power == Power.zero()
+        ), "Non-zero fallback power is not supported for batteries."
 
         self._batteries = connection_manager.get().component_graph.components(
             component_categories={ComponentCategory.BATTERY}
--- a/src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_ev_charger_manager/_ev_charger_manager.py
+++ b/src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_ev_charger_manager/_ev_charger_manager.py
@@ -38,7 +38,9 @@ from ._states import EvcState, EvcStates
 _logger = logging.getLogger(__name__)
 
 
-class EVChargerManager(ComponentManager):
+class EVChargerManager(  # pylint: disable=too-many-instance-attributes
+    ComponentManager
+):
     """Manage ev chargers for the power distributor."""
 
     @override
@@ -47,6 +49,7 @@ class EVChargerManager(ComponentManager):
         component_pool_status_sender: Sender[ComponentPoolStatus],
         results_sender: Sender[Result],
         api_power_request_timeout: timedelta,
+        fallback_power: Power,
     ):
         """Initialize the ev charger data manager.
 
@@ -56,9 +59,13 @@ class EVChargerManager(ComponentManager):
             results_sender: Channel for sending results of power distribution.
             api_power_request_timeout: Timeout to use when making power requests to
                 the microgrid API.
+            fallback_power: The power to assume for an EV charger when it is not
+                reachable.
         """
         self._results_sender = results_sender
         self._api_power_request_timeout = api_power_request_timeout
+        self._fallback_power = fallback_power
+
         self._ev_charger_ids = self._get_ev_charger_ids()
         self._evc_states = EvcStates()
         self._voltage_cache: LatestValueCache[Sample3Phase[Voltage]] = LatestValueCache(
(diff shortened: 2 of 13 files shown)


shsms:
This adds a new argument to the `microgrid.initialize` method called `pv_fallback_power`.  This will be used in the PV power formula calculation as well.

The problem is that `Power.zero()` is probably not the best default value for the field.  And until we have the new API, we don't have access to the rated bounds from the component info, so unless the component streamed some amount of data, we don't know what a max value could be for this field.

And then there is the issue of inverters with varying power ratings at the same location.

Do we keep it optional or make it mandatory?  Should we allow setting fallback power for individual PV inverters or just one average value?

Any ideas?  @cwasicki and @frequenz-floss/python-sdk-team ?

llucax (commented):
In general LGTM, but replying to your message:

> This adds a new argument to the microgrid.initialize method called pv_fallback_power. This will be used in the PV power formula calculation as well.

I wonder if this can't be set on a per-pool base instead for the whole microgrid. It seems weird to have this specific config at the microgrid level.

> The problem is that Power.zero() is probably not the best default value for the field. And until we have the new API, we don't have access to the rated bounds from the component info, so unless the component streamed some amount of data, we don't know what a max value could be for this field.
> 
> And then there is the issue of inverters with varying power ratings at the same location.

Not sure about this, if in the future we can get the rated power, maybe we can introduce a sentinel like `MAX_RATED_POWER` to say we want to set the fallback to the rated power, and maybe `MAX_CURRENT_BOUND` to follow the current bounds?

> Do we keep it optional or make it mandatory? Should we allow setting fallback power for individual PV inverters or just one average value?

I think long term it would be good to be able to have per-component fallback, again for me ideally at the pool level, so we can do something like:

```py
pv_pool1 = microgrid.new_pv_pool([1, 2], fallback_power={
    1: Power.from_kilowatt(50), 2: MAX_RATED_POWER}
)
# uses the same for all components in the pool
pv_pool2 = microgrid.new_pv_pool(fallback_power=MAX_RATED_POWER)
```

shsms:
> I wonder if this can't be set on a per-pool base instead for the whole microgrid. It seems weird to have this specific config at the microgrid level.

I think per-pool won't work because each pool could then have different fallback values, and then some combinations would mean that lower priority actors are able to exceed what a higher priority actor would have set for a given set of bounds.  Else we'd have to use the power manager to apply restrictions on the fallback power as well, which would get very confusing.

> maybe we can introduce a sentinel like MAX_RATED_POWER to say we want to set the fallback to the rated power

yes, with the new API this would be possible, and would simplify actor code.

> I think long term it would be good to be able to have per-component fallback

supporting everything like in your example sounds like a nice idea.

cwasicki:
>I wonder if this can't be set on a per-pool base instead for the whole microgrid. It seems weird to have this specific config at the microgrid level.

If this is possible I would also prefer setting it on pool level.

> And until we have the new API, we don't have access to the rated bounds from the component info

I didn't know we don't have access to these bounds yet, how far are these away?

> so unless the component streamed some amount of data, 

Is this a suggestion to keep track of the maximum value observed so far? Doesn't sound like a bad idea. We would need the same logic later just using the bounds right, since for unreachable inverters we cannot get the bounds neither.

> Should we allow setting fallback power for individual PV inverters or just one average value?

For an intermediate solution I think one value is fine, in practice we would just be conservative about it.

llucax:
> I think per-pool won't work because each pool could then have different fallback values, and then some combinations would mean that lower priority actors are able to exceed what a higher priority actor would have set for a given set of bounds. Else we'd have to use the power manager to apply restrictions on the fallback power as well, which would get very confusing.

I see, what a shame. I think also it makes things more complicated, because different actors will probably want to assume different outcomes, so when we have one app combining multiple actors this might be problematic. For example PV optimization might want to assume the missing power is MAX to avoid grid feed-in, but peak shaving might want to assume missing power is MIN to avoid peaks. :thinking:

shsms:
> > And until we have the new API, we don't have access to the rated bounds from the component info
> 
> I didn't know we don't have access to these bounds yet, how far are these away?

rated bounds are only available in the data streams in the current version of the api.  In the upcoming version, we'll be able to get rated bounds even if there's no live data available, since the start of the app.

> 
> > so unless the component streamed some amount of data,
> 
> Is this a suggestion to keep track of the maximum value observed so far? Doesn't sound like a bad idea. We would need the same logic later just using the bounds right, since for unreachable inverters we cannot get the bounds neither.

no, if we have received any streaming data, we'll have the rated bounds.  so if components were up during application startup and then go down, then the app will have data.  If the app has to restart when a component is unreachable, then we'll not have the info.

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_battery_manager.py
```diff
@@ -151,6 +151,10 @@ def __init__(
         self._api_power_request_timeout = api_power_request_timeout
         self._fallback_power = fallback_power
 
+        assert (
+            self._fallback_power == Power.zero()
+        ), "Non-zero fallback power is not supported for batteries."
+
```
ela-kotulska-frequenz, line 157: Maybe you could mention this in the `fallback_power` docs 4 lines above?
It only says:
```
            fallback_power: The power to assume a battery has if the battery is not
                reachable.
```

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_ev_charger_manager/_ev_charger_manager.py
```diff
@@ -66,6 +66,10 @@ def __init__(
         self._api_power_request_timeout = api_power_request_timeout
         self._fallback_power = fallback_power
 
+        assert (
+            self._fallback_power == Power.zero()
+        ), "Non-zero fallback power is not supported for EV chargers."
+
```
ela-kotulska-frequenz: Same, maybe you could mention this in the `fallback_power` docs 4 lines above?
It only says:
```
            fallback_power: The power to assume a battery has if the battery is not
                reachable.
```

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
```diff
@@ -263,3 +284,69 @@ def _get_pv_inverter_ids(self) -> collections.abc.Set[int]:
...
+
+            meter_data = meter_cache.get()
+            if now - meter_data.timestamp > MAX_DATA_AGE:
+                unreachable_power += self._fallback_power
+                continue
+
+            unreachable_power = Power.from_watts(
+                unreachable_power.as_watts() + meter_data.active_power
+            )
+        return unreachable_power
```
ela-kotulska-frequenz, line 352: Not sure but from logs I saw components are changing state from not-working to working after  ~10sec . 
That would mean that every 10 sec we would try to distribute whole requested pv power - and (probably) feed -in.

Maybe we should check last message timestamp for all meters connected to the requested inverters - not only the unreachable one ?

ela-kotulska-frequenz (dismissed):

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_battery_manager.py
shsms: Done.

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_ev_charger_manager/_ev_charger_manager.py
shsms: Updated.

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
shsms: If components are not sending data, their state doesn't change to `WORKING` until there's data again.  Automatic change only happens from `UNCERTAIN` to `WORKING`, never from `NOT_WORKING` to `WORKING`.  (And `UNCERTAIN` happens when there's data and no error state in the data, but a request to the API has failed, and `NOT_WORKING` happens 10 seconds after there's no data or immediately after data shows component is in error state.)

I've also tested this code and that's not an issue.

shsms:
Will try to get the formula side done as well, before asking for a review again.

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
ela-kotulska-frequenz: Thanks

ela-kotulska-frequenz (dismissed):

shsms:
We have decided to not go ahead with this approach and instead stick to what was originally planned in https://github.com/frequenz-floss/frequenz-sdk-python/issues/1024

PR #1042, opened 2024-08-21
Merge v1.0.0-rc6xx into v1.x.x

llucax:
This pulls one bug fix from rc6xx that weren't ported to v1.x.x:

* **Fix gap in ring buffer when updating a missing value**

We merge the branches anyway to make it perfectly clear and obvious that now all changes in 6xx were incorporated to v1.x.x.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,4 +2,4 @@
 
 ## Bug Fixes
 
-- Fix PV power distribution excluding inverters that haven't sent any data since the application started.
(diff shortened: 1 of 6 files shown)


llucax:
Oh, it seems we have more issues with `mkdocs` that will not be trivial to fix. So I will pull those fixes to a separate branch and see how we deal with them separately so we don't block this (and other) PR(s). We'll have to force-merge until this is fixed.

shsms (approved):

llucax:
Needs a rebase on top of #1043 to fix the CI.

PR #1050, opened 2024-08-22
Allow setting `api_power_request_timeout` in `microgrid.initialize()`

shsms:
This was supposed to be possible from `microgrid.initialize()`, but it was only available from `microgrid._data_pipeline.initialize()` until now.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -16,4 +16,4 @@
 
 - Bump the `grpclib` dependency to pull a fix for using IPv6 addresses.
 
-- Allow setting `api_power_request_timeout` in `microgrid.initialize()`.  This was supposed to be possible since the last release, but wasn't until now.
+- Allow setting `api_power_request_timeout` in `microgrid.initialize()`.
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,19 @@
 # Frequenz Python SDK Release Notes
 
+## Summary
+
+<!-- Here goes a general summary of what this release is about -->
+
+## Upgrading
+
+<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+
+## New Features
+
+<!-- Here goes the main new features and examples or instructions on how to use them -->
+
 ## Bug Fixes
 
 - Bump the `grpclib` dependency to pull a fix for using IPv6 addresses.
+
+- Allow setting `api_power_request_timeout` in `microgrid.initialize()`.  This was supposed to be possible since the last release, but wasn't until now.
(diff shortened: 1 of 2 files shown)


RELEASE_NOTES.md
```diff
@@ -1,5 +1,19 @@
...
+
+## New Features
+
+<!-- Here goes the main new features and examples or instructions on how to use them -->
+
 ## Bug Fixes
 
 - Bump the `grpclib` dependency to pull a fix for using IPv6 addresses.
+
+- Allow setting `api_power_request_timeout` in `microgrid.initialize()`.  This was supposed to be possible since the last release, but wasn't until now.
```
llucax: I'm not sure if talking about what we should have done in the previous release is very relevant here.

```suggestion
- Allow setting `api_power_request_timeout` in `microgrid.initialize()`.
```

RELEASE_NOTES.md
llucax, line 1: I'm fine with leaving it as it is as it is not too bad and probably very unlikely to happen again, but in general it would be good if we split the clearing of the release notes to the adding of new stuff. While cherry-picking some commits I bumped into one instance where I got conflicts because these 2 were mixed together.

src/frequenz/sdk/microgrid/__init__.py
```diff
@@ -238,7 +240,12 @@
 )
 
 
-async def initialize(server_url: str, resampler_config: ResamplerConfig) -> None:
+async def initialize(
+    server_url: str,
+    resampler_config: ResamplerConfig,
+    *,
+    api_power_request_timeout: timedelta = timedelta(seconds=5.0),
```
llucax, line 247: Thinking about the future direction, maybe it would make sense to have some other object, like `MicrogridApiConfig` (or if this is actually a power manager specific thing, `PowerManagerConfig`). If we do this, adding more config related to either the microgrid API or the power manager can be done in a backwards compatible way (by adding more fields to the new config object) and without adding a lot of other keyword arguments to `initialize()`.

llucax (commented):
All comments optional, but I think start thinking about being able to extend the code without introducing breaking changes is becoming more important as we want to make a final 1.0 soon :tm: .

RELEASE_NOTES.md
shsms: This is not clearing the release notes, that was done separately after the release.  But when I pulled rc701 in here, the empty blocks went away,  so adding those here.

src/frequenz/sdk/microgrid/__init__.py
shsms: true, the next one coming up is `pv_fallback_power` which would be under the same category as well, so will do it there.

llucax (approved):

PR #1051, opened 2024-08-26
Unpin mkdocs-autorefs and fix multple URLs for warnings

llucax:
This PR unpins the `mkdocs-autorefs` plugin, so we can use the latest version, but to do that we need to make some changes to the way we link to sections in the docs.

The new version produces warnings when there are more than one possible URL to resolve an "autoref" link, and since we have strict more enabled, the docs fail to build if we have any warnings.

The simplest way to avoid the warning in some cases is just adding the following options:

```
  show_root_heading: false
  show_root_toc_entry: false
```

This avoids anchors being created in the user guide, so it removes the ambiguity of the link.

To remove some other warnings we add a new `autorefs` plugin option to `resolve_closest`, which allows to resolve the reference to the "closest URL". See the documentation for more information: https://github.com/mkdocstrings/autorefs#non-unique-headings.

This PR was done following suggestions in:

* https://github.com/mkdocstrings/autorefs/issues/52

### DIFF
--- a/src/frequenz/sdk/timeseries/formula_engine/_formula_engine.py
+++ b/src/frequenz/sdk/timeseries/formula_engine/_formula_engine.py
@@ -57,68 +57,39 @@ _operator_precedence = {
 
 
 class FormulaEngine(Generic[QuantityT]):
-    """[`FormulaEngine`][frequenz.sdk.timeseries.formula_engine.FormulaEngine]s are a
-    part of the SDK's data pipeline, and provide a way for the SDK to apply formulas on
-    resampled data streams.
+    """An engine to apply formulas on resampled data streams.
 
-    They are used in the SDK to calculate and stream metrics like
-    [`grid_power`][frequenz.sdk.timeseries.grid.Grid.power],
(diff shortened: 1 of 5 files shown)


llucax:
Ready for review.

src/frequenz/sdk/timeseries/formula_engine/_formula_engine.py
```diff
@@ -57,68 +57,39 @@
...
-
-    battery_pool = microgrid.new_battery_pool(priority=5)
-
-    async for power in battery_pool.power.new_receiver():
-        print(f"{power=}")
-    ```
-
-    ### Composition
+    Example: Streaming the power of a battery pool.
+        ```python
```
shsms, line 66: You're already saying please refer to the module documentation, so why can't they get their examples from there too?  It looks a bit weird to have examples after the line that tells people to go elsewhere.

Besides, the formula engine examples already take so long in CI.  Now it is going to be twice as long. :-((

matthias-wende-frequenz (approved):

shsms (approved):
Just an optional comment, it can be cleaned up later as well.

PR #1081, opened 2024-09-27
Improve resiliency of long-running async tasks

### LINKED ISSUE #1078: SoC does not get updated
### What happened?

An application kept reporting 100% SoC while the real SoC queried from the components was around 20%. It seems the application kept the outdated SoC as there wasn't any update for this metric. Apparently the SoC was no longer reported when new PV inverters were added to the microgrid (the PV inverters were reporting only Error state). It took around 5 days to find out about the issue when the components low SoC alert triggered. Unfortunately there is no logs available.

The SoC is reported and it is up-to-date

### Affected version(s)

Observed in v1.0.0-rc601

### Affected part(s)

Microgrid (API, component graph, etc.) (part:microgrid)

### Extra information

The SoC was reported correctly once the application was restarted

shsms:
Closes #1078

### DIFF
--- a/src/frequenz/sdk/_internal/_asyncio.py
+++ b/src/frequenz/sdk/_internal/_asyncio.py
@@ -5,8 +5,12 @@
 
 
 import asyncio
+import logging
 from abc import ABC
-from typing import Any
+from datetime import timedelta
+from typing import Any, Callable, Coroutine
+
+_logger = logging.getLogger(__name__)
 
 
 async def cancel_and_await(task: asyncio.Task[Any]) -> None:
@@ -26,6 +30,25 @@ async def cancel_and_await(task: asyncio.Task[Any]) -> None:
         await task
     except asyncio.CancelledError:
         pass
+
+
+async def run_forever(
+    async_callable: Callable[[], Coroutine[Any, Any, None]],
+    interval: timedelta = timedelta(seconds=1),
+) -> None:
+    """Run a given function forever, restarting it after any exception.
+
+    Args:
+        async_callable: The async callable to run.
+        interval: The interval between restarts.
+    """
+    interval_s = interval.total_seconds()
+    while True:
+        try:
+            await async_callable()
+        except Exception:  # pylint: disable=broad-except
+            _logger.exception("Restarting after exception")
+            await asyncio.sleep(interval_s)
 
 
 class NotSyncConstructible(AssertionError):
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -16,4 +16,4 @@
 
 ## Bug Fixes
 
-<!-- Here goes notable bug fixes that are worth a special mention or explanation -->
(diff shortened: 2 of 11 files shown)


shsms:
Weirdly the `pytest_max` are getting stuck, but `pytest_min` is working.  I will investigate.

shsms:
Appears to be in the config manager issue.  There have not been any changes recently, other than the updates to FileWatcher in channels.

Looks like this needs to be fixed in the channels repo.

llucax:
We also have some issues with tests getting stuck in https://github.com/frequenz-floss/frequenz-dispatch-python/pull/54, it also seems to be related to an update of the `client-dispatch` dependency that in turn has a `channels` updated dependency. It might be related to the changes in `Timer` rather than in `FileWatcher`.

llucax:
@Marenz was doing some debugging for that, I tried to help but didn't have a lot of time to dedicate to this, but maybe we need to increase its priority if it is affecting more projects.

src/frequenz/sdk/_internal/_asyncio.py
```diff
@@ -28,6 +32,25 @@ async def cancel_and_await(task: asyncio.Task[Any]) -> None:
         pass
 
 
+async def run_forever(
+    async_callable: Callable[[], Coroutine[Any, Any, None]],
+    interval: timedelta = timedelta(seconds=1),
```
llucax, line 37: In the future it might be nice to be able to pass a retry strategy. A good reason to move the `retry` module from `client-base` to `core` (https://github.com/frequenz-floss/frequenz-core-python/issues/34).

llucax (dismissed):
Awesome. We should probably also move `run_forever` to `core` (https://github.com/frequenz-floss/frequenz-core-python/issues/33).

src/frequenz/sdk/_internal/_asyncio.py
shsms: ExponentialBackoff is hard to do with generic functions, because it needs to know if a previous attempt succeeded, so that it can reset its backoff interval.

This is easy to do with streams, we just reset it after every incoming message.  But for functions, we'd need something like a timer to reset the backoff interval, if the function hasn't failed for a certain interval, then we say it succeeded, for example.

llucax: Yeah, maybe we can have a strategy that actually takes into account for how long the task ran, so if last time it ran for a long time it waits very little to restart and if it failed immediately it waits quite a long time to restart :thinking:

llucax:
What now with the damn tests :sob: :angry:

shsms:
same timing issue in battery pool tests as in the previous PR that I had a magic fix for.

I'll look for a better fix that doesn't depend on the clock.

ela-kotulska-frequenz (dismissed):

shsms:
@ela-kotulska-frequenz has fixed the merge-time-CI-problem here: https://github.com/frequenz-floss/frequenz-sdk-python/pull/1085

This needs another approval.

ela-kotulska-frequenz (approved):

PR #1087, opened 2024-10-14
Use frequenz-quantities library

### LINKED ISSUE #852: Split the `Quantity`s to its own repository
### What's needed?

We need to allow cloud applications to be able to use `Quantity`s without depending on the microgrid.  We might also want to use `Quantity`s in API clients too, and they should definitely not depend on the SDK.

### Proposed solution

Split the `Quantity` and all its sub-classes to its own repository:

- [x] Create the new repository `frequenz.quantity` and move the code there
- [x] Release a v1.0.0 version (we want to ensure stability from day one, as we are stabilizing all the base repos).
- [x] Remove the code from the SDK and depend on the new repo

### Use cases

All the cloud apps, maybe all API clients.

ela-kotulska-frequenz:
Remove current quantities definition


RELEASE_NOTES.md
```diff
@@ -6,7 +6,7 @@
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+- Replace `frequenz.sdk.timeseries._quantities` module with the external `frequenz.quantities` package.
```
llucax: I would not use internal module names as this is intended for end-users, and also give a link to the new library.

```suggestion
- Replace `Quantity` and its sub-classes (`Power`, `Current`, etc.) in the `frequenz.sdk.timeseries` module with the external [`frequenz-quantities`](https://pypi.org/project/frequenz-quantities/) package. Please add the new library as a dependency and adapt your imports if you are using these types.
```

src/frequenz/sdk/timeseries/__init__.py
```diff
@@ -34,13 +34,7 @@
...
-from .._internal._channels import ReceiverFetcher
-from ._base_types import UNIX_EPOCH, Bounds, Sample, Sample3Phase
-from ._fuse import Fuse
-from ._moving_window import MovingWindow
-from ._periodic_feature_extractor import PeriodicFeatureExtractor
-from ._quantities import (
+from frequenz.quantities import (
     Current,
     Energy,
     Frequency,
```
llucax: Oh, I see you are re-exporting here. I wouldn't do this, I would just go ahead and remove these from here completely (including `__all__` of course).

llucax (commented):
Really nice to see this happening!

src/frequenz/sdk/timeseries/__init__.py
ela-kotulska-frequenz: Done

RELEASE_NOTES.md
ela-kotulska-frequenz: Fixed

shsms (approved):

PR #1088, opened 2024-10-15
Clean some duplicated and unused code

llucax:


llucax:
This was done as part of the upgrade of to the microgrid API v0.17, but placed in a different PR because it is unrelated.

shsms (approved):

PR #1091, opened 2024-10-24
config: Allow reading from multiple files

### LINKED ISSUE #65: Make `ConfigManager` read configuration from multiple files
### What's needed?

Right now the `ConfigManager` is reading the configuration only from one file, but we need a way to set local config defaults and local config overrides.

### Proposed solution

The following config files should be read, by order of precedence:

1. **Local config defaults** will work as per-site constants, something to override the defaults provided by the code itself.
2. **Cloud sync config** is the configuration file written by the the Cloud Sync service with values taken from the UI.
3. **Local config overrides** will work as an emergency override file, to be able to *lock* some configuration in case some particular installation doesn't want to allow configuring some stuff through the UI for safety or other reasons.

So for example:

1. Defines:
   ```toml
   config1 = 11
   config2 = 12
   config3 = 13
   ```
2. Defines:
   ```toml
   config1 = 21
   config4 = 24
   config5 = 25
   ```
3. Defines:
   ```toml
   config1 = 31
   config3 = 33
   config4 = 34
   ```

Then the resulting config should be:
* ```toml
  config1 = 31
  config2 = 12
  config3 = 33
  config4 = 34
  config5 = 25
  ```

If there are issues reading one of the files (for example I/O error or there are syntax errors in the file), an error should be logged and the config manager should proceed with reading the next file.

llucax:
The `ConfigManagingActor` can now take more than one config file, and will read multiple files one after the other, overriding the previous values.

Fixes #65.


llucax:
FYI:

* https://github.com/frequenz-floss/frequenz-core-python/issues/35

RELEASE_NOTES.md
```diff
@@ -6,11 +6,18 @@
...
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+- The `ConfigManagingActor` now takes multiple configuration files as input, and the argument was renamed from `config_file` to `config_files`. If you are using this actor, please update your code. For example:
+
+   ```python
+   # Old
+   actor = ConfigManagingActor(config_file="config.yaml")
+   # New
+   actor = ConfigManagingActor(config_files=["config.yaml"])
```
Marenz: Couldn't we support both parameters for update-convenience?

src/frequenz/sdk/config/_config_managing.py
```diff
@@ -44,7 +72,11 @@ def __init__(
         """Initialize this instance.
 
         Args:
-            config_path: The path to the TOML file with the configuration.
+            config_paths: The paths to the TOML files with the configuration. Order is
```
Marenz: maybe `Order matters as [..]` sounds nicer? :shrug:

src/frequenz/sdk/config/_config_managing.py
```diff
@@ -44,7 +72,11 @@ def __init__(
...
+                important, as the configuration will be read and updated in the order
+                of the paths, so the last path will override the configuration set by
+                the previous paths. Dict keys will be merged recursively, but other
+                objects (like lists) will be replaced by the value in the last path.
```
Marenz, line 90: Why not appending for lists? I suppose it can get complicated if you want to overwrite a value.. but then the same is true for dict, what if youwant to reset it

src/frequenz/sdk/config/_config_managing.py
```diff
@@ -94,45 +139,72 @@ async def _run(self) -> None:
...
 
+        parent_paths = {p.parent for p in self._config_paths}
+
         # FileWatcher can't watch for non-existing files, so we need to watch for the
-        # parent directory instead just in case a configuration file doesn't exist yet
+        # parent directories instead just in case a configuration file doesn't exist yet
         # or it is deleted and recreated again.
         file_watcher = FileWatcher(
-            paths=[self._config_path.parent],
+            paths=list(parent_paths),
```
Marenz, line 148: Maybe worth making this unique? as in, `list(set(parent_paths))`

src/frequenz/sdk/config/_config_managing.py
```diff
@@ -94,45 +139,72 @@ async def _run(self) -> None:
         """
         await self.send_config()
 
+        parent_paths = {p.parent for p in self._config_paths}
```
Marenz, line 153: `unique_parent_paths = [..]` for more clarity? :)

tests/config/test_config_manager.py
```diff
@@ -116,3 +122,235 @@ async def test_update_relative_path(self, config_file: pathlib.Path) -> None:
...
+            config_file.write_text(
+                """
+                logging_lvl = 'INFO'
+                list_non_strict_bool = ["false", "0", "true"]
+                """
+            )
+
+            config = await config_receiver.receive()
+            assert config is not None
+            # The config now doesn't have the old variables only present in config_file,
```
Marenz: This reads a bit confusing

tests/config/test_config_manager.py
```diff
@@ -116,3 +122,235 @@ async def test_update_relative_path(self, config_file: pathlib.Path) -> None:
...
+                "var_float": "3.14",
+                "list_int": "[1,3]",
+                "list_float": "[1,2.0,3.8,12.3]",
+                "var_off": "off",
+                "item_data": '[{"item_id": 1, "name": "My Item 2"}]',
+                "list_non_strict_bool": ["false", "0", "true"],
+                "dict_str_int": {"a": 1, "b": 2, "c": 4, "d": 3},
+            }
+
+            # Now we only updte logging_lvl in config_file2, it still takes precedence
```
Marenz: `update`

tests/config/test_config_manager.py
```diff
@@ -116,3 +122,235 @@ async def test_update_relative_path(self, config_file: pathlib.Path) -> None:
...
+                "var_float": "3.14",
+                "list_int": "[1,3]",
+                "list_float": "[1,2.0,3.8,12.3]",
+                "var_off": "off",
+                "item_data": '[{"item_id": 1, "name": "My Item 2"}]',
+                "list_non_strict_bool": ["false", "0", "true"],
+                "dict_str_int": {"a": 1, "b": 2, "c": 4, "d": 3},
+            }
+
+            # Now add one variable to config_file not present in confit_file2 and remove
```
Marenz: confit_file2

RELEASE_NOTES.md
llucax: Of course we could, but given that we are pre-1.0.0 final I'm not sure if adding the extra maintenance burden is worth it. People will have to update eventually (I don't think it makes sense to keep both options for ever). Specially since we are moving towards using this only in one app, and also eventually I would like the config managing actor to be initialized automatically by the SDK (much like the `microgrid` is, or how `dispatch` also works, with hiding the actor behind a high level interface). Also is a very small and trivial change for apps to update.

src/frequenz/sdk/config/_config_managing.py
llucax: ChatGPT agrees with me:

> The first version sounds better, as "Order is important" is more formal and clearer than "Order matter," which is grammatically incorrect. "Order matter" should be "Order matters" to be correct, but even then, "Order is important" is a better choice for conveying the significance more clearly.

Maybe it's because copilot wrote that :laughing:

llucax: OK, scratch that, it only preferred it because I missed the trailing "s", with the correct grammar it actually prefers your suggestion :unamused:

src/frequenz/sdk/config/_config_managing.py
llucax: Because appending to lists makes it impossible to un-append, so there is no way to actually override the previous config file. With `dict` it is more work, but you can override any value by setting it back to its default value.

src/frequenz/sdk/config/_config_managing.py
llucax: I don't know, it is a set; is it really necessary to say that items in a set are unique?

tests/config/test_config_manager.py
llucax: Is this better?

```py
# The config no longer contains variables that were only present in config_file.
# config_file2 still takes precedence, so the result remains unchanged except
# for the removal of variables unique to the old config_file.
```

tests/config/test_config_manager.py
llucax: fixed

tests/config/test_config_manager.py
llucax: fixed

src/frequenz/sdk/config/_config_managing.py
llucax: fixed

src/frequenz/sdk/config/_config_managing.py
```diff
@@ -19,21 +20,48 @@
...
+            "var3": 4,
+        }
+        ```
     """
 
     # pylint: disable-next=too-many-arguments
     def __init__(
         self,
-        config_path: pathlib.Path | str,
+        config_paths: abc.Iterable[pathlib.Path | str],
```
daniel-zullo-frequenz: I wonder if it would make more sense for `config_paths` to be a `collections.abc.Sequence` rather than a `collections.abc.Iterable`, since order matters. Otherwise, the user could pass an iterable that might not preserve the expected order through the iteration

src/frequenz/sdk/config/_config_managing.py
```diff
@@ -19,21 +20,48 @@
...
-    reading too.
+    If no configuration file could be read, the actor will raise a exception.
+
+    The configuration files are read in the order of the paths, so the last path will
+    override the configuration set by the previous paths. Dict keys will be merged
+    recursively, but other objects (like lists) will be replaced by the value in the
+    last path.
+
+    Example:
+        If `config1.toml` contains:
```
cwasicki, line 38: Maybe have an example which shows the recursive update of dict keys?

src/frequenz/sdk/config/_config_managing.py
```diff
@@ -44,7 +72,11 @@ def __init__(
         """Initialize this instance.
 
         Args:
-            config_path: The path to the TOML file with the configuration.
+            config_paths: The paths to the TOML files with the configuration. Order
```
cwasicki, line 86: I saw your motivation to have three config files in #65, which make sense. Still having concerns that an arbitrary list of config files can be very difficult to maintain, in particular if a sync'ed version is adding new keys that are not wanted. But I guess we will limit the number of config files through our deployment setup?

src/frequenz/sdk/config/_config_managing.py
llucax: I didn't understand very well what you mean, as an iterable will always returns items in *some order*. But then you made me think about `set`, an in this case it is not guaranteed that items are iterated in the order you created it, so I can see a point in favor of `Sequence`, as passing a set could be buggy in a obscure way. It is noisy in my head that a `Sequence` is much more restrictive than a `Iterable` (for example many iterable stuff is safe to use here, like `dict.keys()`, `map`, `filter`, etc.), but I guess if in doubt is better to go with the restrictive option and relax it in the future (forward-compatible) than the other way around. :+1:

src/frequenz/sdk/config/_config_managing.py
llucax: Yeah, users of this class can decide how many config files to try make sense, we'll have the number we need. This will be determined by the app. There might even be users not using the Frequenz platform that just want to have one and only config file and no overrides.

src/frequenz/sdk/config/_config_managing.py
llucax: updated

src/frequenz/sdk/config/_config_managing.py
llucax: updated

llucax:
Enabled auto-merge.

tests/config/test_config_manager.py
llucax: updated

src/frequenz/sdk/config/_config_managing.py
llucax: I just noticed your `[..]`, I wouldn't make it a list because it won't remove the duplicates and it won't be unique anymore. Even more, in that case it's just `non_unique_parent_paths_that_are_actually_the_same_as_config_paths =  list(self._config_paths)` :thinking: :laughing:

RELEASE_NOTES.md
daniel-zullo-frequenz: The ConfigManagingActor only handles `toml` config files
```suggestion
   actor = ConfigManagingActor(config_file="config.toml")
   # New
   actor = ConfigManagingActor(config_files=["config.toml"])
```

src/frequenz/sdk/config/_config_managing.py
```diff
@@ -19,21 +20,57 @@
...
-    When the actor is started, if a configuration file already exists, then it will be
-    read and sent to the `output` sender before the actor starts monitoring the file
-    for updates. This way users can rely on the actor to do the initial configuration
-    reading too.
+    """An actor that monitors a TOML configuration files for updates.
+
+    When the actor is started the configuration files will be read and sent to the
+    output sender. Then the actor will start monitoring the files for updates.
+
+    If no configuration file could be read, the actor will raise a exception.
```
daniel-zullo-frequenz:
```suggestion
    If no configuration file could be read, the actor will raise an exception.
```

src/frequenz/sdk/config/_config_managing.py
```diff
@@ -19,21 +20,57 @@
...
+
+    The configuration files are read in the order of the paths, so the last path will
```
daniel-zullo-frequenz, line 32: I'd suggest to add a brief note to mention that an update on any config file of the monitored paths will make the ConfigManager to read/load all the files in the monitored paths.

daniel-zullo-frequenz (commented):
I have a couple of minor comments to check for. And LGTM otherwise

src/frequenz/sdk/config/_config_managing.py
daniel-zullo-frequenz: I had a similar reasoning about it, and I thought my comment was clear :angel: We'll provide an example next time.
I'm aware `Sequence` is more restrictive but I believe it's worth it in this case and indeed, it can be relaxed if new use-cases needed.
Thanks!

src/frequenz/sdk/config/_config_managing.py
llucax: updated (added the clarification in another previous paragraph)

daniel-zullo-frequenz (commented):
LGTM, I'll wait for @Marenz or @cwasicki to approve it as they both had comments

cwasicki (approved):

PR #1093, opened 2024-11-01
Store the original `Component` and `Connection` in the component graph

### LINKED ISSUE #789: Fix Component Graph conversion from dict to class
### What happened?

The Component Graph implementation is breaking the types. The Component Graph converts the input `Component`s to `dict` to put the data in to the graph using the external graph library `networkx` which supports `dicts`. Then when calling methods to get the `Component`s back, the Component Graph does not properly convert the `dict` to the strong type `Component`

Code to reproduce the bug:

```python
import frequenz.sdk.microgrid.component_graph as gr
from frequenz.sdk.microgrid.client import Connection
from frequenz.sdk.microgrid.component import (
    Component,
    ComponentCategory,
    GridMetadata,
)
from frequenz.sdk.microgrid.fuse import Fuse
from frequenz.sdk.timeseries import Current

fuse = Fuse(max_current=Current.from_amperes(123.0))
components = {
    Component(1, ComponentCategory.GRID, None, GridMetadata(fuse)),
    Component(2, ComponentCategory.METER),
}
connections = {
    Connection(1, 2),
}

graph = gr._MicrogridComponentGraph(components=components, connections=connections)

components = graph.components()
assert len(components) == 2
for component in components:
    if component.component_id == 1:
        assert component.category == ComponentCategory.GRID
        assert component.metadata is not None
        assert isinstance(component.metadata, dict)  # PASSED
        assert isinstance(component.metadata, GridMetadata)  # FAILED
```

```python
assert isinstance(component.metadata, dict)  # FAILED
assert isinstance(component.metadata, GridMetadata)  # PASSED
```

### Affected version(s)

_No response_

### Affected part(s)

Microgrid (API, component graph, etc.) (part:microgrid)

### Extra information

The conversion from `dict` to `Component`s is done using Component(**node) which does not work well when there are nested `dicts`.

The library marshmallow could be useful to deal with this issue as it does support nested classes from nested dicts.

llucax:
For some reason the Component and Connection objects were converted to `dict` to store them in the component graph. This requires re-building the objects when they need to be extracted, which is wasteful.

It also means that type information is lost, and in the future, when adopting the microgrid API client using v0.17, we have a class hierarchy for components instead of relaying in the category and type, so we would even need to store the type if we wanted to rebuild the objects.

Having nested objects is also an issue, as then the nested properties were kept as `dict`, adding an inconsistency and making it type-unsafe.

With this change we are able to get rid of a hack to retrieve the grid metadata.

Fixes #789.

### DIFF
--- a/src/frequenz/sdk/microgrid/component_graph.py
+++ b/src/frequenz/sdk/microgrid/component_graph.py
@@ -22,11 +22,9 @@ flow of power.
 """
 
 import asyncio
-import dataclasses
 import logging
 from abc import ABC, abstractmethod
 from collections.abc import Callable, Iterable
-from dataclasses import asdict
 
 import networkx as nx
 from frequenz.client.microgrid import (
@@ -40,6 +38,10 @@ from frequenz.client.microgrid import (
 _logger = logging.getLogger(__name__)
 
 # pylint: disable=too-many-lines
+
+
+# Constant to store the actual obejcts as data attached to the graph nodes and edges
+_DATA_KEY = "data"
 
 
 class InvalidGraphError(Exception):
@@ -398,18 +400,17 @@ class _MicrogridComponentGraph(
             Set of the components currently connected to the microgrid, filtered by
                 the provided `component_ids` and `component_categories` values.
         """
-        if component_ids is None:
-            # If any node has not node[1], then it will not pass validations step.
-            selection: Iterable[Component] = map(
-                lambda node: Component(**(node[1])), self._graph.nodes(data=True)
-            )
-        else:
-            valid_ids = filter(self._graph.has_node, component_ids)
-            selection = map(lambda idx: Component(**self._graph.nodes[idx]), valid_ids)
+        selection_ids = (
+            self._graph.nodes
+            if component_ids is None
+            else component_ids & self._graph.nodes
+        )
+        selection: Iterable[Component] = (
+            self._graph.nodes[i][_DATA_KEY] for i in selection_ids
+        )
 
         if component_categories is not None:
-            types: set[ComponentCategory] = component_categories
-            selection = filter(lambda c: c.category in types, selection)
+            selection = filter(lambda c: c.category in component_categories, selection)
 
         return set(selection)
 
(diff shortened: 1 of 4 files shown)


llucax:
Skipping release notes as it is invisible to the user.

src/frequenz/sdk/timeseries/grid.py
```diff
@@ -194,33 +194,29 @@ def initialize(
...
+        case 1:
+            metadata = grid_connections[0].metadata
+            if metadata is not None and metadata.fuse is not None:
+                fuse = Fuse(max_current=Current.from_amperes(metadata.fuse.max_current))
+
+            if metadata is None:
+                _logger.warning(
+                    "Unable to get grid metadata, the grid connection point is "
+                    "considered to have no fuse"
                 )
```
daniel-zullo-frequenz: I think we could avoid checking for metadata and fuse conditions twice as follows:
```python
            if metadata is None:
                _logger.warning(
                    "Unable to get grid metadata, the grid connection point is "
                    "considered to have no fuse"
                )
            elif metadata.fuse is None:
                _logger.warning("The grid connection point does not have a fuse")
            else:
                fuse = Fuse(max_current=Current.from_amperes(metadata.fuse.max_current))
```

src/frequenz/sdk/microgrid/component_graph.py
```diff
@@ -398,18 +400,16 @@ def components(
...
-        else:
-            valid_ids = filter(self._graph.has_node, component_ids)
-            selection = map(lambda idx: Component(**self._graph.nodes[idx]), valid_ids)
+        selection: Iterable[Component]
+        selection_ids = (
+            self._graph.nodes
+            if component_ids is None
+            else component_ids & self._graph.nodes
+        )
+        selection = (self._graph.nodes[i][_DATA_KEY] for i in selection_ids)
```
daniel-zullo-frequenz: Optionally you might want to set the type and assign the value at the same time. And the definition of `selection` above can be deleted
```suggestion
        selection: Iterable[Component] = (self._graph.nodes[i][_DATA_KEY] for i in selection_ids)
```

tests/microgrid/test_graph.py
```diff
@@ -1578,3 +1574,48 @@ def test_without_grid_meters(self) -> None:
...
+) -> None:
+    """Add connections to the test graph.
+
+    Args:
+        graph: The graph to add the connections to.
+        *connections: The connections to add.
+    """
+    graph._graph.add_edges_from(
+        (c.start, c.end, {gr._DATA_KEY: c}) for c in connections
+    )
```
daniel-zullo-frequenz: I don't understand why we have 2 similar functions  `_add_connections()` and `add_connections()`. Same question for `add_components()`. Can you please clarify the usage of these functions? Is it because `create_test_graph()` needs them?

daniel-zullo-frequenz (commented):
I have a few minor comments to check for. LGTM otherwise

daniel-zullo-frequenz:
IIUC this PR is fixing https://github.com/frequenz-floss/frequenz-sdk-python/issues/789. If that's the case, it should be mentioned in the PR description

src/frequenz/sdk/microgrid/component_graph.py
llucax: Yeah, declaring the type is not even necessary anymore, I think it was required by a previous iteration of this refactoring.

llucax: Oh, no, it is still needed.

tests/microgrid/test_graph.py
llucax: Long story short, I was trying out Cursor to help me with this refactor and it added these functions without me noticing 😬

Note to self: review the diffs before creating the PR. 😁

llucax:
Updated with the suggested fixes.

daniel-zullo-frequenz (approved):

PR #1095, opened 2024-11-04
Implement and standardize logging configuration via config file 

### LINKED ISSUE #997: Add support for standardized logging
### What's needed?

We discussed to add support to the SDK to have a standardized way for logging in python actors. 

### Proposed solution

@llucax proposed to support fine-grained logging configs via the `logging.config` package.

### Use cases

The goal is to have a common way to set log levels on a per actor basis via the config files.

_No response_

### Additional context

_No response_

ela-kotulska-frequenz:
* Add LoggerConfig and LoggingConfig to standardize logging configuration.
* Create LoggingConfigUpdater to handle runtime config updates.
* Support individual log level settings for each module.

### DIFF
--- a/src/frequenz/sdk/config/_logging_config_updater.py
+++ b/src/frequenz/sdk/config/_logging_config_updater.py
@@ -1,0 +1,194 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""Read and update logging severity from config."""
+
+import logging
+from collections.abc import Mapping
+from dataclasses import field
+from typing import Annotated, Any, Self, cast
+
+import marshmallow
+import marshmallow.validate
+from frequenz.channels import Receiver
+from marshmallow import RAISE
+from marshmallow_dataclass import class_schema, dataclass
+
+from frequenz.sdk.actor import Actor
+
+_logger = logging.getLogger(__name__)
+
+LogLevel = Annotated[
+    str,
+    marshmallow.fields.String(
+        validate=marshmallow.validate.OneOf(choices=logging.getLevelNamesMapping())
+    ),
+]
+
+
+@dataclass
+class LoggerConfig:
+    """A configuration for a logger."""
+
+    level: LogLevel = field(
+        default="NOTSET",
+        metadata={
+            "metadata": {
+                "description": "Log level for the logger. Uses standard logging levels."
+            },
+            "required": False,
+        },
+    )
+    """The log level for the logger."""
+
+
+@dataclass
+class LoggingConfig:
+    """A configuration for the logging system."""
+
+    root_logger: LoggerConfig = field(
+        default_factory=LoggerConfig,
+        metadata={
+            "metadata": {
+                "description": "Default default configuration for all loggers.",
+            },
+            "required": False,
+        },
+    )
+    """The default log level."""
+
+    loggers: dict[str, LoggerConfig] = field(
+        default_factory=dict,
+        metadata={
+            "metadata": {
+                "description": "Configuration for a logger (the key is the logger name)."
+            },
+            "required": False,
+        },
+    )
+    """The list of loggers configurations."""
+
+    @classmethod
+    def load(cls, configs: Mapping[str, Any]) -> Self:  # noqa: DOC502
+        """Load and validate configs from a dictionary.
+
+        Args:
+            configs: The configuration to validate.
+
+        Returns:
+            The configuration if they are valid.
+
+        Raises:
+            ValidationError: if the configuration are invalid.
+        """
+        schema = class_schema(cls)()
+        return cast(Self, schema.load(configs, unknown=RAISE))
+
+
+class LoggingConfigUpdater(Actor):
+    """Actor that listens for logging configuration changes and sets them.
+
+    Example:
+        `config.toml` file:
+        ```toml
+        [logging.root_logger]
+        level = "INFO"
+
+        [logging.loggers."frequenz.sdk.actor.power_distributing"]
+        level = "DEBUG"
+
+        [logging.loggers."frequenz.channels"]
+        level = "DEBUG"
+        ```
+
+        ```python
+        import asyncio
+        from collections.abc import Mapping
+        from typing import Any
+
+        from frequenz.channels import Broadcast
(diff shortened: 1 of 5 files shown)


Marenz (dismissed):

src/frequenz/sdk/config/_logging_config_updater.py
```diff
@@ -0,0 +1,188 @@
...
+    level: LogLevel = field(
+        default="INFO",
+        metadata={
+            "metadata": {
+                "description": "Log level for the logger. Uses standard logging levels."
+            },
+            "required": False,
+        },
+    )
+    """The log level for the logger."""
```
llucax: I guess this comes from an existing internal implementation, and in any case I'm fine with leaving this as is in this PR, but in the future we might want to consider using [`NOTSET`](https://docs.python.org/3/library/logging.html#logging.NOTSET) as the default level, otherwise when we add more options to the `LoggerConfig` and `level` is not specified, it will *silently* override the parent level with `INFO`, which is not probably what we want. For now, to avoid mistakes, we could also make this config required and remove the default.

src/frequenz/sdk/config/_logging_config_updater.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+@dataclass
+class LoggingConfig:
+    """A configuration for the logging system."""
+
+    default: LoggerConfig = field(
```
llucax: I was wondering this when we originally wrote this, and I'm wondering now too. In the `logging` module, this is actually called the *root* logger. *default* seems more intuitive for end users, but I'm not sure, maybe we should just adhere to the Python `logging` nomenclature, so at least then we have a well known definition to point users to. It also makes it more clear the hierarchical nature of the loggers and their configuration. I think I would go with `root_logger` instead of `default`.

src/frequenz/sdk/config/_logging_config_updater.py
```diff
@@ -0,0 +1,188 @@
...
+        """Initialize this instance.
+
+        Args:
+            config_recv: The receiver to listen for configuration changes.
+            log_format: Use the specified format string in logs.
+            log_datefmt: Use the specified date/time format in logs.
+        """
+        super().__init__()
+        self._config_recv = config_recv
+        self._logging_config: LoggingConfig
```
llucax: Maybe `_current_config`?

tests/config/test_logging_config_updater.py
```diff
@@ -0,0 +1,111 @@
...
+    config_sender = config_channel.new_sender()
+    async with LoggingConfigUpdater(
+        config_recv=config_channel.new_receiver().map(
+            lambda app_config: app_config.get("logging", {})
+        )
+    ) as actor:
+        # Check if default config has been set
+        logging_mock.basicConfig.assert_called_once()
+
+        update_logging_spy = mocker.spy(actor, "_update_logging")
```
llucax, line 85: Nice, I didn't know about `spy`. But with this you are not actually testing the `_update_logging` method itself. I would either change this test or add another test and mock `logging.getLogger` to verify it is being called as expected, or maybe even just instead the real `logging` (basically `l = logging.getLogger(...); asser l.level == expected_level`) to verify the config was actually changed.

llucax (commented):
All comments optional, except for the test, I think it is important to test the `_update_logging()` method.

cwasicki (commented):
LGTM, agree with Luca on renaming default to root.

tests/config/test_logging_config_updater.py
ela-kotulska-frequenz: Done, however things get more complicated. :)
1. python doens't allow to override root level config in unit test (which makes sense because it can mess up other tests).
2. loggings setings for other loggers are not reset after unit test method ends, I had to write teardown method (pytest.fixture) to reset them manually. Please look at test.

Also I found bug - with new config, old loggers were not unset. It is fixed and tested.

I am not testing root_logging level, but I think it is ok, because then we start testing external library. 
I could check if logging.basicConfig was called with correct arguments, but I am not sure it it valuable check.

src/frequenz/sdk/config/_logging_config_updater.py
ela-kotulska-frequenz: done

src/frequenz/sdk/config/_logging_config_updater.py
ela-kotulska-frequenz: done

ela-kotulska-frequenz:
Old version had a but: with new config, old loggers were not unset. It is fixed and tested, now.

llucax:
> Old version had a but: with new config, old loggers were not unset. It is fixed and tested, now.

Ah, good catch!

tests/config/test_logging_config_updater.py
llucax: Yeah, I think for that case you can just mock the `logging.getLogger()` function and just trust that if it is called with the right parameters it will work as expected.

src/frequenz/sdk/config/_logging_config_updater.py
```diff
@@ -0,0 +1,194 @@
...
+        return cast(Self, schema.load(configs, unknown=RAISE))
+
+
+class LoggingConfigUpdater(Actor):
+    """Actor that listens for logging configuration changes and sets them.
+
+    Example:
+        `config.toml` file:
+        ```toml
+        [logging.default]
```
llucax:
```suggestion
        [logging.root_logger]
```

llucax (commented):
Just one missing update of `default` -> `root_logger` and it is good to go!

src/frequenz/sdk/config/_logging_config_updater.py
ela-kotulska-frequenz: Done :) thanks!

src/frequenz/sdk/config/_logging_config_updater.py
ela-kotulska-frequenz: Done

llucax (approved):

llucax:
It's nice to see the config stuff getting some love and being put in the SDK :tada:

PR #1097, opened 2024-11-06
Add a function to load configurations with correct type hints

llucax:
When loading configurations from dictionaries, using `marshmallow_dataclass`, the type hints are not preserved by the `load()` method, leading to repetitive casting every time `load()` needs to be used.

This commit adds a new `load_config()` function to the `frequenz.sdk.config` module that takes care of loading configurations from dictionaries into configuration classes with correct type hints.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -22,8 +22,8 @@
 
 - The `ConfigManagingActor` can now take multiple configuration files as input, allowing to override default configurations with custom configurations.
 - Implement and standardize logging configuration with the following changes:
-   * Add LoggerConfig and LoggingConfig to standardize logging configuration.
-   * Create LoggingConfigUpdater to handle runtime config updates.
+   * Add `LoggerConfig` and `LoggingConfig` to standardize logging configuration.
+   * Create `LoggingConfigUpdater` to handle runtime config updates.
    * Support individual log level settings for each module.
 
 ## Bug Fixes
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -21,6 +21,7 @@
 ## New Features
 
 - The `ConfigManagingActor` can now take multiple configuration files as input, allowing to override default configurations with custom configurations.
+* A new `frequenz.sdk.config.load_config()` function is available to load configurations using `marshmallow_dataclass`es with correct type hints.
 - Implement and standardize logging configuration with the following changes:
    * Add `LoggerConfig` and `LoggingConfig` to standardize logging configuration.
    * Create `LoggingConfigUpdater` to handle runtime config updates.
--- a/src/frequenz/sdk/config/_util.py
+++ b/src/frequenz/sdk/config/_util.py
@@ -1,0 +1,45 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""Utilities to deal with configuration."""
+
+from collections.abc import Mapping
+from typing import Any, TypeVar, cast
+
+from marshmallow_dataclass import class_schema
+
+T = TypeVar("T")
+"""Type variable for configuration classes."""
+
+
+def load_config(
+    cls: type[T],
+    config: Mapping[str, Any],
+    /,
+    **marshmallow_load_kwargs: Any,
+) -> T:
+    """Load a configuration from a dictionary into an instance of a configuration class.
+
+    The configuration class is expected to be a [`dataclasses.dataclass`][], which is
+    used to create a [`marshmallow.Schema`][] schema to validate the configuration
+    dictionary.
+
+    To customize the schema derived from the configuration dataclass, you can use
+    [`marshmallow_dataclass.dataclass`][] to specify extra metadata.
+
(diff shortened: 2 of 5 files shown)


RELEASE_NOTES.md
```diff
@@ -22,8 +22,8 @@
 
 - The `ConfigManagingActor` can now take multiple configuration files as input, allowing to override default configurations with custom configurations.
 - Implement and standardize logging configuration with the following changes:
-   * Add LoggerConfig and LoggingConfig to standardize logging configuration.
-   * Create LoggingConfigUpdater to handle runtime config updates.
+   * Add `LoggerConfig` and `LoggingConfig` to standardize logging configuration.
+   * Create `LoggingConfigUpdater` to handle runtime config updates.
```
ela-kotulska-frequenz, line 27: Thanks a lot!

src/frequenz/sdk/config/_util.py
```diff
@@ -0,0 +1,39 @@
...
+from collections.abc import Mapping
+from typing import Any, TypeVar, cast
+
+from marshmallow_dataclass import class_schema
+
+T = TypeVar("T")
+"""Type variable for configuration classes."""
+
+
+def load_config(
```
ela-kotulska-frequenz, line 15: Maybe load_marshmallow_config ?

llucax: I'm not sure if it is worth adding all of that to the name of the function, after all the idea was to make the loading short and neat, and it is documented what's expected. It will be a runtime error passing anything that is not a `marshmallow_dataclass` (actually no, I think a plain `dataclass` will work too! [^1]).

Also, this won't be able to load using a plain `marshmallow.Schema`, it actually needs a `marshmallow_dataclass.dataclass`, so if we want to make it unambigous it would have to be at least `load_marshmallow_dataclass_config()` :grimacing:. At least for me if I read `load_marshmallow_config()` I would expect it to take a plain `marshmallow.Schema`.

[^1]: [`marshmallow_dataclass.class_schema`](https://lovasoa.github.io/marshmallow_dataclass/html/marshmallow_dataclass.html#marshmallow_dataclass.class_schema) also accept a plain `dataclass` and gets a schema out of it on the fly.

llucax: Actually, I will update the docs and tests to reflect this.

llucax: Updated!

ela-kotulska-frequenz (approved):

src/frequenz/sdk/config/_util.py
```diff
@@ -0,0 +1,45 @@
...
+    arguments.
+
+    Args:
+        cls: The configuration class.
+        config: The configuration dictionary.
+        **marshmallow_load_kwargs: Additional arguments to be passed to
+            [`marshmallow.Schema.load`][].
+
+    Returns:
+        The loaded configuration as an instance of the configuration class.
```
daniel-zullo-frequenz, line 40: I understand the function is internal but IMO the documentation should include the `Raises` section as it is the only way the caller of the function can handle validation errors.

llucax: I agree it could be a good addition, but the docs say very clearly that this is just a wrapper for `marshmallow.Schema.load()`, so I guess users can infer it from there. I can add it in a separate PR if you think it is worth it though.

daniel-zullo-frequenz: > I can add it in a separate PR if you think it is worth it though.

No, it's probably not worth it. Though I'd explicitly add it if the function is exposed to the user at some point

llucax: It is already exposed via `frequenz.sdk.config` though :)

daniel-zullo-frequenz: oops, would you mind to add it in a different PR then? :angel: 

I have some other concern that I'd like to mention before adding any `Raise` docs.

```python
def test_none_settings() -> None:
    """Test loading None settings."""
    config: dict[str, Any] = {"key": "value"}

    text: str = "text"

    with pytest.raises(marshmallow.ValidationError) as exc_info:
        loaded_config = load_config(SimpleConfig, None)  # mypy complains incompatible type "None";

    with pytest.raises(marshmallow.ValidationError) as exc_info:
        loaded_config = load_config(
            SimpleConfig, config.get("unknown", None)
        )  # mypy is happpy with `None` when `get()` is used
```

Since `mypy` doesn’t raise any issues when get() is used, we can assume marshmallow will handle it. Marshmallow will raise `ValidationError` for the `None` case if default marshmallow parameters are used but that might be different when non-default parameters are set, for instance, `unknown=INCLUDE` will make marshmallow to raise a TypeError if None is passed as a `config`.

llucax: Just for the record, we had a quick call and decided it would be impossible to document every pitfall we can have with marshmallow in this function, so we'll leave it like this. If we touch this code again, the docs can be made a bit more clear about this being just a wrapper over `marshmallow.Schema.load()`.

PR #1102, opened 2024-11-15
Check that load_config raises ValidationError if config is None

ela-kotulska-frequenz:


llucax (approved):
LGTM. I'm working a revamped config manager that will handle all of this. I think I will create a draft PR so you can have a look and see what's coming.

PR #1103, opened 2024-11-15
Allow passing a `base_schema` to `load_config()`

llucax:
This PR adds a new argument to `load_config()` to allow passing a `base_schema` to be used when loading the configuration. This is necessary to be able to use custom fields in the schema, like the ones provided by `frequenz.quantities`.

It also drops support for using `marshmallow_dataclas.dataclass` directly. `marshmallow_dataclass.dataclass` is intended to be used only when using `my_dataclass.Schema` to get the schema. But using this is not very convenient when using type hints as they are not well-supported by `marshmallow`, as the `load()` function can't have hints.
    
This is actually why `load_config()` exists in the first place, so we are using `class_schema()` instead, so we don't really need that our types are decorated with `marshmallow_dataclass`, we can use the built-in `dataclass` instead, we just need to add the appropriate metadata if we want more complex validation.

Using `class_shema()` is also necessary to be able to pass a `base_schema`, which we'll need when we want to use schemas with custom fields, like the ones provided by `frequenz.quantities`. 

Finally, it improves `load_config` documentation to make it more explicit that this is just a wrapper to external libraries, so users should read their documentation in full and which functions are used exactly.

### DIFF
--- a/src/frequenz/sdk/config/_util.py
+++ b/src/frequenz/sdk/config/_util.py
@@ -6,6 +6,7 @@
 from collections.abc import Mapping
 from typing import Any, TypeVar, cast
 
+from marshmallow import Schema
 from marshmallow_dataclass import class_schema
 
 T = TypeVar("T")
@@ -16,6 +17,7 @@ def load_config(
     cls: type[T],
     config: Mapping[str, Any],
     /,
+    base_schema: type[Schema] | None = None,
     **marshmallow_load_kwargs: Any,
 ) -> T:
     """Load a configuration from a dictionary into an instance of a configuration class.
@@ -42,13 +44,16 @@ def load_config(
     Args:
         cls: The configuration class.
         config: The configuration dictionary.
+        base_schema: An optional class to be used as a base schema for the configuration
+            class. This allow using custom fields for example. Will be passed to
+            [`marshmallow_dataclass.class_schema`][].
         **marshmallow_load_kwargs: Additional arguments to be passed to
             [`marshmallow.Schema.load`][].
 
     Returns:
         The loaded configuration as an instance of the configuration class.
     """
-    instance = class_schema(cls)().load(config, **marshmallow_load_kwargs)
+    instance = class_schema(cls, base_schema)().load(config, **marshmallow_load_kwargs)
     # We need to cast because `.load()` comes from marshmallow and doesn't know which
     # type is returned.
     return cast(T, instance)
--- a/src/frequenz/sdk/config/_util.py
+++ b/src/frequenz/sdk/config/_util.py
@@ -22,13 +22,22 @@ def load_config(
 
     The configuration class is expected to be a [`dataclasses.dataclass`][], which is
     used to create a [`marshmallow.Schema`][] schema to validate the configuration
-    dictionary.
+    dictionary using [`marshmallow_dataclass.class_schema`][] (which in turn uses the
+    [`marshmallow.Schema.load`][] method to do the validation and deserialization).
 
-    To customize the schema derived from the configuration dataclass, you can use
-    [`marshmallow_dataclass.dataclass`][] to specify extra metadata.
+    To customize the schema derived from the configuration dataclass, you can use the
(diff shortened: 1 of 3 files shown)


src/frequenz/sdk/config/_util.py
```diff
@@ -22,13 +22,22 @@ def load_config(
 
     The configuration class is expected to be a [`dataclasses.dataclass`][], which is
     used to create a [`marshmallow.Schema`][] schema to validate the configuration
-    dictionary.
+    dictionary using [`marshmallow.Schema.load`][] (which in turn uses the
+    [`marshmallow.Schema.load`][] method to do the validation and deserialization).
```
ela-kotulska-frequenz: > using [`marshmallow.Schema.load`][] (which in turn uses the [`marshmallow.Schema.load`]

Wrong name?

src/frequenz/sdk/config/_util.py
```diff
@@ -4,22 +4,30 @@
...
 from marshmallow import Schema
 from marshmallow_dataclass import class_schema
 
-T = TypeVar("T")
+
+class Dataclass(Protocol):
+    """A protocol for dataclasses."""
+
+    __dataclass_fields__: ClassVar[dict[str, Any]]
+    """The fields of the dataclass."""
```
ela-kotulska-frequenz, line 24: It took me a while to understand why it checks if ` __dataclass_fields__` is available .
Is it because `is_dataclass` method uses it?
Are you sure we can rely on this field? 

Maybe instead we could check that in `load_config` method during runtime? This way you will use only public interface.
```
from dataclasses import is_dataclass
def load_config( cls: type[T], ...) -> T:
   if not is_dataclass(cls): raise TypeError(...)
   ....
```

But if you prefer to leave it like this, then could you please add one more test, to make sure  `marshmallow_dataclass` has not `__dataclass_fields__`?

src/frequenz/sdk/config/_util.py
llucax: Nope: https://marshmallow.readthedocs.io/en/stable/marshmallow.schema.html#marshmallow.schema.Schema.load

(this would have been caught by the docs tests if it were wrong)

llucax: Yes, the first one should be `class_schema`. Will update.

src/frequenz/sdk/config/_util.py
llucax: > It took me a while to understand why it checks if  __dataclass_fields__ is available .

Good point, I will add a comment.

>  Is it because is_dataclass method uses it?

No, I added the protocol to be able to do that, but the protocol is a way to tell the type system you want to accept dataclasses. Not sure how familiar are you with protocols but they are basically interfaces, [structural subtyping](https://mypy.readthedocs.io/en/stable/protocols.html) is called.

As dataclasses don't inherit from any class, we can't use a base class as the type for arguments for functions. `is_dataclass` is a dynamic check. This is sort of a hack, and assumes that anything that defines __dataclass_fields__ is a dataclass, which is not strictly true, but I think it should work pretty well in practice.

>  Are you sure we can rely on this field?

Well, if you are going to go legal, then no, we can't rely on it, at least according to [this](https://github.com/python/mypy/issues/15974#issuecomment-1694781006). I think in practice [everybody seems to be relying on it](https://github.com/python/mypy/issues/15974#issuecomment-1694993493) so I think we should be safe to do so for the time being.

llucax: Added a comment to try to clarify this a bit.

llucax:
Updated.

src/frequenz/sdk/config/_util.py
ela-kotulska-frequenz: > could you please add one more test, to make sure marshmallow_dataclass has not __dataclass_fields__?

I just checked and dataclass.is_dataclass returns True if class is  marshmallow_dataclass.dataclass...

ela-kotulska-frequenz (approved):

PR #1111, opened 2024-11-21
Update to repo-config v0.11.0

llucax:
Update the files using the migration script with the following changes:

* Group artifact-related actions so they are updated together.
* Add `asyncio_default_fixture_loop_scope` `pytest` option as not specifying this is deprecated.
* Disable some `pylint` checks that are already checked by `mypy`.

We also bump the build dependencies, so the distribution files use names that are compatible with PEP625, which will soon be required by PyPI and fix `pylint` warnings in examples that were previously ignored as they were not checked by `pylint` due to a sybil bug.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -57,3 +57,11 @@ updates:
     labels:
       - "part:tooling"
       - "type:tech-debt"
+    groups:
+      compatible:
+        update-types:
+          - "minor"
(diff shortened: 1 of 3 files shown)


llucax:
Skipping release notes as there are no visible changes.

ela-kotulska-frequenz (approved):

PR #1112, opened 2024-11-21
Fix ConfigLoggingUpdater not changing root logging level

ela-kotulska-frequenz:
Setting log level for root logger with method `logging.basicConfig` during runtime is not working. Instead logging.getLogger() is used.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -14,4 +14,4 @@
 
 ## Bug Fixes
 
-<!-- Here goes notable bug fixes that are worth a special mention or explanation -->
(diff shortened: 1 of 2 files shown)


llucax (dismissed):
Interesting.

> This function does nothing if the root logger already has handlers configured, unless the keyword argument force is set to True.

https://docs.python.org/3/library/logging.html#logging.basicConfig

 So the format and date format will also not be affected if someone called `basicConfig` before the actor is instantiated. Maybe we should document this behaviour.

ela-kotulska-frequenz:
> Maybe we should document this behaviour.

Good idea. Documentation added

llucax (approved):

PR #1116, opened 2024-11-25
Fix `ConfigManagingActor` raising unhandled exceptions when file doesn't exist

ela-kotulska-frequenz:
This was detected when I edited config file with `vim` and `config.toml.swp` file was created and deleted.

### DIFF
--- a/src/frequenz/sdk/config/_config_managing.py
+++ b/src/frequenz/sdk/config/_config_managing.py
@@ -107,14 +107,11 @@ class ConfigManagingActor(Actor):
         self._force_polling: bool = force_polling
         self._polling_interval: timedelta = polling_interval
 
-    def _read_config(self) -> abc.Mapping[str, Any]:
+    def _read_config(self) -> abc.Mapping[str, Any] | None:
         """Read the contents of the configuration file.
 
         Returns:
             A dictionary containing configuration variables.
-
-        Raises:
-            ValueError: If config file cannot be read.
         """
         error_count = 0
         config: dict[str, Any] = {}
@@ -138,14 +135,18 @@ class ConfigManagingActor(Actor):
                 error_count += 1
 
         if error_count == len(self._config_paths):
-            raise ValueError(f"{self}: Can't read any of the config files")
+            _logger.error(
+                "%s: Can't read any of the config files, ignoring config update.", self
+            )
+            return None
 
         return config
 
     async def send_config(self) -> None:
         """Send the configuration to the output sender."""
         config = self._read_config()
-        await self._output.send(config)
+        if config is not None:
+            await self._output.send(config)
 
     async def _run(self) -> None:
         """Monitor for and send configuration file updates.
--- a/src/frequenz/sdk/config/_config_managing.py
+++ b/src/frequenz/sdk/config/_config_managing.py
@@ -74,7 +74,6 @@ class ConfigManagingActor(Actor):
         self,
         config_paths: abc.Sequence[pathlib.Path | str],
         output: Sender[abc.Mapping[str, Any]],
-        event_types: abc.Set[EventType] = frozenset(EventType),
         *,
         name: str | None = None,
         force_polling: bool = True,
@@ -89,7 +88,6 @@ class ConfigManagingActor(Actor):
                 the previous paths. Dict keys will be merged recursively, but other
                 objects (like lists) will be replaced by the value in the last path.
             output: The sender to send the configuration to.
-            event_types: The set of event types to monitor.
             name: The name of the actor. If `None`, `str(id(self))` will
                 be used. This is used mostly for debugging purposes.
             force_polling: Whether to force file polling to check for changes.
@@ -106,7 +104,6 @@ class ConfigManagingActor(Actor):
             for config_path in config_paths
         ]
         self._output: Sender[abc.Mapping[str, Any]] = output
-        self._event_types: abc.Set[EventType] = event_types
         self._force_polling: bool = force_polling
         self._polling_interval: timedelta = polling_interval
 
@@ -166,17 +163,32 @@ class ConfigManagingActor(Actor):
         # or it is deleted and recreated again.
         file_watcher = FileWatcher(
             paths=list(parent_paths),
-            event_types=self._event_types,
+            event_types={EventType.CREATE, EventType.MODIFY},
             force_polling=self._force_polling,
             polling_interval=self._polling_interval,
         )
 
         try:
             async for event in file_watcher:
+                if not event.path.exists():
+                    _logger.error(
+                        "%s: Received event %s, but the watched path %s doesn't exist.",
+                        self,
+                        event,
+                        event.path,
+                    )
+                    continue
                 # Since we are watching the whole parent directories, we need to make
                 # sure we only react to events related to the configuration files we
                 # are interested in.
-                if not any(event.path.samefile(p) for p in self._config_paths):
+                #
+                # pathlib.Path.samefile raises error if any path doesn't exist so we need to
+                # make sure the paths exists before calling it. This could happen as it is not
+                # required that all config files exist, only one is required but we don't know
+                # which.
+                if not any(
+                    event.path.samefile(p) for p in self._config_paths if p.exists()
+                ):
                     continue
 
                 match event.type:
@@ -195,8 +207,9 @@ class ConfigManagingActor(Actor):
                         )
                         await self.send_config()
                     case EventType.DELETE:
-                        _logger.info(
-                            "%s: The configuration file %s was deleted, ignoring...",
+                        _logger.error(
+                            "%s: Unexpected DELETE event for path %s. Please report this "
+                            "issue to Frequenz.",
                             self,
                             event.path,
                         )
--- a/src/frequenz/sdk/config/_config_managing.py
+++ b/src/frequenz/sdk/config/_config_managing.py
@@ -130,6 +130,15 @@ class ConfigManagingActor(Actor):
             except ValueError as err:
                 _logger.error("%s: Can't read config file, err: %s", self, err)
                 error_count += 1
+            except OSError as err:
+                # It is ok for config file to don't exist.
+                _logger.error(
+                    "%s: Error reading config file %s (%s). Ignoring it.",
+                    self,
+                    err,
+                    config_path,
+                )
+                error_count += 1
 
         if error_count == len(self._config_paths):
             raise ValueError(f"{self}: Can't read any of the config files")
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,7 +6,8 @@
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+- The `ConfigManagingActor` now only reacts to `CREATE` and `MODIFY` events. `DELETE` is not supported anymore and are ignored.
+- Remove the `event_types` argument from the `ConfigManagingActor` constructor.
 
 ## New Features
 
@@ -15,3 +16,8 @@
 ## Bug Fixes
 
 - Fix a bug in the resampler that could end up with an *IndexError: list index out of range* exception when a new resampler was added while awaiting the existing resampler to finish resampling.
+
+- Fix bugs with `ConfigManagingActor`:
+  - Raising unhandled exceptions when any file in config directory was deleted.
+  - Raising unhandled exception if not all config files exist.
+  - Eliminate recursive actor crashes when all config files were missing.
--- a/tests/config/test_config_manager.py
+++ b/tests/config/test_config_manager.py
@@ -8,9 +8,12 @@ from collections import defaultdict
 from collections.abc import Mapping, MutableMapping
 from dataclasses import dataclass
 from typing import Any
+from unittest.mock import MagicMock
 
 import pytest
 from frequenz.channels import Broadcast
+from frequenz.channels.file_watcher import Event, EventType
+from pytest_mock import MockerFixture
 
 from frequenz.sdk.config import ConfigManagingActor
 from frequenz.sdk.config._config_managing import _recursive_update
@@ -264,6 +267,96 @@ class TestActorConfigManager:
                 "var_off": "on",
                 "dict_str_int": {"a": 1, "b": 2, "c": 4},
             }
+
+    async def test_actor_works_if_not_all_config_files_exist(
+        self, config_file: pathlib.Path
+    ) -> None:
+        """Test ConfigManagingActor works if not all config files exist."""
+        config_channel: Broadcast[Mapping[str, Any]] = Broadcast(
+            name="Config Channel", resend_latest=True
+        )
+        config_receiver = config_channel.new_receiver()
+
+        # This file does not exist
+        config_file2 = config_file.parent / "config2.toml"
+
+        async with ConfigManagingActor(
+            [config_file, config_file2],
+            config_channel.new_sender(),
+            force_polling=False,
+        ):
+            config = await config_receiver.receive()
+            assert config is not None
+            assert config.get("var2") is None
+
+            number = 5
+            config_file.write_text(create_content(number=number))
+
+            config = await config_receiver.receive()
+            assert config is not None
+            assert config.get("var2") == str(number)
+
+            # Create second config file that overrides the value from the first one
+            number = 42
+            config_file2.write_text(create_content(number=number))
+
+            config = await config_receiver.receive()
+            assert config is not None
+            assert config.get("var2") == str(number)
+
+    async def test_actor_does_not_crash_if_file_is_deleted(
+        self, config_file: pathlib.Path, mocker: MockerFixture
+    ) -> None:
+        """Test ConfigManagingActor does not crash if a file is deleted."""
+        config_channel: Broadcast[Mapping[str, Any]] = Broadcast(
+            name="Config Channel", resend_latest=True
+        )
+        config_receiver = config_channel.new_receiver()
+
+        number = 5
+        config_file2 = config_file.parent / "config2.toml"
+        config_file2.write_text(create_content(number=number))
+
+        # Not config file but existing in the same directory
+        any_file = config_file.parent / "any_file.txt"
+        any_file.write_text("content")
+
+        file_watcher_mock = MagicMock()
+        file_watcher_mock.__anext__.side_effect = [
+            Event(EventType.DELETE, any_file),
+            Event(EventType.MODIFY, config_file2),
+            Event(EventType.MODIFY, config_file),
+        ]
+
+        mocker.patch(
+            "frequenz.channels.file_watcher.FileWatcher", return_value=file_watcher_mock
+        )
+
+        async with ConfigManagingActor(
+            [config_file, config_file2],
+            config_channel.new_sender(),
+            force_polling=False,
+        ) as actor:
+            send_config_spy = mocker.spy(actor, "send_config")
+
+            config = await config_receiver.receive()
+            assert config is not None
+            assert config.get("var2") == str(number)
+            send_config_spy.assert_called_once()
+            send_config_spy.reset_mock()
+
+            # Remove file and send DELETE events
+            any_file.unlink()
+            config_file2.unlink()
+            number = 101
+            config_file.write_text(create_content(number=number))
+
+            config = await config_receiver.receive()
+            assert config is not None
+            assert config.get("var2") == str(number)
+            # Config should be updated only once on MODIFY event
+            # DELETE events are ignored
+            send_config_spy.assert_called_once()
 
 
 @dataclass(frozen=True, kw_only=True)


src/frequenz/sdk/config/_config_managing.py
```diff
@@ -164,10 +164,21 @@ async def _run(self) -> None:
...
+                            "%s: The configuration file %s was deleted, ignoring...",
+                            self,
+                            event.path,
+                        )
+                    continue
+
                 # Since we are watching the whole parent directories, we need to make
                 # sure we only react to events related to the configuration files we
                 # are interested in.
-                if not any(event.path.samefile(p) for p in self._config_paths):
```
llucax, line 170: Can't we check here if `event.path` also exists? I think is harder to reason about the code if we split where the types of events are handled.

llucax (commented):
Oh, good catch!

I think it would be good to add a test case for this, as it is not trivial to see why DELETE is handled separately,, but maybe we need to introduce a integration test, not sure if it would be easy to do with mocking.

src/frequenz/sdk/config/_config_managing.py
ela-kotulska-frequenz: I tried, but it won't cover the case when `event.path` was deleted but is in `self._config_paths` - we should print LOG in that case.

ela-kotulska-frequenz: By the way I think we should print `warning` log in this case, not `info` log. I will change it.

ela-kotulska-frequenz: > I think it would be good to add a test case for this, as it is not trivial to see why DELETE is handled separately,

It should be possible to write unit test. I will do this.

llucax: > I tried, but it won't cover the case when event.path was deleted but is in self._config_paths - we should print LOG in that case.

Yeah, OK, but it is still tricky, the check for `event.path in self._config_paths` could be a bit naive, `event.path` could be textually different from the path ` self._config_paths` if it is a symlink or something, I think. Maybe it would be safer to keep a map between the user-specified file name and the canonical file name it resolves to.

If we always consider the path in `event.path` to match the path in `self._config_paths` with a simple text comparison we don ´t need to use `samefile()` in the if condition either.

> By the way I think we should print warning log in this case, not info log. I will change it.

I think this depends on how config files are updated. In the long run I think they should be updated by creating a new temporary file to create the new file and then use `rename` (as in the system call). In that case I guess the polling will only detect a modification and no removal, but when using inotify, I think we observe a `DELETE` + `CREATE`, so in that case a `DELETE` should be normal. But I agree that `info()` doesn't make a lot of sense, if should be debug (or nothing) if it is normal, or warning if it's not.

I guess a safe option would be to add a timer when it is deleted and if it is not created again after a short while, then print a warning, but maybe that's a bit overkill at this stage.

We could run some tests or make it a warning and if it happens in normal situations we remove it in a future PR/release.

ela-kotulska-frequenz: > I think it would be good to add a test case for this,

Tried and it is not so easy...  I thought about something like this
```
config_file  = create config file with some data
receive config file and check data
            
# Rename config file and change config values
new_file = config_file.rename(config_file.parent / "other_file.toml")
new_file.write_text(new_content)
            
# Rename back
config_file = new_file.rename(new_file.parent / config_file_name)
receive updated config file and check if updated config fields
```
But it does all this faster then inotify detects change and it is not working :D 
I could add timeout in the meantime, but I worry it might  be unstable test and fail sometimes.

ela-kotulska-frequenz: > Yeah, OK, but it is still tricky, the check for event.path in self._config_paths could be a bit naive, event.path could be textually different from the path  self._config_paths if it is a symlink or something, I think

Oww you are right. I didn't thought about it...

> Maybe it would be safer to keep a map between the user-specified file name and the canonical file name it resolves to.

Ok I could do that, however it won't be `one line` solution, because I think it is ok if not all config files exists on startup.

ela-kotulska-frequenz:
Just found few more problems and fixed.

I am still thinking how to fix @llucax  comments and will continue working on it today.

src/frequenz/sdk/config/_config_managing.py
llucax: > But it does all this faster then inotify detects change and it is not working :D
> I could add timeout in the meantime, but I worry it might be unstable test and fail sometimes.

Yeah, we are doing something like that in the channels repo:

* https://github.com/frequenz-floss/frequenz-channels-python/blob/b3c98e594310edde687bde4d9b39f4ac370ed676/tests/test_file_watcher_integration.py#L55

No sure if it would be useful to do something similar here. The thing is this is very filesystem/OS dependant, so it would be probably very hard to do tests that cover all the possibilities, but maybe at least testing for the most common case we'll have (linux, polling) should be better than nothing.

llucax: > Ok I could do that, however it won't be one line solution, because I think it is ok if not all config files exists on startup.

Yeah, I know, and since it changes the whole logic it will required some careful testing. Maybe we should just completely ignore the delete event for now (as in not register to `DELETE` events at all) for now and open a new issue to improve with the learning from this bug/PR in the future. We want to ignore `DELETE` anyways, as it should only be a bug/accident to remove the config file, we will just not be able to print a warning for it.

llucax:
> I am still thinking how to fix @llucax comments and will continue working on it today.

@ela-kotulska-frequenz let me know what do you think about completely ignoring `DELETE` for now. I think that should be super simple, it won't complicate the code logic, should fix the issue with minimal changes and only with a small downside (not logging deletes). If we go for it I would open a new issue about this and  add a comment in the code explaining why we are ignoring `DELETE` and a link to the issue.

ela-kotulska-frequenz:
> let me know what do you think about completely ignoring DELETE for now.

I agree! :) Will do proper changes.

ela-kotulska-frequenz:
* updated all commits and changed order to make PR more readable.
* wrote unit tests.  FileWatcher is well tested (in FileWatcher's repo), so I mocked it to tests scenario when files are deleted.
* @llucax  in new unit tests I found another potential bug: Deletion of config file is ignored and config are not updated. It worked like this before and I left it like this.  Should it work like this?

llucax:
> @llucax in new unit tests I found another potential bug: Deletion of config file is ignored and config are not updated. It worked like this before and I left it like this. Should it work like this?

I think so, I think a complete and permanent deletion of the config file should be an error, and when facing that error I think the best option is to continue with the old config, like we do with other errors. I don't see any valid reason to remove a config file completely and permanently. If the file is created again we should be notified and pick up the changes. If we are not notified about the re-creation, then we have a problem, but maybe we can test that (if it is not tested already).

src/frequenz/sdk/config/_config_managing.py
```diff
@@ -106,9 +106,42 @@ def __init__(
...
         self._output: Sender[abc.Mapping[str, Any]] = output
-        self._event_types: abc.Set[EventType] = event_types
-        self._force_polling: bool = force_polling
-        self._polling_interval: timedelta = polling_interval
+
+        # FileWatcher instance should be create here instead of in _run().
+        # Otherwise, if the actor restarts, we will lost all the events that happened
+        # while the actor was down.
+        self._file_watcher = self._create_file_watcher(
+            force_polling, polling_interval, event_types
```
llucax: What I was thinking when I said ignoring the `DELETE` event completely was to actually change the `event_types` default from `frozenset(EventType)` to `EventType.CREATE | EventType.MODIFY`, but actually when thinking about it, if the logic is so tied to the events we handle, we should probably just **remove** the `event_types` argument from the constructor completely and just hardcode it to `CREATE|MODIFY`, users shouldn't really decide about this for the config manager.

src/frequenz/sdk/config/_config_managing.py
```diff
@@ -106,9 +106,42 @@ def __init__(
             for config_path in config_paths
         ]
         self._output: Sender[abc.Mapping[str, Any]] = output
-        self._event_types: abc.Set[EventType] = event_types
-        self._force_polling: bool = force_polling
-        self._polling_interval: timedelta = polling_interval
+
+        # FileWatcher instance should be create here instead of in _run().
+        # Otherwise, if the actor restarts, we will lost all the events that happened
+        # while the actor was down.
```
llucax: Really? Why would we want to pick up on old events? If there is an error isn't it better to just completely start over? Create the file watcher from scratch, read all config files and continue monitoring them as usual?

What if there is a bug in the file watcher itself and re-creating it fixes it?

For me it feels we should still be creating the file watcher in the run.

src/frequenz/sdk/config/_config_managing.py
```diff
@@ -130,6 +163,12 @@ def _read_config(self) -> abc.Mapping[str, Any]:
             except ValueError as err:
                 _logger.error("%s: Can't read config file, err: %s", self, err)
                 error_count += 1
+            except FileNotFoundError:
+                # It is ok for config file to don't exist.
+                _logger.info(
+                    "%s: Config file %s not found. Ignoring it.", self, config_path
+                )
+                error_count += 1
```
llucax: Mmm, interesting. Why not an error? I also wonder if we shouldn't do this actually for any exception raised while reading, parsing and validating the config files. The goal of this is to be as resilient as possible in case of a partial failure (we can only read part of the config. And it is really tricky, sometimes I think if one config file read failed, we should probably skip the update completely for all files, as we might get into an inconsistent config.

I'm starting to feel maybe it would be better to restart the actor if there is an issue reading **any** of the requested files (i.e. requiring users to create at least and empty config for the files that are requested in the constructor).

But maybe this is a topic for a different PR. But I would make this a `_logger.error()` in this PR. And maybe catch all exceptions, not just `ValueError` and `FileNotFound` (at least all `OSError`, not just `FileNotFound` as well as any errors that could be railed by `tomlib.load()`, maybe it only raises `ValueError`).

src/frequenz/sdk/config/_config_managing.py
```diff
@@ -150,51 +189,45 @@ async def _run(self) -> None:
...
+                        self,
+                        event.path,
+                    )
+                    await self.send_config()
+                case EventType.DELETE:
+                    # Deletion should be handled in the if statement above by checking
+                    # if event.path exists.
+                    raise RuntimeError(
+                        "DELETE event should be handled in the if statement above"
+                    )
```
llucax: I would just log an error here. I guess if we can have a race if a file is created and deleted very quickly, that the if doesn't catch it. But if we don't listed to `DELETE` at all, then we don't have this problem.

src/frequenz/sdk/config/_config_managing.py
```diff
@@ -150,51 +189,45 @@ async def _run(self) -> None:
...
+            # pathlib.Path.samefile raises error if any path doesn't exist so we need to
+            # make sure `event.path` and `p` exists before calling it.
+            #
+            # On EventType.DELETE event the event.path won't exists and we have no way
+            # to check if this path is in self._config_paths so we need to skip this
+            # event.
+            if not event.path.exists() or not any(
+                event.path.samefile(p) for p in self._config_paths if p.exists()
+            ):
+                continue
```
llucax: This is also much clearer if we don't subscribe to `DELETE`, then this can only happen on `CREATE` or `MODIFY`. If the file not there when the event was triggered, it means there was a race, and the file was created or changed and right after that it was deleted, so we can more safely (or more clearly) ignore the event, and hope a subsequent `CREATE` will be triggered again when the file is re-created.

If any of `event.path` or `self._config_paths` do not exist, I would log an error though. It might be a race, but I think we should be notified a config file was removed. We can then see if we get a lot of spurious errors if we want to remove it, o maybe set a timer and if the file doesn't come back after some time to then log an error.

llucax (commented):
Damn, this got really tricky. Sorry for not being more clear about removing `DELETE` from the events we watch. I think doing that could simplify the code even further.

src/frequenz/sdk/config/_config_managing.py
ela-kotulska-frequenz: Case before this update:
file was deleted -> signal DELETE -> ConfigManager crashed -> it took 2 sec to restart -AND new file was created but we didn't detect it because now FIleWatcher exist -> new File Watcher was created

ela-kotulska-frequenz: Correct scenario after this commit :
file was deleted -> signal DELETE ->  ConfigManager crashed BUT we don't remove FileWatcher and it still watches the files -> it took 2 sec to restart -> We continue with previous FileWatcher and handle CREATE event.

src/frequenz/sdk/config/_config_managing.py
ela-kotulska-frequenz: Currently it works like this:
Every time any config file change - we read all config files. But not all config files might exists and this cover this case.

Luca, this PR was to fix existing bugs, not redesign ConfigManger. I guess it needs redesign and rethinking but maybe in separate PR and after we agree how it should work.

src/frequenz/sdk/config/_config_managing.py
ela-kotulska-frequenz: I thought it is ok if not all files in `self._config_paths` exist. 

Regarding `event.path` ok I can do it

ela-kotulska-frequenz: My current understanding is : at least one file in `self._config_paths` must exists.  If it should be differently then we should change more things in this class.

src/frequenz/sdk/config/_config_managing.py
ela-kotulska-frequenz: > we should probably just remove the event_types argument from the constructor completely and just hardcode it to CREATE|MODIFY,

Done

src/frequenz/sdk/config/_config_managing.py
ela-kotulska-frequenz: So here I log info, if any config file is doesn't exist. 
And (code 3 lines below) raises ValueError if no config file exist. (at least one must exist)

src/frequenz/sdk/config/_config_managing.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/config/_config_managing.py
llucax: > AND new file was created but we didn't detect it because now FIleWatcher exist -> new File Watcher was created

I don't get this. You mean the file that was just removed got created again? Why do you mean by "now FileWatcher exist"? In the state previous to this PR the file watcher doesn't exist any more after the actor crashes, right?

src/frequenz/sdk/config/_config_managing.py
llucax: > Luca, this PR was to fix existing bugs, not redesign ConfigManger. I guess it needs redesign and rethinking but maybe in separate PR and after we agree how it should work.

Yes, 100% agreed, this is also why I said re-evaluating what to do on a failure to read a config file is for another PR.

But catching other types of errors is still a bug that IMHO is in scope for this PR. Since you are catching a new `FileNotFound` why not just catching all `OSError`s?

> So here I log info, if any config file is doesn't exist.
> And (code 3 lines below) raises ValueError if no config file exist. (at least one must exist)

Yeah, but that's only if we can't read **any** config file, if only one failed, no `ValueError` will be raised, and the missing file won't be noticed if we are only monitoring WARNING and up. Also it is more consistent to how we handle the TOML parsing `ValueError`. We log an error if TOML parsing fails. I don't understand why the inconsistency of making one issue emit an error log and the new one just a info.

src/frequenz/sdk/config/_config_managing.py
llucax: > I thought it is ok if not all files in self._config_paths exist.

Yeah, true. As things are currently working it would be valid if only one file is missing. So if at some point we require all, we can update this too.

llucax: * https://github.com/frequenz-floss/frequenz-sdk-python/issues/1120

src/frequenz/sdk/config/_config_managing.py
llucax: * https://github.com/frequenz-floss/frequenz-sdk-python/issues/1120

src/frequenz/sdk/config/_config_managing.py
```diff
@@ -130,6 +157,12 @@ def _read_config(self) -> abc.Mapping[str, Any]:
             except ValueError as err:
                 _logger.error("%s: Can't read config file, err: %s", self, err)
                 error_count += 1
+            except FileNotFoundError:
+                # It is ok for config file to don't exist.
+                _logger.info(
+                    "%s: Config file %s not found. Ignoring it.", self, config_path
```
llucax: Just to make it more explicit what I'm suggesting to change in this PR:
```suggestion
            except OSError as err:
                # It is ok for config file to don't exist.
                _logger.error(
                    "%s: Error reading config file %s (%s). Ignoring it.", self, config_path, err,
```

src/frequenz/sdk/config/_config_managing.py
```diff
@@ -150,51 +183,56 @@ async def _run(self) -> None:
...
+            # Since we are watching the whole parent directories, we need to make
+            # sure we only react to events related to the configuration files we
+            # are interested in.
+            #
+            # pathlib.Path.samefile raises error if any path doesn't exist so we need to
+            # make sure `event.path` and `p` exists before calling it.
+            #
+            # On EventType.DELETE event the event.path won't exists and we have no way
+            # to check if this path is in self._config_paths so we need to skip this
+            # event.
```
llucax: This now looks outdated, maybe something like:
```suggestion
            # pathlib.Path.samefile raises error if any path doesn't exist so we need to
            # make sure the paths exists before calling it. This could happen as it is not
            # required that all config files exist, only one is required but we don't know which.
```

src/frequenz/sdk/config/_config_managing.py
```diff
@@ -150,51 +183,56 @@ async def _run(self) -> None:
...
+                case EventType.MODIFY:
+                    _logger.info(
+                        "%s: The configuration file %s was modified, sending update...",
+                        self,
+                        event.path,
+                    )
+                    await self.send_config()
+                case EventType.DELETE:
+                    # Deletion should be handled in the if statement above by checking
+                    # if event.path exists.
```
llucax: Also outdated, I would just remove as the log message makes this clear enough IMHO.
```suggestion
```

llucax (commented):
I made some completely new comments to make it more clear what should be addressed in this PR, the rest if for a future PR, I already created an issue for that.

Besides the new comment, for me it is still not clear why it is better to create the file watcher in the constructor and not the `_run()`.

src/frequenz/sdk/config/_config_managing.py
ela-kotulska-frequenz: fixed

src/frequenz/sdk/config/_config_managing.py
ela-kotulska-frequenz: fixed

src/frequenz/sdk/config/_config_managing.py
ela-kotulska-frequenz: Ok the whole discussion was because I had impression that having not all config files is not an error.
But it looks like it is unexpected so as you suggested I changed severity of this logging to  error. :) 
Sorry for my misunderstanding!

src/frequenz/sdk/config/_config_managing.py
ela-kotulska-frequenz: Ok I gave it a second thought and this was over complicated and also not working solution.
What I wanted was to watch for files all the time - even if actor crashes. Because during actor crash file can be replaced and we will skip this change. But I just realized we send new config at the beginning of `_run` method. So new config will be send on every actor re-run. So this change is not needed.

I removed this commit. But found new problem related with above and fixed it in last(new) commit.

ela-kotulska-frequenz:
Just changed commit message in last commit :)

src/frequenz/sdk/config/_config_managing.py
ela-kotulska-frequenz: fixed

src/frequenz/sdk/config/_config_managing.py
llucax: No it is OK, originally it was thought like that, only one config file should be enough, but I think that's assuming too much, and is hard to know if missing files are errors or expected, so I opened the new issue to change that in the future.

llucax:
Adding label for breaking changes and assigning to rc1400 as this now breaks the API, but we can do a quick release after it is merged anyway.

src/frequenz/sdk/config/_config_managing.py
llucax: Oh, I see in the last commit you are changing this. I would not address the new issue (#1120) in this PR, I think it needs a bit more thought and it is not strictly related to the bug fixes in this PR.

llucax (approved):
LGTM. Approving.

I'm still undecided if it wouldn't be better to remove the last commit and leave that to be fixed separate after some more thought. But I think it is OK to move forward as it is too and reconsider in the future when #1120 is addressed.

PR #1117, opened 2024-11-25
Fix IndexError: list index out of range in resampler

llucax:
When awaiting the resamplers to finish resampling, we need to make a copy of the resamplers dict, because new resamplers could be added or removed while we are awaiting, which would cause the results to be out of sync.

Here is a traceback of an instance of this bug:

```
2024-11-25 12:45:26,052,52 WARNING  [_broadcast.py:416] Broadcast receiver [Broadcast:GrpcStreamBroadcaster-raw-component-data-2006:_Receiver] is full. Oldest message was dropped.
2024-11-25 12:45:26,052,52 ERROR    [_resampling.py:195] The resample() function got an unexpected error, restarting...
  File "/home/marenz/frequenz/sdk/src/frequenz/sdk/microgrid/_resampling.py", line 179, in _log_resampling_task_error
    resampling_task.result()
  File "/home/marenz/frequenz/sdk/src/frequenz/sdk/timeseries/_resampling.py", line 509, in resample
    {
  File "/home/marenz/frequenz/sdk/src/frequenz/sdk/timeseries/_resampling.py", line 514, in <dictcomp>
    if isinstance(results[i], (Exception, asyncio.CancelledError))
                  ~~~~~~~^^^
IndexError: list index out of range
```

### DIFF
--- a/src/frequenz/sdk/timeseries/_resampling.py
+++ b/src/frequenz/sdk/timeseries/_resampling.py
@@ -512,11 +512,11 @@ class Resampler:
             exceptions = cast(
                 dict[Source, Exception | asyncio.CancelledError],
                 {
-                    source: results[i]
-                    for i, source in enumerate(resampler_sources)
+                    source: result
+                    for source, result in zip(resampler_sources, results)
                     # CancelledError inherits from BaseException, but we don't want
                     # to catch *all* BaseExceptions here.
-                    if isinstance(results[i], (Exception, asyncio.CancelledError))
+                    if isinstance(result, (Exception, asyncio.CancelledError))
                 },
             )
             if exceptions:
--- a/src/frequenz/sdk/timeseries/_resampling.py
+++ b/src/frequenz/sdk/timeseries/_resampling.py
@@ -495,6 +495,11 @@ class Resampler:
                     self._config.resampling_period,
                 )
 
+            # We need to make a copy here because we need to match the results to the
+            # current resamplers, and since we await here, new resamplers could be added
+            # or removed from the dict while we awaiting the resampling, which would
+            # cause the results to be out of sync.
+            resampler_sources = list(self._resamplers)
             results = await asyncio.gather(
                 *[r.resample(self._window_end) for r in self._resamplers.values()],
                 return_exceptions=True,
@@ -508,7 +513,7 @@ class Resampler:
                 dict[Source, Exception | asyncio.CancelledError],
                 {
                     source: results[i]
-                    for i, source in enumerate(self._resamplers)
+                    for i, source in enumerate(resampler_sources)
                     # CancelledError inherits from BaseException, but we don't want
                     # to catch *all* BaseExceptions here.
                     if isinstance(results[i], (Exception, asyncio.CancelledError))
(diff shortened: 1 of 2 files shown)


src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -495,8 +495,13 @@ async def resample(self, *, one_shot: bool = False) -> None:
                     self._config.resampling_period,
                 )
 
+            # We need to make a copy here because we need to match the results to the
+            # current resamplers, and since we await here, new resamplers could be added
+            # or removed from the dict while we awaiting the resampling, which would
+            # cause the results to be out of sync.
+            resamplers = self._resamplers.copy()
```
ela-kotulska-frequenz: Copying keys should be enough:
`resampler_keys = [self_.resamplers.keys()]`

src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -508,7 +513,7 @@ async def resample(self, *, one_shot: bool = False) -> None:
                 dict[Source, Exception | asyncio.CancelledError],
                 {
                     source: results[i]
-                    for i, source in enumerate(self._resamplers)
+                    for i, source in enumerate(resamplers)
                     # CancelledError inherits from BaseException, but we don't want
                     # to catch *all* BaseExceptions here.
                     if isinstance(results[i], (Exception, asyncio.CancelledError))
```
ela-kotulska-frequenz: Not related, but it looks like perfect case for using zip :) 
`{source: result for source, result in zip(resamplers_keys, results) if isinstance(result, (Exception, asyncio.CancelledError))`

src/frequenz/sdk/timeseries/_resampling.py
llucax: No, we actually use the value (`source`) too, to set in which resampler the error was found.

src/frequenz/sdk/timeseries/_resampling.py
llucax: Yeah, I could try if it helps mypy do the type-narrowing, but I doubt it.

src/frequenz/sdk/timeseries/_resampling.py
ela-kotulska-frequenz: But : 
* `self._resamplers: dict[Source, _StreamingHelper] = {}` - so you will copy source.
* and `enumerate(resamplers)` - enumerates over keys.

llucax: Right, right, right. I got confused because we are using the `values()` right after the copy, but when we access those `values()` there was no `await` yet, so we can use the original `dict` there. :+1:

src/frequenz/sdk/timeseries/_resampling.py
llucax: It worked, not sure if maybe it also worked now without the `zip()` too with newer `mypy` versions or not, but nice that we don't need the cast anymore!

llucax:
Updated to:
* Copy only the keys
* Use `zip()`
* Remove the unnecessary cast

ela-kotulska-frequenz (approved):
Thanks!

PR #1119, opened 2024-11-27
Small debugging enhancements

Marenz:
- **Fix wrong term in battery formula documentation**
- **Give more receivers/tasks a name for easier debugging**
- **Add a nice str representation for `Sample`**
- **Actually log the error that caused a source to be removed**
- **Add more details to "no relevant sample found" warning log**
- **Add change log**

### DIFF
--- a/src/frequenz/sdk/microgrid/_data_pipeline.py
+++ b/src/frequenz/sdk/microgrid/_data_pipeline.py
@@ -479,7 +479,8 @@ class _DataPipeline:  # pylint: disable=too-many-instance-attributes
                 channel_registry=self._channel_registry,
                 data_sourcing_request_sender=self._data_sourcing_request_sender(),
                 resampling_request_receiver=channel.new_receiver(
-                    limit=_REQUEST_RECV_BUFFER_SIZE
+                    limit=_REQUEST_RECV_BUFFER_SIZE,
+                    name=channel.name + " Receiver",
                 ),
                 config=self._resampler_config,
             )
--- a/src/frequenz/sdk/microgrid/_data_sourcing/microgrid_api_source.py
+++ b/src/frequenz/sdk/microgrid/_data_sourcing/microgrid_api_source.py
@@ -461,7 +461,8 @@ class MicrogridApiSource:
             self.comp_data_tasks[comp_id].cancel()
 
         self.comp_data_tasks[comp_id] = asyncio.create_task(
-            run_forever(lambda: self._handle_data_stream(comp_id, category))
+            run_forever(lambda: self._handle_data_stream(comp_id, category)),
+            name=f"{type(self).__name__}._update_stream({comp_id=}, {category.name})",
         )
 
     async def add_metric(self, request: ComponentMetricRequest) -> None:
--- a/src/frequenz/sdk/microgrid/_resampling.py
+++ b/src/frequenz/sdk/microgrid/_resampling.py
@@ -179,7 +179,11 @@ class ComponentMetricsResamplingActor(Actor):
             resampling_task.result()
         except ResamplingError as error:
             for source, source_error in error.exceptions.items():
-                _logger.error("Error resampling source %s, removing source...", source)
+                _logger.error(
+                    "Error resampling source %s, removing source",
+                    source,
+                    exc_info=source_error,
+                )
                 removed = self._resampler.remove_timeseries(source)
                 if not removed:
                     _logger.error(
--- a/src/frequenz/sdk/timeseries/_base_types.py
+++ b/src/frequenz/sdk/timeseries/_base_types.py
@@ -34,6 +34,14 @@ class Sample(Generic[QuantityT]):
     value: QuantityT | None = None
     """The value of this sample."""
 
+    def __str__(self) -> str:
+        """Return a string representation of the sample."""
+        return f"{type(self).__name__}({self.timestamp}, {self.value})"
+
+    def __repr__(self) -> str:
+        """Return a string representation of the sample."""
+        return f"{type(self).__name__}({self.timestamp=}, {self.value=})"
+
 
 @dataclass(frozen=True)
 class Sample3Phase(Generic[QuantityT]):
--- a/src/frequenz/sdk/timeseries/_resampling.py
+++ b/src/frequenz/sdk/timeseries/_resampling.py
@@ -744,13 +744,42 @@ class _ResamplingHelper:
         # resort to some C (or similar) implementation.
         relevant_samples = list(itertools.islice(self._buffer, min_index, max_index))
         if not relevant_samples:
-            _logger.warning("No relevant samples found for: %s", self._name)
+            self._log_no_relevant_samples(minimum_relevant_timestamp, timestamp)
+
         value = (
             conf.resampling_function(relevant_samples, conf, props)
             if relevant_samples
             else None
         )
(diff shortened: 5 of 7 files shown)


src/frequenz/sdk/microgrid/_data_pipeline.py
```diff
@@ -454,7 +454,9 @@ def _data_sourcing_request_sender(self) -> Sender[ComponentMetricRequest]:
                 name="Data Pipeline: Data Sourcing Actor Request Channel"
             )
             actor = DataSourcingActor(
-                request_receiver=channel.new_receiver(limit=_REQUEST_RECV_BUFFER_SIZE),
+                request_receiver=channel.new_receiver(
+                    limit=_REQUEST_RECV_BUFFER_SIZE, name=channel.name + " Receiver"
+                ),
```
llucax: Wait! This shouldn't be necessary. Maybe you are using `name` directly? If you use `str(receiver)` or `repr(receiver)` the channel name (and the word "receiver") will be already used:

* https://github.com/frequenz-floss/frequenz-channels-python/blob/b3c98e594310edde687bde4d9b39f4ac370ed676/src/frequenz/channels/_broadcast.py#L472-L483
* https://github.com/frequenz-floss/frequenz-channels-python/blob/b3c98e594310edde687bde4d9b39f4ac370ed676/src/frequenz/channels/_broadcast.py#L291-L303

src/frequenz/sdk/microgrid/_data_sourcing/microgrid_api_source.py
```diff
@@ -461,7 +461,8 @@ async def _update_streams(
             self.comp_data_tasks[comp_id].cancel()
 
         self.comp_data_tasks[comp_id] = asyncio.create_task(
-            run_forever(lambda: self._handle_data_stream(comp_id, category))
+            run_forever(lambda: self._handle_data_stream(comp_id, category)),
+            name=f"run_forever:cid={comp_id}",
```
llucax: This looks confusing. We have a `run_forever` function. Why not something like:
```suggestion
            name=f"{type(self).__name__}._update_stream({comp_id=}, {category.name})",
```

src/frequenz/sdk/timeseries/_base_types.py
```diff
@@ -34,6 +34,10 @@ class Sample(Generic[QuantityT]):
     value: QuantityT | None = None
     """The value of this sample."""
 
+    def __str__(self) -> str:
+        """Return a string representation of the sample."""
+        return f"Sample({str(self.timestamp)}, {str(self.value)})"
```
llucax: Why calling `str` explicitly? This is what should be done when you interpolate anyways. Also it's better to use `type(self).__name__` in case someone subclasses it and doesn't override the `str`. Finally, it would be nice to have a `repr` too:

```suggestion
    def __str__(self) -> str:
        """Return a string representation of the sample."""
        return f"{type(self).__name__}({self.timestamp}, {self.value})"

    def __repr__(self) -> str:
        """Return a string representation of the sample."""
        return f"{type(self).__name__}({self.timestamp=}, {self.value=})"
```

src/frequenz/sdk/microgrid/_resampling.py
```diff
@@ -179,7 +179,11 @@ def _log_resampling_task_error(self, resampling_task: asyncio.Task[None]) -> Non
             resampling_task.result()
         except ResamplingError as error:
             for source, source_error in error.exceptions.items():
-                _logger.error("Error resampling source %s, removing source...", source)
+                _logger.error(
+                    "Error resampling source %s, removing source: %s",
+                    source,
+                    source_error,
+                )
```
llucax: Do you think this is enough? Maybe we can show the full backtrace with:

```suggestion
                _logger.error(
                    "Error resampling source %s, removing source",
                    source,
                    exc_info=source_error,
                )
```

src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -746,14 +746,46 @@ def resample(self, timestamp: datetime) -> Sample[Quantity]:
         # resort to some C (or similar) implementation.
         relevant_samples = list(itertools.islice(self._buffer, min_index, max_index))
         if not relevant_samples:
-            _logger.warning("No relevant samples found for: %s", self._name)
+            self._log_no_relevant_samples(minimum_relevant_timestamp, timestamp)
```
llucax, line 747: We are probably always going to enable warnings, but just in case and since that function seems to be non-trivial, maybe you want to [check if the log level is enabled](https://docs.python.org/3/library/logging.html#logging.Logger.isEnabledFor) so you don't call it at all if it's not:

```py
        if not relevant_samples and _logger.isEnabledFor(logging.WARNING):
```

Or maybe it makes more sense to add the check inside the function, to have the levels in the same place (at the expense of a function call that will just return).

src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -746,14 +746,46 @@ def resample(self, timestamp: datetime) -> Sample[Quantity]:
...
+            ("No relevant samples found for: %s\n%-*s %s - %s\n%-*s %s"),
+            self._name,
+            label_width,
+            "Requested:",
+            minimum_relevant_timestamp,
+            timestamp,
+            label_width,
+            buffer_label,
+            buffer_info,
+        )
```
llucax: Maybe to avoid misalignment if the len varies in number of digits and to make the empty buffer message less verbose?

```suggestion
        if self._buffer:
            buffer_info = f"{self._buffer[0].timestamp} - {self._buffer[-1].timestamp} ({len(self._buffer)} samples)"
        else:
            buffer_info = "Empty"

        _logger.warning(
            "No relevant samples found for: %s\n  Requested: %s - %s\n     Buffer: %s",
            self._name,
            minimum_relevant_timestamp,
            timestamp,
            buffer_info,
        )
```

src/frequenz/sdk/microgrid/_data_pipeline.py
Marenz: I was using .name directly, yes

Marenz: Removed

src/frequenz/sdk/timeseries/_resampling.py
```diff
@@ -744,14 +744,43 @@ def resample(self, timestamp: datetime) -> Sample[Quantity]:
...
+    def _log_no_relevant_samples(
+        self, minimum_relevant_timestamp: datetime, timestamp: datetime
+    ) -> None:
+        """Log that no relevant samples were found.
+
+        Args:
+            minimum_relevant_timestamp: Minimum timestamp that was requested
+            timestamp: Timestamp that was requested
+        """
+        if not _logger.isEnabledFor(logging.DEBUG):
```
llucax: This should be `WARNING`!
```suggestion
        if not _logger.isEnabledFor(logging.WARNING):
```

llucax (dismissed):
LGTM except for the `DEBUG` -> `WARNING`. Feel free to force-merge after it is fixed.

src/frequenz/sdk/timeseries/_resampling.py
Marenz: Shit. So much for my FCR deployment.

PR #1129, opened 2024-12-10
Change name of the BackgroundService to include name of the class

ela-kotulska-frequenz:
This makes it more readable.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -26,6 +26,8 @@
         + The class is now immutable.
         + The constructor now accepts only keyword arguments.
 
+- The Actor class now generates a name with the class name and the instance id.
+
 ## New Features
 
 - `LoggingConfigUpdatingActor`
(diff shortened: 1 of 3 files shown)


llucax (commented):
I wouldn't go for this, this will be renamed to `unique_id` and then it doesn't make sense to have the class name in the unique ID. If you want to get a representation of the instance, just use `str(instance)` (or `f"{instance}"` or `logging.debug("%s", instance)`) instead of using the name explicitly (`instance.name`, `f"{instance.name}"`, etc.), that will give you the full class name + unique ID.

This change is [already done in `core`](https://github.com/frequenz-floss/frequenz-core-python/blob/v1.x.x/src/frequenz/core/asyncio/_service.py#L222), I just didn't have time to finish the core library and do the migration. We could also update `BackgroudService` for now but I don't think it is worth it if we'll need to change this once we move to using `core`.

PR #1132, opened 2024-12-12
Logging improvements

shsms:


shsms:
I will add more commits here

shsms:
I hope no one thinks this is overkill.

src/frequenz/sdk/_internal/_logging.py
```diff
@@ -0,0 +1,300 @@
...
+DEFAULT_RATE_LIMIT = timedelta(minutes=15)
+
+# The standard logging.py file uses variadic arguments in the logging methods, but the
+# type hints file has more specific parameters.  pylint is not able to handle this, so
+# we need to suppress the warning.
+#
+# pylint: disable=arguments-differ
+
+
+class RateLimitedLogger:
```
llucax, line 28: Any reason not to inherit from `Logger` so this can be used in places where a plain `Logger` is expected? If you do so you probably only need to implement `log()`.

And maybe this could be implemented as a [`Filter`](https://docs.python.org/3/library/logging.html#filter-objects) instead ([here](https://stackoverflow.com/a/33701485) is an example filter to de-duplicate messages), but not sure, because filters are applied at the [`Handler`](https://docs.python.org/3/library/logging.html#handler-objects) level, so it would apply to everything that's logged or we need to customize the config in such a way that messages that we want to rate-limit must be handled by a different handler.

This looks quite interesting: https://github.com/samuller/log-rate-limit, it is a filter but allows overriding on every log call too, and allows grouping messages in streams, and individual streams can be rate-limited individually.

src/frequenz/sdk/_internal/_logging.py
```diff
@@ -0,0 +1,300 @@
...
+
+_ExcInfoType = (
+    bool
+    | BaseException
+    | tuple[None, None, None]
+    | tuple[type[BaseException], BaseException, TracebackType | None]
+    | None
+)
+
+DEFAULT_RATE_LIMIT = timedelta(minutes=15)
```
llucax, line 19: I would probably just put the literal `timedelta()` in the constructor to avoid the indirection in the docs. If you keep it this way, you should documented via a docstring so it appears in the docs and users can know what's the default.

src/frequenz/sdk/_internal/_logging.py
```diff
@@ -0,0 +1,300 @@
...
+        self._last_log_time: datetime | None = None
+        self._rate_limit: timedelta = rate_limit
+
+    def set_rate_limit(self, rate_limit: timedelta) -> None:
+        """Set the rate limit for the logger.
+
+        Args:
+            rate_limit: Time interval between two log messages.
+        """
+        self._rate_limit = rate_limit
```
llucax, line 63: Why not just making it public?

```suggestion
        self.rate_limit: timedelta = rate_limit
        """The rate limit for the logger."""
```

src/frequenz/sdk/_internal/_logging.py
```diff
@@ -0,0 +1,300 @@
...
+
+    def is_limiting(self) -> bool:
```
llucax, line 65:
```suggestion
    @property
    def is_limiting(self) -> bool:
```

src/frequenz/sdk/_internal/_logging.py
```diff
@@ -0,0 +1,300 @@
...
+        Args:
+            level: Log level.
+            msg: Log message.
+            *args: Arguments for the log message.
+            exc_info: Exception information.
+            stack_info: Stack information.
+            stacklevel: Stack level.
+            extra: Extra information.
+        """
+        if self._rate_limit is None:
```
llucax, line 99: Why the check? It can't be `None`, right?

src/frequenz/sdk/timeseries/battery_pool/_component_metric_fetcher.py
```diff
@@ -120,6 +126,12 @@ async def fetch_next(self) -> ComponentMetricsData | None:
             data = await asyncio.wait_for(
                 self._receiver.receive(), self._max_waiting_time
             )
+            if _missing_data_logger.is_limiting():
+                _missing_data_logger.reset()
+                _missing_data_logger.debug(
+                    "Component %d has started sending data.", self._component_id
+                )
+                _missing_data_logger.reset()
```
llucax, line 134: Couldn't you just log using the regular `_logger` only `reset()` here?
```suggestion
                __logger.debug(
                    "Component %d has started sending data.", self._component_id
                )
                _missing_data_logger.reset()
```

src/frequenz/sdk/timeseries/battery_pool/_component_metric_fetcher.py
```diff
@@ -128,7 +140,9 @@ async def fetch_next(self) -> ComponentMetricsData | None:
             return None
         except asyncio.TimeoutError:
             # Next time wait infinitely until we receive any message.
-            _logger.debug("Component %d stopped sending data.", self._component_id)
+            _missing_data_logger.debug(
+                "Component %d stopped sending data.", self._component_id
```
llucax, line 144: I would rephrase because when it is repeated, one might think that it **just** stopped sending data now. Also since we have instances where the problem was in the data pipeline and not the component itself, maybe is more accurate to say we are not receiving data.

```suggestion
                "No data received for component %d.", self._component_id
```

llucax (commented):
I don't think it is overkill but I'm not convinced about the balance. It seems to me to be too generic for being specific and too specific for being generic.

As a generic solution, I like more the approach from https://github.com/samuller/log-rate-limit, it looks super flexible, allowing even to solve the `reset()` issue by [allowing the next N logs](https://github.com/samuller/log-rate-limit?tab=readme-ov-file#dynamically-override-configuration-options):

```py
# Override the allow_next_n value for a set of logs in the same stream so that this group of logs don't restrict one
# another from occuring consecutively
logger.warning("Test", extra=RateLimit(stream_id="stream2", allow_next_n=2))
logger.info("Extra", extra=RateLimit(stream_id="stream2"))
logger.debug("Info", extra=RateLimit(stream_id="stream2"))
```

And if we want to go more specific, I think it would just keep track of when logs are emitted in the metric fetcher itself, so we can even print more meaningful info, like including when was the last data received (`"No data received for component %d since %s."`).

mkdocs.yml
```diff
@@ -9,7 +9,7 @@ copyright: "Copyright © 2022 Frequenz Energy-as-a-Service GmbH"
 repo_name: "frequenz-sdk-python"
 repo_url: "https://github.com/frequenz-floss/frequenz-sdk-python"
 edit_uri: "edit/v1.x.x/docs/"
-strict: true  # Treat warnings as errors
+# strict: true  # Treat warnings as errors
```
llucax, line 12: This can be removed, https://github.com/marshmallow-code/marshmallow/issues/2739 is fixed.

cwasicki:
This will be very useful for getting to a reasonable monitoring @ela-kotulska-frequenz

PR #1134, opened 2024-12-17
Revamp config management

llucax:
This pull request adds a new `ConfigManager` class that takes care of the *client-side* of the `ConfigManagingActor`, providing the common tasks users of the config system need to do.

The class owns the config channel and provides a convenient method to get new receivers that validate the configuration using a `dataclass` as a `marshmallow` schema, and allow filtering to receive only configuration for a specific key. The receiver also reports errors when the configuration can't be properly loaded, and avoid spurious updates when the configuration didn't change compared to the previous one.

The manager also takes care of instantiating both the `ConfigManagingActor` and the `LoggingConfigUpdatingActor` and provides a utility function to wait for the first configuration to be received.

### DIFF
--- a/src/frequenz/sdk/config/_manager.py
+++ b/src/frequenz/sdk/config/_manager.py
@@ -22,7 +22,7 @@ _logger = logging.getLogger(__name__)
 class InvalidValueForKeyError(ValueError):
     """An error indicating that the value under the specified key is invalid."""
 
-    def __init__(self, msg: str, *, key: str, value: Any) -> None:
+    def __init__(self, msg: str, *, key: Sequence[str], value: Any) -> None:
         """Initialize this error.
 
         Args:
@@ -126,7 +126,12 @@ class ConfigManager(BackgroundService):
 
     def new_receiver(
         self,
-        key: str,
+        # This is tricky, because a str is also a Sequence[str], if we would use only
+        # Sequence[str], then a regular string would also be accepted and taken as
+        # a sequence, like "key" -> ["k", "e", "y"]. We should never remove the str from
+        # the allowed types without changing Sequence[str] to something more specific,
+        # like list[str] or tuple[str] (but both have their own problems).
+        key: str | Sequence[str],
         /,
         *,
         skip_unchanged: bool = True,
@@ -143,6 +148,10 @@ class ConfigManager(BackgroundService):
 
         Only the configuration under the `key` will be received by the receiver. If the
         `key` is not found in the configuration, the receiver will receive `None`.
+
+        If the key is a sequence of strings, it will be treated as a nested key and the
+        receiver will receive the configuration under the nested key. For example
+        `["key", "subkey"]` will get only `config["key"]["subkey"]`.
 
         The value under `key` must be another mapping, otherwise an error
         will be logged and a [`frequenz.sdk.config.InvalidValueForKeyError`][] instance
@@ -164,14 +173,16 @@ class ConfigManager(BackgroundService):
             ```
 
         Args:
-            key: The configuration key to be read by the receiver.
+            key: The configuration key to be read by the receiver. If a sequence of
+                strings is used, it is used as a sub-key.
             skip_unchanged: Whether to skip sending the configuration if it hasn't
                 changed compared to the last one received.
 
         Returns:
             The receiver for the configuration.
         """
-        receiver = self.config_channel.new_receiver(name=f"{self}:{key}", limit=1)
+        recv_name = key if isinstance(key, str) else ":".join(key)
+        receiver = self.config_channel.new_receiver(name=recv_name, limit=1)
 
         def _get_key_or_error(
             config: Mapping[str, Any]
@@ -184,32 +195,54 @@ class ConfigManager(BackgroundService):
         key_receiver = receiver.map(_get_key_or_error)
 
         if skip_unchanged:
-            return key_receiver.filter(WithPrevious(_not_equal_with_logging))
+            # For some reason the type argument for WithPrevious is not inferred
+            # correctly, so we need to specify it explicitly.
+            return key_receiver.filter(
+                WithPrevious[Mapping[str, Any] | InvalidValueForKeyError | None](
+                    lambda old, new: _not_equal_with_logging(
+                        key=key, old_value=old, new_value=new
+                    )
+                )
+            )
 
         return key_receiver
 
 
 def _not_equal_with_logging(
+    *,
+    key: str | Sequence[str],
     old_value: Mapping[str, Any] | InvalidValueForKeyError | None,
     new_value: Mapping[str, Any] | InvalidValueForKeyError | None,
 ) -> bool:
     """Return whether the two mappings are not equal, logging if they are the same."""
     if old_value == new_value:
-        _logger.info("Configuration has not changed, skipping update")
+        _logger.info("Configuration has not changed for key %r, skipping update.", key)
         return False
 
     if isinstance(new_value, InvalidValueForKeyError) and not isinstance(
         old_value, InvalidValueForKeyError
     ):
+        subkey_str = ""
+        if key != new_value.key:
+            subkey_str = f"When looking for sub-key {key!r}: "
         _logger.error(
-            "Configuration for key %r has an invalid value: %r",
+            "%sConfiguration for key %r has an invalid value: %r",
+            subkey_str,
             new_value.key,
             new_value.value,
         )
     return True
 
 
-def _get_key(config: Mapping[str, Any], key: str) -> Mapping[str, Any] | None:
+def _get_key(
+    config: Mapping[str, Any],
+    # This is tricky, because a str is also a Sequence[str], if we would use only
+    # Sequence[str], then a regular string would also be accepted and taken as
+    # a sequence, like "key" -> ["k", "e", "y"]. We should never remove the str from
+    # the allowed types without changing Sequence[str] to something more specific,
+    # like list[str] or tuple[str].
+    key: str | Sequence[str],
+) -> Mapping[str, Any] | None:
     """Get the value from the configuration under the specified key.
 
     Args:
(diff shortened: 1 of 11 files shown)


llucax:
This is a draft because I need to complete/improve documentation and add tests, but the code should be pretty solid, and the existing tests pass (the CI is red because I left a couple of `TODO`s that make both `pylint` and `pytest` fail).

* Based on #1133
* Replaces #1104

src/frequenz/sdk/config/_manager.py
```diff
@@ -133,26 +164,72 @@ def new_receiver(
...
+
+        def _get_key_or_error(
+            config: Mapping[str, Any]
+        ) -> Mapping[str, Any] | InvalidValueForKeyError | None:
+            try:
+                return _get_key(config, key)
+            except InvalidValueForKeyError as error:
+                return error
+
+        key_receiver = receiver.map(_get_key_or_error)
```
ela-kotulska-frequenz: I think it would be more clear (and it would simplify other methods) if you finish processing `key` here completely ( so it doesn't influence other methods )
You can do that by adding one extra filter: " IsNotError()"

```
def isNotErro(config):
   if isinstance(new_value, InvalidValueForKeyError):
      _logger.error
      return False
   return True 
```

 `key_receiver = receiver.map(_get_key_or_error).filter(isNotError)`

src/frequenz/sdk/config/_manager.py
```diff
@@ -133,26 +164,72 @@ def new_receiver(
...
+        config: The configuration to get the value from.
+        key: The key to get the value for.
+
+    Returns:
+        The value under the key, or `None` if the key is not found.
+
+    Raises:
+        InvalidValueForKeyError: If the value under the key is not a mapping.
+    """
+    match config.get(key):
```
ela-kotulska-frequenz: It can raise KeyError, is it ok?
If yes, could you please add comment why as it is not obvious to me :) 
If no, Maybe raising InvalidValueForKeyError for `KeyError`?

src/frequenz/sdk/config/_manager.py
```diff
@@ -175,44 +207,49 @@ def new_receiver(
...
-        key_receiver = receiver.map(_get_key_or_error)
+        receiver = self.config_channel.new_receiver(name=f"{self}:{key}", limit=1).map(
+            lambda config: _load_config_with_logging_and_errors(
+                config,
+                config_class,
+                key=key,
+                base_schema=base_schema,
+                marshmallow_load_kwargs=marshmallow_load_kwargs,
+            )
+        )
```
ela-kotulska-frequenz, line 251: The same, please add extra filter `IsNotException`, then this part will be completely seprate from others and you won't have to change  `_not_equal_with_logging` :)

src/frequenz/sdk/config/_managing_actor.py
```diff
@@ -185,8 +187,8 @@ async def _run(self) -> None:
             async for event in file_watcher:
                 if not event.path.exists():
                     _logger.error(
-                        "%s: Received event %s, but the watched path %s doesn't exist.",
-                        self,
+                        "[%s] Received event %s, but the watched path %s doesn't exist.",
+                        self.name,
```
ela-kotulska-frequenz, line 204: is  `self.name` really needed in these logs? 
We don't use it like this anywhere and Logger includes name of the file in logs.

src/frequenz/sdk/config/_manager.py
llucax: This was changed in subsequent commits, does this comment still applies? Maybe you can make it in the end full diff instead so the context makes more sense?

src/frequenz/sdk/config/_manager.py
llucax: No, `.get()` returns `None` if not found, so it should be fine.

src/frequenz/sdk/config/_manager.py
llucax: I don't fully understand your suggestion here, can you maybe post a small diff with the suggested changes?

Just for context, I tried not to apply many filters on purpose, to avoid the extra overhead, but if it really makes the code cleaner, I guess we can afford it anyways, config updates should be pretty seldom.

src/frequenz/sdk/config/_managing_actor.py
llucax: Yeah, I guess you are right. The idea was that there could be many config managing actors, and without the name we don't know which one is actually emitting the logs, but since this will be probably always a singleton in one app, maybe is not really useful. Sometimes it is a delicate balance between the use we give this now and the use other users might give it in the future. If someone, for some reason, needs to instantiate more than one config managing actor, having the name is definitely useful.

BTW, the `name` exists solely for this (logging), but for singleton actors is not helping as there is no ambiguity.

src/frequenz/sdk/config/_manager.py
ela-kotulska-frequenz: I meant something like this:
```
        receiver = self.config_channel.new_receiver(name=f"{self}:{key}", limit=1)
        receiver = receiver.map(_get_key_or_error).filter(is_not_exception)
        receiver = receiver.map(_load_config_or_error).filter(is_not_exception)
        if skip_unchanged:
            receiver = receiver.filter(WithPrevious[DataclassT | None](
                lambda old, new: _not_equal_with_logging(
                    key=key, old_value=old, new_value=new
                )
            ))
```
Then all following method will simplify , you don't have to pass Exception as argument and check for exception. And typing will simplify, too because filter(is_not_exception) will stop as soon any exception appears.

But I just found outthat the receiver returns InvalidValueForKeyError exception, why is it necessary?
`Receiver[Mapping[str, Any] | InvalidValueForKeyError | None]:`

llucax: The original PR did what you are suggesting (completely hide config loading errors from the downstream users), but I thought doing that might not be the best approach, maybe downstream actors would want to do something with this information (an error and what kind of error while trying to get the new configuration), like falling back to some safe default config or something, the current approach of keeping the old configuration is only one possible way to handle errors, but errors are still there and people might want to handle them differently.

If actors want to keep the old config if there are errors retrieving the new config, they can do it explicitly, it is just 2 extra lines of code. I also think it is good to see at the actor level more explicitly that we are handling possible errors.

I would at some point even make the `ConfigManagingActor` also send exceptions via the channel, like errors reading the file or parsing the TOML, that right now are completely being hidden to the users.

This was the main motivation for this.

llucax:
If there are no major complains about this approach, I will move forward to clean it up so we can merge it soon. I will start doing that, so if anyone wants to review this, please do it sooner than later @frequenz-floss/python-sdk-team .

src/frequenz/sdk/config/_manager.py
ela-kotulska-frequenz: here :) 
https://github.com/frequenz-floss/frequenz-sdk-python/pull/1134#discussion_r1891424080

src/frequenz/sdk/config/_manager.py
ela-kotulska-frequenz: Ok, you are right, it is better to return it.  Thanks for explanation!

llucax:
> * Replaces #1104 

Maybe too late for this, but these are the main changes compared to the old PR (I had them written down and forgot to put them in the PR when I created it):

- Make `unknown=EXCLUDE` the default in `ConfigManager`
- Make config manager a background service
- Remove auto start (requires the constructor to be run in an async context, not useful)
- Remove `skip_none`
- Make passing of marshmallow arguments explicit
- Make `key` mandatory
- Make wrong value for key an error we send and send all errors
- Catch unexpected exceptions
- Add a new `wait_for_first` utility function

src/frequenz/sdk/config/_managing_actor.py
llucax: I'm undecided about this, so unless there is a strong push-back I prefer to leave it as it is now, we can change it in a future PR.

llucax:
OK, this is ready for review. I completed the release notes, added tests, added comprehensive documentation, including recommended practices when writing actors and apps (please review!) and a user guide, and fixed a bug in the actor (and manager) when passing a string as a config file (now both accept also a single file).

I prepared the release notes to release after this is merged.

src/frequenz/sdk/config/_manager.py
```diff
@@ -0,0 +1,118 @@
...
+            config_paths: The paths to the TOML files with the configuration. Order
+                matters, as the configuration will be read and updated in the order
+                of the paths, so the last path will override the configuration set by
+                the previous paths. Dict keys will be merged recursively, but other
+                objects (like lists) will be replaced by the value in the last path.
+            force_polling: Whether to force file polling to check for changes.
+            name: A name to use when creating actors. If `None`, `str(id(self))` will
+                be used. This is used mostly for debugging purposes.
+            polling_interval: The interval to poll for changes. Only relevant if
+                polling is enabled.
```
ela-kotulska-frequenz: Not very related, but why `force_polling` and `polling_interval` should be configurable here?
1. It depends from the system whether we need `force_polling` or not.
2. `polling_interval` is ok I guess,  but description `Only relevant if polling is enabled.`  seems not correct.  Even if `force_pooling` set to False, file polling will still be used as a fallback when the underlying file system does not support event-based notifications.

src/frequenz/sdk/config/_manager.py
```diff
@@ -313,6 +315,99 @@ def new_receiver(  # pylint: disable=too-many-arguments
...
+                        _logger.info("%s: Received first configuration.", receiver_name)
+                        return config
+                    case unexpected:
+                        assert (
+                            False
+                        ), f"{receiver_name}: Unexpected value received: {unexpected!r}."
+    except asyncio.TimeoutError:
+        _logger.error("%s: No configuration received in time.", receiver_name)
+        raise
+    raise ReceiverStoppedError(receiver)
```
ela-kotulska-frequenz, line 360: About `raise ReceiverStoppedError(receiver)`
If we plan to start and stop actors dynamically then I guess it will be be normal for receiver to stop. For example if sender stopped then receiver will probably stop, too.
And user can't really do anything if receiver stopped.
Maybe it would also be ok to just log warning, and return?

ela-kotulska-frequenz (commented):
LGTM :)

src/frequenz/sdk/config/_manager.py
llucax: In general, since we are building a SDK that anyone could use on any stack, IMHO it makes sense to allow full customization. We only use `force_polling` as a safer choice as `inotify` doesn't work well on all filesystems and mount options, but if you know what you are doing, `inotify` could be more efficient and someone might want to use it instead of force-polling.

I think the *Only relevant if polling is enabled* is correct, it doesn't say *Only relevant if `force_polling` is enabled*, it talks about *polling* in general, and as you say, if there is no better way to monitor for file changes, *polling* will be enabled as a fallback even if `force_polling` is not enabled.

I would leave it as it is, it is very niche and I think it is not worth spending a lot of time writing a ton of documentation about it, if someone needs to tune this, they will probably need to go read the docs up to `watchfiles`, as it is not trivial to figure out how this works and all the subtle behavior changes that can come up from different backends and options.

src/frequenz/sdk/config/_manager.py
llucax: Mmm, I'm not sure how would that work. If we return, it means the receiver is ready, but it is not, it is stopped. Normally this exception will bubble up and will end up restarting actors (or exiting the application if it was at the app level), which I guess is what we want? At least if the receiver is re-created when an actor is restarted.

I think there is a bigger topic here, how do we want to handle actor restarts in terms of completely resetting the state (like re-creating receivers) or not. Because of things like this, I think we should probably re-create the state, otherwise a stopped receiver will never be fixed unless each and every actor handle the exception and re-creates the receiver manually.

But this is a much wider topic and I think it is out of the scope of this PR.

ela-kotulska-frequenz (approved):

PR #1140, opened 2025-01-02
Don't allow checking if `None` is within the bounds

llucax:
This was never supported, a `ValueError` would be raised if `None` was passed to `__contains__`. Now we assert to make it clear that it's not allowed.

In the future we should probably change `Bounds` to always allow `None` as bounds and make `_T` simply bound to `Comparable`, as supporting `None` depending on the generic type makes it very hard to provide correct type hints.


llucax:
This doesn't change behavior, we can decide if we want to change it in a separate PR, but I will merge this for now to unblock the merging of the mypy update.

PR #1141, opened 2025-01-02
Fix cross-arch CI tests

### LINKED ISSUE #1142: Cross-arch tests are failing, package `python3.12-distutils` not found
### What happened?

```
160.1 E: Couldn't find any package by glob 'python3.12-distutils'
160.1 E: Couldn't find any package by regex 'python3.12-distutils'
```

The tests to pass.

### Affected version(s)

_No response_

### Affected part(s)

### Extra information

It seems that package is not available for arm64 for some reason.

llucax:
We only need to do this when the cache directory exists, otherwise the step fails with an error.

Also install `python-venv` instead of `python-distutils`, as the `distutils` package was removed in Python 3.12 (and deadsnakes).

### DIFF
--- a/.github/containers/nox-cross-arch/arm64-ubuntu-20.04-python.Dockerfile
+++ b/.github/containers/nox-cross-arch/arm64-ubuntu-20.04-python.Dockerfile
@@ -19,7 +19,7 @@ RUN apt-get update -y && \
         curl \
         git \
         python${PYTHON_VERSION} \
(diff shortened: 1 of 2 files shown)


llucax:
This is passing the CI in my fork, so I will force-merge as it is blocking all other PRs.

https://github.com/llucax/frequenz-sdk-python/actions/runs/12584299521

PR #1146, opened 2025-01-15
Implement the `ShiftingMatryoshka` algorithm for the PowerManager

### LINKED ISSUE #1125: PowerManager algorithm that can shift or limit bounds
### What's needed?

Currently the power manager has a 2-group approach, where each group has a Matryoshka implementation that supports only limiting of bounds.  And the first group shifts the second group's bounds.

We need a more flexible implementation where each actor is able to decide whether they want to shift or limit the bounds of the lower priority actors.

The following changes are necessary to achieve this:

- The priority becomes optional.
- A proposal with just a target power and _without_ bounds would shift the other actors.
- A proposal with target bounds and optional target power would limit other actors with no priority or lower priorities.

### Proposed solution

_No response_

### Use cases

_No response_

_No response_

### Additional context

_No response_

shsms:
This PR removes the `set_operating_point` feature and replaces the power manager's original `Matryoshka` algorithm with the new `ShiftingMatryoshka` algorithm.

With the new algorithm, power proposals from actors are added to get the target power for the components, and higher-priority actors can limit the bounds available to lower-priority actors.


src/frequenz/sdk/microgrid/_power_managing/_power_managing_actor.py
```diff
@@ -63,15 +65,7 @@ def __init__(  # pylint: disable=too-many-arguments
...
             algorithm: The power management algorithm to use.
-
-        Raises:
-            NotImplementedError: When an unknown algorithm is given.
         """
-        if algorithm is not Algorithm.MATRYOSHKA:
-            raise NotImplementedError(
-                f"PowerManagingActor: Unknown algorithm: {algorithm}"
-            )
-
```
llucax, line 75: This change is in the commit that says it is changing the default algorithm, I would move it into the commit adding the new algorithm instead.

src/frequenz/sdk/microgrid/_power_managing/_power_managing_actor.py
```diff
@@ -84,9 +78,17 @@ def __init__(  # pylint: disable=too-many-arguments
...
+            case Algorithm.MATRYOSHKA:
+                self._algorithm: BaseAlgorithm = Matryoshka(
+                    max_proposal_age=timedelta(seconds=60.0)
+                )
+            case Algorithm.SHIFTING_MATRYOSHKA:
+                self._algorithm = ShiftingMatryoshka(
+                    max_proposal_age=timedelta(seconds=60.0)
+                )
+            case _:
+                assert_never(algorithm)
```
llucax: This change is in the commit that says it is changing the default algorithm, I would move it into the commit adding the new algorithm instead.

llucax (commented):
I just gave this a quick look, only to the structural changes.

I wonder why removing the old algorithm if we can support more than one. I can understand if it is to remove the maintenance cost, but I would delay the removal until we collected more experience.

src/frequenz/sdk/microgrid/__init__.py
```diff
@@ -174,53 +174,132 @@
...
+When an actor requests a power that's outside its available bounds, the closest
+available becomes the target power.
+
+#### Shifting the bounds for lower priority actors
+
+When an actor A calls the `propose_power` method with a power, but without any
+bounds (or passes None values for both lower and upper bounds), the lower
+priority actors would see bounds shifted by the power proposed by actor A.
+After lower priority actor B sets a power, it will get shifted back by the power
+set by actor A.  This has the effect of *adding* the powers set by actors A and B.
```
llucax: I find it a bit confusing to focus on the bounds shifting aspect in this mode instead of focusing on the result (the adding of all power requests). I would write the docs reversing this focus, and saying that when no bounds are requested, the power requests are simply added, and to achieve that, bounds need to be shifted.

src/frequenz/sdk/microgrid/__init__.py
```diff
@@ -174,53 +174,132 @@
...
+Irrespective of what any actor sets, the final power won't exceed the available
+battery bounds.
+
+*Example 2*:
+
+| actor | priority | available bounds | request bounds | request power | unshifted power |
+|-------|----------|------------------|----------------|---------------|-----------------|
+| A     | 3        | -100kW .. 100kW  | None           | 20kW          |                 |
+| B     | 2        | -120kW .. 80kW   | None           | -20kW         | 0kW             |
+|       |          |                  |                | target power  | 0kW             |
```
llucax: I also didn't it find trivial what's the role of the priority in this mode. I guess all it does is affected the order in which bounds are shifted, right? So if priorities of actor A and B are swapped, this will look like this?

```
| actor | priority | available bounds | request bounds | request power | unshifted power |
|-------|----------|------------------|----------------|---------------|-----------------|
| B     | 3        | -100kW .. 100kW  | None           | -20kW         | 0kW             |
| A     | 2        | -80kW .. 120kW   | None           | 20kW          |                 |
|       |          |                  |                | target power  | 0kW             |
```

And in any case, in this mode changing priorities would change nothing, right? Just the bounds that are seen by the actors, but which in any case are "virtual bounds" that the actual power request might not even be able to follow.

This confuses me a bit, because it looks like shifting bounds in this case doesn't make a lot of sense, is more like actors are blind, and can request power without having bounds in mind at all, as there is no accounting for how much power goes into each actor. Or is there any scenarios where bounds are important that I'm missing?

src/frequenz/sdk/microgrid/__init__.py
```diff
@@ -174,53 +174,132 @@
...
+| pri | available bounds | req bounds     | req power | s/l    | power in range | unshifted power |
+|-----|------------------|----------------|-----------|--------|----------------|-----------------|
+| 7   | -100kW .. 100kW  | None           | 10kW      | shifts | 10kW           | 10kW            |
+| 6   | -110kW .. 90kW   | -110kW .. 80kW | 10kW      | limits |                |                 |
+| 5   | -110kW .. 80kW   | -100kW .. 80kW | 80kW      | limits |                |                 |
+| 4   | -100kW .. 80kW   | None           | -120kW    | shifts | -100kW         | -90kW           |
+| 3   | 0kW .. 180kW     | None           | 60kW      | shifts | 60kW           | -30kW           |
+| 2   | -60kW .. 120kW   | -40kW .. 30kW  | 20kW      | limits |                |                 |
+| 1   | -40kW .. 30kW    | None           | 12kW      | shifts | 12kW           | -18kW           |
+| 0   | -52kW .. 18kW    | -40kW .. -10kW | -10kW     | limits | -10kW          | -28kW           |
```
llucax: Is this right? In the previous steps when limiting power, the requested power wasn't taken into account, but in this case it is. This is because it is the last power request? Maybe it is worth clarifying?

src/frequenz/sdk/microgrid/__init__.py
merlin-esser-frequenz: > but which in any case are "virtual bounds" that the actual power request might not even be able to follow.

what do you mean with "might not even be able to follow"?
It is true, that actor A could not charge 120 kW without actor B requesting -20 kW. Is that what you call virtual?

src/frequenz/sdk/microgrid/__init__.py
merlin-esser-frequenz: For me it doesnt look correct.
Priority 3 is the highest priority?
what exactly is "unshifted power"?

src/frequenz/sdk/microgrid/__init__.py
llucax: Yeah, what I mean by "virtual" is more like "fake", because any actor down the line can change what the end result is, they are not really limiting much. Is more like an information say "at this point we know what you can request is around these lines".

src/frequenz/sdk/microgrid/_power_managing/_power_managing_actor.py
shsms: Created a separate commit for integrating the new algorithm into the power manager.

src/frequenz/sdk/microgrid/_power_managing/_power_managing_actor.py
shsms: Created a separate commit for integrating the new algorithm into the power manager.

src/frequenz/sdk/microgrid/__init__.py
shsms: I had this feeling the whole time I was writing it, but somehow my commitment to the "shifting" narrative was stronger.  Updated.

src/frequenz/sdk/microgrid/__init__.py
shsms: > I guess all it does is affected the order in which bounds are shifted, right?

yes.

> And in any case, in this mode changing priorities would change nothing, right? Just the bounds that the actors see, but which in any case are "virtual bounds" that the actual power request might not even be able to follow.

I think priority can still have a significant influence because the highest priority actors have the first say. They can say min_bound = max_bound = power = 50kW and block everyone out.  If a lower-priority actor does the same limiting, it can't block the shifting effect of the higher-priority actor.  So, having a higher priority definitely gives an actor more influence.

That said, I've been thinking about an alternative algorithm where there's always shifting and additional limiting if bounds are specified. That algorithm sounds more consistent to me, unlike this PR, where the presence of bounds decides whether to add power or limit it. I'll see if I can make some examples before the next call.

src/frequenz/sdk/microgrid/__init__.py
shsms: > This is because it is the last power request?

It is because nobody is overriding the last actor's (limiting) request.  We see the same thing in the limiting-only examples as well.

> Priority 3 is the highest priority?

No, 7 is the highest priority.

>  what exactly is "unshifted power"?

I've renamed it to "cumulative power".

shsms: But this too looks a bit weird and complicated.  If we had shifting for all actors and optional limiting based on whether bounds are specified, that would be a much nicer interface.  I feel like that would be sufficient, but we can look into it a bit with some examples during the next call.

shsms:
> I wonder why removing the old algorithm if we can support more than one. I can understand if it is to remove the maintenance cost, but I would delay the removal until we collected more experience.

This was because pylint was complaining about duplicate lines, plus it is a separate commit, which should be easy to revert if necessary.  If you prefer, I can drop that commit, but we'll need to find a way to silence pylint either by refactoring the tests or pylint config.

But if we remove the current PR's limiting-only option like I mentioned in some comments above, this would stop being an issue as well, so we can revisit this after we decide that.

src/frequenz/sdk/microgrid/__init__.py
llucax: > I think priority can still have a significant influence because the highest priority actors have the first say. They can say min_bound = max_bound = power = 50kW and block everyone out. If a lower-priority actor does the same limiting, it can't block the shifting effect of the higher-priority actor. So, having a higher priority definitely gives an actor more influence.

Yeah, but what I mean is to really affect what downstream actors can do, you need to set bounds, if you don't, then actors downstream can do whatever they want, the reported bounds don't say much, as they are not narrowed.

> That said, I've been thinking about an alternative algorithm where there's always shifting and additional limiting if bounds are specified. That algorithm sounds more consistent to me, unlike this PR, where the presence of bounds decides whether to add power or limit it. I'll see if I can make some examples before the next call.

This sounds nice. Yeah, my feeling is this approach seems maybe too much influenced by the previous algorithm, and the 2 modes (with and without bounds settings) doesn't look very cohesive, so maybe it is worth thinking about the problem from scratch instead of as a patch to the current algorithm, so very cool if you already did this and have something to share!

llucax:
> This was because pylint was complaining about duplicate lines, plus it is a separate commit, which should be easy to revert if necessary. If you prefer, I can drop that commit, but we'll need to find a way to silence pylint either by refactoring the tests or pylint config.

We should definitely not remove code because pylint complains about duplication :D

Can't we just add a `# pylint: disable=<duplication>` at the beginning of the file(s) reporting duplication?

src/frequenz/sdk/microgrid/__init__.py
llucax: > I've renamed it to "cumulative power".

Maybe also mention that this is the final power the power manager will apply? For me it still feels not perfectly clear.

shsms:
This is ready for review again.

src/frequenz/sdk/microgrid/_power_managing/_shifting_matryoshka.py
```diff
@@ -0,0 +1,303 @@
...
+and their preferred power to determine the target power for the components.
+
+The preferred power of lower priority actors will take precedence as long as they
+respect the bounds set by higher priority actors.  If lower priority actors request
+power values outside the bounds set by higher priority actors, the target power will
+be the closest value to the preferred power that is within the bounds.
+
+When there is only a single proposal for a set of components, its preferred power would
+be the target power, as long as it falls within the system power bounds for the
+components.
```
llucax: I would move this docs to the class, as the module is private, it will never be rendered in the documentation website otherwise (unless you are explicitly including it in `docs/`.

src/frequenz/sdk/microgrid/_power_managing/_shifting_matryoshka.py
```diff
@@ -0,0 +1,303 @@
...
+        self._max_proposal_age_sec = max_proposal_age.total_seconds()
+        self._component_buckets: dict[frozenset[int], set[Proposal]] = {}
+        self._target_power: dict[frozenset[int], Power] = {}
+
+    def _calc_targets(
+        self,
+        proposals: set[Proposal],
+        system_bounds: SystemBounds,
+        priority: int | None = None,
+    ) -> tuple[Power | None, Bounds[Power]]:
```
llucax, line 69: I'm shaking while about bounds due to PTSD after fighting with bounds updates for the v0.17 microgrid API. I hope it is not too hard to merge :scream:

src/frequenz/sdk/microgrid/_power_managing/_shifting_matryoshka.py
```diff
@@ -0,0 +1,303 @@
...
+
+            if proposal_lower >= upper_bound:
+                proposal_lower = upper_bound
+                proposal_upper = upper_bound
+            elif proposal_upper <= lower_bound:
+                proposal_lower = lower_bound
+                proposal_upper = lower_bound
+
+            lower_bound = max(lower_bound, proposal_lower)
+            upper_bound = min(upper_bound, proposal_upper)
```
llucax: Are all these checks for `proposal_power`, `proposal_lower` and `proposal_upper` intentionally covering for an eventual `0.0` besides `None` or was the intention to only match `None`? Since in the code below, you are actually checking for `None` explicitly, I guess it is intentional, but maybe it would be best to add some comment to say it explicitly (either way).

src/frequenz/sdk/microgrid/__init__.py
```diff
@@ -174,53 +174,128 @@
...
-   | 3              | No           | 1000, -4000..2500     | -3000..3000      |
-   | 2              | No           | 2500                  | -3000..2500      |
-   | 1              | Yes          | None                  | -3000..3000      |
+| Actor | Priority | System Bounds   | Requested Bounds | Requested    | Aggregate |
+|       |          |                 |                  | Power        | Power     |
+|-------|----------|-----------------|------------------|--------------|-----------|
+| A     | 3        | -100kW .. 100kW | None             | 20kW         | 20kW      |
+| B     | 2        | -120kW .. 80kW  | None             | 50kW         | 70kW      |
+| C     | 1        | -170kW .. 30kW  | None             | 50kW         | 100kW     |
+|       |          |                 |                  | target power | 100kW     |
```
llucax: Maybe like [in the test](https://github.com/frequenz-floss/frequenz-sdk-python/pull/1146/files#diff-4312e57dc65aad6abe1ffd3ac6ff96fc64b446b1e0a5acfad6a13e5af1e98120R582-R596), adding a column with the "adjusted power" could add some extra clarity.

src/frequenz/sdk/microgrid/__init__.py
```diff
@@ -174,53 +174,128 @@
 controlling batteries, power could be distributed based on the `SoC` of the
 individual batteries, to keep the batteries in balance.
 
-### Resolving conflicting power proposals
+### How to work with other actors
```
llucax, line 177: All (or at least some of) the old documentation here should still be useful for people interested in the old `Matryoshka` algorithm, so not sure if replacing it is the best choice. I wonder if the docs shouldn't be moved to the algorithms themselves, and either referenced here (or see if there is a way to include it with mkdocs, but this won't work/look nice when reading the docs from the editor). Maybe in here we can just say more generically that there are 2 available algorithms, say which is the default, and then just link to the algorithm docs for details? If the plan is to remove the old algo soon, I guess it is fine to move forward as it is, but if we really want to support more than one algorithm, the docs should be structured in a way that we can explain all of them and switch the default easily without rewriting the whole `__init__.py` docs.

llucax (commented):
Didn't do a super in-depth review, in particular I didn't try to validate or fully understand the algorithm code, but I did follow the examples/new tests to understand the algorithm and it looks good to me and it was easier to follow (not sure if because I'm now more familiar with the algorithm or because the new wording focused on shifting :D).

Just a few minor comments, otherwise LGTM.

One final question though. The old matryoshka algorithm still works as before the operating point setting/group/shifting was added, right? The removal of the `set_operating_point` option only affects that?

src/frequenz/sdk/microgrid/_power_managing/_shifting_matryoshka.py
shsms: oh wow, I forgot to update these docs.  These are for the old matryoshka algorithm.  I will do that and move it to the class.

llucax: Hehe, right, I was reviewing commit by commit and commented on this when I saw you added the file but then didn't realized you didn't update it afterwards.

src/frequenz/sdk/microgrid/_power_managing/_shifting_matryoshka.py
shsms: > intentionally covering for an eventual `0.0` besides `None` or was the intention to only match `None`

I don't understand what that means.  Of course, I can add comments for these, though.

src/frequenz/sdk/microgrid/__init__.py
shsms: The code for the old algorithm is still there, but users cannot choose it.  Chances are we'll not need the old algorithm anymore.  We could introduce new algorithms in the future just for specific component types, like EVs, etc.  But we don't have to change the docs structure until we offer multiple algorithms for them to choose from.

src/frequenz/sdk/microgrid/_power_managing/_shifting_matryoshka.py
shsms: updated

src/frequenz/sdk/microgrid/_power_managing/_shifting_matryoshka.py
shsms: it should be fine because you're handling it in the compatibility layer from the client right?

src/frequenz/sdk/microgrid/_power_managing/_shifting_matryoshka.py
shsms: added more comments, also fixed a bug.

src/frequenz/sdk/microgrid/__init__.py
shsms: done

src/frequenz/sdk/microgrid/_power_managing/_shifting_matryoshka.py
```diff
@@ -0,0 +1,311 @@
...
+            if proposal_lower >= upper_bound:
+                proposal_lower = upper_bound
+                proposal_upper = upper_bound
+            elif proposal_upper <= lower_bound:
+                proposal_lower = lower_bound
+                proposal_upper = lower_bound
+
+            # Clamp the available bounds by the proposal bounds.
+            lower_bound = max(lower_bound, proposal_lower)
+            upper_bound = min(upper_bound, proposal_upper)
```
ela-kotulska-frequenz, line 134: If I understand correctly, we can replace all these lines with.
```
# If proposal bounds and bounds cover, clamp the available bounds by the proposal bounds.
if proposal_lower < upper_bound and proposal_upper > lower_bound:
           lower_bound = max(lower_bound, proposal_lower)
           upper_bound = min(upper_bound, proposal_upper)
```

ela-kotulska-frequenz (commented):
I am sorry but I still don't understand new approach fully. :D

For example (test based on `test_shifting_matryoshka.py`):
```test_shifting_matryoshka.py
    system_bounds = _base_types.SystemBounds(
        timestamp=datetime.now(tz=timezone.utc),
        inclusion_bounds=timeseries.Bounds(
            lower=Power.from_watts(-1000.0), upper=Power.from_watts(1000.0)
        ),
        exclusion_bounds=None,
    )

    tester = StatefulTester(batteries, system_bounds)
    tester.tgt_power(priority=7, power=150.0, bounds=(100.0, 200.0), expected=150.0)
    tester.tgt_power(priority=6, power=50.0, bounds=(0, 150.0), expected=200.0)
```


Two actors (not aware of each other and not aware of any bounds) sets power and bounds.
 I expected: `target_power = 100` and `bounds = (100..150)` , 
 Actual result:  `target_power = 200`. 
So two things happen:
1. target_power is bigger then any power proposal
2. Actors with priority 6 should narrow the upper bound to 150 but it doesn't.

Is it expected and why?

Another question: How actor should should be aware of other actor and sdk bounds? I am asking because some actors don't look at any bounds, now. They assume PowerManager will update bounds. 
But with this algorithm actor should be aware of the power proposed by another actors , because his proposed power will be combined with it.

shsms:
> I am sorry but I still don't understand new approach fully. :D

Okay, that's good because now I know I was right to be skeptical about there being no questions when I explained it last Tuesday.  But it was a nice few days when I thought that I must have done a really good job explaining and that we finally have created a fully consistent, easy-to-understand algorithm. :D  We can go through it again on Monday.

> Two actors (not aware of each other and not aware of any bounds) sets power and bounds. I expected: `target_power = 100` and `bounds = (100..150)` , Actual result: `target_power = 200`.

The original matryoshka algorithm would have produced 100 here.  The new algorithm adds the request powers from all the actors.

> So two things happen:
> 
>    1. target_power is bigger then any power proposal

Yes, it is the sum of the two actors.  That is how the shifting matryoshka algorithm differs from the older algorithm, which was just limiting.

> 
>    2. Actors with priority 6 should narrow the upper bound to 150 but it doesn't.
> 

Actor pri 6 narrows the bounds only for itself and for actors with a lower pri than 6.  Actor pri 7 sets 150.  And that becomes the 0 point for Actor pri 6.  Because Actor 7 limits bounds to 100..200, according to the new zero point, the available range for Actor 6 becomes -50..50.  When Actor 6 sets power to 50, it has it is within its available range.

You could try setting other powers from 6.  higher than 50, lower than 50, and negative values.

> 
> Is it expected and why?
> 
> Another question: How actor should should be aware of other actor and sdk bounds? I am asking because some actors don't look at any bounds, now. They assume PowerManager will update bounds. But with this algorithm actor should be aware of the power proposed by another actors , because his proposed power will be combined with it.

No, they are still independent.  They don't have access to what's going on with the actual batteries, or what other actors are setting.

src/frequenz/sdk/microgrid/_power_managing/_shifting_matryoshka.py
shsms: Just tried your suggestion, and it is missing some cases.

We can talk about exactly what my code is doing during the Monday call as well.

shsms: (you can also try it if you want, you'll see the issues in the tests)

ela-kotulska-frequenz: Right... sorry :)

ela-kotulska-frequenz (approved):

src/frequenz/sdk/microgrid/_power_managing/_shifting_matryoshka.py
llucax: Half and half, what was hard was how to cope with bounds being None/NaN/0, as sometimes 0 means None/NaN because we were blindly reading the protobuf value, and the default in protobuf (if the value is missing) is 0 for `int`. So these semantic differences I think are not handled centrally in the compatibility layer, but I'm not sure right now.

src/frequenz/sdk/microgrid/_power_managing/_shifting_matryoshka.py
llucax: I was talking specifically about using a nullable int as a boolean implicitly, the `next_proposal.bounds.lower or lower_bound`,  `f proposal_power and ...`, etc. In these cases both `0` and `None` will be considered `False`, so the question was if this was considered. Sometimes I feel it might be good to have a linting rule that requires writing `if (proposal_power == 0 or proposal_power is None) and (...` (but only sometimes, because other times I guess it can be overkill :laughing: )

shsms: Oh damn, you're right, we can't `continue` here if `proposal_power == 0`.  Stupid Python and its weird implicit booleans.  I will make a PR next week.

> Sometimes I feel it might be good to have a linting rule that requires writing `if (proposal_power == 0 or proposal_power is None) and (...` (but only sometimes, because other times I guess it can be overkill 😆 )

That sounds like a great idea. It is not overkill. Expressing our intentions clearly is always the way to go.

src/frequenz/sdk/microgrid/_power_managing/_shifting_matryoshka.py
shsms: We should expect that the API's bounds are always correct. If we get `None/NaN/0` from the API, then those should be treated as 0.

src/frequenz/sdk/microgrid/_power_managing/_shifting_matryoshka.py
llucax: Upvote so it is ready when we switch to ruff :)

* https://github.com/astral-sh/ruff/issues/10949

shsms: Looks like I had already changed this before merging: https://github.com/frequenz-floss/frequenz-sdk-python/blob/6d106d149aaa3c686370d997fa4793f601806b0e/src/frequenz/sdk/microgrid/_power_managing/_shifting_matryoshka.py#L136

So nothing to do.

PR #1151, opened 2025-01-22
Remove "Checking battery" log message

shsms:
It is not providing any useful information or details about its
context, and it is too noisy.


shsms:
Extracted from #1132.  Will keep that PR only for rate-limiting, which could take a while.

llucax (approved):

PR #1155, opened 2025-01-24
Drop empty Matryoshka buckets

### LINKED ISSUE #1148: `propose_power(None)` doesn't clear old buckets from the power manager
### What happened?

We proposed `None` power and still got "Overlapping component IDs are not supported".

When `None` proposals are received, the power manager should no longer keep any information about this source so that subsequent requests with overlapping component IDs can be accepted.

### Affected version(s)

_No response_

### Affected part(s)

I don't know (part:❓)

### Extra information

_No response_

shsms:
If buckets remain for ever in the power manager's Matryoshka instances, proposals with overlapping component IDs will never get accepted.  But if empty buckets go away as soon as they become empty, then it becomes possible to propose requests for components without conflicting with old unused component ID buckets.

### DIFF
--- a/src/frequenz/sdk/microgrid/_power_managing/_matryoshka.py
+++ b/src/frequenz/sdk/microgrid/_power_managing/_matryoshka.py
@@ -199,7 +199,15 @@ class Matryoshka(BaseAlgorithm):
             bucket = self._component_buckets.setdefault(component_ids, set())
             if proposal in bucket:
                 bucket.remove(proposal)
-            bucket.add(proposal)
+            if (
+                proposal.preferred_power is not None
+                or proposal.bounds.lower is not None
+                or proposal.bounds.upper is not None
+            ):
+                bucket.add(proposal)
+            elif not bucket:
+                del self._component_buckets[component_ids]
+                _ = self._target_power.pop(component_ids, None)
 
         # If there has not been any proposal for the given components, don't calculate a
         # target power and just return `None`.
--- a/src/frequenz/sdk/microgrid/_power_managing/_matryoshka.py
+++ b/src/frequenz/sdk/microgrid/_power_managing/_matryoshka.py
@@ -292,10 +292,17 @@ class Matryoshka(BaseAlgorithm):
         Args:
             loop_time: The current loop time.
         """
-        for bucket in self._component_buckets.values():
-            to_delete = []
-            for proposal in bucket:
+        buckets_to_delete: list[frozenset[int]] = []
+        for component_ids, proposals in self._component_buckets.items():
+            to_delete: list[Proposal] = []
+            for proposal in proposals:
                 if (loop_time - proposal.creation_time) > self._max_proposal_age_sec:
                     to_delete.append(proposal)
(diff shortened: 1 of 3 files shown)


shsms:
I will update #1146 to incorporate these changes after this is merged.

src/frequenz/sdk/microgrid/_power_managing/_matryoshka.py
```diff
@@ -199,7 +199,15 @@ def calculate_target_power(
...
-            bucket.add(proposal)
+            if (
+                proposal.preferred_power is not None
+                or proposal.bounds.lower is not None
+                or proposal.bounds.upper is not None
+            ):
+                bucket.add(proposal)
+            elif not bucket:
+                del self._component_buckets[component_ids]
+                _ = self._target_power.pop(component_ids, None)
```
llucax, line 210: I don't think you need the `_ = `. Also is it a `pop` instead of a `del` because `component_id` might not be in `self._target_power`?

tests/actor/_power_managing/test_matryoshka.py
```diff
@@ -424,3 +428,58 @@ async def test_matryoshka_drop_old_proposals() -> None:
     tester.tgt_power(
         priority=1, power=20.0, bounds=(20.0, 50.0), expected=25.0, must_send=True
     )
+
+    # When all proposals are too old, they are dropped, and the buckets are dropped as
+    # well.  After that, sending a request for a different but overlapping bucket will
+    # succeed.  And it will fail until then.
+    with pytest.raises(
+        NotImplementedError,
+        match=re.escape(
```
llucax, line 437: TIL: `re.escape`. :D

llucax (approved):

src/frequenz/sdk/microgrid/_power_managing/_matryoshka.py
shsms: I have been using an extra pedantic lsp server that warns about discarding return values.  And because I have been hit by such bugs before, I decided to not disable it.

And yes, pop because there's one case where component IDs won't be in target_power, and that's when the first request ever sent to a battery pool is a `None, (None, None)`.

llucax: Which linter is that? I did a quick search and didn't found any linter that has that feature. I agree is a good lint to have, otherwise it looks very arbitrary, as we have many other places where values are silently discarded. It would be nice to enabled it for everyone so we don't end up with inconsistent code (that it might also be very noisy for the people using the pendantic tool).

llucax:
This is getting a weird error saying `bash: nox: command not found` when trying to install dependencies :-/

src/frequenz/sdk/microgrid/_power_managing/_matryoshka.py
shsms: I've been using https://docs.basedpyright.com/latest/ and enjoying it a lot.  I've been using it as an LSP server, but it is also available as a CLI tool.

PR #1159, opened 2025-02-04
Add a `wait_for_samples` method to the `MovingWindow`

### LINKED ISSUE #967: [MovingWindow] Add a trigger that fires after having received a fixed number of samples

shsms:
Closes #967

### DIFF
--- a/src/frequenz/sdk/timeseries/_moving_window.py
+++ b/src/frequenz/sdk/timeseries/_moving_window.py
@@ -355,14 +355,24 @@ class MovingWindow(BackgroundService):
             asyncio.create_task(self._resampler.resample(), name="resample")
         )
 
-    def count_valid(self) -> int:
-        """
-        Count the number of valid samples in this `MovingWindow`.
+    def count_valid(
+        self, *, since: datetime | None = None, until: datetime | None = None
+    ) -> int:
+        """Count the number of valid samples in this `MovingWindow`.
+
+        If `since` and `until` are provided, the count is limited to the samples between
+        (and including) the given timestamps.
+
+        Args:
+            since: The timestamp from which to start counting.  If `None`, the oldest
+                timestamp of the buffer is used.
+            until: The timestamp until (and including) which to count.  If `None`, the
+                newest timestamp of the buffer is used.
 
         Returns:
             The number of valid samples in this `MovingWindow`.
         """
-        return self._buffer.count_valid()
+        return self._buffer.count_valid(since=since, until=until)
 
     def count_covered(self) -> int:
         """Count the number of samples that are covered by the oldest and newest valid samples.
--- a/src/frequenz/sdk/timeseries/_moving_window.py
+++ b/src/frequenz/sdk/timeseries/_moving_window.py
@@ -343,8 +343,7 @@ class MovingWindow(BackgroundService):
         assert self._resampler is not None
 
         async def sink_buffer(sample: Sample[Quantity]) -> None:
-            if sample.value is not None:
-                self._buffer.update(sample)
+            self._buffer.update(sample)
 
         resampler_channel = Broadcast[Sample[Quantity]](name="average")
         self._resampler_sender = resampler_channel.new_sender()
--- a/src/frequenz/sdk/timeseries/_moving_window.py
+++ b/src/frequenz/sdk/timeseries/_moving_window.py
@@ -374,14 +374,25 @@ class MovingWindow(BackgroundService):
         """
         return self._buffer.count_valid(since=since, until=until)
 
-    def count_covered(self) -> int:
+    def count_covered(
+        self, *, since: datetime | None = None, until: datetime | None = None
+    ) -> int:
         """Count the number of samples that are covered by the oldest and newest valid samples.
+
+        If `since` and `until` are provided, the count is limited to the samples between
+        (and including) the given timestamps.
+
+        Args:
+            since: The timestamp from which to start counting.  If `None`, the oldest
+                timestamp of the buffer is used.
+            until: The timestamp until (and including) which to count.  If `None`, the
+                newest timestamp of the buffer is used.
 
         Returns:
             The count of samples between the oldest and newest (inclusive) valid samples
                 or 0 if there are is no time range covered.
         """
-        return self._buffer.count_covered()
+        return self._buffer.count_covered(since=since, until=until)
 
     @overload
     def __getitem__(self, key: SupportsIndex) -> float:
--- a/src/frequenz/sdk/timeseries/_moving_window.py
+++ b/src/frequenz/sdk/timeseries/_moving_window.py
@@ -190,6 +190,8 @@ class MovingWindow(BackgroundService):
             align_to=align_to,
         )
 
+        self._condition_new_sample = asyncio.Condition()
+
     def start(self) -> None:
         """Start the MovingWindow.
 
@@ -318,6 +320,44 @@ class MovingWindow(BackgroundService):
             start, end, force_copy=force_copy, fill_value=fill_value
         )
 
+    async def wait_for_samples(self, n: int) -> None:
+        """Wait until the next `n` samples are available in the MovingWindow.
+
+        This function returns after `n` new samples are available in the MovingWindow,
+        without considering whether the new samples are valid.  The validity of the
+        samples can be verified by calling the
+        [`count_valid`][frequenz.sdk.timeseries.MovingWindow.count_valid] method.
+
+        Args:
+            n: The number of samples to wait for.
+
+        Raises:
+            ValueError: If `n` is less than or equal to 0 or greater than the capacity
+                of the MovingWindow.
+        """
+        if n == 0:
+            return
+        if n < 0:
+            raise ValueError("The number of samples to wait for must be 0 or greater.")
+        if n > self.capacity:
+            raise ValueError(
+                "The number of samples to wait for must be less than or equal to the "
+                + f"capacity of the MovingWindow ({self.capacity})."
+            )
+        start_timestamp = (
+            # Start from the next expected timestamp.
+            self.newest_timestamp + self.sampling_period
+            if self.newest_timestamp is not None
+            else None
+        )
+        while True:
+            async with self._condition_new_sample:
+                # Every time a new sample is received, this condition gets notified and
+                # will wake up.
+                _ = await self._condition_new_sample.wait()
+            if self.count_covered(since=start_timestamp) >= n:
+                return
+
     async def _run_impl(self) -> None:
         """Awaits samples from the receiver and updates the underlying ring buffer.
 
@@ -331,6 +371,9 @@ class MovingWindow(BackgroundService):
                     await self._resampler_sender.send(sample)
                 else:
                     self._buffer.update(sample)
+                    async with self._condition_new_sample:
+                        # Wake up all coroutines waiting for new samples.
+                        self._condition_new_sample.notify_all()
 
         except asyncio.CancelledError:
             _logger.info("MovingWindow task has been cancelled.")
@@ -344,6 +387,9 @@ class MovingWindow(BackgroundService):
 
         async def sink_buffer(sample: Sample[Quantity]) -> None:
             self._buffer.update(sample)
+            async with self._condition_new_sample:
+                # Wake up all coroutines waiting for new samples.
+                self._condition_new_sample.notify_all()
 
         resampler_channel = Broadcast[Sample[Quantity]](name="average")
         self._resampler_sender = resampler_channel.new_sender()
--- a/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
+++ b/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
@@ -678,13 +678,29 @@ class OrderedRingBuffer(Generic[FloatArray]):
             // self._sampling_period.total_seconds()
         )
 
-    def count_valid(self) -> int:
-        """Count the number of valid items that this buffer currently holds.
+    def count_valid(
+        self, *, since: datetime | None = None, until: datetime | None = None
+    ) -> int:
+        """Count the number of valid items in this buffer.
+
+        If `since` and `until` are provided, the count is limited to the items between
+        (and including) the given timestamps.
+
+        Args:
+            since: The timestamp from which to start counting.  If `None`, the oldest
+                timestamp in the buffer is used.
+            until: The timestamp until (and including) which to count.  If `None`, the
+                newest timestamp in the buffer is used.
 
         Returns:
             The number of valid items in this buffer.
         """
-        if self._timestamp_newest == self._TIMESTAMP_MIN:
+        if since is None or since < self._timestamp_oldest:
+            since = self._timestamp_oldest
+        if until is None or until > self._timestamp_newest:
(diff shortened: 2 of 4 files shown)


src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -678,13 +678,29 @@ def count_covered(self) -> int:
...
+    ) -> int:
+        """Count the number of valid items in this buffer.
+
+        If `since` and `until` are provided, the count is limited to the items between
+        the given timestamps.
+
+        Args:
+            since: The timestamp from which to start counting.  If `None`, the oldest
+                timestamp in the buffer is used.
+            until: The timestamp until which to count.  If `None`, the newest timestamp
```
cwasicki: Better clarify that `until` is inclusive since usually the end timestamp (or index) is exclusive in python.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -318,6 +320,34 @@ def window(
...
+
+        Raises:
+            ValueError: If `n` is less than or equal to 0 or greater than the capacity
+                of the MovingWindow.
+        """
+        if n <= 0:
+            raise ValueError(
+                "The number of samples to wait for must be greater than 0."
+            )
+        if n > self.capacity:
```
cwasicki, line 342: Theoretically I don't see why this is required but also don't have a good example where it hurts.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -318,6 +320,34 @@ def window(
...
+            raise ValueError(
+                "The number of samples to wait for must be less than or equal to the "
+                + f"capacity of the MovingWindow ({self.capacity})."
+            )
+        start_timestamp = self.newest_timestamp
+        if n < self.capacity:
+            n += self.count_valid(since=start_timestamp)
+        while True:
```
cwasicki, line 353: I am not familiar with the Condition concept and cannot follow what is happening here, maybe you could add some explanatory comments here and in the tests below.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -318,6 +320,34 @@ def window(
             start, end, force_copy=force_copy, fill_value=fill_value
         )
 
+    async def wait_for_samples(self, n: int) -> None:
+        """Wait until the next `n` samples are available in the MovingWindow.
```
cwasicki, line 324: I guess this is _valid_ samples? If so, I am actually not sure if we want this or something time-based i.e. allow that not all samples are valid. However, for our current use-case this also works since we would set n=1 anyway.

src/frequenz/sdk/timeseries/_moving_window.py
shsms: We are counting the number of items in the buffer, for reporting.  If we have to support n > capacity, we'll need a different implementation.  But I couldn't think of a usecase for allowing n > capacity.

src/frequenz/sdk/timeseries/_moving_window.py
shsms: That's interesting.  _valid_ means that any data was received.  If a component is missing data, resampler will send `None` and that is not a valid value.

If a component is sending only `None`, should this function return after `n` `None`s are received?  I'm guessing it should?

cwasicki: I see, in many scenarios I wouldn't distinguish between missing or None values. I think it shouldn't return after n Nones but after `n` new time steps which have at least 1 real value. However, we could leave this also for later.

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
shsms: done

src/frequenz/sdk/timeseries/_moving_window.py
shsms: I've added a few lines, hope that's sufficient.

src/frequenz/sdk/timeseries/_moving_window.py
shsms: I've updated it to return after `n` samples are received.  Whether they were valid or not needs to be checked with a call to `count_valid`.  I've also updated the docs to state this.

cwasicki: Oh, I see the confusion, I understood that it triggers when `n` new _output_ samples have been "received", i.e. there are timestamps in the resulting moving window. But this is about _input_ samples. So even if we receive 100 samples, if these are all older than newest timestamp we wouldn't get any new timestamp in the window but updated data points of older timestamps. 

This makes sense to me, would stress that in the doc though, e.g the valid samples part is confusing IMO since this is indeed about the new samples.

cwasicki: Maybe:
```
       """Wait until the next `n` samples have been received in the MovingWindow.

        This function returns after `n` input samples have been received, without considering
        whether the received samples are valid or which timestamp they have. The validity of 
        the samples in the updated moving window can be verified by calling the
        [`count_valid`][frequenz.sdk.timeseries.MovingWindow.count_valid] method.
```

tests/timeseries/test_moving_window.py
```diff
@@ -210,21 +213,234 @@ async def test_access_empty_window() -> None:
...
+    """Test waiting for samples in the window."""
+    window, sender = init_moving_window(timedelta(seconds=10))
+    async with window:
+        task = asyncio.create_task(window.wait_for_samples(5))
+        await asyncio.sleep(0)
+        assert not task.done()
+        await push_logical_meter_data(sender, range(0, 5))
+        await asyncio.sleep(0)
+        # After pushing 5 values, the `wait_for_samples` task should be done.
+        assert task.done()
```
cwasicki, line 385: Now looking at the indended usage example here, would it be possible to wrap this into a receiver that sends the moving window content each time the method triggers? We could implement this in downstream apps of course, but I guess in the end this would be the pattern we would use.

src/frequenz/sdk/timeseries/_moving_window.py
shsms: No, it would wait until there are `n` output samples, but some output samples could be nan.

It does consider the timestamps when the samples are received.  It expects `n` "new" samples to be available in the buffer before it returns.

The current tests only cover cases where there is no resampling in the moving window.  I'll rectify that.

tests/timeseries/test_moving_window.py
shsms: Yup, can do that.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -318,6 +320,44 @@ def window(
...
+            n: The number of samples to wait for.
+
+        Raises:
+            ValueError: If `n` is less than or equal to 0 or greater than the capacity
+                of the MovingWindow.
+        """
+        if n <= 0:
+            raise ValueError(
+                "The number of samples to wait for must be greater than 0."
+            )
```
llucax: I wonder if returning immediately couldn't be a more useful alternative, in cases where maybe `n` is calculated and could result in 0, in which case you actually don't need to wait for anything. I guess @cwasicki can hint which approach could be more convenient for the use cases he has in mind.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -318,6 +320,44 @@ def window(
...
+        if n > self.capacity:
+            raise ValueError(
+                "The number of samples to wait for must be less than or equal to the "
+                + f"capacity of the MovingWindow ({self.capacity})."
+            )
```
llucax, line 346: This could also just silently wait for `self.capacity` instead, like slicing when you do `[:10]` for an array that has less than 10 items. It can be more confusing if you wanted to wait for 10 and got 5 instead but it is something at least familiar in Python. Again maybe @cwasicki can hint which approach would be more intuitive for data scientists.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -318,6 +320,44 @@ def window(
...
+            # Start from the next expected timestamp.
+            self.newest_timestamp + self.sampling_period
+            if self.newest_timestamp is not None
+            else None
+        )
+        while True:
+            async with self._condition_new_sample:
+                # Every time a new sample is received, this condition gets notified and
+                # will wake up.
+                _ = await self._condition_new_sample.wait()
```
llucax, line 357: Just thinking out loud, and I don't even think we need to think about it for this PR, but maybe this could be done more efficiently by reversing the logic, and only set the condition when the counter is set (it could be even be a simple `Event` in this case). This would mean we would need to save the "waiters" in the instance, and then notify the appropriate waiter, so maybe that's a bit costly too, but I guess the most common case will be having only one or very few waiters. :thinking:

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -318,6 +320,44 @@ def window(
             start, end, force_copy=force_copy, fill_value=fill_value
         )
 
+    async def wait_for_samples(self, n: int) -> None:
```
llucax, line 323: Maybe:
```suggestion
    async def wait_for_samples(self, n: int, /) -> None:
```
as using it as a keyword doesn't really bring anything.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -355,23 +401,44 @@ async def sink_buffer(sample: Sample[Quantity]) -> None:
             asyncio.create_task(self._resampler.resample(), name="resample")
         )
 
-    def count_valid(self) -> int:
-        """
-        Count the number of valid samples in this `MovingWindow`.
+    def count_valid(
+        self, since: datetime | None = None, until: datetime | None = None
```
llucax: Here I think is more important to use keyword-only as both have the same time and could be mixed up.
```suggestion
        self, *, since: datetime | None = None, until: datetime | None = None
```

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -355,23 +401,44 @@ async def sink_buffer(sample: Sample[Quantity]) -> None:
...
 
         Returns:
             The number of valid samples in this `MovingWindow`.
         """
-        return self._buffer.count_valid()
+        return self._buffer.count_valid(since, until)
 
-    def count_covered(self) -> int:
+    def count_covered(
+        self, since: datetime | None = None, until: datetime | None = None
```
llucax: Here too:
```suggestion
        self, *, since: datetime | None = None, until: datetime | None = None
```

src/frequenz/sdk/timeseries/_moving_window.py
cwasicki: Good idea, that would save a try block

src/frequenz/sdk/timeseries/_moving_window.py
cwasicki: I thought of it that if n > capacity the n would still be respected, but calculated in terms of time steps since this was triggered. But I think the different understanding here shows already that this could be confusing and we can limit it until we have a case where we need it.

tests/timeseries/test_moving_window.py
cwasicki: As discussed, don't think we need it for this PR but can be added in future if we prefer that pattern.

src/frequenz/sdk/timeseries/_moving_window.py
llucax: Yeah, true, best to start with a safe approach, for my behaviour you can still easily get is via `mw.wait_for_samples(min(n, mw.capacity))` I guess, so it doesn't look like it is adding too much value.

src/frequenz/sdk/timeseries/_moving_window.py
llucax: Not sure about negative values though, as it is not perfectly clear what it means (as opposed to "wait for zero samples" which for me translates more naturally to "don't wait"). Although if we don't raise I think the only possible meaning is to also not wait. If we raise, I guess if you could get a negative you can always use `mw.wait_for_samples(max(0, calculated_n_that_can_be_negative))` so it wouldn't be that bad either.

src/frequenz/sdk/timeseries/_moving_window.py
shsms: Updated to clarify that it returns after `n` new samples in the moving window, and not just input samples, as discussed.

src/frequenz/sdk/timeseries/_moving_window.py
shsms: done, still raises for < 0.

src/frequenz/sdk/timeseries/_moving_window.py
shsms: The most common case is also with n=1 and resampling_interval=15 minutes.

src/frequenz/sdk/timeseries/_moving_window.py
shsms: Not doing it, because what it someone wants to.

src/frequenz/sdk/timeseries/_moving_window.py
shsms: done

src/frequenz/sdk/timeseries/_moving_window.py
shsms: done

shsms:
Also added tests for `wait_for_samples` in moving windows with a resampler.

src/frequenz/sdk/timeseries/_moving_window.py
```diff
@@ -318,6 +320,44 @@ def window(
             start, end, force_copy=force_copy, fill_value=fill_value
         )
 
+    async def wait_for_samples(self, n: int) -> None:
+        """Wait until the next `n` samples are available in the MovingWindow.
+
+        This function returns after `n` samples are available in the MovingWindow,
```
cwasicki: Would say `after n new samples are available`

cwasicki (dismissed):
LGTM, just minor clarification proposal.

src/frequenz/sdk/timeseries/_moving_window.py
shsms: done

llucax (approved):
LGTM. I disabled auto-merge to give @cwasicki the opportunity to approve too, as he knows what's needed best.

shsms:
He had already approved.

PR #1165, opened 2025-02-18
Stop formula engine

ela-kotulska-frequenz:
This PR introduces 4 new changes:
* make formula `stop` method public
* Add stop method to MetricFetcher - which stops receivers and FallbackMetricFetcher
* Fix bug that raised many exception when formulas stopped
* We can stop formula, now. So we can stop  FallbackMetricFetcher when primary metric fetcher is good again.

All this simplified the code a lot.

### DIFF
--- a/src/frequenz/sdk/timeseries/formula_engine/_formula_engine.py
+++ b/src/frequenz/sdk/timeseries/formula_engine/_formula_engine.py
@@ -117,9 +117,10 @@ class FormulaEngine(Generic[QuantityT]):
         if self._task is None:
             return
         await cancel_and_await(self._task)
+
         _, fetchers = self._builder.finalize()
         for fetcher in fetchers.values():
-            fetcher.stream.close()
+            await fetcher.stop()
 
     @classmethod
     def from_receiver(
--- a/src/frequenz/sdk/timeseries/formula_engine/_formula_engine.py
+++ b/src/frequenz/sdk/timeseries/formula_engine/_formula_engine.py
@@ -313,7 +313,7 @@ class FormulaEngine(Generic[QuantityT]):
             try:
                 msg = await evaluator.apply()
             except asyncio.CancelledError:
-                _logger.exception("FormulaEngine task cancelled: %s", self._name)
+                _logger.debug("FormulaEngine task cancelled: %s", self._name)
                 raise
             except Exception as err:  # pylint: disable=broad-except
                 _logger.warning(
@@ -582,7 +582,7 @@ class FormulaEngine3Phase(Generic[QuantityT]):
                     phase_3.value,
                 )
             except asyncio.CancelledError:
-                _logger.exception("FormulaEngine task cancelled: %s", self._name)
+                _logger.debug("FormulaEngine task cancelled: %s", self._name)
                 break
             else:
                 await sender.send(msg)
--- a/src/frequenz/sdk/timeseries/formula_engine/_formula_engine.py
+++ b/src/frequenz/sdk/timeseries/formula_engine/_formula_engine.py
@@ -112,7 +112,7 @@ class FormulaEngine(Generic[QuantityT]):
         self._channel: Broadcast[Sample[QuantityT]] = Broadcast(name=self._name)
         self._task: asyncio.Task[None] | None = None
 
-    async def _stop(self) -> None:
+    async def stop(self) -> None:
         """Stop a running formula engine."""
         if self._task is None:
             return
@@ -428,7 +428,7 @@ class FormulaEngine3Phase(Generic[QuantityT]):
             FormulaEngine[QuantityT],
         ] = phase_streams
 
-    async def _stop(self) -> None:
+    async def stop(self) -> None:
         """Stop a running formula engine."""
         if self._task is None:
             return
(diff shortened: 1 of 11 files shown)


src/frequenz/sdk/timeseries/formula_engine/_formula_engine.py
```diff
@@ -313,7 +313,7 @@ async def _run(self) -> None:
             try:
                 msg = await evaluator.apply()
             except asyncio.CancelledError:
-                _logger.exception("FormulaEngine task cancelled: %s", self._name)
+                _logger.info("FormulaEngine task cancelled: %s", self._name)
```
llucax: Not sure if this is really useful information to have unless one is debugging. I would make it a `debug`.

llucax (commented):
LGTM but the description about the `close` issue should be removed from the OR description so it doesn't go to the merge commit and it needs release notes as the formula engine is public.

src/frequenz/sdk/timeseries/formula_engine/_formula_engine.py
ela-kotulska-frequenz: Done

ela-kotulska-frequenz:
One more error described in PR message

src/frequenz/sdk/timeseries/formula_engine/_formula_steps.py
```diff
@@ -529,7 +529,15 @@ async def _fetch_next(self) -> Sample[QuantityT] | None:
         try:
             next_value = await self._stream.receive()
         except ReceiverError[Any] as err:
```
llucax: My bet would be that `ReceiverError[Any]` is not really a `ReceiverError` but some `type` object or something like that. Maybe try replacing by just `ReceiverError`?

I just did a quick test and it is that:

```py
from typing import Any, Generic, TypeVar

T = TypeVar("T")

class Ex(Exception, Generic[T]):
    pass

try:
    raise Ex("foo")
except Ex[Any] as e:
    print(e)
```

This triggers the same issue, removing the `[Any]` makes it work.

ela-kotulska-frequenz: You are right! :) It was ReceiverStoppedError .

ela-kotulska-frequenz:
Last commit is very complicated,  sorry for that!
But those things are really related and one implied the other and I couldn't separate it.

src/frequenz/sdk/timeseries/formula_engine/_formula_steps.py
llucax: What I mean is the generic type specification was wrong, a `ReceiverStoppedError` is still a `ReceiverError`, but you can't use the type parameter in `except` clauses.

src/frequenz/sdk/timeseries/formula_engine/_formula_steps.py
```diff
@@ -417,86 +433,69 @@ def stream_name(self) -> str:
...
+                    "Stream for fallback metric fetcher %s closed.",
+                    fallback_fetcher.name,
+                )
+            else:
+                _logger.error(
+                    "Failed to fetch next value from %s. Fallback stream closed.",
+                    self._name,
+                )
+            return None
+        except ReceiverError[Any] as err:
```
llucax: This is still wrong, and if for any reason an exception that is not a `ReceiverStoppedError` is raised, we'll get the same error about the exception not inheriting from `BaseException`.

It should be:
```suggestion
        except ReceiverError as err:
```

src/frequenz/sdk/timeseries/formula_engine/_formula_steps.py
```diff
@@ -506,34 +505,62 @@ async def fetch_next(self) -> Sample[QuantityT] | None:
...
+            primary_value = await self._stream.receive()
+        except ReceiverStoppedError:
+            if self._is_stopped:
+                _logger.debug("Stream for metric fetcher %s closed.", self._name)
+                return None
+            _logger.error(
+                "Failed to fetch next value from %s. Primary stream closed.",
+                self._name,
+            )
         except ReceiverError[Any] as err:
```
llucax: Even if this is a pr-existing bug, we should fix it too.
```suggestion
        except ReceiverError as err:
```

llucax (commented):
I'll leave the changes in formula engine to @shsms as he is more familiar with that code base.

src/frequenz/sdk/timeseries/formula_engine/_formula_steps.py
ela-kotulska-frequenz: Right, sorry I didn't understand it. Fixed :)

src/frequenz/sdk/timeseries/formula_engine/_formula_steps.py
ela-kotulska-frequenz: Done

src/frequenz/sdk/timeseries/formula_engine/_formula_steps.py
ela-kotulska-frequenz: done

ela-kotulska-frequenz:
Thanks @shsms for separating these commits. Is much more readable now! :)

shsms (approved):

PR #1166, opened 2025-02-24
Stop raising CancelledError when actor is cancelled

ela-kotulska-frequenz:
Cancelling BackgroundService should not raise exception. But we see exception when we stop actor, after calling `actor.wait()`.

### DIFF
--- a/src/frequenz/sdk/actor/_run_utils.py
+++ b/src/frequenz/sdk/actor/_run_utils.py
@@ -38,18 +38,21 @@ async def run(*actors: Actor) -> None:
         done_tasks, pending_tasks = await asyncio.wait(
             pending_tasks, return_when=asyncio.FIRST_COMPLETED
         )
-
         # This should always be only one task, but we handle many for extra safety
         for task in done_tasks:
-            # Cancellation needs to be checked first, otherwise the other methods
-            # could raise a CancelledError
-            if task.cancelled():
+            # BackgroundService returns a BaseExceptionGroup containing multiple
+            # exceptions. The 'task.result()' statement raises these exceptions,
+            # and 'except*' is used to handle them as a group. If the task raises
+            # multiple different exceptions, 'except*' will be invoked multiple times,
+            # once for each exception group.
+            try:
+                task.result()
+            except* asyncio.CancelledError:
                 _logger.info("Actor %s: Cancelled while running.", task.get_name())
-            elif exception := task.exception():
-                _logger.error(
+            except* BaseException:  # pylint: disable=broad-exception-caught
+                _logger.exception(
                     "Actor %s: Raised an exception while running.",
                     task.get_name(),
-                    exc_info=exception,
                 )
             else:
                 _logger.info("Actor %s: Finished normally.", task.get_name())
--- a/src/frequenz/sdk/actor/_run_utils.py
+++ b/src/frequenz/sdk/actor/_run_utils.py
@@ -49,7 +49,7 @@ async def run(*actors: Actor) -> None:
                 task.result()
             except* asyncio.CancelledError:
                 _logger.info("Actor %s: Cancelled while running.", task.get_name())
-            except* BaseException:  # pylint: disable=broad-exception-caught
+            except* Exception:  # pylint: disable=broad-exception-caught
                 _logger.exception(
                     "Actor %s: Raised an exception while running.",
                     task.get_name(),
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -19,3 +19,5 @@
 - Fixed bug with formulas raising exception when stopped.
 
 - Fix a bug that raised `CancelledError` when actor was started with `frequenz.sdk.actor.run` and stopped.
+
+- Stop catching `BaseException` in `frequenz.sdk.actor.run`. Only `CancelledError` and `Exception` are caught now.
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -17,3 +17,5 @@
 ## Bug Fixes
 
 - Fixed bug with formulas raising exception when stopped.
+
+- Fix a bug that raised `CancelledError` when actor was started with `frequenz.sdk.actor.run` and stopped.
--- a/tests/actor/test_actor.py
+++ b/tests/actor/test_actor.py
@@ -78,31 +78,6 @@ class RaiseExceptionActor(BaseTestActor):
         print(f"{self} done (should not happen)")
 
 
-class RaiseBaseExceptionActor(BaseTestActor):
-    """A faulty actor that raises a BaseException as soon as it receives a message."""
-
-    def __init__(
-        self,
-        recv: Receiver[int],
-    ) -> None:
-        """Create an instance.
-
-        Args:
-            recv: A channel receiver for int data.
-        """
-        super().__init__(name="test")
-        self._recv = recv
-
-    async def _run(self) -> None:
-        """Start the actor and crash upon receiving a message."""
-        print(f"{self} started")
-        self.inc_restart_count()
-        async for _ in self._recv:
-            print(f"{self} is about to crash")
-            raise MyBaseException("This is a test")
-        print(f"{self} done (should not happen)")
-
-
 ACTOR_INFO = ("frequenz.sdk.actor._actor", 20)
 ACTOR_ERROR = ("frequenz.sdk.actor._actor", 40)
 RUN_INFO = ("frequenz.sdk.actor._run_utils", 20)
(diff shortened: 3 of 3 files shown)


shsms:
Looks like the release notes were not cleared earlier after the previous release.  Could you base on top of this: https://github.com/frequenz-floss/frequenz-sdk-python/pull/1167

ela-kotulska-frequenz:
Done, thanks!

llucax (changes requested):
I think `wait()` is working as it should. `wait()` is used to implement `__await__` and `__await__` doesn't swallow `CancelledError`s in a `Task`, so I think we should be consistent with that.

In `BackgroundService` there is `stop()` though, which is basically:

```py
self.cancel()
try:
    await self.wait()
except CancelledError:
    pass
```

So there is already a convenience method to stop a background service without caring about cancellation. With `wait()` we can't assume the user doesn't want to be notified about `CancelledError`, as they might be `await`ing on the service without actually ever calling `cancel()`, so getting a `CancelledError` might be an actual unexpected situation/error.

ela-kotulska-frequenz:
Right! Thanks for checking.
It was problem with the method that waited for actor. I just made a fix :)

src/frequenz/sdk/actor/_run_utils.py
```diff
@@ -55,3 +56,13 @@ async def run(*actors: Actor) -> None:
                 _logger.info("Actor %s: Finished normally.", task.get_name())
 
     _logger.info("All %s actor(s) finished.", len(actors))
+
+
+def _was_cancelled(task: asyncio.Task[Any]) -> bool:
+    if task.cancelled():
```
llucax: ~Oh, interesting, so a task that stops because a cancellation error was raised in its code is not considered cancelled? So I guess cancelled will only be true if we call `.cancel()` to the internal task created only to await the actor finalization. If this is the case, `task.cancelled()` should never be true, as these tasks are internal to this function and we never call `.cancel()`. Maybe it is worth adding a note about this.~

---

[...after doing some experiments...]

No, that doesn't seem to be the case:

```py
import asyncio

async def cancelled_task() -> None:
    await asyncio.sleep(10)

async def wrapper_task() -> None:
    task = asyncio.create_task(cancelled_task())
    await asyncio.sleep(1)
    task.cancel()
    await task

async def main() -> None:
    pending = {asyncio.create_task(wrapper_task())}
    while pending:
        done, pending = await asyncio.wait(pending, return_when=asyncio.FIRST_COMPLETED)
        for task in done:
            if task.cancelled():
                print("Wrapper task was cancelled")
            elif exception := task.exception():
                match exception:
                    case asyncio.CancelledError:
                        print("Sub-task was cancelled")
                    case _:
                        print(f"Wrapper task raised an exception: {exception}")

asyncio.run(main())
```
This prints:
```
Wrapper task was cancelled
```

---

So the issue is the `BaseExceptionGroup` (or [exception groups](https://peps.python.org/pep-0654/) in general). When a `CancelledError` is wrapped in a group, then this exception is not properly translated to a wrapper task cancellation, and instead the task re-raises the whole group.

```py
import asyncio

async def cancelled_task() -> None:
    try:
        await asyncio.sleep(10)
    except asyncio.CancelledError as exc:
        raise BaseExceptionGroup("Sub-task was cancelled", [exc])

async def wrapper_task() -> None:
    task = asyncio.create_task(cancelled_task())
    await asyncio.sleep(1)
    task.cancel()
    await task

async def main() -> None:
    pending = {asyncio.create_task(wrapper_task())}
    while pending:
        done, pending = await asyncio.wait(pending, return_when=asyncio.FIRST_COMPLETED)
        for task in done:
            if task.cancelled():
                print("Wrapper task was cancelled")
            elif exception := task.exception():
                match exception:
                    case asyncio.CancelledError:
                        print("Sub-task was cancelled")
                    case _:
                        print(f"Wrapper task raised an exception: {exception!r}")

asyncio.run(main())
```
This prints:
```
Wrapper task raised an exception: BaseExceptionGroup('Sub-task was cancelled', [CancelledError()])                                                                                                                                                                                                                         
```

---

So to handle this properly, we need to use [`except*`](https://docs.python.org/3/reference/compound_stmts.html#except-star) clauses, so we can [split](https://docs.python.org/3/library/exceptions.html#BaseExceptionGroup.split) of the exception we want to catch as [subgroup](https://docs.python.org/3/library/exceptions.html#BaseExceptionGroup.subgroup)s.

```py
import asyncio
from collections.abc import Coroutine

async def cancelled_task() -> None:
    await asyncio.sleep(10)

async def cancelled_tasks() -> None:
    try:
        await asyncio.sleep(10)
    except asyncio.CancelledError as exc:
        raise BaseExceptionGroup(
            "Sub-tasks have errors",
            [
                exc,
                Exception("Some other exception"),
                asyncio.CancelledError("another CancelledError"),
            ],
        )

async def wrapper_task(coro: Coroutine[None, None, None]) -> None:
    task = asyncio.create_task(coro)
    await asyncio.sleep(1)
    task.cancel()
    await task

async def main() -> None:
    pending = {
        asyncio.create_task(wrapper_task(cancelled_task()), name="cancelled_task"),
        asyncio.create_task(wrapper_task(cancelled_tasks()), name="cancelled_tasks"),
    }
    while pending:
        done, pending = await asyncio.wait(pending, return_when=asyncio.FIRST_COMPLETED)
        for task in done:
            if task.cancelled():
                print(f"{task.get_name()} was cancelled")
            else:
                try:
                    await task
                except* asyncio.CancelledError as exc:
                    print(f"{task.get_name()} has some cancellations: {exc!r}")
                except* Exception as exc:
                    print(
                        f"Wrapper task for {task.get_name()} raised an exception: {exc!r}"
                    )

asyncio.run(main())
```
This prints:
```
cancelled_task was cancelled                                                                                                                                                                                                                                                                                               
cancelled_tasks has some cancellations: BaseExceptionGroup('Sub-tasks have errors', [CancelledError(), CancelledError('another CancelledError')])                                                                                                                                                                          
Wrapper task for cancelled_tasks raised an exception: ExceptionGroup('Sub-tasks have errors', [Exception('Some other exception')])                                                                                                                                                                                         
```

ela-kotulska-frequenz:
Wow... this was tricky. I didn't know about `except*`...

src/frequenz/sdk/actor/_run_utils.py
```diff
@@ -38,18 +38,22 @@ async def run(*actors: Actor) -> None:
...
+                await task
+            except* asyncio.CancelledError:
                 _logger.info("Actor %s: Cancelled while running.", task.get_name())
-            elif exception := task.exception():
+            except* BaseException as err:
                 _logger.error(
                     "Actor %s: Raised an exception while running.",
                     task.get_name(),
-                    exc_info=exception,
+                    exc_info=err,
```
llucax: You could change this with `_logger.exception()` now, as it is inside a `try/except`.

src/frequenz/sdk/actor/_run_utils.py
```diff
@@ -38,18 +38,22 @@ async def run(*actors: Actor) -> None:
...
+            # exceptions. The 'await' statement raises these exceptions, and 'except*'
+            # is used to handle them as a group. If the task raises multiple different
+            # exceptions, 'except*' will be invoked multiple times, once for each
+            # exception group.
+            try:
+                await task
+            except* asyncio.CancelledError:
                 _logger.info("Actor %s: Cancelled while running.", task.get_name())
-            elif exception := task.exception():
+            except* BaseException as err:  # pylint: disable=broad-exception-caught
```
llucax: This would be a change of behaviour, but I think we should only catch `Exception` here, we probably don't want to catch stuff like [`KeyboardInterrupt`](https://docs.python.org/3/library/exceptions.html#KeyboardInterrupt), [`GeneratorExit`](https://docs.python.org/3/library/exceptions.html#GeneratorExit) or [`SystemExit`](https://docs.python.org/3/library/exceptions.html#SystemExit). `CancelledError` is probably the only [`BaseException`](https://docs.python.org/3/library/exceptions.html#BaseException) we ever want to catch.

src/frequenz/sdk/actor/_run_utils.py
```diff
@@ -38,18 +38,22 @@ async def run(*actors: Actor) -> None:
...
-            # Cancellation needs to be checked first, otherwise the other methods
-            # could raise a CancelledError
-            if task.cancelled():
+            # BackgroundService returns a BaseExceptionGroup containing multiple
+            # exceptions. The 'await' statement raises these exceptions, and 'except*'
+            # is used to handle them as a group. If the task raises multiple different
+            # exceptions, 'except*' will be invoked multiple times, once for each
+            # exception group.
+            try:
+                await task
```
llucax: I'm thinking that in this case it might be better to just call [`task.result()`](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.result) so we avoid the `await` (entering the async loop) when we know the task is finished anyway and we don't need to `await` it.

src/frequenz/sdk/actor/_run_utils.py
ela-kotulska-frequenz: done

src/frequenz/sdk/actor/_run_utils.py
ela-kotulska-frequenz: I think so,too.  But we had dedicated tests for BaseException.
Removed in last commit.

src/frequenz/sdk/actor/_run_utils.py
ela-kotulska-frequenz: done

llucax (approved):
Oh, damn, so `BaseException` was not actually handled, just logged and bubbled up. That is fine too, but also not that important, as a base exception, if it is not only used for flow control (like exiting from the program), will probably be printed with a backtrace, so we should know it came from an actor.

So I think both approaches are OK (logging or not logging `BaseException`s), I'm approving and leaving it up to you.

Thanks! This was a very tricky bug.

llucax:
(disabled auto-merge in case you want to keep logging base exceptions, but after disabling it, I think the current approach is more correct, logging them will probably just be noise anyway)

ela-kotulska-frequenz:
No no, If I understand correctly:
Previously we handled BaseException , that means if there is code after `await run(*actors)`, then it was executed.
Now we raise BaseException. That means if there is code after `await run(*actors)` then it will not be executed.

And I think new approach is correct, because we stop as soon as something raise BaseException.

Issue #632, opened 2023-09-04
Make actor automatic restarting more resilient


llucax:
### What's needed?

The `Actor._run_loop()` method is restarting the actor if an exception in `_run()`. This is not ideal because if there is an unexpected exception, then the actor might be in a bad state, so it would be better to be able to restart it externally, so the state can be reset too.

### Proposed solution

~Move restart logic from `actor.Actor._run_loop()` to `actor.run()`, so even the task running the *run loop* can be re-created.~

~We probably will also need some `reset()` method, or some way to even re-create the whole actor from scratch, to make sure some bad state is not preserved between restarts.~

We already have quite a bit of code dealing with the state not being reset, which can actually simplify things when handling a large group of tasks and only one fails, then we don't need to re-create all the tasks that didn't have any issues. So most likely we will just add a `reset_status()` method for the user to override with any cleanup they need when an automatic restart is done.

### Additional context

This have a low risk of needing breaking changes, so if possible we should have a deeper look at it before 1.0 is released.

#### Related issues

The solution to this issue needs to have in mind the following related issues:

* #906 
* #826 
* #819

matthias-wende-frequenz:
Add actor heartbeat functionality. 

Each actor has, depending an a constructor flag, a heartbeat channel. If it doesn't write into it after a configurable time it gets externally restarted.

PR #1169, opened 2025-02-27
Add RestartActorException

ela-kotulska-frequenz:
Actor should be able to restart `_run` method
For example to update formula and start listening on new receiver.
Restarting actor should not print exception.

### DIFF
--- a/src/frequenz/sdk/actor/_actor.py
+++ b/src/frequenz/sdk/actor/_actor.py
@@ -13,6 +13,10 @@ from ._background_service import BackgroundService
 _logger = logging.getLogger(__name__)
 
 
+class RestartActorException(Exception):
+    """Exception raised when an actor should be restarted."""
+
+
 class Actor(BackgroundService, abc.ABC):
     """A primitive unit of computation that runs autonomously.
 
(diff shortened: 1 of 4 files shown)


src/frequenz/sdk/actor/_actor.py
```diff
@@ -58,19 +62,14 @@ def start(self) -> None:
     async def _run(self) -> None:
         """Run this actor's logic."""
 
-    async def _delay_if_restart(self, iteration: int) -> None:
-        """Delay the restart of this actor's n'th iteration.
-
-        Args:
-            iteration: The current iteration of the restart.
-        """
+    async def _delay(self) -> None:
```
llucax, line 65: Maybe make it a bit more explicit?

```suggestion
    async def _delay_restart(self) -> None:
```

llucax (commented):
I understand where this comes from, but I'm not sure if it is the path we want to follow in the future (https://github.com/frequenz-floss/frequenz-sdk-python/issues/632).

That said, this is how things work now and I can see how having a convenient way to restart is helpful, but an exception is also not the only way to restart, and have some implications. Mainly, **anyone, at any level in the call stack, can trigger an actor restart**.

An alternative that can prevent this, would be to use a `asyncio.Event` and await the event or the actor to finish, and then restart it if we got the event. Then adding the `Actor` class a `restart()` method that just triggers the event.

Even when Python already uses this exception mechanism for exits (like `SystemExit` or `GeneratorExit`), I'm not sure this is the right approach, because in those cases they are `BaseException`s and nobody is supposed to catch them (and body will, unless they explicitly catch `BaseException`),but with `RestartActorException`, it shouldn't inherit from `BaseException` (as per [Python docs](https://docs.python.org/3/library/exceptions.html#BaseException)), but if it doesn't, it also mean anyone using `except Exception`, will probably avoid restarting the actor (unwillingly), so it has its risks.

All that said, since restarting actors like this should probably change in the future, I'm not sure how important is to get this perfectly right, but unless we really need the exception, I think it would be safer to go for a `restart()` method with an event.

ela-kotulska-frequenz:
Closed, to discuss later how we should implement it

PR #1182, opened 2025-03-24
Upgrade to microgrid client v0.7

llucax:
Use the new microgrid client `ComponentId` and `MicrogridId`.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -39,6 +39,8 @@
       LoggingConfig(root_logger=RootLoggerConfig(level="ERROR"))
       ```
 
+* The SDK now depends on the `frequenz-client-microgrid` v0.7.x series. The main change is now all component and microgrid IDs need to be passed using the wrapper classes `ComponentId`/`MicrogridId` instead of `int`.
+
 ## New Features
 
 <!-- Here goes the main new features and examples or instructions on how to use them -->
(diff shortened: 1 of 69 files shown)


llucax:
Draft to avoid merging accidentally because it depends on https://github.com/frequenz-floss/frequenz-client-microgrid-python/pull/122 and a release, but it is really ready for a review.

llucax:
Ready for review now.

shsms (dismissed):
You could make a few more commits; it's fine if they don't pass the tests individually.  Or at least make fixup commits for review and squash before merging.  Otherwise, it is too hard to keep track of where you are if you have to take several breaks while reviewing.

llucax:
I can only see breaking the commit on a per-file basis, but that doesn't really bring much, if you want to resume the review after a while and splitting on a file-by-file basis is enough, then you can just use GitHub feature of marking files as read in the review.

image

If you think there was another way to split the commit, I'm interested!

shsms:
> I can only see breaking the commit on a per-file basis, but that doesn't really bring much, if you want to resume the review after a while and splitting on a file-by-file basis is enough, then you can just use GitHub feature of marking files as read in the review.

I see, didn't know that.  I'll try.

shsms:
Now an amd test is also failing :-|

llucax:
Oh dear, failing in Python 3.12 which I'm not using to test locally. Is the component graph tests, so I wonder if it is the `set[int]` ordering coming to bite us in a different way in 3.12 (maybe I missed some fix that worked in 3.11). I will need to have a look.

llucax:
Ready for another try at merging, thanks @shsms for debugging the test failures! :heart:

shsms (approved):

PR #1186, opened 2025-04-09
Update to repo-config v0.13

llucax:
We now also treat warnings as errors, although we had to make one exception because of the clean-up issues we have in the SDK.


llucax:
Related to:

* https://github.com/frequenz-floss/frequenz-sdk-python/issues/982

llucax:
I already switched PyPI back to publish from the `ci.yaml` workflow, so we should merge this before releasing.

Marenz (approved):

PR #1199, opened 2025-04-15
Don't stop pools

shsms:
When there are multiple pool instances controlling the same components and one of them stops, that would stop all the pools.

This PR adds an emergency fix to prevent this bug, to buy some time to come up with a better solution.

### DIFF
--- a/src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool.py
+++ b/src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool.py
@@ -217,7 +217,12 @@ class EVChargerPool:
 
     async def stop(self) -> None:
         """Stop all tasks and channels owned by the EVChargerPool."""
-        await self._pool_ref_store.stop()
+        # This was closing the pool_ref_store, which is not correct, because those are
+        # shared.
+        #
+        # This method will do until we have a mechanism to track the resources created
+        # through it.  It can also eventually cleanup the pool_ref_store, when it is
+        # holding the last reference to it.
 
     @property
     def _system_power_bounds(self) -> ReceiverFetcher[SystemBounds]:
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -14,4 +14,5 @@
 
 ## Bug Fixes
 
-<!-- Here goes notable bug fixes that are worth a special mention or explanation -->
+- Fixes a bug where battery pool metrics would stop for one actor when another actor managing the same components stops its pool.
+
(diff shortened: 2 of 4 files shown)


src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool.py
```diff
@@ -217,7 +217,6 @@ def power_distribution_results(self) -> ReceiverFetcher[_power_distributing.Resu
 
     async def stop(self) -> None:
         """Stop all tasks and channels owned by the EVChargerPool."""
-        await self._pool_ref_store.stop()
```
ela-kotulska-frequenz, line 220: Could you please add comment why this methods are empty?
It looks like bug now, because method description and definition does different thing.

ela-kotulska-frequenz (commented):
Wow but the formulas are shared, too.
So we have the same problem if 2 actors uses the same formula (like BatteryPool.power()) and one of them stops it.
And the same for `microgrid.producer().power` - one actor stops it, it won't work for other actors.

So this PR won't fix the problem :/
Maybe quick fix would be to not share pools & formulas? (I don't know how much work it would be).
Or have counter (how many is active and how many stopped

llucax:
:wave: reference counting :wave:

src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool.py
shsms: Done

llucax (approved):
Approving because I guess if you are making a PR with such a hack, you really need it urgently, but I think the release notes are a bit misleading. You fixed a bug by introducing a new bug (pools are not stopped) :laughing:

shsms:
> And the same for `microgrid.producer().power` - one actor stops it, it won't work for other actors.

This is not a problem.  Users only want to close the resources that they have created.  So when they make a `new_receiver()` and have a reference to the receiver, they will stop it.  The same is true with `new_battery_pool` because they are creating it.

We don't have the idiom of holding references to built-in formulas.  When people create custom formulas, they can stop those; that's fine.

Also, when formulas are closed by mistake, it's immediately noticeable. But when the battery pool is closed by mistake and we're only subscribing to SOC or capacity, for example, it's much more dangerous because we'll just assume the value hasn't changed.

> Maybe quick fix would be to not share pools & formulas?

Because there are so many metrics, that would quickly become a noticeable overhead.  Additionally, the pools share access to the power manager, etc.

I think we should just look for a proper solution, hopefully next week.

PR #1202, opened 2025-04-22
Use the additive `ShiftingMatryoshka` algorithm only for batteries

shsms:
And revert back to plain Matryoshka for PV and EV chargers.

### DIFF
--- a/src/frequenz/sdk/microgrid/_data_pipeline.py
+++ b/src/frequenz/sdk/microgrid/_data_pipeline.py
@@ -19,6 +19,8 @@ from datetime import timedelta
 from frequenz.channels import Broadcast, Sender
 from frequenz.client.microgrid import ComponentCategory, InverterType
 
+from frequenz.sdk.microgrid._power_managing._base_classes import Algorithm
+
 from .._internal._channels import ChannelRegistry
 from ..actor._actor import Actor
 from ..timeseries import ResamplerConfig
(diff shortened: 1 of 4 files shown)


llucax (approved):
Nice to see how simple this change was :tada:

src/frequenz/sdk/microgrid/_data_pipeline.py
```diff
@@ -103,16 +105,19 @@ def __init__(
...
+            power_manager_algorithm=Algorithm.MATRYOSHKA,
             component_category=ComponentCategory.EV_CHARGER,
         )
         self._pv_power_wrapper = PowerWrapper(
             self._channel_registry,
             api_power_request_timeout=api_power_request_timeout,
+            power_manager_algorithm=Algorithm.MATRYOSHKA,
             component_category=ComponentCategory.INVERTER,
             component_type=InverterType.SOLAR,
         )
```
ela-kotulska-frequenz, line 123: If I remember correctly, original Matryoshka algorithm had something like `operating point` adjustment.
How it works now?

ela-kotulska-frequenz (commented):
LGTM.
Just have one question.

src/frequenz/sdk/microgrid/_data_pipeline.py
llucax: That was like a partition, so all actors that were in the operating point group would shift the other actors or something like that. I think now is like the operating point group is always empty. But maybe @shsms can check I'm not mistaken.

shsms: That was mainly for battery use cases and is gone now.  We only have a single matryoshka instance for PV and EV chargers.

ela-kotulska-frequenz: OK, thanks a lot! :)

PR #1212, opened 2025-05-02
Reset components to default power when all proposals are withdrawn

shsms:
Previously, when all proposals for a set of components were withdrawn with `pool.propose_power(None)`, the power manager would just stop setting powers for those components, and the last set powers remain until they get reset by the API service or some other process.  This is not desirable, and instead the components should get set back to their respective default powers, according to the table below.  This is implemented in this PR.

| component category | default power                             |
|--------------------|-------------------------------------------|
| Battery            | 0.0                                       |
| PV                 | Minimum power (aka max production power)  |
| EV Chargers        | Maximum power (aka max consumption power) |

### DIFF
--- a/tests/timeseries/_ev_charger_pool/test_ev_charger_pool_control_methods.py
+++ b/tests/timeseries/_ev_charger_pool/test_ev_charger_pool_control_methods.py
@@ -24,7 +24,7 @@ from frequenz.sdk.microgrid._power_distributing._component_pool_status_tracker i
     ComponentPoolStatusTracker,
 )
 from frequenz.sdk.timeseries import ResamplerConfig, Sample3Phase
-from frequenz.sdk.timeseries.ev_charger_pool import EVChargerPool, EVChargerPoolReport
+from frequenz.sdk.timeseries.ev_charger_pool import EVChargerPoolReport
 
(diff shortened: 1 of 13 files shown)


tests/timeseries/_ev_charger_pool/test_ev_charger_pool_control_methods.py
```diff
@@ -140,9 +140,26 @@ async def _init_ev_chargers(self, mocks: _Mocks) -> None:
                 0.05,
             )
 
+    async def _recv_reports_until(
+        self,
+        bounds_rx: Receiver[EVChargerPoolReport],
+        check: typing.Callable[[EVChargerPoolReport], bool],
```
llucax, line 146: nitpick: `check` -> `predicate`? (oh, I see this is in tests, so nitpick³)

llucax (dismissed):

llucax:
One more thing, maybe this is worth mentioning in the docs too, in the part explaining the power manager and how proposing power works.

shsms:
> One more thing, maybe this is worth mentioning in the docs too, in the part explaining the power manager and how proposing power works.

True, I've added a new commit.

llucax (approved):

PR #1215, opened 2025-05-09
Allow create PVPool instances in locations without PV

### LINKED ISSUE #1130: Using PV pool in locations without PV cause error
### What happened?

```
2024-12-10T09:31:59+0100 ERROR    frequenz.sdk._internal._asyncio:50: Restarting after exception
  File "/frequenz-app-edge/.venv2/lib/python3.11/site-packages/frequenz/sdk/_internal/_asyncio.py", line 48, in run_forever
    await async_callable()
  File "/frequenz-app-edge/.venv2/lib/python3.11/site-packages/frequenz/sdk/timeseries/pv_pool/_system_bounds_tracker.py", line 108, in _run
    pv_data_rx = merge(
                 ^^^^^^
  File "/frequenz-app-edge/.venv2/lib/python3.11/site-packages/frequenz/channels/_merge.py", line 92, in merge
    raise ValueError("At least one receiver must be provided")
ValueError: At least one receiver must be provided
```

This was from a setup without PV inverters.

It shouldn't crash or cause errors. However, the `propose_power` method could print a warning when there are no components.

### Affected version(s)

_No response_

### Affected part(s)

### Extra information

_No response_

shsms:
Closes #1130

### DIFF
--- a/src/frequenz/sdk/timeseries/pv_pool/_pv_pool_reference_store.py
+++ b/src/frequenz/sdk/timeseries/pv_pool/_pv_pool_reference_store.py
@@ -97,15 +97,20 @@ class PVPoolReferenceStore:
             name=f"System Bounds for PV inverters: {component_ids}",
             resend_latest=True,
         )
-        self.bounds_tracker: PVSystemBoundsTracker = PVSystemBoundsTracker(
-            self.component_ids,
-            self.status_receiver,
-            self.bounds_channel.new_sender(),
-        )
-        self.bounds_tracker.start()
+
+        self.bounds_tracker: PVSystemBoundsTracker | None = None
+        # In locations without PV inverters, the bounds tracker will not be started.
+        if self.component_ids:
+            self.bounds_tracker = PVSystemBoundsTracker(
+                self.component_ids,
+                self.status_receiver,
+                self.bounds_channel.new_sender(),
+            )
+            self.bounds_tracker.start()
 
     async def stop(self) -> None:
         """Stop all tasks and channels owned by the PVInverterPool."""
         await self.formula_pool.stop()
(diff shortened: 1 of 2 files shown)


src/frequenz/sdk/timeseries/pv_pool/_pv_pool_reference_store.py
```diff
@@ -97,15 +97,20 @@ def __init__(  # pylint: disable=too-many-arguments
...
+                self.bounds_channel.new_sender(),
+            )
+            self.bounds_tracker.start()
 
     async def stop(self) -> None:
         """Stop all tasks and channels owned by the PVInverterPool."""
         await self.formula_pool.stop()
-        await self.bounds_tracker.stop()
+        if self.bounds_tracker is not None:
+            await self.bounds_tracker.stop()
```
llucax, line 115: Not introduced by this PR, but just FYI, as I mentioned in other insteances and at least if this is a `BackgroundService`, stopping of sub-services should be done separately via `cancel()` and `wait()`, and leave `stop()` being just an alias for `self.cancel(); await self`.

llucax (approved):
Approving as this, if it is really an issue, it is not introduced by this PR.

src/frequenz/sdk/timeseries/pv_pool/_pv_pool_reference_store.py
shsms: It is not a bug the way it is now, at least not in this layer, because this is in the `*reference_store`, which is supposed to be shared already.

But we will have to look at the outer layers, where we just removed the call to `*reference_store.stop()` as a temporary fix.

PR #1216, opened 2025-05-12
Ensure root component is as close as possible to the grid component

shsms:
The `find_first_descendant_component` accepts a `root_category` and finds
a component matching that category to begin traversing from.  This PR
ensures that the matching root component that's found is as close as
possible to the GRID component.

### DIFF
--- a/src/frequenz/sdk/microgrid/component_graph.py
+++ b/src/frequenz/sdk/microgrid/component_graph.py
@@ -311,16 +311,11 @@ class ComponentGraph(ABC):
     def find_first_descendant_component(
         self,
         *,
-        root_category: ComponentCategory,
         descendant_categories: Iterable[ComponentCategory],
     ) -> Component:
         """Find the first descendant component given root and descendant categories.
 
-        This method searches for the root component within the provided root
-        category. If multiple components share the same root category, the
-        first found one is considered as the root component.
-
-        Subsequently, it looks for the first descendant component from the root
+        This method looks for the first descendant component from the GRID
         component, considering only the immediate descendants.
 
         The priority of the component to search for is determined by the order
@@ -328,13 +323,12 @@ class ComponentGraph(ABC):
         highest priority.
 
         Args:
-            root_category: The class of the root component to search for.
             descendant_categories: The descendant classes to search for the first
                 descendant component in.
 
         Returns:
             The first descendant component found in the component graph,
-                considering the specified `root` and `descendants` categories.
+                considering the specified `descendants` categories.
         """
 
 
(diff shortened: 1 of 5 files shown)


src/frequenz/sdk/microgrid/component_graph.py
```diff
@@ -853,10 +853,31 @@ def find_first_descendant_component(
...
+                ),
+                None,
+            )
+            if grid_component is None:
+                raise ValueError("Grid component not found in component graph!")
+
+            root_component = self.find_first_descendant_component(
+                root_category=ComponentCategory.GRID,
+                descendant_categories=[root_category],
+            )
```
llucax: Alternative with less duplication for finding the grid component:

```suggestion
        grid_component = next(
            (
                comp
                for comp in self.components(
                    component_categories={ComponentCategory.GRID}
                )
            ),
            None,
        )
        if root_category == ComponentCategory.GRID:
            root_component = grid_component
        else:
            if grid_component is None:
                raise InvalidGraphError("Grid component not found in component graph")

            root_component = self.find_first_descendant_component(
                root_category=ComponentCategory.GRID,
                descendant_categories=[root_category],
            )
```

src/frequenz/sdk/microgrid/component_graph.py
```diff
@@ -853,10 +853,31 @@ def find_first_descendant_component(
...
+                (
+                    comp
+                    for comp in self.components(
+                        component_categories={ComponentCategory.GRID}
+                    )
+                ),
+                None,
+            )
+            if grid_component is None:
+                raise ValueError("Grid component not found in component graph!")
```
llucax: I would raise a `InvalidGraphError` as there is really nothing wrong with the values passed to the function, is just the component graph is not suitable for this operation, right?
```suggestion
                raise InvalidGraphError("Grid component not found in component graph!")
```

llucax (commented):
The other comments are optional, but I also wonder what happens with islands, we are supposed to support islands. Isn't there any way to find descendants if there is no grid connection?

shsms:
I'm actually wondering why we need to support this `root_category == Meter`, etc.  The root category is the GRID.  The two places where this function is being used are specifying `GRID` as the root component.

Unless you're against, I'll remove this parameter.  But if we keep it instead, we should reject everything that's not the GRID.  The rest are not "root", because there can be multiple of them.  But there can only be one GRID component in the graph.

shsms:
Ideally this function is replaced by a bfs, but I will not do that now.

src/frequenz/sdk/microgrid/component_graph.py
shsms: done

src/frequenz/sdk/microgrid/component_graph.py
shsms: dropped this section.

llucax (approved):

llucax:
> Ideally this function is replaced by a bfs, but I will not do that now.

There was some bfs implemented by @matthias-wende-frequenz somewhere, IIRC.

shsms:
> There was some bfs implemented by @matthias-wende-frequenz somewhere, IIRC.

No, he made a dfs

PR #1221, opened 2025-05-16
Update tests to support latest time-machine

shsms:
And migrate to latest time-machine version v2.16.0

This new version of `time-machine` mocks the monotonic clock of
asyncio.

This was causing the
`test_battery_status.py::TestBatteryStatus::test_async_battery_status`
test to fail, because `channels.Timer`s weren't working properly.

To fix this, this commit switches this test file to use an
`async_solipsism` event loop instead, which `time-machine` doesn't
affect.

But introducing the `async_solipsism` event loop caused the
`TestBatteryStatusRecovery::test_stale_data` test to fail, because it
needed the clock time to make progress, which wasn't happening because
`sleep` was returning immediately.

To fix this, this commit also updates that test to use `time-machine`
and moves the time forward manually, so the test can pass.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,7 +84,7 @@ dev-pytest = [
   "frequenz-repo-config[extra-lint-examples] == 0.13.3",
   "pytest-mock == 3.14.0",
   "pytest-asyncio == 0.26.0",
-  "time-machine == 2.12.0",
+  "time-machine == 2.16.0",
   "async-solipsism == 0.7",
   "hypothesis == 6.131.9",
 ]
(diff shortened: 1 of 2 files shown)


llucax (commented):
LGTM, but can you explain in the commit and/or PR what was going on exactly? Why this test hangs with the new version of time-machine?

shsms:
> LGTM, but can you explain in the commit and/or PR what was going on exactly? Why this test hangs with the new version of time-machine?

done

llucax (approved):

PR #1223, opened 2025-05-22
Fix issue with actor restarting instead of stopping

ela-kotulska-frequenz:
When an actor threw an exception during cancellation,
it restarted rather than stopping as expected.
Now it stops and raises and just prints exception.

### DIFF
--- a/src/frequenz/sdk/actor/_actor.py
+++ b/src/frequenz/sdk/actor/_actor.py
@@ -44,6 +44,15 @@ class Actor(BackgroundService, abc.ABC):
         This is mostly used for testing purposes and shouldn't be set in production.
     """
 
+    def __init__(self, *, name: str | None = None) -> None:
+        """Create actor instance.
+
+        Args:
+            name: The name of this background service.
+        """
+        super().__init__(name=name)
+        self._is_cancelled = False
+
     def start(self) -> None:
         """Start this actor.
 
@@ -51,6 +60,8 @@ class Actor(BackgroundService, abc.ABC):
         """
         if self.is_running:
             return
+
+        self._is_cancelled = False
         self._tasks.clear()
         self._tasks.add(asyncio.create_task(self._run_loop()))
 
@@ -94,6 +105,17 @@ class Actor(BackgroundService, abc.ABC):
                 _logger.info("Actor %s: Cancelled.", self)
                 raise
             except Exception:  # pylint: disable=broad-except
+                if self._is_cancelled:
+                    # If actor was cancelled, but any tasks have failed with an exception
+                    # other than asyncio.CancelledError, those exceptions are combined
+                    # in an ExceptionGroup or BaseExceptionGroup.
+                    # We have to handle that case separately to stop actor instead
+                    # of restarting it.
+                    _logger.exception(
+                        "Actor %s: Raised an unhandled exception during stop.", self
+                    )
+                    break
+
                 _logger.exception("Actor %s: Raised an unhandled exception.", self)
                 limit_str = "∞" if self._restart_limit is None else self._restart_limit
                 limit_str = f"({n_restarts}/{limit_str})"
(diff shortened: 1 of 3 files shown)


src/frequenz/sdk/actor/_actor.py
```diff
@@ -44,13 +44,31 @@ class Actor(BackgroundService, abc.ABC):
...
+
     def start(self) -> None:
         """Start this actor.
 
         If this actor is already running, this method does nothing.
         """
         if self.is_running:
             return
+
+        if self._is_cancelled:
```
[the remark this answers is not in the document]
ela-kotulska-frequenz: Done ! :)

src/frequenz/sdk/actor/_actor.py
```diff
@@ -44,13 +44,33 @@ class Actor(BackgroundService, abc.ABC):
...
+                be started again.
         """
         if self.is_running:
             return
+
+        if self._is_cancelled:
+            raise RuntimeError(
+                "Actor %s: was canceled and can't be re-run."
+                "Please create another actor instance.",
+            )
```
llucax: I wouldn't go for this at the actor level yet. We have many actors that assume they can be restarted and count on this. I would like to make a big overhaul of how we handle tasks and actors, I would avoid making any breaking changes until then as adapting all existing code to any breaking changes here will be very costly.

src/frequenz/sdk/actor/_actor.py
```diff
@@ -44,13 +44,33 @@ class Actor(BackgroundService, abc.ABC):
...
     """
 
+    def __init__(self, *, name: str | None = None) -> None:
+        """Create actor instance.
+
+        Args:
+            name: The name of this background service.
+        """
+        super().__init__(name=name)
+        self._is_cancelled = False
```
llucax, line 54: Can't we save the main actor task instead and use `self._main_task.cancelled()` as a single source of truth?

tests/actor/test_actor.py
```diff
@@ -53,6 +53,31 @@ async def _run(self) -> None:
...
+        self.inc_restart_count()
+        if BaseTestActor.restart_count == 1:
+            # This actor should not restart
+            # If it does we just return to avoid infinite await on `stop`
+            return
+        try:
+            async for msg in self._recv:
+                await self._sender.send(msg)
+        except asyncio.CancelledError as exc:
+            raise RuntimeError("Actor should stop.") from exc
```
llucax, line 78: I'm not sure we need to support this, this feels wrong, if a task is cancelled, it should not raise some other exception. Cancellation should always be supported and do nothing if it was already cancelled.

llucax (commented):
This looks sketchy to me, it looks like supporting a random exception to be raised during cancellation is just a fix for a but we just created by raising this exception.

tests/actor/test_actor.py
ela-kotulska-frequenz: > if a task is cancelled, it should not raise some other exception.

Yes but bugs happens and anything can raise exception. The problem is in this case we restart the actor and `await actor.stop()` never ends... :/

src/frequenz/sdk/actor/_actor.py
ela-kotulska-frequenz: Checked and in that case `self._main_task.cancelled()` returns False, because task was not cancelled successfully.

src/frequenz/sdk/actor/_actor.py
ela-kotulska-frequenz: removed

src/frequenz/sdk/actor/_actor.py
```diff
@@ -94,6 +114,12 @@ async def _run_loop(self) -> None:
                 _logger.info("Actor %s: Cancelled.", self)
                 raise
             except Exception:  # pylint: disable=broad-except
+                if self._is_cancelled:
```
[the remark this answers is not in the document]
ela-kotulska-frequenz: done

llucax (approved):

shsms (approved):

PR #1230, opened 2025-06-02
Update microgrid client

Marenz:
Blocked by:
* ~The release of base-client v0.11~
* ~The release of microgrid client~

PR #1235, opened 2025-06-17
Fix test typing

florian-wagner-frequenz:
Previously there was a test which violated type-hints. As we can not keep having the test _and_ type hints on that test, the test itself was rewritten to test a semantically slightly different but similar edge case.
The original test case is of course prevented when using type-hints (i.e not passing `None` to a function that doesn't accept `None`)

### DIFF
--- a/tests/config/test_util.py
+++ b/tests/config/test_util.py
@@ -54,13 +54,14 @@ def test_load_config_dataclass(
     [SimpleConfig, MmSimpleConfig],
     ids=["dataclass", "marshmallow_dataclass"],
 )
-def test_load_config_load_None(
+@pytest.mark.parametrize("config", [dict(), None], ids=["empty", "none"])
+def test_load_config_load_empty(
     config_class: type[SimpleConfig] | type[MmSimpleConfig],
+    config: dict[str, Any] | None,
 ) -> None:
-    """Test that load_config raises ValidationError if the configuration is None."""
-    config: dict[str, Any] = {}
+    """Test that load_config raises ValidationError if the configuration is empty."""
     with pytest.raises(marshmallow.ValidationError):
-        _ = load_config(config_class, config.get("loggers", None))
+        _ = load_config(config_class, config)  # type: ignore[arg-type]
 
 
 @pytest.mark.parametrize(


florian-wagner-frequenz:
Fixes the underlying problem preventing #1234 from proceeding

tests/config/test_util.py
```diff
@@ -54,13 +54,13 @@ def test_load_config_dataclass(
...
-def test_load_config_load_None(
+def test_load_config_load_empty(
     config_class: type[SimpleConfig] | type[MmSimpleConfig],
 ) -> None:
-    """Test that load_config raises ValidationError if the configuration is None."""
+    """Test that load_config raises ValidationError if the configuration is empty."""
     config: dict[str, Any] = {}
     with pytest.raises(marshmallow.ValidationError):
-        _ = load_config(config_class, config.get("loggers", None))
+        _ = load_config(config_class, config)
```
llucax: This was a very weird way to test for passing `None`, I wonder if we were consciously exploiting `mypy` failure to figure out `config.get("loggers", None)` can return `None` :laughing: 

In any case, you can also test if this does the right thing at runtime by adding a type ignore:

```py
_ = load_config(config_class, None)  # type: ignore[arg-type]
```

We could parametrize this to test both:

```py
@pytest.mark.parametrize(
    "config_class",
    [SimpleConfig, MmSimpleConfig],
    ids=["dataclass", "marshmallow_dataclass"],
)
@pytest.mark.parametrize("config", [{}, None], ids=["empty", "none"])
def test_load_config_load_empty(
    config_class: type[SimpleConfig] | type[MmSimpleConfig],
    config: dict[str, Any] | None,
) -> None:
    """Test that load_config raises ValidationError if the configuration is empty."""
    with pytest.raises(marshmallow.ValidationError):
        _ = load_config(config_class, config)  # type: ignore[arg-type]
```

RELEASE_NOTES.md
```diff
@@ -14,4 +14,4 @@
 
 ## Bug Fixes
 
-<!-- Here goes notable bug fixes that are worth a special mention or explanation -->
+* Fixed a bug where a test was testing a case explicitly forbidden by type hints
```
llucax: We should't include this in the release notes as it doesn't affect users, only our internal tests.

llucax (commented):
The suggested improvements to the tests are completely optional, not approving only because I think we should not add stuff that users don't care about in the release notes.

RELEASE_NOTES.md
florian-wagner-frequenz: Aright, my worry is that I will get an error from the CI then, but I can of course just force-merge.

tests/config/test_util.py
florian-wagner-frequenz: I would argue we shouldn't test things explicitly forbidden by type hints.
That said, we were testing for this case before, so I will add it.

RELEASE_NOTES.md
Marenz: If you add the label `cmd:skip-release-notes` it will not require release note changes

tests/config/test_util.py
Marenz: Unfortunately they are just hints ;)

florian-wagner-frequenz: I am aware, but if we start testing them, then a principled stance would be to test such things _everywhere_ and I don't see us testing for "None" arguments in a general fashion :wink:

RELEASE_NOTES.md
florian-wagner-frequenz: Ah, thank you

Marenz (dismissed):

llucax (dismissed):

Marenz:
> removed this pull request from the [merge queue](https://github.com/frequenz-floss/frequenz-sdk-python/queue/v1.x.x) due to no response for status checks [1 hour ago](https://github.com/frequenz-floss/frequenz-sdk-python/pull/1235#event-18207243497)

hmm maybe my fix https://github.com/frequenz-floss/frequenz-sdk-python/pull/1230/commits/45c4ad784449a1eb37b76b2917a9a9cead2ad280
wasn't a fix?

florian-wagner-frequenz:
Trying once more in the hope that this is spurious. Otherwise I would leave it to @Marenz to figure out what is going on

florian-wagner-frequenz:
Yup, your problem now @Marenz

llucax:
We believe the issue should be fixed by #1237, so you can rebase and try again now.

src/frequenz/sdk/actor/_actor.py
```diff
@@ -104,7 +104,16 @@ async def _run_loop(self) -> None:
             except asyncio.CancelledError:
                 _logger.info("Actor %s: Cancelled.", self)
                 raise
-            except Exception:  # pylint: disable=broad-except
+            except Exception as exc:  # pylint: disable=broad-except
+                if isinstance(exc, RuntimeError) and "no running event loop" in str(
+                    exc
+                ):
```
shsms: Maybe just a `loop.is_closed` check here, instead of string match?

Marenz: just calling `asyncio.get_running_eventloop()` will throw exactly the exception we're catching here.
We do it this way also in `run_forever`, I just copied it from luca

llucax (commented):
It looks to me that we have a more fundamental problem if this fixes the infinite loop issue. Both the actor and the battery status tracker loops should exit when there is a `CancelledError`, so it looks like we are forgetting to cancel at some point, or worse, someone is eating up the `CancelledError` in the middle.

This hack will uncover this more fundamental issue, which could come to bite us in other ways, so I would try to find the real root cause for this instead of patching it just to pass the tests.

llucax:
Also probably not the best idea to hijack this PR to fix the CI, why didn't you create a new PR like last time?

Marenz:
> Also probably not the best idea to hijack this PR to fix the CI, why didn't you create a new PR like last time?

Because this was easier to test, and well, it's working now.. it wasn't reliably reproducable with the other PR which is how we ended up in this situation in the first place ;)

> so I would try to find the real root cause for this

This is fixing the real root cause.
I wrote it on slack last week, but here is the analysis:

> I think I got it. So the run loop in _batter_status_tracker will never quit except for a cancelled exception. On shutdown, it will get the eventloop closed error and will try again..and again..and again.. and because the eventloop is closed, it's never giving up control to let other tasks properly clean up, including the task that would trigger the cancelexception for it (componentpool_status_tracker.py the _run method with the async exit stack context lib thing).

I didn't write it explicitly, but the exact same thing is also happening in the actor-restart part. The actor stops because of the RuntimeError: Eventloop is closed error makes it try to restart before any chancel exception can be done.

llucax:
> Because this was easier to test, and well, it's working now.. it wasn't reliably reproducable with the other PR which is how we ended up in this situation in the first place ;)

You could have created another PR with the same commits as this PR for testing. Now is like the original PR got completely lost in the noise, and we are also spamming @florian-wagner-frequenz unnecessarily :P

Also discussing via slack, but for the records, what I mean about root cause is that these loops should finish, before the event loop is closed, via a `CancelledError`. If this is not happening it means that something of these is happening:

* We are missing a `cancel()`
* We are missing an `await` on the cancelled task
* We are doing the above, but somehow after the loop was closed (in `__del__` for example).

We should find and fix when the loop is being closed without every task being properly stopped and awaited, instead of just hiding when it happens pretending that nothing is wrong... :scream:

Marenz:
> and we are also spamming florian-wagner-frequenz unnecessarily :P

Nah, he announced that he unsubscribed and gave over ownership to me :) (that is, until you highlighted him again now :P )
His part is reviewed and approved, nothing to get lost. 

But yes, I suppose for future historians this is a less optimal conflation of things that get fixed.

src/frequenz/sdk/_internal/_asyncio.py
```diff
@@ -46,12 +50,10 @@ async def run_forever(
...
-                    "Something went wrong, no running event loop, skipping execution of %s",
-                    async_callable.__name__,
-                )
+        except RuntimeError:
+            try:
+                asyncio.get_running_loop()
+            except RuntimeError:
                 return
         except Exception:  # pylint: disable=broad-except
             if not asyncio.get_event_loop().is_running():
```
llucax: This is actually wrong, `get_event_loop()` is deprecated in Python 3.12, `get_running_loop()` is recommended but then this function will raise if there is no running event loop. Apparently there is no reliable way to query if an event loop is running in the current context that is not calling `get_running_loop()` and catch the `RuntimeError`, so we should probably do that here too, or unify it with `except RuntimeError`, as catching `RuntimeError in particular doesn't help in any way if we are not inspecting the message (which I agree is less reliable).

All this said, and since this is just a safety net to avoid infinite loop in the presence of a **HUGE** bug (async code should never run outside of the context of an event loop), instead of returning I would try to crash the application with a backtrace. Maybe log the exception and then just simply `assert False`, or even `sys.exit(-1)` to ensure this `assert` is not simply swallowed by some dead task that didn't propagated the exception because it wasn't awaited. For me in this case crashing the complete thing makes sense because once there is no more event loop, we can't really do any work, it doesn't make sense to try to continue after knowing there is no loop anymore.

I think this could help us figuring out exactly which part of the code is entering an infinite loop.

src/frequenz/sdk/actor/_actor.py
```diff
@@ -104,7 +104,17 @@ async def _run_loop(self) -> None:
...
+                if isinstance(exc, RuntimeError):
+                    try:
+                        asyncio.get_running_loop()
+                    except RuntimeError:
+                        _logger.exception(
+                            "Something went wrong, no running event loop,"
+                            " not trying to restart %s again.",
+                            self,
+                        )
+                        break
```
llucax: Same here, `assert False` instead of `break`.

llucax (commented):
Looks good to me in general, but:

1. After you verified it works, please open a new PR with these fixes, these fixes are too important to be sneaked in a PR that was trying to fix something much more trivial.
2. It would be good to verify that this works even without handling the case where the loop is gone in `run_forever()` and `Actor._run_loop()`, but I guess if we make them crash instead of continuing as if nothing happened, that also works, as `pytest` will fail/error the tests that crashed.

Marenz (approved):

PR #1237, opened 2025-06-18
Attempt to fix ci-gets-stuck in shutdown loop 

Marenz:
In an attempt to find the dependency that causes the CI to get stuck (the _max test works so it must be a min dependency)

### DIFF
--- a/src/frequenz/sdk/_internal/_asyncio.py
+++ b/src/frequenz/sdk/_internal/_asyncio.py
@@ -46,7 +46,20 @@ async def run_forever(
     while True:
         try:
             await async_callable()
+        except RuntimeError as exc:
+            if "no running event loop" in str(exc):
+                _logger.exception(
+                    "Something went wrong, no running event loop, skipping execution of %s",
+                    async_callable.__name__,
+                )
+                return
         except Exception:  # pylint: disable=broad-except
+            if not asyncio.get_event_loop().is_running():
+                _logger.exception(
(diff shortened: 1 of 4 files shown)


src/frequenz/sdk/microgrid/_power_distributing/_component_status/_battery_status_tracker.py
```diff
@@ -262,7 +262,9 @@ async def _run(
         inverter = inverter_receiver
         set_power_result = set_power_result_receiver
 
-        while True:
+        # To prevent entering a shutdown-loop, we explicitly check if there is
+        # an eventloop running
+        while asyncio.get_event_loop().is_running():
```
llucax: Shouldn't this stop when there is an exception (like there is no loop)? Or is this being run in some other `run_forever()` or something like that?

Marenz: It's more about preventing it to start in the first place. The proper way is probably to do this check in select in the channels repo

llucax (approved):

shsms (approved):

src/frequenz/sdk/microgrid/_power_distributing/_component_status/_battery_status_tracker.py
llucax: But I think it is fine, if there is no loop an exception is raised in `select()`, the issue is we are blindly looping on broken code forever.

PR #1242, opened 2025-06-26
Improve resampler structure and performance

llucax:
This pull request splits the `resampling` module into many submodules as it was getting very large

If also uses tuples internally to store and calculate samples, which improves performance and simplifies the logic, as `Sample`s can have `None` values, but the values in the internal buffer can never be `None`, and we don't need to use a `Quantity` as we know we are always working with the same units.

It also exposes the `ResamplingFunction` and `SourceProperties` types, converts `ResamplingFunction` to a `Protocol`, and replaces `average` as the default resampling function with Python's `statistics.fmean`, among other small improvements and fixes.


llucax:
This is an extract of some assorted improvement that were in #999, plus a few new ones. It is a preparation for the final version of #999.

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -11,8 +11,9 @@
 
 import numpy as np
 import numpy.typing as npt
+from frequenz.core.datetime import UNIX_EPOCH
```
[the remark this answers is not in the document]
llucax: Not true :unamused:

benchmarks/timeseries/resampling.py
```diff
@@ -9,13 +9,13 @@
...
 
-from frequenz.sdk.timeseries import ResamplerConfig, Sample
+from frequenz.sdk.timeseries import ResamplerConfig
 from frequenz.sdk.timeseries._resampling._base_types import SourceProperties
 from frequenz.sdk.timeseries._resampling._resampler import _ResamplingHelper
 
 
 def nop(  # pylint: disable=unused-argument
-    samples: Sequence[Sample[Quantity]],
+    samples: Sequence[tuple[datetime, Quantity]],
```
Marenz: I wonder if a `NoNoneSample` type or so might be a nice abstraction here. I can imagine there are actually many places that would enjoy not having to check for `None` after they initially received the data. I can certainly use it in FCR as well.

For now its fine.

tests/timeseries/test_resampling.py
```diff
@@ -55,6 +54,12 @@ async def source_chan() -> AsyncIterator[Broadcast[Sample[Quantity]]]:
     await chan.close()
 
 
+def astuple(sample: Sample[Quantity]) -> tuple[datetime, float]:
```
Marenz: `astuple` splitted into the wrong commit I think?

Marenz (dismissed):
LGTM. `astuple` is added one commit later than the one that uses it, but not very important imo.

tests/timeseries/test_resampling.py
shsms: this appears to be an unused implementation. you're using it from `dataclasses` somewhere else.

src/frequenz/sdk/timeseries/_resampling/_config.py
```diff
@@ -134,7 +114,9 @@ class ResamplerConfig:
         passed to the resampling function.
     """
 
-    resampling_function: ResamplingFunction = average
+    resampling_function: ResamplingFunction = lambda samples, _, __: statistics.fmean(
+        [s[1] for s in samples]
```
shsms: according to the docs this can be an iterable: https://docs.python.org/3/library/statistics.html#statistics.fmean

```suggestion
        s[1] for s in samples

File: benchmarks/timeseries/resampling.py (hunk 2 above)
Comment (llucax) [reply benchmarks/timeseries/resampling.py]: Yeah, in any case here we want to get rid of even the `Quantity`, we shouldn't really use `Sample` either, at least until we can use `Sample[float]`.

If we go with the `None`/not `None` sample, maybe it can be `OptionalSample` (`None` values allowed) and `Sample` (can't have a `None` value) to follow Python's old `Optional` as an alias of `| None`. Another option, which I'm not sure if it is doable or not, is `Sample[Power]` vs `Sample[Power | None]`. I have the feeling we tried to make that work and failed, but maybe I'm just remembering wrongly.

File: tests/timeseries/test_resampling.py (hunk 3 above)
Comment (llucax) [reply tests/timeseries/test_resampling.py]: Oh, damn, I will have a look.

Comment (llucax) [reply tests/timeseries/test_resampling.py]: Ah, yeah, the commits are correct. I first used the "standard" `dataclasses.astuple` and then changed the implementation to convert to `float` too. At first I used a different name for the function, but then thought re-using `astuple` could make the diff smaller and, at the end, we want to convert to a tuple too, but maybe it is just too confusing, so I can rename the custom `astuple` to something less confusing.

**Comment by llucax** [comment]
Updated with fixup commits.

**Review by shsms — APPROVED** [review]

**Comment by llucax** [comment]
Now that @shsms approved, I will fixup and enable auto-merge, if you want to make any further comments you need to hurry @Marenz  :D (you can always disable auto-merge to get some extra time).

File: src/frequenz/sdk/timeseries/_resampling/_config.py [hunk 5]
```diff
@@ -134,7 +114,9 @@ class ResamplerConfig:
         passed to the resampling function.
     """
 
-    resampling_function: ResamplingFunction = average
+    resampling_function: ResamplingFunction = lambda samples, _, __: statistics.fmean(
```
cwasicki, line 118: Are we sure there are no NaNs?

llucax: Yes, no `NaN`s, no `None`s, ~no `inf`s~. That's filtered out before they are fed to the resampling function.

llucax: Sorry, `inf` is still allowed:

https://github.com/frequenz-floss/frequenz-sdk-python/blob/c1bc91717334c4c84cd22955fd3e56bf23780dd2/src/frequenz/sdk/timeseries/_resampling/_resampler.py#L512-L514

PR #1245, opened 2025-07-14
Log reason for  service stop

ela-kotulska-frequenz:
Actor has `info` log because this log is important to see workflow. Log is printed only if `msg` is not None, to avoid log duplication (actor prints that it is cancelled in run method - and this log was useful many times.)

Background service has debug log, because `info` would log many misleading logs during normal runtime, which might be confusing. And those logs are only useful during debugging.

### DIFF
--- a/src/frequenz/sdk/actor/_actor.py
+++ b/src/frequenz/sdk/actor/_actor.py
@@ -153,5 +153,8 @@ class Actor(BackgroundService, abc.ABC):
         Args:
             msg: The message to be passed to the tasks being cancelled.
         """
+        if msg is not None:
+            _logger.info("Actor %s cancelled, reason: %s", self, msg)
+
         self._is_cancelled = True
         return super().cancel(msg)
--- a/src/frequenz/sdk/actor/_background_service.py
+++ b/src/frequenz/sdk/actor/_background_service.py
@@ -6,8 +6,11 @@
 import abc
 import asyncio
 import collections.abc
+import logging
 from types import TracebackType
 from typing import Any, Self
+
+_logger = logging.getLogger(__name__)
 
 
 class BackgroundService(abc.ABC):
@@ -167,6 +170,10 @@ class BackgroundService(abc.ABC):
         Args:
             msg: The message to be passed to the tasks being cancelled.
         """
+        _logger.debug(
+            "Service %s cancelled%s", self, f": {msg}" if msg is not None else ""
+        )
+
         for task in self._tasks:
             task.cancel(msg)
 


ela-kotulska-frequenz:
I need to fix the test, but PR is ready to review

src/frequenz/sdk/actor/_actor.py
```diff
@@ -153,5 +153,7 @@ def cancel(self, msg: str | None = None) -> None:
         Args:
             msg: The message to be passed to the tasks being cancelled.
         """
+        if msg is not None:
```
Marenz, line 156: Isn't this calling the super method and then its logged twice?

ela-kotulska-frequenz: In Actor has info log.
In Backgroud has debug log.

Having info log in BackgroundService will create lots of unnecessary logs (that we probably don't need).


Do you have different idea?

Marenz: Ah, well in this case it seems fine then. Didn't notice the different levels

Marenz (dismissed):

src/frequenz/sdk/actor/_actor.py
shsms: Cancel sounds like a rare event, and it's useful to know. INFO doesn't trigger any alerts. It wouldn't be too bad if we just had INFO in the background service.

src/frequenz/sdk/actor/_background_service.py
```diff
@@ -167,6 +170,8 @@ def cancel(self, msg: str | None = None) -> None:
         Args:
             msg: The message to be passed to the tasks being cancelled.
         """
+        _logger.info("Service %s cancelled, reason: %s", self.name, msg)
```
llucax: Maybe it would be a bit easier to the eyes like this?

```suggestion
        _logger.info("Service %s cancelled%s", self.name, msg, f": {msg}" if msg is not None else "")
```

llucax (commented):
I'm not sure if this should be info or debug also, it seems like it might be too fine grained detail for a normal run, as we are already logging when it actually stops, so cancel would only be interesting if there is a bug and the cancel is not resulting in a stop.

src/frequenz/sdk/actor/_actor.py
llucax: Oh, I didn't see this comment. For me this still sounds like a debug, because I guess it is only useful when debugging, when things are working you should see an INFO for the actor stopping, and that should be enough. That said, I'm OK to merge if there is enough consensus.

src/frequenz/sdk/actor/_actor.py
```diff
@@ -153,5 +153,8 @@ def cancel(self, msg: str | None = None) -> None:
...
+            _logger.info("Actor %s cancelled, reason: %s", self, msg)
```
shsms, line 157: looks like this came back somehow.

ela-kotulska-frequenz:
Actor has `info` log because this log is important to see workflow. Log is printed only if `msg` is not None, to avoid log duplication (actor prints that it is cancelled in run method - and this log was useful many times.)

Background service has debug log, because `info` would log many misleading logs during normal runtime, which might be confusing. And those logs are only useful during debugging.

src/frequenz/sdk/actor/_actor.py
ela-kotulska-frequenz: Actor has info log because this log is important to see workflow. Log is printed only if msg is not None, to avoid log duplication (actor prints that it is cancelled in run method - and this log was useful many times.)

Background service has debug log, because info would log many misleading logs during normal runtime, which might be confusing. And those logs are only useful during debugging.

llucax: I wouldn't put the if, I think it might be more confusing, as one might expect seeing a log entry for all cancelled operations once you saw one in the logs, you would have to be very involved in the code to know it is only logged if there is a message.

I feel pretty strongly about this one as I think it could do real harm, so for me it is either always or never.

Then again, I would prefer not to have a special case in actor with info, but I can live with that one.

llucax:
> Actor has `info` log because this log is important to see workflow. Log is printed only if `msg` is not None, to avoid log duplication (actor prints that it is cancelled in run method - and this log was useful many times.)

Can you give some examples? I'm still very doubtful of adding a conditional log (only logging if there is a message) as it could be misleading, specially during a stressful debugging session. Is it really a pattern we use that we cancel without a message when we want stop something but use a message when something else happens? I can hardly remember any case where we use a cancellation message (unless it is passed automatically by `stop()` for example).

PR #1249, opened 2025-07-30
Add a wall clock attached timer

llucax:
This timer uses the wall clock to trigger ticks and handles discrepancies between the wall clock and monotonic time. Since sleeping is performed using monotonic time, differences between the two clocks can occur.

When the wall clock progresses slower than monotonic time, it is referred to as *compression* (wall clock time appears in the past relative to monotonic time). Conversely, when the wall clock progresses faster, it is called *expansion* (wall clock time appears in the future relative to monotonic time). If these differences exceed a configured threshold, a warning is emitted.

If the difference becomes excessively large, it is treated as a *time jump*. Time jumps can occur, for example, when the wall clock is adjusted by NTP after being out of sync for an extended period. In such cases, the timer resynchronizes with the wall clock and triggers an immediate tick.

### DIFF
--- a/src/frequenz/sdk/timeseries/_resampling/_wall_clock_timer.py
+++ b/src/frequenz/sdk/timeseries/_resampling/_wall_clock_timer.py
@@ -1,0 +1,849 @@
+# License: MIT
+# Copyright © 2025 Frequenz Energy-as-a-Service GmbH
+
+"""A timer attached to the wall clock for the resampler."""
+
+from __future__ import annotations
+
+import asyncio
+import logging
+import math
+from collections.abc import Sequence
+from dataclasses import dataclass
+from datetime import datetime, timedelta, timezone
+from typing import Literal, Self, assert_never
+
+from frequenz.channels import Receiver, ReceiverStoppedError
+from frequenz.core.datetime import UNIX_EPOCH
+from typing_extensions import override
+
+_logger = logging.getLogger(__name__)
+
+_TD_ZERO = timedelta()
+
+
+@dataclass(frozen=True, kw_only=True)
+class WallClockTimerConfig:
+    """Configuration for a [wall clock timer][frequenz.sdk.timeseries.WallClockTimer]."""
+
+    align_to: datetime | None = UNIX_EPOCH
+    """The time to align the timer to.
+
+    The first timer tick will occur at the first multiple of the timer's interval after
+    this value.
+
+    It must be a timezone aware `datetime` or `None`. If `None`, the timer aligns to the
+    time it is started.
+    """
+
+    async_drift_tolerance: timedelta | None = None
+    """The maximum allowed difference between the requested and the real sleep time.
+
+    The timer will emit a warning if the difference is bigger than this value.
+
+    It must be bigger than 0 or `None`. If `None`, no warnings will ever be emitted.
+    """
+
+    wall_clock_drift_tolerance_factor: float | None = None
+    """The maximum allowed relative difference between the wall clock and monotonic time.
+
+    The timer will emit a warning if the relative difference is bigger than this value.
+    If the difference remains constant, the warning will be emitted only once, as the
+    previous drift is taken into account. If there is information on the previous drift,
+    the previous and current factor will be used to determine if a warning should be
+    emitted.
+
+    It must be bigger than 0 or `None`. If `None`, no warnings will be ever emitted.
+
+    Info:
+        The calculation is as follows:
+
+        ```py
+        tolerance = wall_clock_drift_tolerance_factor
+        factor = monotonic_elapsed / wall_clock_elapsed
+        previous_factor = previous_monotonic_elapsed / previous_wall_clock_elapsed
+        if abs(factor - previous_factor) > tolerance:
+            emit warning
+        ```
+
+        If there is no previous information, a `previous_factor` of 1.0 will be used.
+    """
+
+    wall_clock_jump_threshold: timedelta | None = None
+    """The amount of time that's considered a wall clock jump.
+
+    When the drift between the wall clock and monotonic time is too big, it is
+    considered a time jump and the timer will be resynced to the wall clock.
+
+    This value determines how big the difference needs to be to be considered a
+    jump.
+
+    Smaller values are considered wall clock *expansions* or *compressions* and are
+    always gradually adjusted, instead of triggering a resync.
+
+    Must be bigger than 0 or `None`. If `None`, a resync will never be triggered due to
+    time jumps.
+    """
+
+    def __post_init__(self) -> None:
+        """Check that config values are valid.
+
+        Raises:
+            ValueError: If any value is out of range.
+        """
+        if self.align_to is not None and self.align_to.tzinfo is None:
+            raise ValueError(
+                f"align_to ({self.align_to}) should be a timezone aware datetime"
+            )
+
+        def _is_strictly_positive_or_none(value: float | timedelta | None) -> bool:
+            match value:
+                case None:
+                    return True
+                case timedelta() as delta:
+                    return delta > _TD_ZERO
+                case float() as num:
+                    return math.isfinite(num) and num > 0.0
+                case int() as num:
+                    return num > 0
+                case _ as unknown:
+                    assert_never(unknown)
+
+        if not _is_strictly_positive_or_none(self.async_drift_tolerance):
+            raise ValueError(
+                "async_drift_tolerance should be positive or None, not "
+                f"{self.async_drift_tolerance!r}"
+            )
+        if not _is_strictly_positive_or_none(self.wall_clock_drift_tolerance_factor):
+            raise ValueError(
+                "wall_clock_drift_tolerance_factor should be positive or None, not "
+                f"{self.wall_clock_drift_tolerance_factor!r}"
+            )
+        if not _is_strictly_positive_or_none(self.wall_clock_jump_threshold):
+            raise ValueError(
+                "wall_clock_jump_threshold should be positive or None, not "
+                f"{self.wall_clock_jump_threshold!r}"
+            )
+
+    @classmethod
+    def from_interval(  # pylint: disable=too-many-arguments
+        cls,
+        interval: timedelta,
+        *,
+        align_to: datetime | None = UNIX_EPOCH,
+        async_drift_tolerance_factor: float = 0.1,
+        wall_clock_drift_tolerance_factor: float = 0.1,
+        wall_clock_jump_threshold_factor: float = 1.0,
+    ) -> Self:
+        """Create a timer configuration based on an interval.
+
+        This will set the tolerance and threshold values proportionally to the interval.
+
+        Args:
+            interval: The interval between timer ticks. Must be bigger than 0.
+            align_to: The time to align the timer to. See the
+                [`WallClockTimer`][frequenz.sdk.timeseries.WallClockTimer] documentation
+                for details.
+            async_drift_tolerance_factor: The maximum allowed difference between the
+                requested and the real sleep time, relative to the interval.
+                `async_drift_tolerance` will be set to `interval * this_factor`.  See
+                the [`WallClockTimer`][frequenz.sdk.timeseries.WallClockTimer]
+                documentation for details.
+            wall_clock_drift_tolerance_factor: The maximum allowed relative difference
+                between the wall clock and monotonic time. See the
+                [`WallClockTimer`][frequenz.sdk.timeseries.WallClockTimer] documentation
+                for details.
+            wall_clock_jump_threshold_factor: The amount of time that's considered a
+                wall clock jump, relative to the interval. This will be set to
+                `interval * this_factor`. See the
+                [`WallClockTimer`][frequenz.sdk.timeseries.WallClockTimer] documentation
+                for details.
+
+        Returns:
+            The created timer configuration.
+
+        Raises:
+            ValueError: If any value is out of range.
+        """
+        if interval <= _TD_ZERO:
+            raise ValueError(f"interval must be bigger than 0, not {interval!r}")
+
+        return cls(
+            align_to=align_to,
+            wall_clock_drift_tolerance_factor=wall_clock_drift_tolerance_factor,
+            async_drift_tolerance=interval * async_drift_tolerance_factor,
+            wall_clock_jump_threshold=interval * wall_clock_jump_threshold_factor,
+        )
+
+
+@dataclass(frozen=True, kw_only=True)
+class ClocksInfo:
+    """Information about the wall clock and monotonic clock and their drift.
+
+    The `monotonic_requested_sleep` and `monotonic_elapsed` values must be strictly
+    positive, while the `wall_clock_elapsed` can be negative if the wall clock jumped
+    back in time.
+    """
+
+    monotonic_requested_sleep: timedelta
+    """The requested monotonic sleep time used to gather the information (must be positive)."""
+
+    monotonic_time: float
+    """The monotonic time right after the sleep was done."""
+
+    wall_clock_time: datetime
+    """The wall clock time right after the sleep was done."""
+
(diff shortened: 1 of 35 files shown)


src/frequenz/sdk/timeseries/_resampling/_wall_clock_timer.py
```diff
@@ -0,0 +1,846 @@
...
+        # too, so it shouldn't change the logical "frozenness" of the class.
+        if math.isnan(self.wall_clock_factor):
+            wall_clock_elapsed = self.wall_clock_elapsed
+            if wall_clock_elapsed <= _TD_ZERO:
+                _logger.warning(
+                    "The monotonic clock advanced %s, but the wall clock stayed still or "
+                    "jumped back (elapsed: %s)! Hopefully this was just a singular jump in "
+                    "time and not a permanent issue with the wall clock not moving at all. "
+                    "For purposes of calculating the wall clock factor, a fake elapsed time "
+                    "of one tenth of the elapsed monotonic time will be used.",
```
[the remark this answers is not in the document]
llucax: Stayed still means "did not advance". Is this not clear for any humans out there?

src/frequenz/sdk/timeseries/_resampling/_wall_clock_timer.py
```diff
@@ -0,0 +1,846 @@
...
+                    self.monotonic_elapsed,
+                    wall_clock_elapsed,
+                )
+                wall_clock_elapsed = self.monotonic_elapsed * 0.1
```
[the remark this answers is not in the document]
llucax: Damn, I had a comment about it, I think I lost it after a refactor. Will comment on it again.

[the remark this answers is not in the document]
llucax: Oh, right, it is in the warning message. Not sure if it needs further clarification as a code comment, I think it is pretty self explanatory.

tests/timeseries/_resampling/wall_clock_timer/util.py
```diff
@@ -0,0 +1,198 @@
...
+    object.__setattr__(
+        approx_tick_info,
+        "expected_tick_time",
+        approx_time(expected.expected_tick_time, abs=abs),
+    )
+    object.__setattr__(approx_tick_info, "sleep_infos", approx_sleeps)
+
+    return approx_tick_info
+
+
```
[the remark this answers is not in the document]
llucax: I tried cleaner approaches but I don't think there is a way around it, using `replace()` is not an option because the `__postinit__()` will run too. We really need to bypass checks to be able to replace floats with `pytest.approx()` objects.

src/frequenz/sdk/timeseries/_resampling/_wall_clock_timer.py
```diff
@@ -0,0 +1,846 @@
...
+    def next_tick_time(self) -> datetime | None:
+        """The wall clock time when the next tick should happen, or `None` if it is not running."""
+        return None if self._closed else self._next_tick_time
+
+    def reset(self) -> None:
+        """Reset the timer to start timing from now (plus an optional alignment).
+
+        If the timer was closed, or not started yet, it will be started.
+        """
+        self._closed = False
```
shsms, line 469: this is interesting, that we can now reopen closed receivers.

src/frequenz/sdk/timeseries/_resampling/_wall_clock_timer.py
```diff
@@ -0,0 +1,846 @@
...
+        iterator interface will raise a
+        [`ReceiverStoppedError`][frequenz.channels.ReceiverStoppedError].
+
+        You can restart the timer with `reset()`.
+        """
+        self._closed = True
+        # We need to make sure it's not None, otherwise `ready()` will start it
+        self._next_tick_time = datetime.now(timezone.utc)
+
+    def _should_resync(self, info: ClocksInfo | timedelta | None) -> bool:
```
shsms, line 491: why not just call with `clocks_info.wall_clock_jump` and type `info` as just `timedelta | None`, instead of doing runtime type checks with `isinstance` in this function?

src/frequenz/sdk/timeseries/_resampling/_wall_clock_timer.py
```diff
@@ -0,0 +1,846 @@
...
+        while wall_clock_time_to_next_tick > _TD_ZERO:
+            prev_clocks_info = self._clocks_info
+            # We don't assign directly to self._clocks_info because its type is
+            # ClocksInfo | None, and sleep() returns ClocksInfo, so we can avoid some
+            # None checks further in the code with `clocks_info` (and we make the code
+            # more succinct).
+            clocks_info = await self._sleep(
+                wall_clock_time_to_next_tick, prev_clocks_info=prev_clocks_info
+            )
+            should_resync = self._should_resync(clocks_info)
```
shsms, line 564: It looks like you're calling `_should_resync` with the same arguments at the end of the previous `_sleep` call.  It might be better to put that in `clocks_info`, so you don't have to call the same function again.

src/frequenz/sdk/timeseries/_resampling/_wall_clock_timer.py
```diff
@@ -0,0 +1,846 @@
...
+                    "time and not a permanent issue with the wall clock not moving at all. "
+                    "For purposes of calculating the wall clock factor, a fake elapsed time "
+                    "of one tenth of the elapsed monotonic time will be used.",
+                    self.monotonic_elapsed,
+                    wall_clock_elapsed,
+                )
+                wall_clock_elapsed = self.monotonic_elapsed * 0.1
+            super().__setattr__(
+                "wall_clock_factor", self.monotonic_elapsed / wall_clock_elapsed
+            )
```
shsms, line 256: why the setattr?

src/frequenz/sdk/timeseries/_resampling/_wall_clock_timer.py
```diff
@@ -0,0 +1,846 @@
...
+        ```py
+        tolerance = wall_clock_drift_tolerance_factor
+        factor = monotonic_elapsed / wall_clock_elapsed
+        previous_factor = previous_monotonic_elapsed / previous_wall_clock_elapsed
+        if abs(factor - previous_factor) > tolerance:
+            emit warning
+        ```
+
+        If there is no previous information, a `previous_factor` of 1.0 will be used.
+    """
```
shsms, line 70: maybe good to mention this needs to be lower than `wall_clock_jump_threshold` and reverse on the other one?

shsms (commented):
Finished the timer commit.  Will continue with the tests.

src/frequenz/sdk/timeseries/_resampling/_wall_clock_timer.py
llucax: Not sure if this is true for all receivers (I don't think we have a standard way to do it in the base `Receiver`, but the `Timer` in channels could always also be *reopened* (before it was restarted, so the timer could be stopped and then started again). This timer was based on `Timer` so it has the same *feature*, is not that it is something introduced in this PR.

src/frequenz/sdk/timeseries/_resampling/_wall_clock_timer.py
llucax: Mmm, maybe, but not 100% convinced as `ClocksInfo` is part of the public interface, so either we provide a very niche function that users are not really intended to use or we need to keep it private and then add `# type: ignore` or similar to ask linters not to complain about using private methods.

src/frequenz/sdk/timeseries/_resampling/_wall_clock_timer.py
llucax: Yeah, good catch, I can also check for this in the `__post_init__` and raise otherwise.

src/frequenz/sdk/timeseries/_resampling/_wall_clock_timer.py
llucax: Because it is a frozen dataclass, I can add a comment to clarify.

src/frequenz/sdk/timeseries/_resampling/_wall_clock_timer.py
llucax: I think mainly because because I use `self._should_resync(self._clocks_info)` and on the first run `self.clocks_info` could be `None` itself, so if I remove `ClocksInfo` from here, I would have to do `self._should_resync(None if self._clocks_info is None else self._clocks_info.wall_clock_jump)`, or `should_resync: bool = False ifself._clocks_info is None else self._should_resync(self._clocks_info.wall_clock_jump)`, but I feel in this case is sort of putting part of the logic of `_should_resync` in here. 🤷 

A `isinstance` check is probably more expensive than comparing to `None` though, so that might be one reason to change it, let me know if you still feel it is worth changing.

src/frequenz/sdk/timeseries/_resampling/_wall_clock_timer.py
llucax: I will also adjust the assumed `previous_factor` of 1.0 if we don't have info to be smaller than `wall_clock_jump_threshold`.

llucax: Oh, wait, no, this doesn't make a lot of sense. The factor is a relative measure and the threshold is absolute, so there is no easy way to check for this or enforce it. What the code does is *prioritizing* jumps, so if we detect a *jump* we ignore the drift factor and consider it a jump, not a drift.

This is somehow clarified in the `wall_clock_jump_threshold` docs:

```
    When the drift between the wall clock and monotonic time is too big, it is
    considered a time jump and the timer will be resynced to the wall clock.
```

And in `WallClockTimer` class docs:

```
    When the wall clock progresses slower than monotonic time, it is referred to as
    *compression* (wall clock time appears in the past relative to monotonic time).
    Conversely, when the wall clock progresses faster, it is called *expansion*
    (wall clock time appears in the future relative to monotonic time). If these
    differences exceed a configured threshold, a warning is emitted. The threshold
    is defined by the
    [`wall_clock_drift_tolerance_factor`][frequenz.sdk.timeseries.WallClockTimerConfig.wall_clock_drift_tolerance_factor].

    If the difference becomes excessively large, it is treated as a *time jump*.
    Time jumps can occur, for example, when the wall clock is adjusted by NTP after
    being out of sync for an extended period. In such cases, the timer resynchronizes
    with the wall clock and triggers an immediate tick. The threshold for detecting
    time jumps is controlled by the
    [`wall_clock_jump_threshold`][frequenz.sdk.timeseries.WallClockTimerConfig.wall_clock_jump_threshold].
```

Do you think it needs further clarification?

llucax:
> Finished the timer commit. Will continue with the tests.

I just finished going through the comments on your initial review.

About tests, there is a bit of black magic to be able to get nice pytest output. Most of it was done by AI, and I tried to make it as clean as possible but also didn't spend a lot of time on cleaning it and making it super nice, as it is just a testing tool. I think I mentioned this somewhere else, but letting you know just in case.

src/frequenz/sdk/timeseries/_resampling/_wall_clock_timer.py
shsms: Another option with less glitter:

```python
should_rsync = False
if self._clock_info is not None:
    should_rsync = self._should_rsync(self._clocks_info.wall_clock_jump)
```

but I get the modularity issue, so if you're rather not do it, the current approach is fine of course.

src/frequenz/sdk/timeseries/_resampling/_wall_clock_timer.py
shsms: ok, that makes sense, so they are independent things.

src/frequenz/sdk/timeseries/_resampling/_wall_clock_timer.py
llucax: Alright, then I would keep it. I'm happy to refactor/change if there is a clear advantage, but in cases like that this that it is not entirely clear, I would rather take the lazy option and keep it like it is :innocent:

tests/timeseries/_resampling/wall_clock_timer/util.py
```diff
@@ -223,3 +232,175 @@ def __eq__(self, other: object) -> bool:
...
+    mono_delta: timedelta | float
+    """The monotonic time delta to sleep before adjusting the wall clock time, in seconds."""
+
+    wall_time: datetime | float
+    """The new wall clock time to set at that monotonic time, in seconds since the epoch in UTC."""
+
+
+@dataclass(kw_only=True, frozen=True)
+class TimeDriver:
```
shsms, line 252: I wonder if this couldn't have been implemented by using hypothesis and timemachine instead.

shsms (commented):
Need to continue from " Add ticking behavior tests "

tests/timeseries/_resampling/wall_clock_timer/util.py
llucax: Not sure how hypothesis could be used for this, but I made some failed attempt to use `timemachine`, but there is this issue when using solipsism too that the handling of the monotonic clock conflicts. It was also very convenient to be able to only track time changes inside the module being tested, and not the sleeps inside the tests themselves.

I think another aspect where I needed more control and I think this approach helped is the fact that I need the timer to enter the sleep and *while it is sleeping* update the wall clock to simulate the time jumps and/or drifting. And also `timemachine` will manipulate both the monotonic time and wall clock time in sync, if you freeze the time both are frozen and if you advance it also both advanced, and I need to control them independently.

Take all of this with a grain of salt, as it took me quite some time to come up with an approach that worked and was *easy* to work with, so I might be mixing things up or having wrong recollections.

shsms: > Not sure how hypothesis could be used for this

oh I meant solipsism.

The rest of the stuff makes sense.

shsms:
Will pick up from "Add custom comparison for TickInfo objects for pytest" tomorrow.

tests/timeseries/_resampling/wall_clock_timer/util.py
llucax: I am using solipsism.

tests/timeseries/_resampling/wall_clock_timer/conftest.py
```diff
@@ -49,3 +50,106 @@ async def time_driver(datetime_mock: MagicMock) -> TimeDriver:
...
+    return report
+
+
+# We are dealing these fields dynamically, so we make a sanity check here to make sure
+# if something changes, we can catch it early instead of getting some cryptic errors
+# later, deep in the code.
+assert set(TickInfo.__dataclass_fields__.keys()) == {
+    "expected_tick_time",
+    "sleep_infos",
+}
```
shsms: What does this mean: "we are dealing these fields ..."?

shsms (dismissed):
🎉

shsms:
I'll approve again after you squash

tests/timeseries/_resampling/wall_clock_timer/conftest.py
llucax: I guess "dealing **with** these fields ...", will check and rephrase.

llucax: Fixed in the last 2 new fixup commits.

* [`47e9944` (#1249)](https://github.com/frequenz-floss/frequenz-sdk-python/pull/1249/commits/47e9944f54b6bf39a3dd4462db276582c97cf242)
* [`9202d83` (#1249)](https://github.com/frequenz-floss/frequenz-sdk-python/pull/1249/commits/9202d8317060c68514c60342a613cba6b0093191)

shsms (approved):

PR #1261, opened 2025-08-15
Remove unncessary use of `tzdata` in tests

### LINKED ISSUE #1260: Tests fail on Windows due to missing `tzdata` package
### What happened?

While running the tests on my Windows machine, I got errors when loading the "Europe/Berlin" timezone.
On Linux this works because Python reads time zone info from the OS, but on Windows it requires the PyPI `tzdata` package to be installed.

This caused the following tests to fail:

- `tests/microgrid/test_microgrid_api.py::TestMicrogridApi::test_connection_manager`
- `tests/microgrid/test_microgrid_api.py::TestMicrogridApi::test_connection_manager_another_method`

The tests run and pass.

### Affected version(s)

_No response_

### Affected part(s)

### Extra information

_No response_

denini08:
This PR fixes the issue reported in #1260 where tests failed on Windows because the "Europe/Berlin" timezone could not be loaded.

On Linux, zoneinfo can read timezone data from the system, but on Windows it requires the PyPI tzdata package.

### DIFF
--- a/tests/microgrid/test_microgrid_api.py
+++ b/tests/microgrid/test_microgrid_api.py
@@ -4,7 +4,6 @@
 """Tests of MicrogridApi."""
 
 import asyncio
-import zoneinfo
 from asyncio.tasks import ALL_COMPLETED
 from unittest import mock
 from unittest.mock import AsyncMock, MagicMock
@@ -102,7 +101,6 @@ class TestMicrogridApi:
             location=Location(
                 latitude=52.520008,
                 longitude=13.404954,
-                timezone=zoneinfo.ZoneInfo("Europe/Berlin"),
             ),
         )
 
@@ -170,8 +168,6 @@ class TestMicrogridApi:
 
             assert api.microgrid_id == metadata.microgrid_id
             assert api.location == metadata.location
-            assert api.location and api.location.timezone
-            assert api.location.timezone.key == "Europe/Berlin"
 
             # It should not be possible to initialize method once again
             with pytest.raises(AssertionError):
@@ -215,5 +211,3 @@ class TestMicrogridApi:
 
         assert api.microgrid_id == metadata.microgrid_id
         assert api.location == metadata.location
-        assert api.location and api.location.timezone
-        assert api.location.timezone.key == "Europe/Berlin"


llucax:
Hi @denini08, thanks for submitting a PR! :pray: 

I'm not sure what to do about this, I will reply in the issue because there is nothing wrong with the PR itself, but I'm still unsure if we really want to solve this.

llucax:
Hi, thanks for the follow-up!

Tests are failing. Please make sure to run `nox` locally and make sure all checks pass.

Also ideally squash both commits when things are working again.

denini08:
@llucax thanks for the feedback, can you please check now?

tests/microgrid/test_microgrid_api.py
```diff
@@ -170,8 +168,7 @@ async def test_connection_manager(
 
             assert api.microgrid_id == metadata.microgrid_id
             assert api.location == metadata.location
-            assert api.location and api.location.timezone
-            assert api.location.timezone.key == "Europe/Berlin"
+            assert api.location is not None
```
llucax: Nitpick: This check is redundant and might get out of sync. We are already checking that `api.location == metadata.location`, so if at some point someone wants to change `metadata.location` to `None`, this check will fail. I would remove it.

tests/microgrid/test_microgrid_api.py
```diff
@@ -215,5 +212,4 @@ async def test_connection_manager_another_method(
 
         assert api.microgrid_id == metadata.microgrid_id
         assert api.location == metadata.location
-        assert api.location and api.location.timezone
-        assert api.location.timezone.key == "Europe/Berlin"
+        assert api.location is not None
```
llucax: Same.

llucax (commented):
Thanks, LGTM. Just a couple of nitpick comments, but for me this is basically ready to be merged if tests pass.

llucax:
@denini08 we also require [all commits to have verified signatures](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) to be able to merge.

denini08:
@llucax can you please check again?

llucax:
Thanks again for taking the time for this @denini08! I was sort of live-watching as you fixed all the many issue with GPG signing and DCO, I'm sorry the process is so complicated!

All seems to be looking good now! Will merge when ready.

llucax (approved):

llucax:
I took the liberty of updating the PR title to more accurately reflect what it does.

denini08:
> Thanks again for taking the time for this @denini08! I was sort of live-watching as you fixed all the many issue with GPG signing and DCO, I'm sorry the process is so complicated!
> 
> All seems to be looking good now! Will merge when ready.

It's a difficult process, but it was good that I learned some new things

PR #1283, opened 2025-10-16
Update to use the microgrid API v0.18

llucax:
This PR migrates the SDK from the old `frequenz-client-microgrid` using the old microgrid API v0.15 to the new using the API v0.18. The main changes involve updating component models, metrics, and data fetching mechanisms to use the new client library structure.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,8 +29,8 @@ dependencies = [
   # Make sure to update the mkdocs.yml file when
   # changing the version
   # (plugins.mkdocstrings.handlers.python.import)
-  "frequenz-client-microgrid >= 0.9.0, < 0.10.0",
-  "frequenz-client-common >= 0.3.2, < 0.4.0",
+  "frequenz-client-microgrid >= 0.18.0, < 0.19.0",
+  "frequenz-client-common >= 0.3.6, < 0.4.0",
   "frequenz-channels >= 1.6.1, < 2.0.0",
   "frequenz-quantities[marshmallow] >= 1.0.0, < 2.0.0",
   "networkx >= 2.8, < 4",
--- a/src/frequenz/sdk/microgrid/_old_component_data.py
+++ b/src/frequenz/sdk/microgrid/_old_component_data.py
@@ -160,7 +160,7 @@ class ComponentData(ABC):
         from .. import microgrid
 
         components = microgrid.connection_manager.get().component_graph.components(
-            matching_ids={component_id}
+            matching_ids=component_id
         )
         if not components:
             raise ValueError(f"Unable to find component with {component_id}")
--- a/src/frequenz/sdk/microgrid/_old_component_data.py
+++ b/src/frequenz/sdk/microgrid/_old_component_data.py
@@ -1,0 +1,1290 @@
+# License: MIT
+# Copyright © 2025 Frequenz Energy-as-a-Service GmbH
+
+"""Component data types for data coming from a microgrid.
+
+This is a transitional module for migrating from the microgrid API v0.15 to v0.17.
+It maps the new component data types to the old ones, so the rest of the code can
+be updated incrementally.
+
+This module should be removed once the migration is complete.
+"""
+
+# pylint: disable=too-many-lines,fixme
+
+from __future__ import annotations
+
+import logging
+import math
+from abc import ABC, abstractmethod
+from collections.abc import Set
+from dataclasses import dataclass
+from datetime import datetime
+from enum import Enum
+from typing import ClassVar, Self, TypeAlias, TypeGuard, TypeVar, assert_never, cast
+
+from frequenz.channels import Receiver
+from frequenz.client.common.microgrid.components import ComponentId
+from frequenz.client.microgrid import MicrogridApiClient
+from frequenz.client.microgrid.component import (
+    ComponentCategory,
+    ComponentDataSamples,
+    ComponentErrorCode,
+    ComponentStateCode,
+    ComponentStateSample,
+)
+from frequenz.client.microgrid.metrics import Bounds, Metric, MetricSample
+from typing_extensions import override
+
+_logger = logging.getLogger(__name__)
+
+T = TypeVar("T", bound="ComponentData")
+
+PhaseTuple: TypeAlias = tuple[float, float, float]
+
+DATA_STREAM_BUFFER_SIZE: int = 50
+
+
+class TransitionalMetric(Enum):
+    """An enum representing the metrics we had in v0.15 but are not a metric in v0.17."""
+
+    SOC_LOWER_BOUND = "soc_lower_bound"
+    """Lower bound of state of charge."""
+    SOC_UPPER_BOUND = "soc_upper_bound"
+    """Upper bound of state of charge."""
+
+    POWER_INCLUSION_LOWER_BOUND = "power_inclusion_lower_bound"
+    """Power inclusion lower bound."""
+    POWER_EXCLUSION_LOWER_BOUND = "power_exclusion_lower_bound"
+    """Power exclusion lower bound."""
+    POWER_EXCLUSION_UPPER_BOUND = "power_exclusion_upper_bound"
+    """Power exclusion upper bound."""
+    POWER_INCLUSION_UPPER_BOUND = "power_inclusion_upper_bound"
+    """Power inclusion upper bound."""
+
+    ACTIVE_POWER_INCLUSION_LOWER_BOUND = "active_power_inclusion_lower_bound"
+    """Active power inclusion lower bound."""
+    ACTIVE_POWER_EXCLUSION_LOWER_BOUND = "active_power_exclusion_lower_bound"
+    """Active power exclusion lower bound."""
+    ACTIVE_POWER_EXCLUSION_UPPER_BOUND = "active_power_exclusion_upper_bound"
+    """Active power exclusion upper bound."""
+    ACTIVE_POWER_INCLUSION_UPPER_BOUND = "active_power_inclusion_upper_bound"
+    """Active power inclusion upper bound."""
+
+
+@dataclass(kw_only=True)
+class ComponentData(ABC):
+    """A private base class for strongly typed component data classes."""
+
+    component_id: ComponentId
+    """The ID identifying this component in the microgrid."""
+
+    timestamp: datetime
+    """The timestamp of when the data was measured."""
+
+    states: Set[ComponentStateCode | int] = frozenset()
+    """The states of the component."""
+
+    warnings: Set[ComponentErrorCode | int] = frozenset()
+    """The warnings of the component."""
+
+    errors: Set[ComponentErrorCode | int] = frozenset()
+    """The errors of the component."""
+
+    CATEGORY: ClassVar[ComponentCategory] = ComponentCategory.UNSPECIFIED
+    """The category of this component."""
+
+    METRICS: ClassVar[frozenset[Metric]] = frozenset()
+    """The metrics of this component."""
+
+    @abstractmethod
+    def to_samples(self: Self) -> ComponentDataSamples:
+        """Convert the component data to a component data object."""
+
+    @staticmethod
+    def _from_samples(class_: type[T], /, samples: ComponentDataSamples) -> T:
+        """Create a new instance from a component data object."""
+        if not samples.metric_samples:
+            raise ValueError("No metrics in the samples.")
+
+        # FIXME: This might not be true forever, but the service sends all metrics with
+        # the same timestamp for now, and it is very convenient to map the received data
+        # to the old component data metrics packets, which had only one timestamp.
+        # When we move away frome these old_component_data wrappers, we should not
+        # assume only one metric sample can come per telemetry message anymore.
+        timestamp = samples.metric_samples[-1].sampled_at
+        for sample in samples.metric_samples[:-1]:
+            if sample.sampled_at != timestamp:
+                _logger.warning(
+                    "ComponentData has multiple timestamps. Using the last one. Samples: %r",
+                    samples,
+                )
+                break
+
+        if not samples.states:
+            return class_(component_id=samples.component_id, timestamp=timestamp)
+
+        # FIXME: Same as with metric samples, for now we get only one, and it is super
+        # convenient to map to component data, but we should not assume this when moving
+        # away from the legacy component data wrappers.
+        if len(samples.states) > 1:
+            _logger.warning(
+                "ComponentData has more than one state. Using the last one. States: %r",
+                samples.states,
+            )
+
+        return class_(
+            component_id=samples.component_id,
+            timestamp=timestamp,
+            states=samples.states[-1].states,
+            warnings=samples.states[-1].warnings,
+            errors=samples.states[-1].errors,
+        )
+
+    @classmethod
+    @abstractmethod
+    def from_samples(cls, samples: ComponentDataSamples) -> Self:
+        """Create a new instance from a component data object."""
+
+    @classmethod
+    def _check_category(cls, component_id: ComponentId) -> None:
+        """Check if the given component_id is of the expected type.
+
+        Args:
+            component_id: Component id to check.
+
+        Raises:
+            ValueError: if the given id is unknown or has a different type.
+        """
+        # pylint: disable-next=import-outside-toplevel,cyclic-import
+        from .. import microgrid
+
+        components = microgrid.connection_manager.get().component_graph.components(
+            filter_by_ids={component_id}
+        )
+        if not components:
+            raise ValueError(f"Unable to find component with {component_id}")
+        if len(components) > 1:
+            raise ValueError(f"Multiple components with id {component_id}")
+        component = components.pop()
+        if component.category != cls.CATEGORY:
+            raise ValueError(
+                f"Component with {component_id} is a {component.category}, "
+                f"not a {cls.CATEGORY}."
+            )
+
+    @classmethod
+    def subscribe(
+        cls,
+        api_client: MicrogridApiClient,
+        component_id: ComponentId,
+        *,
+        buffer_size: int = DATA_STREAM_BUFFER_SIZE,
+    ) -> Receiver[Self]:
+        """Subscribe to the component data stream."""
+        cls._check_category(component_id)
+
+        def _is_valid(messages: Self | Exception) -> TypeGuard[Self]:
+            return not isinstance(messages, Exception)
+
+        receiver = api_client.receive_component_data_samples_stream(
+            component_id, cls.METRICS, buffer_size=buffer_size
+        )
+
+        return receiver.map(cls._receive_logging_errors).filter(_is_valid)
+
+    # This needs to be a classmethod because otherwise it seems like mypy can't
+    # guarantee that the Self returned by this function is the same Self in the
+    # subscribe() method.
+    @classmethod
+    def _receive_logging_errors(
+        cls, samples: ComponentDataSamples, /
+    ) -> Self | Exception:
+        try:
+            return cls.from_samples(samples)
+        except Exception as exc:  # pylint: disable=broad-except
+            _logger.exception(
+                "Error while creating %r from samples: %r", cls.__name__, samples
+            )
+            return exc
+
+
+@dataclass(kw_only=True)
+class MeterData(ComponentData):
+    """A wrapper class for holding meter data."""
+
+    # FIXME: All of this have now a default of 0.0 because this is what it was doing when
+    # we used the API v0.15, as we accessed the fields without checking if the fields
+    # really existed, so the default protobuf value of 0.0 for floats was used.
+    # When moving away from these legacy component data wrappers, this should not be
+    # a problem anymore, as we will just get the data we get.
+    active_power: float = 0.0
+    """The total active 3-phase AC power, in Watts (W).
+
+    Represented in the passive sign convention.
+
+    * Positive means consumption from the grid.
+    * Negative means supply into the grid.
+    """
+
+    active_power_per_phase: PhaseTuple = (0.0, 0.0, 0.0)
+    """The per-phase AC active power for phase 1, 2, and 3 respectively, in Watt (W).
+
+    Represented in the passive sign convention.
+
+    * Positive means consumption from the grid.
+    * Negative means supply into the grid.
+    """
+
+    reactive_power: float = 0.0
+    """The total reactive 3-phase AC power, in Volt-Ampere Reactive (VAr).
+
+    * Positive power means capacitive (current leading w.r.t. voltage).
+    * Negative power means inductive (current lagging w.r.t. voltage).
+    """
+
+    reactive_power_per_phase: PhaseTuple = (0.0, 0.0, 0.0)
+    """The per-phase AC reactive power, in Volt-Ampere Reactive (VAr).
+
+    The provided values are for phase 1, 2, and 3 respectively.
+
+    * Positive power means capacitive (current leading w.r.t. voltage).
+    * Negative power means inductive (current lagging w.r.t. voltage).
+    """
+
+    current_per_phase: PhaseTuple = (0.0, 0.0, 0.0)
+    """AC current in Amperes (A) for phase/line 1,2 and 3 respectively.
+
+    Represented in the passive sign convention.
+
+    * Positive means consumption from the grid.
+    * Negative means supply into the grid.
+    """
+
+    voltage_per_phase: PhaseTuple = (0.0, 0.0, 0.0)
+    """The ac voltage in volts (v) between the line and the neutral wire for phase/line
+        1,2 and 3 respectively.
+    """
+
+    frequency: float = 0.0
+    """The AC power frequency in Hertz (Hz)."""
+
+    CATEGORY: ClassVar[ComponentCategory] = ComponentCategory.METER
+    """The category of this component."""
+
+    METRICS: ClassVar[frozenset[Metric]] = frozenset(
+        [
+            Metric.AC_ACTIVE_POWER,
+            Metric.AC_ACTIVE_POWER_PHASE_1,
+            Metric.AC_ACTIVE_POWER_PHASE_2,
+            Metric.AC_ACTIVE_POWER_PHASE_3,
+            Metric.AC_REACTIVE_POWER,
+            Metric.AC_REACTIVE_POWER_PHASE_1,
+            Metric.AC_REACTIVE_POWER_PHASE_2,
+            Metric.AC_REACTIVE_POWER_PHASE_3,
+            Metric.AC_CURRENT_PHASE_1,
+            Metric.AC_CURRENT_PHASE_2,
+            Metric.AC_CURRENT_PHASE_3,
+            Metric.AC_VOLTAGE_PHASE_1_N,
+            Metric.AC_VOLTAGE_PHASE_2_N,
+            Metric.AC_VOLTAGE_PHASE_3_N,
+            Metric.AC_FREQUENCY,
+        ]
+    )
+    """The metrics of this component."""
+
+    @override
+    @classmethod
+    def from_samples(cls, samples: ComponentDataSamples) -> Self:
+        """Create a new instance from a component data object."""
+        if not samples.metric_samples:
+            raise ValueError("No metrics in the samples.")
+
+        self = cls._from_samples(cls, samples)
+
+        active_power_per_phase: list[float] = [0.0, 0.0, 0.0]
+        reactive_power_per_phase: list[float] = [0.0, 0.0, 0.0]
+        current_per_phase: list[float] = [0.0, 0.0, 0.0]
+        voltage_per_phase: list[float] = [0.0, 0.0, 0.0]
+
+        for sample in samples.metric_samples:
+            match sample.metric:
+                case Metric.AC_ACTIVE_POWER:
+                    self.active_power = sample.as_single_value() or 0.0
+                case Metric.AC_ACTIVE_POWER_PHASE_1:
+                    active_power_per_phase[0] = sample.as_single_value() or 0.0
+                case Metric.AC_ACTIVE_POWER_PHASE_2:
+                    active_power_per_phase[1] = sample.as_single_value() or 0.0
+                case Metric.AC_ACTIVE_POWER_PHASE_3:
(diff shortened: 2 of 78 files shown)


src/frequenz/sdk/microgrid/_old_component_data.py
```diff
@@ -0,0 +1,1287 @@
...
+        """Convert the component data to a component data object."""
+
+    @staticmethod
+    def _from_samples(class_: type[T], /, samples: ComponentDataSamples) -> T:
+        """Create a new instance from a component data object."""
+        if not samples.metric_samples:
+            raise ValueError("No metrics in the samples.")
+
+        # FIXME: This might not be true forever, but the service for now seems to send
+        # all metrics with the same timestamp for now.
```
llucax: I think this one can be removed, at least IIRC @tiyash-basu-frequenz confirmed this was going to be like this for the time being. In any case, this is a temporary transitional module, at some point we should start using the telemetry messages as they come from the API.

src/frequenz/sdk/microgrid/_old_component_data.py
```diff
@@ -0,0 +1,1287 @@
...
+                _logger.warning(
+                    "ComponentData has multiple timestamps. Using the last one. Samples: %r",
+                    samples,
+                )
+                break
+
+        if not samples.states:
+            return class_(component_id=samples.component_id, timestamp=timestamp)
+
+        # FIXME: Maybe we can have more than one, in which case we need to merge them?
```
llucax: I think @tiyash-basu-frequenz also confirmed this was fine for now, but I don't remember if the idea is that more than one could come in the future.

src/frequenz/sdk/microgrid/_old_component_data.py
```diff
@@ -0,0 +1,1287 @@
...
+
+@dataclass(kw_only=True)
+class MeterData(ComponentData):
+    """A wrapper class for holding meter data."""
+
+    # FIXME: All of this have now a default of 0.0 because this is what it was done when
+    # we used the API v0.15, as we accessed the fields without checking if the fields
+    # really existed, so the defaul protobuf value of 0.0 for floats was used.
+    # We might beed to review this and if they are not present interpret them as None
+    # instea.
```
llucax: This one is tricky, I guess we can leave it as it is now and will have to come back to it forcibly when removing this temporary adap**ta**tion layer.

src/frequenz/sdk/microgrid/_old_component_data.py
```diff
@@ -0,0 +1,1287 @@
...
+        for sample in samples.metric_samples:
+            value = sample.as_single_value() or 0.0
+            match sample.metric:
+                case Metric.BATTERY_SOC_PCT:
+                    self.soc = value
+                    if sample.bounds:
+                        # Update power bounds from the SOC metric bounds,
+                        # FIXME: We assume only one range is present
+                        # If one bound is None, we assume 0 to match the previous
+                        # behavior of v0.15, but this should eventually be fixed
```
llucax: I guess we'll also fix this when removing this adaption layer, not sure if it is worth adding a separate issue for this one.

src/frequenz/sdk/microgrid/_old_component_data.py
llucax, line 1: I guess we can disable pylint `fixme` check for this entire module and keep the FIXMEs as they are now to remember to address these issues when removing this adaption layer.

src/frequenz/sdk/microgrid/_power_distributing/_component_status/_battery_status_tracker.py
```diff
@@ -416,25 +422,32 @@ def _is_battery_state_correct(self, msg: BatteryData) -> bool:
...
-        state = msg.component_state
+        component_states = msg.states
         # pylint: disable-next=protected-access
-        if state not in BatteryStatusTracker._battery_valid_state:
+        valid_states = BatteryStatusTracker._battery_valid_state
+        # FIXME: There was a semantic change here, before the component data had only
+        # one state, and now it has a set of states, for now we are considering the
+        # component to be working if any of its states is valid. Maybe we need to switch
+        # to the negative and consider a component not working if is in any known
+        # invalid state instead.
```
llucax: @shsms Maybe you have a more clear idea of what to do here. Maybe for a first step we can merge as is and create a separate issue to address it in the future?

src/frequenz/sdk/microgrid/_power_distributing/_component_status/_battery_status_tracker.py
```diff
@@ -394,14 +394,20 @@ def _is_inverter_state_correct(self, msg: InverterData) -> bool:
...
-        state = msg.component_state
+        component_states = msg.states
         # pylint: disable-next=protected-access
-        if state not in BatteryStatusTracker._inverter_valid_state:
+        valid_states = BatteryStatusTracker._inverter_valid_state
+        # FIXME: There was a semantic change here, before the component data had only
+        # one state, and now it has a set of states, for now we are considering the
+        # component to be working if any of its states is valid. Maybe we need to switch
+        # to the negative and consider a component not working if is in any known
+        # invalid state instead.
```
llucax: @shsms Maybe you have a more clear idea of what to do here. Maybe for a first step we can merge as is and create a separate issue to address it in the future?

src/frequenz/sdk/microgrid/_old_component_data.py
```diff
@@ -0,0 +1,1287 @@
...
+        """
+        # Old code:
+        # return self.component_state not in (
+        #     EVChargerComponentState.AUTHORIZATION_REJECTED,
+        #     EVChargerComponentState.ERROR,
+        # ) and self.cable_state in (
+        #     EVChargerCableState.EV_LOCKED,
+        #     EVChargerCableState.EV_PLUGGED,
+        # )
+        # TODO: Verify this logic is correct
```
llucax: Maybe you can check this @shsms ?

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_battery_manager.py
```diff
@@ -84,9 +80,10 @@ def _get_battery_inverter_mappings(
 
     for battery_id in battery_ids:
         inverters: set[ComponentId] = set(
-            component.component_id
+            component.id
             for component in component_graph.predecessors(battery_id)
-            if component.category == ComponentCategory.INVERTER
+            # TODO: Shouldn't this be (SolarInverter, HybridInverter)?
```
llucax: @shsms ? Maybe it is better to create a separate issue anyways as this was a pre-existing issue.

src/frequenz/sdk/microgrid/_power_distributing/_component_status/_battery_status_tracker.py
```diff
@@ -371,15 +371,15 @@ def _no_critical_error(self, msg: BatteryData | InverterData) -> bool:
         Returns:
             True if message has no critical error, False otherwise.
         """
-        critical = ErrorLevel.CRITICAL
-        critical_err = next((err for err in msg.errors if err.level == critical), None)
-        if critical_err is not None:
+        # TODO: This was using ErrorLevel.CRITICAL to see if an error was critical or
+        # warning. I guess now we use the separate errors and warnings fields.
```
llucax: Can you have a look at this one too @shsms ?

src/frequenz/sdk/microgrid/component_graph.py
```diff
@@ -539,22 +563,63 @@ def refresh_from(
...
+            _logger.warning("Component %r has a mismatched category!", component)
+
+        # TODO: This is old validation logic, which should probably be removed
+        if component.id == ComponentId(0) and not isinstance(
+            component, GridConnectionPoint
+        ):
+            issues.append(
+                "Component with ID 0 should be a GridConnectionPoint, "
+                f"but it is a {component!r} instead."
+            )
```
llucax: I think this check is safe to remove. I only found a previous check in an unused method `_correct_graph_errors`, removed in 635914ca5683b0195d3c8c0f7af9a2e8a4250298. I think I wrote this code before committing that removal.

FYI @tiyash-basu-frequenz @shsms @ela-kotulska-frequenz, I will remove this check unless someone complains.

src/frequenz/sdk/timeseries/battery_pool/_methods.py
```diff
@@ -253,6 +253,9 @@ async def _send_on_update(self, min_update_interval: timedelta) -> None:
                 await sender.send(result)
 
                 if result is None:
+                    # TODO: This code is unreacheable, as T can't be None. We need to
+                    # figure out which case this was supposed to handle. Maybe we need
+                    # some sentinel value or to allow None in T.
```
llucax: @shsms maybe you can check this one?

llucax (commented):
I think these are all the TODOs/FIXMEs I wrote when doing the update. It would be good to get some feedback.

src/frequenz/sdk/microgrid/_old_component_data.py
shsms: I think for EV chargers, we will have component state and cable state, both of which are necessary for its operation.  But I guess @tiyash-basu-frequenz means that they will both be inside `states[0].states`, which is also a list.  We will only not have multiple sets of states. 

In that case, we just need to update the warning message below.

src/frequenz/sdk/microgrid/_old_component_data.py
```diff
@@ -0,0 +1,1287 @@
...
+        """Check if the given component_id is of the expected type.
+
+        Args:
+            component_id: Component id to check.
+
+        Raises:
+            ValueError: if the given id is unknown or has a different type.
+        """
+        # pylint: disable-next=import-outside-toplevel,cyclic-import
+        from frequenz.sdk import microgrid
```
shsms: why not relative path instead: `from .. import microgrid`

src/frequenz/sdk/microgrid/_old_component_data.py
shsms:
```suggestion
    # FIXME: All of this have now a default of 0.0 because this is what it was doing when
    # we used the API v0.15, as we accessed the fields without checking if the fields
    # really existed, so the default protobuf value of 0.0 for floats was used.
    # We might need to review this and if they are not present interpret them as None
    # instead.
```

Also, it is the "adap**ta**tion" layer.

src/frequenz/sdk/microgrid/_old_component_data.py
```diff
@@ -0,0 +1,1287 @@
...
+                case Metric.AC_FREQUENCY:
+                    self.frequency = sample.as_single_value() or 0.0
+                case unexpected:
+                    _logger.warning(
+                        "Unexpected metric %s in meter data sample: %r",
+                        unexpected,
+                        sample,
+                    )
+
+        self.active_power_per_phase = cast(PhaseTuple, tuple(active_power_per_phase))
```
shsms, line 349: wow, amazing this doesn't happen automatically, given that it is just an alias.

src/frequenz/sdk/microgrid/_old_component_data.py
shsms: the only issue in the logic below is that we're keeping only the latest state - it could be the component state of READY/ERROR,  or cable state of PLUGGED/LOCKED etc.

src/frequenz/sdk/microgrid/_old_component_data.py
```diff
@@ -0,0 +1,1287 @@
...
+                {
+                    ComponentStateCode.EV_CHARGING_CABLE_LOCKED_AT_EV,
+                    ComponentStateCode.EV_CHARGING_CABLE_LOCKED_AT_STATION,
+                }
+                & self.states
+                or {
+                    ComponentStateCode.EV_CHARGING_CABLE_PLUGGED_AT_EV,
+                    ComponentStateCode.EV_CHARGING_CABLE_PLUGGED_AT_STATION,
+                }
+                & self.states
```
shsms: we probably need to drop the `*_AT_STATION` codes.  Cable connected to the station doesn't mean the other end of the cable has a car attached.

src/frequenz/sdk/microgrid/_old_component_data.py
shsms: just a comment about the cable states, written below.

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_battery_manager.py
shsms: We were not doing that because there were still a lot of edge configs from the early days that didn't specify the type of the inverter.  Maybe things have changed now and we can start with a log message if the inverter type is not set.

src/frequenz/sdk/microgrid/_power_distributing/_component_status/_battery_status_tracker.py
shsms: This makes sense I think.

src/frequenz/sdk/microgrid/_power_distributing/_component_status/_battery_status_tracker.py
shsms: Sounds unlikely, unless there's a bug, buf if there are multiple contradictory states for a component for the same timestamp, then we'll have to assume the component is broken.

src/frequenz/sdk/microgrid/_power_distributing/_component_status/_battery_status_tracker.py
shsms: We'll have to assume the component is broken.

src/frequenz/sdk/microgrid/_power_distributing/_component_status/_ev_charger_status_tracker.py
```diff
@@ -87,13 +84,25 @@ def start(self) -> None:
...
+                {
+                    ComponentStateCode.EV_CHARGING_CABLE_LOCKED_AT_EV,
+                    ComponentStateCode.EV_CHARGING_CABLE_LOCKED_AT_STATION,
+                }
+                & ev_data.states
+                or {
+                    ComponentStateCode.EV_CHARGING_CABLE_PLUGGED_AT_EV,
+                    ComponentStateCode.EV_CHARGING_CABLE_PLUGGED_AT_STATION,
+                }
+                & ev_data.states
```
shsms: Here too, we have to drop the `AT_STATION` states.  An ev charger can't be assigned power if it doesn't have an EV attached, and that's what the `is_working` method is really about.

src/frequenz/sdk/timeseries/battery_pool/_methods.py
shsms: looks like it stopped being `T | None` here: https://github.com/frequenz-floss/frequenz-sdk-python/pull/692/commits/bf22d6d669b8570b2a90217c96d38c9bb2813b14

The handling for this case can go away as well.

shsms (commented):
Only read a little bit of the big commit

src/frequenz/sdk/microgrid/_old_component_data.py
llucax: Yeah, we can still have multiple `samples.states[0].states` with the current code, I was referring to multiple.... Mmmm, this is getting confusing having 2 states that are actually different things, I was referring to multiple `samples.states`.

src/frequenz/sdk/microgrid/_old_component_data.py
llucax: I guess no reason.
```suggestion
        from .. import microgrid
```

src/frequenz/sdk/microgrid/_old_component_data.py
llucax: No comment on the original comment? :D

src/frequenz/sdk/microgrid/_old_component_data.py
llucax: I guess it is because there is a dynamic conversion going on `active_power_per_phase` is a list[float]` and it is converted to a tuple dynamically, so this should have type `tuple[float, ...]` (if the mapping `list[float]` -> tuple[float, ...]` even works), not `tuple[float, float, float]`.

src/frequenz/sdk/microgrid/_old_component_data.py
llucax: But this is checking that is connected at both ends, the EV and the station. it seems like the new API is splitting the `LOCKED` and `PLUGGED` into both ends, so the intention here was that we make sure both ends are connected to return true for `is_ev_connected`.

@tiyash-basu-frequenz maybe can also help here.

llucax: Thinking a bit more maybe you mean it is redundant because ia EV charging station can only know the EV end is connected if the station end is also connected? Also I was thinking maybe I should add more variables to make it more readable:

```py
has_errors = ComponentStateCode.ERROR not in self.states
is_authorized = (
                ComponentErrorCode.UNAUTHORIZED not in self.errors
                and ComponentErrorCode.UNAUTHORIZED not in self.warnings
            )
both_ends_are locked = bool(
                {
                    ComponentStateCode.EV_CHARGING_CABLE_LOCKED_AT_EV,
                    ComponentStateCode.EV_CHARGING_CABLE_LOCKED_AT_STATION,
                }
                & self.states)
both_ends_are_plugged = bool(
                {
                    ComponentStateCode.EV_CHARGING_CABLE_PLUGGED_AT_EV,
                    ComponentStateCode.EV_CHARGING_CABLE_PLUGGED_AT_STATION,
                }
                & self.states)
return not has_errors and is_authorized and (both_ends_are_locked or both_ends_are_plugged)
```
This also makes me think that I'm not sure if a combination of "plugged at EV-locked at station" should also not be considered as "connected".

src/frequenz/sdk/microgrid/_power_distributing/_component_status/_battery_status_tracker.py
llucax: <del>In which case? :thinking: </del>

Oh, from the following comment I guess you mean if there is a contradictory state.

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_battery_manager.py
llucax: Yeah, in the current client this should be logged as a warning as soon it is received by the client, and a `UnspecifiedInverter` type will be returned (with `category` `INVERTER` and `type` `UNSPECIFIED`, which is the default when no type is present).

So I guess you suggest to leave it as it is for now, and make it more specific when we confirm we have no inverters without a type anymore.

llucax: Also I guess it would be a topological error to have a non matching inverter type as a predecessor, but we should still try to use it in case it is a configuration error?

src/frequenz/sdk/timeseries/battery_pool/_methods.py
llucax: Great, thanks for checking!

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_battery_manager.py
llucax: According to @tiyash-basu-frequenz everything should be properly configured by now.

src/frequenz/sdk/microgrid/_old_component_data.py
llucax: I got a confirmation that only one state message will come in a telemetry message (sample).

src/frequenz/sdk/microgrid/_old_component_data.py
llucax: Marking as resolved as this is confirmed.

src/frequenz/sdk/microgrid/_power_distributing/_component_status/_ev_charger_status_tracker.py
llucax: Wouldn't it make more sense in this case to just change to something like:

```py
is_connected = ev_data.is_ev_connected()
is_ok = bool({
                ComponentStateCode.READY,
                ComponentStateCode.CHARGING,
                ComponentStateCode.DISCHARGING,
            }
            & ev_data.states)
return is_connected and is_ok
```

?

src/frequenz/sdk/microgrid/_old_component_data.py
llucax: After talking to @tiyash-basu-frequenz I will update to:

```py
    def is_ev_connected(self) -> bool:
        """Check whether an EV is connected to the charger.

        Returns:
            When the charger is not in an error state, whether an EV is connected to
                the charger.
        """
        has_error = ComponentStateCode.ERROR not in self.states
        is_authorized = (
            ComponentErrorCode.UNAUTHORIZED not in self.errors
            and ComponentErrorCode.UNAUTHORIZED not in self.warnings
        )
        is_connected_at_ev = bool(
            {
                ComponentStateCode.EV_CHARGING_CABLE_LOCKED_AT_EV,
                ComponentStateCode.EV_CHARGING_CABLE_PLUGGED_AT_EV,
            }
            & self.states
        )
        is_connected_at_station = bool(
            {
                ComponentStateCode.EV_CHARGING_CABLE_LOCKED_AT_STATION,
                ComponentStateCode.EV_CHARGING_CABLE_PLUGGED_AT_STATION,
            }
            & self.states
        )

        return (
            not has_error
            and is_authorized
            and is_connected_at_ev
            and is_connected_at_station
        )
```

src/frequenz/sdk/microgrid/_power_distributing/_component_status/_battery_status_tracker.py
llucax: OK, I will leave it as it is then. I guess checking for inconsistencies will be something new and probably out of the scope of this PR.

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_battery_manager.py
llucax: In any case, I will keep the old behavior in this PR, as it is only about the microgrid API version upgrade.

* https://github.com/frequenz-floss/frequenz-sdk-python/issues/1285

src/frequenz/sdk/microgrid/component_graph.py
llucax: Removing.

src/frequenz/sdk/microgrid/_power_distributing/_component_status/_ev_charger_status_tracker.py
llucax: I will go for this, @shsms let me know if you think it is wrong.

src/frequenz/sdk/microgrid/_old_component_data.py
shsms: oops, I agree, if a field is not there, it should be None, but can be done separately.

src/frequenz/sdk/microgrid/_old_component_data.py
shsms: There's likely only one cable-state.  So it would either say LOCKED_AT_STATION (meaning there's cable, but no car) or LOCKED_AT_EV (meaning there's car, and therefore cable).

So we can't use AT_STATION states for this check.

src/frequenz/sdk/microgrid/_power_distributing/_component_status/_ev_charger_status_tracker.py
shsms: is_ev_connected needs a change, but otherwise this looks good.

tests/microgrid/power_distributing/_component_status/test_ev_charger_status.py
```diff
@@ -60,9 +60,12 @@ async def test_status_changes(self, mocker: MockerFixture) -> None:
...
                     active_power=0.0,
-                    component_state=EVChargerComponentState.READY,
-                    cable_state=EVChargerCableState.EV_PLUGGED,
-                )
+                    states={
+                        ComponentStateCode.READY,
+                        ComponentStateCode.EV_CHARGING_CABLE_PLUGGED_AT_EV,
+                        ComponentStateCode.EV_CHARGING_CABLE_PLUGGED_AT_STATION,
+                    },
+                ).to_samples()
```
shsms, line 68: cable plugged at EV necessarily means cable LOCKED at station.

And it was not the case with the service to send two states for the cable, and I don't think it needs to change now, it is already obvious, so we'd just be sending redundant information.

tests/microgrid/power_distributing/_component_status/test_ev_charger_status.py
```diff
@@ -121,9 +132,12 @@ async def test_status_changes(self, mocker: MockerFixture) -> None:
...
                     datetime.now(tz=timezone.utc),
                     active_power=0.0,
-                    component_state=EVChargerComponentState.READY,
-                    cable_state=EVChargerCableState.EV_LOCKED,
-                )
+                    states={
+                        ComponentStateCode.READY,
+                        ComponentStateCode.EV_CHARGING_CABLE_PLUGGED_AT_EV,
+                        ComponentStateCode.EV_CHARGING_CABLE_LOCKED_AT_STATION,
+                    },
```
shsms, line 139: this is not an exact match.

Old says EV LOCKED.  new says EV PLUGGED.

shsms (commented):
Still got a number of files to go through.

llucax:
I'm working on splitting the big WIP commit, so maybe you want to wait for that before you continue with the review.

tests/microgrid/power_distributing/_component_status/test_ev_charger_status.py
llucax: Apparently it already changed, at least this is what I understood from a meeting with @tiyash-basu-frequenz

tests/microgrid/power_distributing/_component_status/test_ev_charger_status.py
llucax: Oh, good catch! Will fix.

llucax: The test still passed because actually all permutations are considered as *connected* I guess.

src/frequenz/sdk/microgrid/_old_component_data.py
llucax: I left the comment in the module, we'll have to address some comments there when removing the adaptation layer module anyway.

src/frequenz/sdk/microgrid/_old_component_data.py
llucax: Again, question for @tiyash-basu-frequenz as what I understood from the meeting with him is in v0.18 we should check for both.

llucax:
OK, this should be ready for a final review, I split the big WIP commit into smaller commits. The commit split is still not the best, but it should be much more reviewable now. I left a few unrelated changes/improvements (like the removal of unreachable code) for a follow-up PR with just cleanups/improvements that are not strictly part of the v0.18 update (although there still a few unrelated commits in this PR too).

src/frequenz/sdk/microgrid/_old_component_data.py
shsms: ok, in that case, the new and and and function looks good.

src/frequenz/sdk/microgrid/_old_component_data.py
```diff
@@ -0,0 +1,1290 @@
...
+            not has_error
+            and is_authorized
+            and is_connected_at_ev
+            and is_connected_at_station
+        )
+
+
+def _bound_ranges_to_inclusion_exclusion(
+    bounds: list[Bounds], name: str, sample: MetricSample
+) -> tuple[float, float, float, float]:
```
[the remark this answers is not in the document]
llucax: I have no idea what do you mean by this, but in any case `mypy` doesn't allow us to use implicit returns anyways.

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_battery_manager.py
```diff
@@ -8,21 +8,17 @@
...
 from frequenz.client.common.microgrid.components import ComponentId
-from frequenz.client.microgrid import (
-    ApiClientError,
-    BatteryData,
-    ComponentCategory,
-    InverterData,
-    OperationOutOfRange,
-)
+from frequenz.client.microgrid import ApiClientError, OperationOutOfRange
+from frequenz.client.microgrid.component import Battery, Inverter
```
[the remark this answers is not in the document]
llucax: BS

tests/microgrid/power_distributing/_component_status/test_ev_charger_status.py
llucax: (fixed anyway)

pyproject.toml
```diff
@@ -29,8 +29,9 @@ dependencies = [
   # Make sure to update the mkdocs.yml file when
   # changing the version
   # (plugins.mkdocstrings.handlers.python.import)
-  "frequenz-client-microgrid >= 0.9.0, < 0.10.0",
-  "frequenz-client-common >= 0.3.2, < 0.4.0",
+  #"frequenz-client-microgrid >= 0.18.0, < 0.19.0",
+  "frequenz-client-microgrid @ git+https://github.com/frequenz-floss/frequenz-client-microgrid-python@refs/heads/v0.18.x",
```
shsms: Is there a PR for the client already?  Can't find one.

llucax:
Updated the dependency, I think it would be best to merge when the review is ready, so marking as ready. I just realized I also need to update the release notes.

shsms:
Super, you can remove WIP from the commit message as well.  I will continue reviewing now.

shsms (dismissed):
🎉

llucax:
Whooohooo! Rebased and enabled auto-merge.

llucax:
OK, I needed to [adapt a new test](https://github.com/frequenz-floss/frequenz-sdk-python/compare/578863929a8bb9381692cd0e48cb0bd1f0bdc266..683bcb24695c9103c090841686476bd5bff3a5be), so it will need a new approval.

shsms (approved):

Issue #1080, opened 2024-09-27
New approach for calculating production and consumption powers 


shsms:
A new approach to calculating production and consumption powers that would bypass the problem of identifying where consumption happens in the component graph.

### Batteries or meters WITH battery successors(direct or indirect):

- Their consumption power could be charging power of the battery, so their powers can't be used for the site's consumption power.

- Their production power could be discharging from batteries, so they can't be used for the site's production power.

- When such a component is encountered during traversal, if it has other non-battery successors, those should be used to calculate production and consumption powers.

### Non-battery components or meters WITHOUT battery successors (direct or indirect):

- This includes dangling meters, and meters with only other types of successors, other than battery inverters.

- Production formula should contain a `min(0,meter_power)` component in the formula for each such meter.

- Consumption formula should contain a `max(0,meter_power)` component in the formula for each such meter.

### Validation

- These `min`, `max` formula components should accept an optional validation check.  For example, `min(0, meter_power, x >= 0.0)`.  Then for each value in the data stream, every time the validation fails, we log a message, but we still use the value.

- The log messages should be rate-limited to upto once a minute or something similar, to prevent spamming.

- **Meters without successors**: They are normally assumed to have only consumers, so the *production* component should have a validation of `x == 0`.  Then if we measure production from dangling meters, we will see a log message.

- **CHP**: They are expected to always produce, so their *consumption* component should have a validation of `x == 0`.

- **PV inverters**: They are expected to always produce, but they have small consumption at night.  So a simple `x == 0` validation for consumption will not be enough.

### When there's a grid meter

- When there's a grid meter at a site, we would also add an unmetered component to the formulas.

- This would be calculated by subtracting the powers of all the successors of the grid meter, from the power measured at the grid meter.

- Production formula should contain a `min(0,unmetered_power, x == 0)` component in the formula for each such meter.

- Consumption formula should contain a `max(0,unmetered_power)` component in the formula for each such meter.

shsms:
This is now taken care of by the rust component graph and is no longer the responsibility of the SDK.

PR #1295, opened 2025-11-20
Switch to the rust component graph

shsms:
- Also implements the new `Formula`/`Formula3Phase` classes to replace the old formula engine, for supporting the new formula functions - max/min/coalesce.
- Drops the old component graph and the old formula engine


shsms:
Based on #1283

tests/microgrid/power_distributing/_component_status/test_battery_status.py
```diff
@@ -7,7 +7,7 @@
 
 import asyncio
 import math
-from collections.abc import AsyncIterator, Iterable
+from collections.abc import AsyncIterator, Set
```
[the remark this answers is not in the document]
shsms: Wrong

llucax:
Can you rebase it on top of Based on https://github.com/frequenz-floss/frequenz-sdk-python/pull/1293 instead?

shsms:
> Can you rebase it on top of Based on https://github.com/frequenz-floss/frequenz-sdk-python/pull/1293 instead?

#1293 is making changes to a lot of the files that are being deleted by this PR.  I'd rather take this in first, then drop the changes on these files, then we only need to review required changes.

shsms:
Oh, maybe those changes are from the api PR.  don't know.  Let's take the API PR first then, we can decide.

shsms:
It was actually trivial to rebase on top of #1293.  Done.

shsms:
Damn, this causes my tests to fail.  Switched back to based on #1283.

llucax:
This needs a rebase now.

shsms:
Rebased

src/frequenz/sdk/timeseries/formulas/_ast.py
```diff
@@ -0,0 +1,205 @@
...
+        return expr
+
+
+@dataclass
+class Sub(Node):
+    """Subtraction operation node."""
+
+    left: Node
+    right: Node
+
```
ela-kotulska-frequenz, line 143: Just curious, why `Add`, `Sub`, `Mul` are a `Node` , not a `Function`?

src/frequenz/sdk/timeseries/formulas/_formula_evaluator.py
```diff
@@ -0,0 +1,129 @@
...
+                timestamp = next(
+                    comp.latest_sample.timestamp
+                    for comp in self._components
+                    if comp.latest_sample is not None
+                )
+
+                res = self._root.evaluate()
+                next_sample = Sample(
+                    timestamp, None if res is None else self._create_method(res)
+                )
```
ela-kotulska-frequenz, line 76: The `evaluate` method can returns `math.nan` (from Div node). 
And you know that only if you read the code.

Maybe write in documentation that `_create_method`,  shoudl cover`math.nan` that case?

src/frequenz/sdk/timeseries/formulas/_formula_evaluator.py
```diff
@@ -0,0 +1,129 @@
...
+            )
+            if ex := next((e for e in fetch_results if isinstance(e, Exception)), None):
+                if isinstance(ex, (StopAsyncIteration, ReceiverStoppedError)):
+                    _logger.debug(
+                        "input streams closed; stopping formula evaluator. (%s)",
+                        self._root,
+                    )
+                    await self._output_channel.close()
+                    return
+                raise ex
```
ela-kotulska-frequenz, line 104: Could you explain these part of the code?
You first calculate and send the result, and then check if streams are still open.
Why not first check and then (if everything works) calculate the result and send it?

src/frequenz/sdk/timeseries/formulas/_formula.py
```diff
@@ -0,0 +1,383 @@
...
+            #     f"Formula evaluator for '{self._root}' is not running. Please "
+            #     + "call `start()` on the formula before using it.",
+            # )
+            # _logger.warning(
+            #     "Formula evaluator for '%s' is not running. Starting it.  "
+            #     + "Please call `start()` on the formula before using it."
+            #     self._root,
+            # )
+            self.start()
+        return self._channel.new_receiver(limit=limit)
```
ela-kotulska-frequenz, line 93: remove commented code :)

src/frequenz/sdk/timeseries/formulas/_formula_evaluator.py
ela-kotulska-frequenz: explained during the meeting

src/frequenz/sdk/timeseries/formulas/_formula_evaluator.py
ela-kotulska-frequenz: Explained

src/frequenz/sdk/timeseries/formulas/_ast.py
```diff
@@ -0,0 +1,205 @@
...
+        """Evaluate the expression and return its numerical value."""
+
+    @abc.abstractmethod
+    def format(self, wrap: bool = False) -> str:
+        """Return a string representation of the node."""
+
+    @override
+    def __str__(self) -> str:
+        """Return the string representation of the node."""
+        return self.format()
```
llucax, line 41: Idea for the future: add `__format__()` with some format modifier so we can do something like f`{node:w}` to call `format(wrap=True)`.

src/frequenz/sdk/timeseries/formulas/_ast.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+@dataclass
+class TelemetryStream(Node, Generic[QuantityT]):
+    """A AST node that retrieves values from a component's telemetry stream."""
+
+    source: str
```
llucax, line 48: The properties are missing docs, I think mkdocs will not show them if they are undocumented (also it is not clear to me what `source` is).

src/frequenz/sdk/timeseries/formulas/_ast.py
```diff
@@ -0,0 +1,205 @@
...
+        """Return a string representation of the node."""
+
+    @override
+    def __str__(self) -> str:
+        """Return the string representation of the node."""
+        return self.format()
+
+
+@dataclass
+class TelemetryStream(Node, Generic[QuantityT]):
```
llucax, line 45: Could we use `float` here instead to make it simpler? Or do you need quantities because these streams will be somehow operated on and you already need, at this level, to know the quantity to disallow invalid operations between different quantities?

src/frequenz/sdk/timeseries/formulas/_ast.py
```diff
@@ -0,0 +1,205 @@
...
+        """Fetch the next value for this component and store it internally."""
+        self._latest_sample = await anext(self.stream)
+
+
+@dataclass
+class FunCall(Node):
+    """A function call in the formula."""
+
+    function: Function
+    args: list[Node]
```
llucax, line 81: Same about properties docs for the rest of the file.

src/frequenz/sdk/timeseries/formulas/_ast.py
```diff
@@ -0,0 +1,205 @@
...
+from ._functions import Function
+
+_logger = logging.getLogger(__name__)
+
+
+@dataclass
+class Node(abc.ABC):
+    """An abstract syntax tree node representing a formula expression."""
+
+    span: tuple[int, int] | None
```
llucax, line 28: Add docstring. What is `span`?

src/frequenz/sdk/timeseries/formulas/_ast.py
```diff
@@ -0,0 +1,205 @@
...
+from typing_extensions import override
+
+from ..._internal._math import is_close_to_zero
+from .._base_types import QuantityT, Sample
+from ._functions import Function
+
+_logger = logging.getLogger(__name__)
+
+
+@dataclass
```
llucax, line 24: I suggest making all these dataclasses `kw_only` for extra clarity. If not all, at least the ones taking 2 arguments of the same type, like `Add` (`left`, `right`).

src/frequenz/sdk/timeseries/formulas/_ast.py
llucax: I guess `Function` is the actual function, it would be like `value` in `Constant`, while `FunCall` is the calling of the function and the node in the AST.

src/frequenz/sdk/timeseries/formulas/_functions.py
```diff
@@ -0,0 +1,97 @@
...
+        """Create a function instance from its name."""
+        match name.upper():
+            case "COALESCE":
+                return Coalesce()
+            case "MAX":
+                return Max()
+            case "MIN":
+                return Min()
+            case _:
+                raise ValueError(f"Unknown function name: {name}")
```
llucax, line 37: I found this a bit confusing, as we usually use `from_xxx()` when we have a `__init__()`. I guess here you cannot use `__init__()` because this is a *factor* and needs to return a sub-type.

Any reason not to just make it a free function instead, like `parse_function(name: str) -> Function`?

I'm also fine to leave it as it is, I'm just curious about the choice.

src/frequenz/sdk/timeseries/formulas/_peekable.py
```diff
@@ -0,0 +1,53 @@
...
+
+from collections.abc import Iterator
+from typing import Generic, TypeVar
+
+from typing_extensions import override
+
+_T = TypeVar("_T")
+
+
+class Peekable(Generic[_T], Iterator[_T]):
```
llucax, line 16: Oh, `Peekable` is back! 🎉 

I think this is a good candidate for a new `itertools` module in [core](https://github.com/frequenz-floss/frequenz-core-python/).

There is also [`peekable`](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.peekable) (and [`spy`](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.spy)) in [`more-itertools`](https://more-itertools.readthedocs.io/en/stable/), which seems to be [widely used and well maintained](https://snyk.io/advisor/python/more-itertools) and we could consider using.

src/frequenz/sdk/timeseries/formulas/_peekable.py
```diff
@@ -0,0 +1,53 @@
...
+    """Create a Peekable iterator from an existing iterator."""
```
llucax, line 17: Maybe it is worth clarifying that the `Peekable` takes ownership of the iterator, otherwise if someone iterates it externally too it might *lose* data (removed from the iterator and stored in the internal buffer).

src/frequenz/sdk/timeseries/formulas/_peekable.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def __init__(self, iterator: Iterator[_T]):
+        """Initialize this instance.
+
+        Args:
+            iterator: The underlying iterator to wrap.
+        """
+        self._iterator: Iterator[_T] = iterator
+        self._buffer: _T | None = None
```
llucax, line 26: I think we should use a sentinel here, there is always the chance that a user wants to do `Peekable([1, None, 2])` and then we are doomed.

src/frequenz/sdk/timeseries/formulas/_lexer.py
```diff
@@ -0,0 +1,126 @@
...
+
+from collections.abc import Iterator
+
+from typing_extensions import override
+
+from . import _token
+from ._peekable import Peekable
+
+
+class Lexer(Iterator[_token.Token]):
```
llucax, line 17: This also doesn't render nicely in the docs, so I would recommend using `from ._token import Token` (unless this whole thing is internal, then it is fine.

src/frequenz/sdk/timeseries/formulas/_token.py
```diff
@@ -0,0 +1,71 @@
...
+
+    value: str
+    """The string value of the token."""
+
+
+@dataclass
+class Component(Token):
+    """An electrical component token."""
+
+    id: str
```
llucax, line 26: docstring

tests/timeseries/_formulas/test_lexer.py
```diff
@@ -0,0 +1,68 @@
...
+    close_paren = next(lexer)
+    assert isinstance(close_paren, _token.CloseParen)
+    assert close_paren.value == ")"
+    assert close_paren.span == (31, 31)
+
+    try:
+        _ = next(lexer)
+        assert False, "Expected StopIteration"
+    except StopIteration:
+        pass
```
llucax, line 68: `with pytest.raises(StopIteration)`?

tests/timeseries/_formulas/test_lexer.py
llucax, line 1: Maybe add at least one error case?

llucax:
FYI: I started reviewing :eyes:

src/frequenz/sdk/timeseries/formulas/_formula_evaluator.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+class FormulaEvaluatingActor(Generic[QuantityT], Actor):
+    """An evaluator for a formula represented as an AST."""
+
+    def __init__(  # pylint: disable=too-many-arguments
+        self,
+        *,
+        root: _ast.Node,
+        components: list[_ast.TelemetryStream[QuantityT]],
```
llucax, line 31: I got a bit confused about this name, maybe `streams`?

src/frequenz/sdk/timeseries/formulas/_formula_evaluator.py
llucax, line 104: I think you can potentially simplify this exception handling if you use a `TaskGroup`, something like:


```suggestion
            try:
                async with TaskGroup() as tg:
                    for comp in self._components:
                        tg.create_task(comp.fetch_next())
            except* (StopAsyncIteration, ReceiverStoppedError) as exc:
                _logger.debug(
                    "input streams closed; stopping formula evaluator. (%s): %s",
                    self._root,
                    exc,
                )
                await self._output_channel.close()
                # no return here, otherwise it will interrupt the handling of other exceptions
            if self._output_channel.is_closed:
                return
```

This should create one task per component to fetch and wait until all finished if there are no errors.

If there are errors, all other pending tasks will be cancelled **immediately** (gather will wait until all awaitables finish then `return_exceptions=True`).

If all errors are stop errors, it will log and exit. If there are any other errors, the exception group will bubble up (be re-raised).

src/frequenz/sdk/timeseries/formulas/_formula_evaluator.py
```diff
@@ -0,0 +1,129 @@
...
+                raise ex
+
+
+async def synchronize_receivers(
+    components: list[_ast.TelemetryStream[QuantityT]],
+) -> None:
+    """Synchronize the given telemetry stream receivers."""
+    _ = await asyncio.gather(
+        *(comp.fetch_next() for comp in components),
+    )
```
llucax, line 113: Here since you don't need to catch exceptions, gather should work well, but maybe using a `TaskGroup` for consistency could be a good idea.

Thinking more about it, shouldn't the output channel be closed too if this happens? Maybe you need to do the exception handling in the whole `_run()` to cover for this.

src/frequenz/sdk/timeseries/formulas/_formula_evaluator.py
llucax, line 1: I generally wonder if adding a timeout wouldn't be a good idea while fetching, as an extra safety net, but I guess if we don't receive it means the resampler is not working and everything will break anyway, right?

src/frequenz/sdk/timeseries/formulas/_formula.py
```diff
@@ -0,0 +1,383 @@
...
+            sub_formula.start()
+        self._evaluator.start()
+
+    @override
+    async def stop(self, msg: str | None = None) -> None:
+        """Stop the formula evaluator."""
+        await BackgroundService.stop(self, msg)
+        for sub_formula in self._sub_formulas:
+            await sub_formula.stop(msg)
+        await self._evaluator.stop(msg)
```
llucax, line 108: You should probably override `cancel()` + `await()` instead of `stop()`.

tests/timeseries/_formulas/test_formulas.py
```diff
@@ -0,0 +1,651 @@
...
+        """Test simple formulas."""
+        await self.run_test(
+            "#2 - #4 + #5",
+            "[f](#2 - #4 + #5)",
+            [
+                ([10.0, 12.0, 15.0], 13.0),
+                ([15.0, 17.0, 20.0], 18.0),
+            ],
+        )
+        await self.run_test(
```
llucax, line 85: I would convert this into `pytest.mark.parametrize()` instead of writing tests sequentially.

If you do so, `pytest` will tell you exactly which test case failed, and keep running other cases in the same "group" instead of aborting on the first failure. For example:

```py
@dataclass
FormulaCase:
        formula_str: str
        expected: str
        io_pairs: list[tuple[list[float | None], float | None]]

@pytest.mark.parametrize(
    "case",
    [
        FormulaCase(
            "#2 - #4 + #5",
            "[f](#2 - #4 + #5)",
            [
                ([10.0, 12.0, 15.0], 13.0),
                ([15.0, 17.0, 20.0], 18.0),
            ],
        ),
        ...,
    ],
    id=lambda c: c.formula_str,
)
async def test_simple(self, case: FormulaCase) -> None:
    await self.run_test(case)
```

src/frequenz/sdk/timeseries/formulas/_formula_3_phase.py
llucax, line 1: As I said, I'm fine with merging this as is, but I think we need to generalize it to something like `FormulaTuple` or something, where we can chose which metrics to put as a tupe, so we can also use it for metrics that have phase-to-phase variants (so are a 6-values tuple).

See:

* https://github.com/frequenz-floss/frequenz-sdk-python/discussions/988

`FormulaTuple` could eventually get more operations, including operations to reduce it to one value, like `TUPLE_MIN(t) -> 1` where `t = (1, 2, 3)` (we could have `AVG`, maybe `TUPLE_AT(t, 2) -> 3` and stuff like that (or introduce some sort of method and indexing notation, like `t.MIN()`, `t[2]`, etc).

src/frequenz/sdk/timeseries/formulas/_formula_3_phase_evaluator.py
llucax, line 1: Same about generalization.

tests/timeseries/_formulas/test_formulas_3_phase.py
```diff
@@ -0,0 +1,153 @@
...
+                    assert phase_value.base_value == expected_value
+
+    async def test_composition(self) -> None:
+        """Test a simple 3-phase formula."""
+        await self.run_test(
+            [
+                ([(1.0, 2.0, 3.0), (2.0, 3.0, 4.0)], (6.0, 10.0, 14.0)),
+                ([(-5.0, 10.0, 15.0), (10.0, 15.0, -20.0)], (10.0, 50.0, -10.0)),
+                ([(None, 2.0, 3.0), (2.0, None, 4.0)], (None, None, 14.0)),
+            ],
```
llucax, line 113: Same comment about `parametrize()`.

src/frequenz/sdk/timeseries/formulas/_formula_pool.py
```diff
@@ -0,0 +1,282 @@
...
+
+NON_EXISTING_COMPONENT_ID = sys.maxsize
+"""The component ID for non-existent components in the components graph.
+
+The non-existing component ID is commonly used in scenarios where a formula
+requires a component ID but there are no available components in the graph to
+associate with it. Thus, the non-existing component ID is subscribed instead so
+that the formula can send `None` or `0` values at the same frequency as the
+other streams.
+"""
```
llucax, line 35: Can we find a better solution for this? I already bumped into this weird CID in production and it was very confusing. Can't you just check the component graph and ignore any component that doesn't exist?

If not, what about extending the resampler to cover for this specific use case? Like allowing the component ID to be `None` and just send `None`s for the `None` component?

src/frequenz/sdk/timeseries/formulas/_formula_pool.py
llucax, line 1: I wonder if this can be made more generic, like being able to create formulas for arbitrary metrics (and associated quantity) instead.

src/frequenz/sdk/timeseries/formulas/_ast.py
shsms: Because functions are one more layer removed, there's a FunCall node, then the Function.  Nice to group actual function calls like that, but don't have to do that for the fundamental operations.

src/frequenz/sdk/timeseries/formulas/_formula.py
shsms: Done at the end of my brand new PR that's based on this one: https://github.com/frequenz-floss/frequenz-sdk-python/pull/1322

src/frequenz/sdk/_internal/_graph_traversal.py
llucax, line 1: So far `frequenz.sdk._internal` is used by super generic stuff, any reason not to put this in `frequenz.sdk.microgrid` instead?

src/frequenz/sdk/_internal/_graph_traversal.py
```diff
@@ -0,0 +1,241 @@
...
+    Args:
+        component: The component to check.
+
+    Returns:
+        `True` if the component is an EV charger, `False` otherwise.
+    """
+    return isinstance(component, EvCharger)
+
+
+def is_battery_chain(
```
llucax, line 72: For functions taking a `ComponentGraph`, wouldn't it make sense to create a new `ComponentGraph` class in the SDK that inherit from `frequenz.microgrid_component_graph.ComponentGraph` and adds the new methods to keep the OO interfacce and not having half the stuff as method and half as free functions?

src/frequenz/sdk/_internal/_graph_traversal.py
```diff
@@ -0,0 +1,241 @@
...
+
+def is_ev_charger(component: Component) -> bool:
+    """Check if the component is an EV charger.
+
+    Args:
+        component: The component to check.
+
+    Returns:
+        `True` if the component is an EV charger, `False` otherwise.
+    """
```
llucax, line 68: Any reasons for these wrappers? Why not just using `isinstance()` directly?

pyproject.toml
```diff
@@ -30,12 +30,12 @@ dependencies = [
   # changing the version
   # (plugins.mkdocstrings.handlers.python.import)
   "frequenz-client-microgrid >= 0.18.0, < 0.19.0",
+  "frequenz-microgrid-component-graph >= 0.2.0, < 0.3",
```
llucax, line 33: Ideally I would like to add only dependencies that are at `> 1.0.0` in the SDK. The v0.x approach proved to be confusing and error prone. Can we make a 1.0 release for `frequenz-microgrid-component-graph` soon? I guess this is only expected to be used inside the SDK, so it is not that bad, but still. I think we can still do a lot of incremental improvements with backwards compatible changes, and at some point it is fine to bump to v2.0.0 if we need to do some cleanup.

pyproject.toml
llucax, line 74: This should be removed too.

tests/microgrid/test_grid.py
```diff
@@ -5,56 +5,28 @@
...
-from tests.utils.graph_generator import GraphGenerator
 
-from ..timeseries._formula_engine.utils import equal_float_lists, get_resampled_stream
+from ..timeseries._formulas.utils import equal_float_lists, get_resampled_stream
 from ..timeseries.mock_microgrid import MockMicrogrid
 
 _MICROGRID_ID = MicrogridId(1)
 
 
-async def test_grid_1(mocker: MockerFixture) -> None:
```
llucax, line 32: I guess when getting island support back the test will change, but an alternative is to skip using `@pytest.mark.skip("reason")`.

tests/microgrid/test_grid.py
```diff
@@ -64,7 +36,9 @@ async def test_grid_2(mocker: MockerFixture) -> None:
     components = {grid_1, meter_2}
     connections = {ComponentConnection(source=grid_1.id, destination=meter_2.id)}
 
-    graph = gr._MicrogridComponentGraph(  # pylint: disable=protected-access
+    graph = gr.ComponentGraph[
+        Component, ComponentConnection, ComponentId
+    ](  # pylint: disable=protected-access
```
llucax, line 41: Not needed anymore, right?
```suggestion
    ](
```

src/frequenz/sdk/microgrid/connection_manager.py
```diff
@@ -51,7 +57,9 @@ def api_client(self) -> MicrogridApiClient:
 
     @property
     @abstractmethod
-    def component_graph(self) -> ComponentGraph:
+    def component_graph(
+        self,
+    ) -> ComponentGraph[Component, ComponentConnection, ComponentId]:
```
llucax, line 62: Maybe it is worth adding a `TypeAlias`, but I guess if we do the suggestion about having a component graph in the SDK that inherits from the rust component graph, we can also already specialize it. So:

```py
class ComponentGraph(microgrid_component_graph.ComponentGraph[Component, ComponentConnection, ComponentId]):
    # extra methods
```

Then all these type hints also get much shorter.

src/frequenz/sdk/microgrid/connection_manager.py
```diff
@@ -156,7 +173,11 @@ async def _update_client(self, server_url: str) -> None:
 
     async def _initialize(self) -> None:
         self._microgrid = await self._client.get_microgrid_info()
-        await self._graph.refresh_from_client(self._client)
+        components, connections = await asyncio.gather(
+            self._client.list_components(),
+            self._client.list_connections(),
+        )
+        self._graph = ComponentGraph(set(components), set(connections))
```
llucax, line 180: If you want to gether even further :P
```suggestion
        self._microgrid, components, connections = await asyncio.gather(
            self._client.get_microgrid_info()
            self._client.list_components(),
            self._client.list_connections(),
        )
        self._graph = ComponentGraph(set(components), set(connections))
```

src/frequenz/sdk/timeseries/_grid_frequency.py
```diff
@@ -54,7 +55,8 @@ def __init__(
         """
         if not source:
             component_graph = connection_manager.get().component_graph
-            source = component_graph.find_first_descendant_component(
+            source = find_first_descendant_component(
+                component_graph,
```
llucax, line 59: Here if we go for the SDK `ComponentGraph` inheriting from the rust one, we don't need to update all these method calls, no need for new imports, no need for code updates.

tests/utils/component_graph_utils.py
llucax, line 1: This file also would not need any changes if we subclass, etc. But I'll stop now :P

tests/microgrid/test_graph.py
llucax, line 1: I guess all these tests are done in the `frequenz-microgrid-component-graph` repo, right?

src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py
```diff
@@ -90,11 +91,11 @@ def start_formula(
         self,
         formula: str,
         metric: Metric,
-        *,
-        nones_are_zeros: bool = False,
-    ) -> FormulaEngine[Quantity]:
+    ) -> Formula[Quantity]:
         """Start execution of the given formula.
 
+        TODO: link to formula syntax.
```
llucax, line 97: Maybe create an issue for this instead of leaving a TODO in the docs?

tests/timeseries/_battery_pool/test_battery_pool.py
```diff
@@ -75,7 +81,7 @@ def get_components(
     return {
         component.id
         for component in mock_microgrid.component_graph.components(
-            matching_types=component_type
+            matching_types=[component_type]
```
llucax, line 84: Oh, the rust component graph doesn't support passing a single element yet 😢

tests/timeseries/_battery_pool/test_battery_pool.py
```diff
@@ -589,118 +595,28 @@ async def test_batter_pool_power_two_batteries_per_inverter(
...
     mockgrid = MockMicrogrid(
-        graph=GraphGenerator().to_graph(
+        graph=graph_gen.to_graph(
             (
                 ComponentCategory.METER,
-                [ComponentCategory.INVERTER, ComponentCategory.INVERTER],
+                [
+                    graph_gen.component(ComponentCategory.INVERTER, InverterType.SOLAR),
+                    graph_gen.component(ComponentCategory.INVERTER, InverterType.SOLAR),
+                ],
```
llucax, line 619: Why 2 inverters now?

tests/timeseries/_formulas/__init__.py
llucax, line 1: I think this one was added in the wrong commit, but just FYI, it is fine to leave as it is for me.

tests/timeseries/test_logical_meter.py
```diff
@@ -77,69 +79,3 @@ async def test_pv_power_no_pv_components(self, mocker: MockerFixture) -> None:
 
             await mockgrid.mock_resampler.send_non_existing_component_value()
             assert (await pv_power_receiver.receive()).value == Power.zero()
-
-    async def test_pv_power_with_failing_meter(self, mocker: MockerFixture) -> None:
```
llucax, line 81: I also guess you removed fallback tests because these are done in the component graph/formula rust-based repos?

tests/timeseries/_battery_pool/test_battery_pool.py
```diff
@@ -510,15 +516,15 @@ async def test_battery_pool_power(mocker: MockerFixture) -> None:
         # send meter power [grid_meter, battery1_meter, battery2_meter]
         await mockgrid.mock_resampler.send_meter_power([100.0, 2.0, 3.0])
         await mockgrid.mock_resampler.send_bat_inverter_power([20.0, 30.0])
-        assert (await power_receiver.receive()).value == Power.from_watts(5.0)
+        assert (await power_receiver.receive()).value == Power.from_watts(50.0)
```
llucax, line 519: About these changes, maybe it would be nice to explain in the commit message **why** this was actually changed, it says what changed but not why and it is not entirely clear to me, as meters are usually more precise and have better data rates, right?

> Test priority of component powers in formulas over meter powers
> 
> The new formulas use component metrics as primary and only when those
> are not available, uses any corresponding meters as fallback.

src/frequenz/sdk/timeseries/formulas/__init__.py
llucax, line 1: If `FormulaEngine` and `FormulaEngine3Phase` were part of the public interface, and you say the interface didn't really changed (so the class were only *renamed*), can we keep a deprecated alias with the old names so we can make this a non-breaking change?

llucax (commented):
So, yeah, 45 comments 😬 

Of course most of it are minor and some are just about the same issue.

To summarize:
- Missing docstrings, I would add them to this PR, as it should be very low effort and important.
- `__format__()`, I would create an issue about it if you think it is a good idea, not necessary at all for this PR.
- `TelemetryStream` using float: Same, only discuss if it makes sense and if it does can be a separate issue.
- `@dataclass` with `kw_only`, I would do it for this PR, as it should also be low effort, but not very important so also fine to create an issue.
- `Function.from_string()` vs `parse_function`: Also curious to discuss, but OK to leave as it is.
- `Peekable`: I think support for `None` should be fixed in this PR, not sure if by using `peekable` from `more-itertools` or fix ours. I have mixed feelings, pulling a dependency for such a small thing seems overkill, but then this lib has a lot of other useful stuff, so maybe having it as a dependency maybe is not bad. Similar to [asyncstdlib](https://asyncstdlib.readthedocs.io/en/stable/index.html), at some point maybe it is good to have it in the toolbox by default.
- Docstring rendering fixes should probably be applied in this PR too
- `with pytest.raises(StopIteration)` also probably good to change in this PR, as it is low effort
- Adding error case to lexer, and having more tests in general (is it me or the new stuff is a bit undertested? Or is it tested indirectly?) could be done as separate issue/PR.
- Rename `components` -> `stream` can be done in this PR if you agree, but optional
- About using `TaskGroup`, I would push for it. According to Python docs of `gather()`:

    > Note
    >
    > A new alternative to create and run tasks concurrently and wait for their completion is [asyncio.TaskGroup](https://docs.python.org/3/library/asyncio-task.html#asyncio.TaskGroup). TaskGroup provides stronger safety guarantees than gather for scheduling a nesting of subtasks: if a task (or a subtask, a task scheduled by a task) raises an exception, TaskGroup will, while gather will not, cancel the remaining scheduled tasks).

- Overriding `cancel()` + `await()` instead of `stop()` is important, but if things work for now can also be done in a separate issue/PR.
- Use of `pytest.mark.parametrize` also of course optional, but I would do it (even if in a separate PR/issue), AI should do it well in a few seconds and error reporting will be better when things fail.
- `Formala3Phase`, also fine to leave as it is in this PR, but I think we need to take the opportunity in this refactoring to get it right ASAP.
- `NON_EXISTING_COMPONENT_ID` same, I would love to see it go, probably not in this PR, but I would like to use this refactoring impulse to get rid of it ASAP.
- Making somehow formulas more generic to take a metrici + quantity: definitely not for this PR, but I would like to discuss
- Move `_graph_traversal`, probably for this PR
- Having a `ComponentGraph` that inherits (and specializes) from the new rust component graph: I would definitely do it. I know many advantages are now pointless because you updated the code already, but still, I think it makes things much easier than this long type hints and the mix of methods and free functions we have in this PR.
- I would also remove the `is_xxx()` wrappers over `isinstance()` in this PR unless there is a very good reason to have them.
- Remove `types-networkx` in this PR.
- *Why 2 inverters now?* would be nice to know and document better
- And also try to keep the deprecated aliases `FormulaEngine` and `FormulaEngine3Phase`, if not in this PR add them later, but before the release so hopefully we can make it a non-breaking change.

There are a few other small comments and questions that are completely optional.

src/frequenz/sdk/timeseries/formulas/_ast.py
shsms: https://github.com/frequenz-floss/frequenz-sdk-python/issues/1324

src/frequenz/sdk/timeseries/formulas/_ast.py
shsms: All operations become QuantityT in the next PR, because we need `Sample`s with timestamp everywhere, so no more float.  Need to look into performance later.  Done here: https://github.com/frequenz-floss/frequenz-sdk-python/pull/1322/changes/7c827ae9e2c58e80fb52babe6240c9e52f5f55af

src/frequenz/sdk/timeseries/formulas/_ast.py
shsms: Will do docs in next PR, because some of these params have moved a bit.  span is only partially implemented.  It is the position of the node in the source string formula.

shsms: Will link here when done.

src/frequenz/sdk/timeseries/formulas/_ast.py
shsms: Done in the next PR already: https://github.com/frequenz-floss/frequenz-sdk-python/pull/1322/changes/dce0a8aa767330715cbae34792407dcf45e72315

src/frequenz/sdk/timeseries/formulas/_functions.py
shsms: It makes sense to me, but it can be changed separately if necessary,  please create an issue if you prefer something else.

src/frequenz/sdk/timeseries/formulas/_peekable.py
shsms: Just a synchronous thing made only for the formulas.  The previous parser had a peekable as well, but implemented differently.  This one is more abstract.

src/frequenz/sdk/timeseries/formulas/_peekable.py
shsms: It is just internal to formulas at the moment, if we ever pull it out, we can document it better.   But I've made an issue anyway: https://github.com/frequenz-floss/frequenz-sdk-python/issues/1325

src/frequenz/sdk/timeseries/formulas/_peekable.py
shsms: It is not used for types with `| None`, and is internal, so we don't need it for this PR, but I agree it needs to be fixed, I've made an issue: https://github.com/frequenz-floss/frequenz-sdk-python/issues/1326

src/frequenz/sdk/timeseries/formulas/_lexer.py
shsms: It is internal.

tests/timeseries/_formulas/test_lexer.py
shsms: True, didn't think of that.

tests/timeseries/_formulas/test_lexer.py
shsms: Error handling is very bad right now in the lexer and the parser.  And the spans are incorrect.  I'll pick that up next week.

src/frequenz/sdk/timeseries/formulas/_formula_evaluator.py
shsms: These are not needed after the next PR and are going away here: 623844291d506cd2a0c36fc49f2cd8b88bdfacdc

src/frequenz/sdk/timeseries/formulas/_formula_evaluator.py
shsms: This stuff has gone away from here in the next PR.  This might still be useful elsewhere, so I'll keep in mind.

src/frequenz/sdk/timeseries/formulas/_formula_evaluator.py
shsms: Yes, this is strongly tied to the resampler.

src/frequenz/sdk/timeseries/formulas/_formula_evaluator.py
shsms: This is what this function looks like in its latest form: https://github.com/frequenz-floss/frequenz-sdk-python/blob/623844291d506cd2a0c36fc49f2cd8b88bdfacdc/src/frequenz/sdk/timeseries/formulas/_formula_evaluator.py#L46-L71

src/frequenz/sdk/timeseries/formulas/_formula.py
shsms: made an issue here: https://github.com/frequenz-floss/frequenz-sdk-python/issues/1327

tests/timeseries/_formulas/test_formulas.py
shsms: I agree, these were copied over from the old formula engine.  Will have to make nice later.

src/frequenz/sdk/timeseries/formulas/_formula_3_phase.py
shsms: there are 3 phase-to-phase voltages, separate from 3 phase-to-neutral voltages.  They can't be grouped together into a 6 value tuple, because there's no standardized ordering between them.  But there is, in each of the 3 value groups.

shsms: We can still discuss this though, because there seem to be other reasons why we should do this.

src/frequenz/sdk/timeseries/formulas/_formula_3_phase_evaluator.py
shsms: Not sure which generalization you mean.

tests/timeseries/_formulas/test_formulas_3_phase.py
shsms: Answered above, this is just copying over old tests.  Need to rewrite later.

src/frequenz/sdk/timeseries/formulas/_formula_pool.py
shsms: When someone asks for a battery formula where there are no batteries, we need a timestamp source.  That's why this exists.

We shouldn't have to see this in production logs.  We should add an exception in the resampler, to not log if this component id doesn't get any data.

src/frequenz/sdk/timeseries/formulas/_formula_pool.py
shsms: Not sure what you have in mind,  but please feel free to make an issue.

src/frequenz/sdk/_internal/_graph_traversal.py
shsms: Maybe that's a good idea.  Only the `find_first_descendant_component` function remains, used only by stuff in the `timeseries` module.  Hopefully that can eventually be replaced with the coalesce formulas. for voltage and frequency.

But I'd leave it there for now, it will go away soon.

src/frequenz/sdk/_internal/_graph_traversal.py
shsms: these are no longer necessary, they were just there for the step-by-step transition to the new component graph.  They have been removed, only `find_first_descendant_component` remains, it will also go away.

src/frequenz/sdk/_internal/_graph_traversal.py
shsms: They have been removed.

pyproject.toml
shsms: We can, will look into it for the next release.

pyproject.toml
shsms: done here: a713f87a2af46038d4c1b828ead6637ed0a2d599

tests/microgrid/test_grid.py
shsms: That would make sense for a bug or for something that's in the pipeline already.  But I haven't heard anything about islands to think they're coming even in the medium-term.

src/frequenz/sdk/microgrid/connection_manager.py
shsms: As soon as the component types reach common,  the external type can use those and doesn't have to have any generics.

tests/microgrid/test_grid.py
shsms: Done here: a076b683afa74f6e0c8fa6b8d1ad1aa7c3f9eab4

src/frequenz/sdk/microgrid/connection_manager.py
shsms: Added an issue: https://github.com/frequenz-floss/frequenz-sdk-python/issues/1328

src/frequenz/sdk/timeseries/_grid_frequency.py
shsms: Actually the rust component graph should expose the required methods.  It does, for this one the `coalesce` formulas.  Need to look into how to upgrade and drop this one.

tests/microgrid/test_graph.py
shsms: Yup.  Also the formula generation tests.

src/frequenz/sdk/timeseries/formulas/_peekable.py
llucax: After seeing `more-itertools` I think we should keep it internal if we prefer not to add the dependency, pulling it out to `core` doesn't make sense if there is a stable alternative out there.

src/frequenz/sdk/timeseries/formulas/_peekable.py
llucax: If it is internal and we only specialized to `str`, then I would remove the genericity out of it, then it is clear `None` is not a valid value and only represents a missing value.

src/frequenz/sdk/timeseries/formulas/_formula_pool.py
llucax: Still, the solution seems pretty hacky to me, and technically someone could go pick `int.max` for a component ID, right? If we can have something more specific to get a timestamp source, I think we should do that instead. I will add it as a topic for a meeting.

src/frequenz/sdk/microgrid/connection_manager.py
llucax: oooohhhhhhhhh

src/frequenz/sdk/timeseries/formulas/_peekable.py
shsms: No, it is used to peek into the `Token` stream in the parser also.

src/frequenz/sdk/timeseries/logical_meter/_logical_meter.py
shsms: done here: 77b3541ee90a0b2047bf814a8d01bfe204015fde

tests/timeseries/_battery_pool/test_battery_pool.py
shsms: I added that, might have missed converting so back to without brackets.  Going to leave it as it is though.

tests/timeseries/_battery_pool/test_battery_pool.py
shsms: there were two already, both were in 1 line.  Now we're specifying type for both.

tests/timeseries/_formulas/__init__.py
shsms: Oops, is possible.

tests/timeseries/test_logical_meter.py
shsms: Yes, fallbacks are determined in the component graph.  And lazy subscription to fallback components is implemented and tested in the next PR.  There it becomes a formula feature, rather than specific to individual formula generators.

src/frequenz/sdk/timeseries/formulas/_peekable.py
llucax: OK, I think I would lean more towards using the one in `more-itertools` then.

src/frequenz/sdk/timeseries/formulas/_formula_3_phase_evaluator.py
llucax: The `Formula3Phase` vs `FormulaTuple`.

tests/timeseries/_battery_pool/test_battery_pool.py
shsms: This is because we decided meters could have unknown loads attached, so only if component is not available, we fallback to the meter.  This was planned here: https://github.com/frequenz-floss/frequenz-sdk-python/issues/1080 but it is no longer the responsibility of the SDK.

src/frequenz/sdk/timeseries/formulas/_formula_pool.py
llucax: I need to take some time to think about it, maybe it doesn't make sense. I need to see the code as whole to evaluate it, just an idea that came up to mind. But I think I was imagining something along this line:

```py
    def from_formula(
        self,
        channel_key: str,
        formula_str: str,
        value_ctor: Callable[[float], QuantityT],
        metric: Metric
    ) -> Formula[QuantityT]: ...
```

Instead of the many:
```py
    def from_power_formula(self, channel_key: str, formula_str: str) -> Formula[Power]: ...

    def from_reactive_power_formula(
        self, channel_key: str, formula_str: str
    ) -> Formula[ReactivePower]: ...
```

Etc.

src/frequenz/sdk/timeseries/formulas/__init__.py
shsms: Any ideas how?

With newer python, we could do `type FormulaEngine[W] = Formula[W]`.

src/frequenz/sdk/timeseries/formulas/_formula_pool.py
llucax: And at some point I think we should have a general mapping `Metric` -> `Quantity` subclass, so at some point passing only the metric should be enough.

src/frequenz/sdk/timeseries/formulas/__init__.py
shsms: Or I can just rename the thing back to FormulaEngine.  Just thought it was weird because technically we can call everything an engine, resampling engine, etc.

src/frequenz/sdk/timeseries/formulas/_ast.py
shsms: Done here: d86187fab21024070f3c8f76d6766c9f97c32ac0

shsms:
This is ready again @llucax .

And so is https://github.com/frequenz-floss/frequenz-sdk-python/pull/1322

src/frequenz/sdk/timeseries/formulas/__init__.py
llucax: I guess with [`TypeAlias`](https://typing.python.org/en/latest/spec/aliases.html), not sure if it works with generics too, but also not sure if deprecated works with it, so probably the easiest approach would be to use inheritance:

```py
@deprecated("...")
class FormulaEngine(Formula[T]): pass
```

Then you should be able to pass a `FormulaEngine` to every place a `Formula` is expected and get the deprecation message both in the interpreter and type checkers.

llucax (approved):
I think the backwards compatibility thing is really important, but we can do anytime before the release.

src/frequenz/sdk/timeseries/formulas/__init__.py
shsms: We can actually rename the whole thing back to `FormulaEngine`.  It was convenient to have two names during development, but `FormulaEngine` makes it more easily identifiable, different from the abstract concept of a formula.

PR #1299, opened 2025-11-24
Don't repeat zero commands to inverters

### LINKED ISSUE #1259: Battery power distributor shouldn't send 0W commands at 0% SoC
### What's needed?

When the SoC is 0%, the power manager throttles discharge requests to 0W, and the power distributor sends out 0W commands to the API.

It shouldn't send 0W commands when the SoC is 0%, so that it doesn't interfere with the battery's recovery.

Alternatively, it should remember if the previous request was for 0W and successful, and if so, it shouldn't repeat it at all.  Only non-0 commands should be repeated.

### Proposed solution

_No response_

### Use cases

_No response_

_No response_

### Additional context

_No response_

shsms:
Closes #1259

### DIFF
--- a/src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_battery_manager.py
+++ b/src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_battery_manager.py
@@ -183,6 +183,9 @@ class BatteryManager(ComponentManager):  # pylint: disable=too-many-instance-att
         )
         """The distribution algorithm used to distribute power between batteries."""
 
+        self._last_set_powers: dict[ComponentId, Power] = {}
+        """The last power value set to each battery's inverter."""
+
     @override
     def component_ids(self) -> collections.abc.Set[ComponentId]:
         """Return the set of component ids."""
@@ -638,7 +641,14 @@ class BatteryManager(ComponentManager):  # pylint: disable=too-many-instance-att
                 api.set_power(inverter_id, power.as_watts())
             )
             for inverter_id, power in distribution.distribution.items()
+            if power != Power.zero()
+            or self._last_set_powers.get(inverter_id) != Power.zero()
         }
+
+        if not tasks:
+            if _logger.isEnabledFor(logging.DEBUG):
+                _logger.debug("Not repeating 0W commands to batteries.")
+            return Power.zero(), set()
 
         if _logger.isEnabledFor(logging.DEBUG):
             battery_distribution = {
@@ -727,6 +737,8 @@ class BatteryManager(ComponentManager):  # pylint: disable=too-many-instance-att
             if failed:
                 failed_power += distribution[inverter_id]
                 failed_batteries.update(battery_ids)
+            else:
+                self._last_set_powers[inverter_id] = distribution[inverter_id]
 
         return failed_power, failed_batteries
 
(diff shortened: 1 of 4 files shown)


src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_battery_manager.py
```diff
@@ -271,21 +274,31 @@ async def _distribute_power(
         for inverter_id, dist in distribution.distribution.items():
             for battery_id in self._inv_bats_map[inverter_id]:
                 battery_ids.add(battery_id)
-            battery_distribution_logging[self._inv_bats_map[inverter_id]] = dist
+            if not (
+                dist == Power.zero()
+                and self._last_set_power.get(inverter_id) == Power.zero()
+            ):
```
cwasicki: Would it be possible to clean the distribution items first so that this check isn't needed in two places that are quite far apart.

cwasicki (commented):
LGTM

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_battery_manager.py
```diff
@@ -652,6 +665,12 @@ async def _set_distributed_power(
                 api.set_power(inverter_id, power.as_watts())
             )
             for inverter_id, power in distribution.distribution.items()
+            # Don't repeat setting zero power to inverters, to not interfere
+            # with any lower-level users of the inverters.
+            if not (
+                power == Power.zero()
+                and self._last_set_power.get(inverter_id) == Power.zero()
+            )
         }
```
ela-kotulska-frequenz: I thought actors set lower and upper power limit, to not interfere with lower-level actors. 
And if actor won't use battery, then it can set None (then lower level actor will have full control).

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_battery_manager.py
shsms: True.  The problem is that the distribution is created in a different class that doesn't have access to whether previous commands were successful or not.  And I don't want to modify the distribution object after it is created and returned by the other class.

But maybe it is less weird to pass the last_set_power information to that class each time.  I'll try that.

matthias-wende-frequenz: Good catch. That should be possible and safe. (as long as nobody starts using _set_distributed_power from another place)

```
if inverter_id in distribution.distribution:
    del distribution.distribution[inverter_id]
```

matthias-wende-frequenz (dismissed):

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_battery_manager.py
shsms: Yes, that's when they're using the battery pool, etc.  And the `None` requests are handled by the power manager.  Then the power manager sends a request to the power distributor which is what this PR is changing.

The power distributor can't send None values to the API, it has to send zero values.

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_battery_manager.py
shsms: I'll just move the log line to the place where the power is set, such that there are checks only in one place.

ela-kotulska-frequenz (dismissed):

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_battery_manager.py
ela-kotulska-frequenz: Ok thanks a lot!

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_battery_manager.py
shsms: Ok, ready for a look again

llucax (approved):
Just a pedantic comment: technically all classes implicitly inherit from `object`, so they have a parent :)

PR #1323, opened 2025-12-12
Make battery power manager algorithm configurable

shsms:
With this, `SHIFTING_MATRYOSHKA` remains the default algorithm for
batteries, but can now be changed to the original `MATRYOSHKA`
algorithm when calling `microgrid.initialize()`.

### DIFF
--- a/src/frequenz/sdk/microgrid/__init__.py
+++ b/src/frequenz/sdk/microgrid/__init__.py
@@ -383,6 +383,7 @@ from ._data_pipeline import (
     producer,
     voltage_per_phase,
 )
+from ._power_managing import PowerManagerAlgorithm
 
 
 async def initialize(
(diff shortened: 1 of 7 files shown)


shsms:
This means we need to adjust the docs slightly as well, but will pick that up separately later.

src/frequenz/sdk/microgrid/__init__.py
```diff
@@ -383,13 +383,16 @@
...
+from ._power_managing import PowerManagerAlgorithm
 
 
 async def initialize(
     server_url: str,
     resampler_config: ResamplerConfig,
     *,
     api_power_request_timeout: timedelta = timedelta(seconds=5.0),
+    # pylint: disable-next: line-too-long
+    battery_power_manager_algorithm: PowerManagerAlgorithm = PowerManagerAlgorithm.SHIFTING_MATRYOSHKA,  # noqa: E501
```
llucax, line 395: I guess this is needed somehow soon, but in the future I think it might make more sense to have copy what we have with the `resampler_config and add a `power_manager_config: PowerManagerConfig`, and in there it can have something like `power_manager_config.battery.algorithm` for example.

llucax (approved):
If not too urgent, I would add a `power_manager_config: PowerManagerConfig` instead from the start, otherwise I'm fine with merge as is for now.

cwasicki (approved):

src/frequenz/sdk/microgrid/__init__.py
shsms: https://github.com/frequenz-floss/frequenz-sdk-python/issues/1329

PR #1322, opened 2025-12-11
Implement lazy subscription for Coalesce function

shsms:

PR #1330, opened 2025-12-16
Revert unreleased interface changes 

shsms:
These changes were introduced in the PRs #1295 and #1322, and are being reverted here.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,21 +1,11 @@
 # Frequenz Python SDK Release Notes
 
-## Summary
-
-<!-- Here goes a general summary of what this release is about -->
-
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
-
-- The `FormulaEngine` is now replaced by a newly implemented `Formula` type.  This doesn't affect the high level interfaces.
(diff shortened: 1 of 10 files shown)


shsms:
Based on #1322

shsms:
No release notes, because this is reverting unreleased changes.

Marenz (dismissed):
LGTM, only complain might be that the last commit also sneaks in dependency updates

shsms:
@Marenz the microgrid client is a patch update, and the component graph dep is only being introduced in this release cycle, so we can sneak it in.

Could I get another approval with the release notes?

ela-kotulska-frequenz (approved):

llucax (approved):

llucax:
> @Marenz the microgrid client is a patch update, and the component graph dep is only being introduced in this release cycle, so we can sneak it in.

I guess he means they are in an unrelated commit. I agree it would have been nicer to split it into their own commit(s) if they were not related/necessary to the changes in that commit. But just a minor thing, so approving.

shsms:
> I agree it would have been nicer to split it into their own commit(s) if they were not related/necessary to the changes in that commit.

They absolutely are.  Graph repo had new subclassability feature needed for that commit.  Graph update also includes wind turbines, for which we need client 18.1.  So both are related to that commit.

PR #1340, opened 2026-01-05
Validate component IDs when creating component pools

shsms:


src/frequenz/sdk/timeseries/battery_pool/_battery_pool_reference_store.py
```diff
@@ -81,12 +81,24 @@ def __init__(  # pylint: disable=too-many-arguments
...
+        Raises:
+            ValueError: If any of the specified batteries is not present in the
+                microgrid.
         """
         self._batteries: frozenset[ComponentId]
+        all_batteries = self._get_all_batteries()
         if batteries_id:
             self._batteries = frozenset(batteries_id)
+            if not self._batteries.issubset(all_batteries):
+                unknown_ids = self._batteries - all_batteries
```
llucax, line 94: You could save a line of code and a couple cycles by reversing this:
```suggestion
            if unknown_ids := self._batteries - all_batteries:
```
(same for the rest)

RELEASE_NOTES.md
```diff
@@ -1,5 +1,17 @@
 # Frequenz Python SDK Release Notes
 
-## Bug fixes
+## Summary
 
-- `FormulaEngine` and `FormulaEngine3Phase` are now type aliases to `Formula` and `Formula3Phase`, fixing a typing issue introduced in `v1.0.0-rc2202`.
+<!-- Here goes a general summary of what this release is about -->
+
+## Upgrading
```
llucax, line 7: I would also add a note here, as this is effectively a breaking change. Users will need to update their code to deal with the exception, or validate the IDs before passing them to `new_xxx_pool()`.

llucax (commented):
Assigning to rc2300 because this is a breaking change. Also I think we need to mention it in the upgrading section of the release notes.

src/frequenz/sdk/timeseries/battery_pool/_battery_pool_reference_store.py
shsms: that doesn't look like a subset check.

shsms:
It is of course not a breaking change, because it was broken much worse earlier when they sent non-battery to battery pool.  This is a more predictable meaningful error.  Just a bug fix.

llucax:
It is a breaking change in the sense that raising a new exception is an interface change, but I see your point that passing the wrong ID would also make the user code explode, only in a different and worse way, so OK. Rolling-back the breaking change label and milestone.

llucax (dismissed):

llucax:
Added a commit to add the missing `Raises` section to the docstrings of methods now raising a  `ValueError`.

Needs approval from someone else now. Enabling auto-merge too, as Sahas will be away for a few weeks.

llucax:
Tests still failing for the same reasons as #1343, see comments in that PR for details.

Marenz (dismissed):

llucax:
Rebased to get the CI fix, and needed to merge some conflicts for that, so it needs a new approval.

ela-kotulska-frequenz (approved):

PR #1343, opened 2026-01-12
Fix off-by-one in `OrderedRingBuffer.count_covered` and add consistency test

malteschaaf:
### Summary

This PR fixes a subtle off-by-one issue in the `OrderedRingBuffer.count_covered` method and adds a test to ensure window count consistency in `MovingWindow`.  

### Changes

1. **Fix `count_covered` calculation**  
   - Switched from floating-point division to integer `timedelta` division (`timedelta // timedelta`).  
   - This eliminates rounding errors in sample count calculations and ensures accurate coverage for all window sizes and sampling periods.

2. **Add `test_moving_window_length`**  
   - New test verifies that the moving window length is consistent without resampling.  
   - Confirms that the count of covered samples matches the expected window capacity.  
   - Helps prevent regressions related to off-by-one errors in the future.

### Motivation

Floating-point division could produce off-by-one errors for certain window sizes and sampling periods, causing `count_covered` to be inaccurate. Switching to integer-based `timedelta` division ensures deterministic and exact counts. The new test provides coverage to catch similar issues.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -14,4 +14,4 @@
 
 ## Bug Fixes
 
-<!-- Here goes notable bug fixes that are worth a special mention or explanation -->
+- Fixed an off-by-one calculation in `OrderedRingBuffer.count_covered` by switching to integer timedelta division, ensuring accurate sample counting for all window sizes and sampling periods.
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,17 @@
 # Frequenz Python SDK Release Notes
 
-## Bug fixes
+## Summary
 
-- `FormulaEngine` and `FormulaEngine3Phase` are now type aliases to `Formula` and `Formula3Phase`, fixing a typing issue introduced in `v1.0.0-rc2202`.
+<!-- Here goes a general summary of what this release is about -->
+
+## Upgrading
+
+<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+
+## New Features
+
+<!-- Here goes the main new features and examples or instructions on how to use them -->
+
+## Bug Fixes
+
+<!-- Here goes notable bug fixes that are worth a special mention or explanation -->
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ dependencies = [
   # changing the version
   # (plugins.mkdocstrings.handlers.python.import)
   "frequenz-client-microgrid >= 0.18.1, < 0.19.0",
-  "frequenz-microgrid-component-graph >= 0.3.2, < 0.4",
+  "frequenz-microgrid-component-graph >= 0.3.2, < 0.3.4",
   "frequenz-client-common >= 0.3.6, < 0.4.0",
   "frequenz-channels >= 1.6.1, < 2.0.0",
   "frequenz-quantities[marshmallow] >= 1.0.0, < 2.0.0",
--- a/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
+++ b/src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
@@ -705,10 +705,7 @@ class OrderedRingBuffer(Generic[FloatArray]):
             The count of samples between the oldest and newest (inclusive) valid samples
                 or 0 if there are is no time range covered.
         """
-        return int(
-            self._covered_time_range(since, until).total_seconds()
-            // self._sampling_period.total_seconds()
-        )
+        return self._covered_time_range(since, until) // self._sampling_period
 
     def count_valid(
         self, *, since: datetime | None = None, until: datetime | None = None
--- a/tests/timeseries/test_moving_window.py
+++ b/tests/timeseries/test_moving_window.py
@@ -560,6 +560,33 @@ async def test_resampling_window(fake_time: time_machine.Coordinates) -> None:
             assert 4.9 < value < 5.1
 
 
+async def test_moving_window_length(fake_time: time_machine.Coordinates) -> None:
+    """Test moving window length without resampling."""
+    channel = Broadcast[Sample[Quantity]](name="net_power")
+    sender = channel.new_sender()
+
+    window_size = timedelta(seconds=1)
+    input_sampling = timedelta(seconds=0.1)
+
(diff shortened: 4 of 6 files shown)


cwasicki (dismissed):

src/frequenz/sdk/timeseries/_ringbuffer/buffer.py
```diff
@@ -705,10 +705,7 @@ def count_covered(
             The count of samples between the oldest and newest (inclusive) valid samples
                 or 0 if there are is no time range covered.
         """
-        return int(
-            self._covered_time_range(since, until).total_seconds()
-            // self._sampling_period.total_seconds()
-        )
+        return int((self._covered_time_range(since, until) // self._sampling_period))
```
llucax: Nitpick: now `int()` can be dropped, integer division of timedeltas is integer (as it divides `int`s as pointed out in the commit message).
```suggestion
        return self._covered_time_range(since, until) // self._sampling_period
```

We can still keep it for extra clarity, but it is not necessary, strictly speaking.

tests/timeseries/test_moving_window.py
```diff
@@ -560,6 +560,34 @@ async def test_resampling_window(fake_time: time_machine.Coordinates) -> None:
             assert 4.9 < value < 5.1
 
 
+# pylint: disable=redefined-outer-name
+async def test_moving_window_length(fake_time: time_machine.Coordinates) -> None:
```
llucax:
```suggestion
async def test_moving_window_length(
        fake_time: time_machine.Coordinates  # pylint: disable=redefined-outer-name
    ) -> None:
```

So it is not disabled for the whole file.

RELEASE_NOTES.md
```diff
@@ -3,3 +3,4 @@
 ## Bug fixes
 
 - `FormulaEngine` and `FormulaEngine3Phase` are now type aliases to `Formula` and `Formula3Phase`, fixing a typing issue introduced in `v1.0.0-rc2202`.
+- Fixed an off-by-one calculation in `OrderedRingBuffer.count_covered` by switching to integer timedelta division, ensuring accurate sample counting for all window sizes and sampling periods.
```
llucax: We forgot to clear the release notes since last release. Can you add a commit to the beginning resetting the release notes? (`cp .github/RELEASE_NOTES.template.md RELEASE_NOTES.md`)?

llucax (commented):
Thanks a lot! LGTM, only a couple of minor comments.

Also the CI is not passing, it is weird because the issues look unrelated to this PR, but maybe you can fix them (removing the `type: ignore` should be enough, maybe mypy or numpy got some hidden improvement in some patch release).

```
nox > mypy 
src/frequenz/sdk/timeseries/_periodic_feature_extractor.py:412: error: Unused "type: ignore" comment  [unused-ignore]
benchmarks/timeseries/periodic_feature_extractor.py:75: error: Unused "type: ignore" comment  [unused-ignore]
```

tests/timeseries/test_moving_window.py
malteschaaf: Seems like it's not needed at all...

cwasicki (dismissed):

llucax:
Very weird, now tests are failing in some apparently unrelated tests (I don't think these tests use the moving window). :thinking:

llucax:
I just re-run the tests and they fail the same, so that's *good news* (I was more scare about this being a flaky failure). I guess the next step would be to verify tests pass without these changes. If the failure is consistent, then probably these changes are somehow affecting those tests and we need to find why and fix it.

@malteschaaf are the tests passing locally?

malteschaaf:
> @malteschaaf are the tests passing locally?

The tests fail locally as well but no matter the change (they fail with and without any of my changes)

llucax:
OK, this is definitely some indirect dependency bump issue. Locally using `nox -s pytest_min` works but `nox -s pytest_max` doesn't. Probably the best way to approach this is to see in the CI what was the last successful build and see the output for the `pip freeze` command to get a list of dependencies. Then do the same in one failing build and then compare the two to see what are the differences and find which one is breaking stuff.

@malteschaaf do you think you can take care of this?

malteschaaf:
Seems like `frequenz-microgrid-component-graph` v0.3.4 caused this

I'm not sure how to proceed from here though...
I feel like just pushing a commit restricting the max version of `frequenz-microgrid-component-graph` isn't the best way.

cwasicki (approved):

llucax:
> I feel like just pushing a commit restricting the max version of `frequenz-microgrid-component-graph` isn't the best way.

I'll have a look to see if there is anything obvious in the new version that is wrong. Thanks for the research!

llucax:
OK, it looks like v0.3.4 introduced breaking changes (coming from the [rust library v0.4.0](https://github.com/frequenz-floss/frequenz-microgrid-component-graph-rs/releases/tag/v0.4.0)):

> ### Upgrading
> - The PV, battery, CHP, EV and Wind Turbine formulas now prefer meters as the primary components and fallback to inverters or other corresponding components only when the meters are not available.
> 
>   This makes a big difference in performance when there are multiple PV inverters behind a single meter, for example.
> 
>   This behaviour can be changed with the newly introduced meter preference config flags.
> 
> - Consumer formulas don't consider phantom loads by default anymore.  The original behaviour is still available through a config flag.

I will mark that release as broken and wait for Sahas to come back to see how we move forward (we probably need to re-release it as v0.4.0 and then update the SDK tests to pass using the new logic).

pyproject.toml
```diff
@@ -30,7 +30,7 @@ dependencies = [
   # changing the version
   # (plugins.mkdocstrings.handlers.python.import)
   "frequenz-client-microgrid >= 0.18.1, < 0.19.0",
-  "frequenz-microgrid-component-graph >= 0.3.2, < 0.4",
+  "frequenz-microgrid-component-graph >= 0.3.2, < 0.3.4",
```
llucax, line 33: This should have not be committed 😭 

We are finally not going this route. I made a follow-up PR.

llucax: * https://github.com/frequenz-floss/frequenz-sdk-python/pull/1346

malteschaaf: Sorry 😬

llucax: No worries and keep the contributions coming!

PR #1344, opened 2026-01-14
Enhance ResamplerConfig with closed and label options and add comprehensive tests

malteschaaf:
This PR introduces the `WindowSide` enum and two new configuration options—`closed` and `label`—to the `ResamplerConfig` class, providing a clearer and more flexible way to define the behavior of the resampling process. Comprehensive tests have been added to ensure correctness.  

### Changes:

1. **Introduce `WindowSide` enum**
   - Added the `WindowSide` enum with values `LEFT` and `RIGHT`:
     - `LEFT`: Represents the start of a resampling window.
     - `RIGHT`: Represents the end of a resampling window.
   - Exposed `WindowSide` in the `__init__.py` of the timeseries module for public use.

2. **Add `closed` option to ResamplerConfig**
   - Introduced the `closed` parameter with options `WindowSide.RIGHT` (default) and `WindowSide.LEFT`.
   - Updated resampling logic to respect the `closed` configuration:
     - `RIGHT`: Includes samples at the end of the window, excludes those at the start.
     - `LEFT`: Includes samples at the start of the window, excludes those at the end.
   - Updated documentation to explain the behavior of the `closed` parameter.

3. **Add `label` option to ResamplerConfig**
   - Introduced the `label` parameter with options `WindowSide.RIGHT` (default) and `WindowSide.LEFT`.
   - Updated resampling logic to respect the `label` configuration:
     - `RIGHT`: Timestamp of the resampled data corresponds to the end of the resampling window.
     - `LEFT`: Timestamp of the resampled data corresponds to the start of the resampling window.
   - Adjusted logic for setting `sample_time` based on the `label` configuration.
   - Updated documentation accordingly.

4. **Add tests for `closed` and `label`**
   - Verified `closed` behavior with both `WindowSide.RIGHT` and `WindowSide.LEFT`, including additional sample points.
   - Verified `label` behavior to ensure timestamps of resampled data correspond correctly to the start or end of the window.
   - Added assertions to confirm correct resampling outputs and updates to source properties and buffer lengths.


src/frequenz/sdk/timeseries/_resampling/_config.py
```diff
@@ -94,15 +94,15 @@ class ResamplerConfig:
     It must be a positive time span.
     """
 
-    max_data_age_in_periods: float = 3.0
```
cwasicki, line 97: I agree that this value is leading to unexpected behavior, but this would be a very intrusive change which would result in many more NaN values in deployments. If we want to change it we could make it a required parameter for migration purpose, and later introduce the new default.

src/frequenz/sdk/timeseries/_resampling/_resampler.py
```diff
@@ -437,7 +437,7 @@ def resample(self, timestamp: datetime) -> Sample[Quantity]:
         )
         minimum_relevant_timestamp = timestamp - period * conf.max_data_age_in_periods
 
-        min_index = bisect(
```
cwasicki, line 440: I think the behavior how to resample, i.e. left or right open and the labeling should be config parameters.

src/frequenz/sdk/timeseries/_resampling/_config.py
malteschaaf: Sounds good.

src/frequenz/sdk/timeseries/_resampling/_resampler.py
malteschaaf: I think we should make left or right opened configurable with the corresponding label, such as:
right open: [t,t+1) -> labeled as t
left open: (t-1,t] -> labeled as t (the current behavior)
Or do you want to allow the user to additionally do something like:
right open, label in the end: [t,t+1) -> labeled as t+1
left open, label in the beginning:(t-1,t] -> labeled as t-1
Because I think the later could lead to a lot of confusion (not sure if its ever needed)

cwasicki: I think the latter is also reasonable options (see e.g. https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.resample.html), but don't see a strong reason to implement this now if not needed. If it's well-documented, the users can also adjust the timestamps trivially. So your proposal sounds good to me.

src/frequenz/sdk/timeseries/_resampling/_config.py
llucax: I would even first run some experiments before changing the API here. We can just change the value in the edge app for now, or even make it configurable via env vars (like the resampling period) to allow for easy, per-site, experimentation.

src/frequenz/sdk/timeseries/_resampling/_resampler.py
llucax: Yeah, whatever we do, we should probably be much more explicit of how output samples are calculated and structured.

src/frequenz/sdk/timeseries/_resampling/_config.py
llucax: Also, as Christoph pointed out, this will lead to more NaNs. A middle approach, that will avoid the NaNs and degrade the quality only when data is missing is to change the resampling function.

The resampler was designed to be very flexible, so we can keep the 3 windows of data, and check **inside** the resampling function, if there is data for the last window, if there is, we can limit the function to use only data from the last window, and if there isn't, we can go back in time to find some data.

This can even be a composable function, like `use_latest_window_only(average)`, so we can even exchange the underlying resampling function, basically separating buffer/history management from resampling.

Not sure if this is still useful from the DS point of view, or if it is best to just see a NaN if data is missing. From the control PoV, I think it makes sense to work on slightly old data rather than halting (or entering in emergency mode immediately). But even this is something we might want to discuss, as it wasn't reviewed since we first coded the resampler, and I think now we have a much better understanding of the use cases and requirements.

llucax (commented):
In general LGTM, I think this would need quite some testing, and we probably want to introduce this using a config option to change the behavior, specially if configurability is something we want to allow in the future (if we don't see having this level of configurability useful, we can hide it in a feature flag using a env var like we did with the new wall clock timer). As we experimented recently with the changes in the component graph, subtle changes in behavior can go easily unnoticed and cause issues.

src/frequenz/sdk/timeseries/_resampling/_config.py
malteschaaf: Sounds reasonable to me! I removed this change from this PR and we can discuss it further before any changes are made.

malteschaaf:
Ready for Review 🧐

**Changes implemented:**
- Removed modifications to the `max_data_age_in_periods` default value
- Made the openness of the resampling window configurable via the `closed` parameter:
    - `"right"`: Includes the end of the window, excludes the start. (Default: same behavior as before)
    - `"left"`: Includes the start of the window, excludes the end.
- Added configurability for timestamp labeling with the label parameter:
    - `"start"`: Resampled timestamp represents the start of the resampling window.
    - `"end"`: Resampled timestamp represents the end of the resampling window. (Default: same behavior as before)
- Added comprehensive tests to verify the behavior of the `closed` and `label` parameters.

src/frequenz/sdk/timeseries/_resampling/_config.py
```diff
@@ -126,6 +126,23 @@ class ResamplerConfig:
...
+    closed: Literal["right", "left"] = "right"
+    """Indicates which side of the resampling window is closed.
+
+    If "right", the resampling window includes samples with timestamps
+    equal to the end of the window, but not those equal to the start.
+    If "left", the resampling window includes samples with timestamps
+    equal to the start of the window, but not those equal to the end.
+    """
+
+    label: Literal["start", "end"] = "end"
```
cwasicki: Why not use "left" and "right" here, too?

malteschaaf: I thought I would use the same as pandas does but seems like i looked in the wrong line (convention uses start and end)

I will change it

malteschaaf:
Ready for Review 🧐

**Changes implemented:**
- Changed label from `Literal["start","end"]` to `Literal["left","right"]`

cwasicki (commented):
LGTM

src/frequenz/sdk/timeseries/_resampling/_config.py
```diff
@@ -126,6 +126,23 @@ class ResamplerConfig:
     value.
     """
 
+    closed: Literal["right", "left"] = "right"
```
daniel-zullo-frequenz: Have you considered defining/using an enum instead? I think an enum would be more suitable for comparisons and less error prone

malteschaaf: Yes I did and then saw that pandas is using this literal logic so i thought that might be the better way. Maybe someone needs to explain to me why it would be less error prone to use an enum

src/frequenz/sdk/timeseries/_resampling/_config.py
```diff
@@ -126,6 +126,23 @@ class ResamplerConfig:
...
+    If "left", the resampling window includes samples with timestamps
+    equal to the start of the window, but not those equal to the end.
+    """
+
+    label: Literal["right", "left"] = "right"
+    """Indicates the timestamp label of the resampled data.
+
+    If "right", the timestamp of the resampled data corresponds to the end
+    of the resampling window. If "left", the timestamp corresponds to
+    the start of the resampling window.
```
daniel-zullo-frequenz: I personally find it more clear to use start/end directly in the code rather than left/right  with the clarification mapping in the docs

malteschaaf: That's probably because it was named start/end before and when i switched to left/right for consistency between `closed` and `label` as @cwasicki suggested above, I didn't overthink the documentation again. I will adjust the docs.

malteschaaf:
**Ready for Review again** 

Changes:
- Intorduced `WindowSide` enum to represent the sides of a resampling window and used that for the configuration of the `closed` and `label` parameters.
- Updated the documentation for the `closed` and `label` config parameters, removing mapping `left` and `right` to "start" and "end"

daniel-zullo-frequenz (commented):
LGTM. I'll leave to Luca to review it and approve it

src/frequenz/sdk/timeseries/_resampling/_config.py
daniel-zullo-frequenz: we had a chat about it

llucax:
Sorry for the delay, will have a look now.

src/frequenz/sdk/timeseries/_resampling/_config.py
```diff
@@ -44,6 +45,16 @@
...
 
 
+class WindowSide(Enum):
+    """Represents a side of a resampling window."""
+
+    LEFT = "left"
+    """The left side of the resampling window."""
+
+    RIGHT = "right"
+    """The right side of the resampling window."""
```
llucax, line 55: Super nitpick: Why not `START` and `END`? For me it sounds like right and left might be more confusing, it depends on how the tuple (start,end) is defined. But if for (data) scientists left/right is the common nomenclature, let's go with it.

src/frequenz/sdk/timeseries/_resampling/_config.py
```diff
@@ -126,6 +137,29 @@ class ResamplerConfig:
...
+    within the range (start, end], where `start` and `end` are the boundaries of
+    the window.
+
+    If `WindowSide.LEFT`, the resampling window is closed on the left side and
+    open on the right, meaning it includes samples with timestamps within the
+    range [start, end), where `start` and `end` are the boundaries of the
+    window.
+    """
+
+    label: WindowSide = WindowSide.RIGHT
```
llucax, line 154: `label` also sounds weird, for me a label is text. I would find it more obvious if it is called something like `resampled_timestamp_side`.

src/frequenz/sdk/timeseries/_resampling/_resampler.py
```diff
@@ -437,12 +438,20 @@ def resample(self, timestamp: datetime) -> Sample[Quantity]:
...
+                key=lambda s: s[0],
+            )
+            max_index = bisect_left(self._buffer, timestamp, key=lambda s: s[0])
+        else:
+            min_index = bisect(
+                self._buffer,
+                minimum_relevant_timestamp,
+                key=lambda s: s[0],
+            )
+            max_index = bisect(self._buffer, timestamp, key=lambda s: s[0])
```
llucax: Please use `match`, so we get exhaustion checks (unlikely, I know, but if someone adds `MIDDLE` in the future and we forget to handle that, `match` will remind us). Also, it looks to me that the code for both branches is exactly the same, so you could reduce it to:

```suggestion
        match self._config.closed:
            case WindowSide.LEFT:
                bisect_fun = bisect_left
            case WindowSide.RIGHT:
                bisect_fun = bisect
            case unhandled:
                assert_never(unhandled)
            min_index = bisect_fun(
                self._buffer,
                minimum_relevant_timestamp,
                key=lambda s: s[0],
            )
            max_index = bisect_fun(self._buffer, timestamp, key=lambda s: s[0])
```

src/frequenz/sdk/timeseries/_resampling/_resampler.py
```diff
@@ -458,7 +467,13 @@ def resample(self, timestamp: datetime) -> Sample[Quantity]:
...
             else None
         )
-        return Sample(timestamp, None if value is None else Quantity(value))
+
+        sample_time = (
+            timestamp - conf.resampling_period
+            if self._config.label == WindowSide.LEFT
+            else timestamp
+        )
+        return Sample(sample_time, None if value is None else Quantity(value))
```
llucax:
```suggestion
        match self._config.label:
            case WindowSide.LEFT:
                timestamp -= conf.resampling_period
            case WindowSide.RIGHT:
                pass
            case unhandled:
                assert_never(unhandled)
            return Sample(timestamp, None if value is None else Quantity(value))
```

RELEASE_NOTES.md
```diff
@@ -12,7 +12,18 @@
 
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+- **`WindowSide` Enum**: Introduced the `WindowSide` enum to represent the sides of a resampling window:
+  - `LEFT`: Represents the start of the resampling window.
+  - `RIGHT`: Represents the end of the resampling window.
+  - The `WindowSide` enum is publicly available for import from the timeseries module.
```
llucax: I would remove this, this is just a helper class, no need to put it in the release notes.

RELEASE_NOTES.md
```diff
@@ -12,7 +12,18 @@
...
+
+- **`closed` Parameter**: Added the `closed` parameter to `ResamplerConfig`, allowing users to configure which side of the resampling window is included. This parameter uses the new `WindowSide` enum:
+  - `WindowSide.RIGHT`: Includes the end of the window, excludes the start. (Default)
+  - `WindowSide.LEFT`: Includes the start of the window, excludes the end.
+
+- **`label` Parameter**: Added the `label` parameter to `ResamplerConfig`, allowing users to configure the timestamp labeling of the resampled data. This parameter uses the new `WindowSide` enum:
+  - `WindowSide.RIGHT`: The resampled timestamp represents the end of the resampling window. (Default)
+  - `WindowSide.LEFT`: The resampled timestamp represents the start of the resampling window.
```
llucax: I would put more prominently that this affects the resampler, like:

> - `Resampler`: The resampler can now be configured to have the resampling window closed to the right (default) or left, and to also set the resampler timestamp to the right (default) or left end of the window being resampled. You can configure setting the new options `closed` and `label` in the `ResamplerConfig`.

llucax (commented):
LGTM, a few very nitpick comments, and one that is moderately important to me (using `match` instead of `if`). I know it might look irrelevant, but it looks to me like one of these "64k is enough for everyone" situation, where it seems impossible that anyone could add a new enum member... until someone does.

src/frequenz/sdk/timeseries/_resampling/_config.py
cwasicki: I suggested it above, to me that's less confusing since this is used in pandas https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.resample.html

src/frequenz/sdk/timeseries/_resampling/_config.py
cwasicki: Same, see https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.resample.html

malteschaaf:
READY FOR REVIEW

- Rebased
- Changed Release Notes according to comment
- `match` enum instead of `if`
- Updated the docstring of the `resample` method of the `StreamingHelper` class to better describe how the timestamp parameter is used in resampling calculations

llucax (approved):
🙏

malteschaaf:
Unfortunately I have no rights to merge myself @llucax

llucax:
Merged, thanks again for the great contribution!

PR #1352, opened 2026-02-02
Add auto-dependabot GitHub Action workflow

llucax:
This workflow was created by the repo-config migration script when upgrading to v0.14.0 but we missed to add it to the commit.


llucax:
Force-merging as this was added to many repos already.

PR #1359, opened 2026-02-19
Improve formula validation

### LINKED ISSUE #1356: Improve formula validation
### What's needed?

Formulas need to be fully validated and error messages must be clear even for non-technical users.

### Proposed solution

- Add any missing validation to the formula parser+lexer.
- Improve error messages.
- Add tests for all validation (currently missing).

### Use cases

Formulas defined by users who are unfamiliar with the formula grammar and functions.

_No response_

### Additional context

_No response_

simonvoelcker:
This PR improves formula validation and adds test coverage for all parser and lexer validation errors.

### Changes

1. Switch to conventional token span start/end semantics: [start, end). Spans are only used internally now, though (see (2)).
2. Introduced `FormulaError` and `FormulaSyntaxError` classes. `FormulaSyntaxError.__str__()` produces a human-readable error message, highlighting the offending span in the context of the original formula. This should be fit for non-technical end users.
3. Removed `ASTNode.span` property as it was unused.
4. Added tests and adapted existing ones to the changes described above.

### DIFF
--- a/src/frequenz/sdk/timeseries/formulas/_exceptions.py
+++ b/src/frequenz/sdk/timeseries/formulas/_exceptions.py
@@ -1,0 +1,53 @@
+# License: MIT
+# Copyright © 2026 Frequenz Energy-as-a-Service GmbH
+
+"""Exception classes for formulas parser and evaluator."""
+
+
+class FormulaError(Exception):
+    """Base class for exceptions in this module."""
+
+
+class FormulaSyntaxError(FormulaError):
+    """Exception raised for syntax errors in the formula."""
+
+    def __init__(
+        self, *, formula: str, span: tuple[int, int] | None, message: str
+    ) -> None:
+        """Initialize this instance."""
+        self._formula: str = formula
+        self._span: tuple[int, int] | None = span
+        self._message: str = message
+
+    def __str__(self) -> str:
+        """Return a string representation of the error.
+
+        The span (if present) will be used to highlight the error location.
+        For long formulas, the beginning and/or end will be truncated as needed.
+        """
+        if self._span is None:
+            return self._message
+
+        formula_line = self._formula
+        span_padding = " " * self._span[0]
+        span_highlight = "^" * (self._span[1] - self._span[0])
+        error_line = f"{span_padding}{span_highlight} {self._message}"
+
+        char_limit: int = 80 - 11  # Account for "  Formula: " prefix in final output
+
+        if len(error_line) > char_limit:
+            # Remove characters from the left to fit error line within char limit
+            num_chars_to_truncate = len(error_line) - char_limit
+            error_line = error_line[num_chars_to_truncate:]
+            # Shift formula line by the same amount and insert leading ellipsis
+            formula_line = f"... {formula_line[num_chars_to_truncate+4:]}"
+
+        if len(formula_line) > char_limit:
+            # Truncate the formula line, insert ellipsis
+            formula_line = f"{formula_line[:char_limit-4]} ..."
+
+        return (
+            "Formula syntax error:\n"
+            f"  Formula: {formula_line}\n"
+            f"           {error_line}"
+        )
--- a/src/frequenz/sdk/timeseries/formulas/_functions.py
+++ b/src/frequenz/sdk/timeseries/formulas/_functions.py
@@ -85,19 +85,14 @@ class Function(abc.ABC, Generic[QuantityT]):
         )
 
     @classmethod
-    def from_string(
-        cls, name: str, params: list[AstNode[QuantityT]]
-    ) -> Function[QuantityT]:
-        """Create a function instance from its name."""
-        match name.upper():
-            case "COALESCE":
-                return Coalesce(params)
-            case "MAX":
-                return Max(params)
-            case "MIN":
-                return Min(params)
-            case _:
-                raise ValueError(f"Unknown function name: {name}")
+    def function_class_by_name(cls, name: str) -> type[Function[QuantityT]] | None:
+        """Return the function class corresponding to the given name."""
+        known_functions: dict[str, type[Function[QuantityT]]] = {
+            "COALESCE": Coalesce,
+            "MAX": Max,
+            "MIN": Min,
+        }
+        return known_functions.get(name.upper())
 
 
 @dataclass
--- a/src/frequenz/sdk/timeseries/formulas/_lexer.py
+++ b/src/frequenz/sdk/timeseries/formulas/_lexer.py
@@ -11,6 +11,7 @@ from collections.abc import Iterator
 from typing_extensions import override
 
 from . import _token
+from ._exceptions import FormulaSyntaxError
 from ._peekable import Peekable
 
 
@@ -74,7 +75,11 @@ class Lexer(Iterator[_token.Token]):
             _ = next(self._iter)  # consume '#'
             comp_id = self._read_integer()
             if not comp_id:
-                raise ValueError(f"Expected integer after '#' at position {pos}")
+                raise FormulaSyntaxError(
+                    formula=self._formula,
+                    span=(pos + 1, pos + 2),
+                    message="Expected integer after '#'",
+                )
             end_pos = pos + len(comp_id) + 1  # account for '#'
             return _token.Component(
                 span=(pos, end_pos),
@@ -120,4 +125,8 @@ class Lexer(Iterator[_token.Token]):
             end_pos = pos + len(symbol)
             return _token.Symbol(span=(pos, end_pos), value=symbol)
 
-        raise ValueError(f"Unexpected character '{char}' at position {pos}")
+        raise FormulaSyntaxError(
+            formula=self._formula,
+            span=(pos, pos + 1),
+            message="Unexpected character",
+        )
--- a/src/frequenz/sdk/timeseries/formulas/_lexer.py
+++ b/src/frequenz/sdk/timeseries/formulas/_lexer.py
@@ -75,52 +75,49 @@ class Lexer(Iterator[_token.Token]):
             comp_id = self._read_integer()
             if not comp_id:
                 raise ValueError(f"Expected integer after '#' at position {pos}")
-            end_pos = pos + len(comp_id)
+            end_pos = pos + len(comp_id) + 1  # account for '#'
             return _token.Component(
-                span=(
-                    pos + 1,
-                    end_pos + 1,  # account for '#'
-                ),
+                span=(pos, end_pos),
                 id=comp_id,
                 value=self._formula[pos:end_pos],
             )
 
         if char == "+":
             _, char = next(self._iter)  # consume operator
-            return _token.Plus(span=(pos + 1, pos + 1), value=char)
+            return _token.Plus(span=(pos, pos + 1), value=char)
 
         if char == "-":
             _, char = next(self._iter)
-            return _token.Minus(span=(pos + 1, pos + 1), value=char)
+            return _token.Minus(span=(pos, pos + 1), value=char)
 
         if char == "*":
             _, char = next(self._iter)
-            return _token.Mul(span=(pos + 1, pos + 1), value=char)
+            return _token.Mul(span=(pos, pos + 1), value=char)
 
         if char == "/":
             _, char = next(self._iter)
-            return _token.Div(span=(pos + 1, pos + 1), value=char)
+            return _token.Div(span=(pos, pos + 1), value=char)
 
         if char == "(":
             _, char = next(self._iter)
-            return _token.OpenParen(span=(pos + 1, pos + 1), value=char)
+            return _token.OpenParen(span=(pos, pos + 1), value=char)
 
         if char == ")":
             _, char = next(self._iter)
-            return _token.CloseParen(span=(pos + 1, pos + 1), value=char)
+            return _token.CloseParen(span=(pos, pos + 1), value=char)
 
         if char == ",":
             _, char = next(self._iter)
-            return _token.Comma(span=(pos + 1, pos + 1), value=char)
+            return _token.Comma(span=(pos, pos + 1), value=char)
 
         if char.isdigit():
             num = self._read_number()
             end_pos = pos + len(num)
-            return _token.Number(span=(pos + 1, end_pos), value=num)
+            return _token.Number(span=(pos, end_pos), value=num)
 
         if char.isalpha():
             symbol = self._read_symbol()
             end_pos = pos + len(symbol)
-            return _token.Symbol(span=(pos + 1, end_pos), value=symbol)
+            return _token.Symbol(span=(pos, end_pos), value=symbol)
 
         raise ValueError(f"Unexpected character '{char}' at position {pos}")
--- a/src/frequenz/sdk/timeseries/formulas/_parser.py
+++ b/src/frequenz/sdk/timeseries/formulas/_parser.py
@@ -18,6 +18,7 @@ from frequenz.sdk.timeseries._base_types import QuantityT
 
 from . import _ast, _token
 from ._base_ast_node import AstNode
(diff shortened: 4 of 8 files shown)


src/frequenz/sdk/timeseries/formulas/_exceptions.py
```diff
@@ -0,0 +1,57 @@
...
+class FormulaError(Exception):
+    """Base class for exceptions in this module."""
+
+
+class FormulaSyntaxError(FormulaError):
+    """Exception raised for syntax errors in the formula."""
+
+    formula: str
+    span: tuple[int, int]
+    message: str
```
shsms: These seem to be unused, can be dropped.

src/frequenz/sdk/timeseries/formulas/_exceptions.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def __init__(
+        self, *, formula: str, span: tuple[int, int] | None, message: str
+    ) -> None:
+        """Initialize this instance."""
+        self._formula = formula
+        self._span = span
+        self._message = message
```
shsms: It would be nice to type these, because although this works with mypy, most newer type checkers assume that there can be more variants.  For example, if you say `self.v = None` and later do `self.v = 10`, newer type checkers allow that, mypy doesn't.  So it is useful to hint these instance variables.

```suggestion
        self._formula: str = formula
        self._span: tuple[int, int] = span
        self._message: str = message
```

src/frequenz/sdk/timeseries/formulas/_exceptions.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def __str__(self) -> str:
+        """Return a string representation of the error.
+
+        The span (if present) will be used to highlight the error location.
+        For long formulas, the beginning or end will be truncated as needed.
```
shsms:
```suggestion
        For long formulas, the beginning and/or end will be truncated as needed.

File: src/frequenz/sdk/timeseries/formulas/_exceptions.py [hunk 4]
```diff
@@ -0,0 +1,57 @@
...
+        if len(error_line) > char_limit:
+            # Remove characters from the left to fit error line within char limit
+            num_chars_to_truncate = len(error_line) - char_limit
+            error_line = error_line[num_chars_to_truncate:]
+            # Shift formula line by the same amount and insert leading ellipsis
+            formula_line = f"... {formula_line[num_chars_to_truncate+4:]}"
+
+        if len(formula_line) > char_limit:
+            # Truncate the formula line, insert ellipsis
+            formula_line = f"{formula_line[:char_limit-4]} ..."
```
ela-kotulska-frequenz, line 47: Why not using bold text?

> print("\033[1mThis is bold\033[0m")

src/frequenz/sdk/timeseries/formulas/_functions.py
```diff
@@ -85,19 +85,14 @@ async def unsubscribe(self) -> None:
...
-                return Min(params)
-            case _:
-                raise ValueError(f"Unknown function name: {name}")
+    def function_class_by_name(cls, name: str) -> type[Function[QuantityT]] | None:
+        """Return the function class corresponding to the given name."""
+        known_functions: dict[str, type[Function[QuantityT]]] = {
+            "COALESCE": Coalesce,
+            "MAX": Max,
+            "MIN": Min,
+        }
```
ela-kotulska-frequenz, line 94: Maybe `known_functions` could be class variable, now?

src/frequenz/sdk/timeseries/formulas/_lexer.py
```diff
@@ -74,7 +75,11 @@ def __next__(self) -> _token.Token:  # pylint: disable=too-many-branches
             _ = next(self._iter)  # consume '#'
             comp_id = self._read_integer()
             if not comp_id:
-                raise ValueError(f"Expected integer after '#' at position {pos}")
+                raise FormulaSyntaxError(
+                    formula=self._formula,
+                    span=(pos + 1, pos + 2),
+                    message="Expected integer",
```
ela-kotulska-frequenz: Consider `Expected integer after '#'`, because this explains why integer is expected :)

src/frequenz/sdk/timeseries/formulas/_parser.py
```diff
@@ -80,8 +82,10 @@ def _parse_term(self) -> AstNode[QuantityT] | None:
             next_factor = self._parse_factor()
 
             if next_factor is None:
-                raise ValueError(
-                    f"Expected factor after operator at span: {token.span}"
+                raise FormulaSyntaxError(
+                    formula=self._formula,
+                    span=(token.span[1], token.span[1] + 1),
+                    message="Expected expression",
                 )
```
ela-kotulska-frequenz, line 94: Expected `expression` or `factor`? The same for `unary`

src/frequenz/sdk/timeseries/formulas/_parser.py
```diff
@@ -136,49 +146,82 @@ def _parse_bracketed(self) -> AstNode[QuantityT] | None:
...
 
         token: _token.Token | None = self._lexer.peek()
         if token is None or not isinstance(token, _token.OpenParen):
-            raise ValueError(
-                f"Expected '(' after function name at position {fn_name.span}"
+            raise FormulaSyntaxError(
+                formula=self._formula,
+                span=(fn_name.span[1], fn_name.span[1] + 1),
+                message="Expected '('",
             )
```
ela-kotulska-frequenz, line 203: I would keep `Expected '(' after function name`, because this explains why `(` is expected.

src/frequenz/sdk/timeseries/formulas/_functions.py
```diff
@@ -85,19 +85,14 @@ async def unsubscribe(self) -> None:
...
+        return known_functions.get(name.upper())
```
shsms, line 95: It is usually nice to do these in a separate commit first, before introducing the exceptions.  Then it is clear what parts are affected, and you can provide an explanation for why it is necessary.

Now I have to search for the uses and make guesses about the reason.

ela-kotulska-frequenz (commented):
Sorry I was on holiday and could not check earlier. 
Looks very good only small comments / questions. :)

ela-kotulska-frequenz:
I will also request Sahas for quick look, because he knows formula better

src/frequenz/sdk/timeseries/formulas/_exceptions.py
simonvoelcker: Good idea :+1:

tests/timeseries/_formulas/test_formulas.py
```diff
@@ -333,6 +344,160 @@ async def test_max_min_coalesce(self) -> None:
...
+                "^^^ Unknown function name",
+            ),
+            (
+                "foo(1)",
+                "^^^ Unknown function name",
+            ),
+            (
+                "max(1,,2)",
+                "      ^ Expected argument",
+            ),
```
shsms, line 425: Maybe we can add some more cases:

- `"1 2"`
- `"1, 2"`
- `"max(1, 2,)"`
- `"max(1, 2))"`
- `"max(1, 2), "`

src/frequenz/sdk/timeseries/formulas/_functions.py
simonvoelcker: It could be, yes. If I see it correctly, that would entail reordering the classes though, because `Coalesce`, `Max` and `Min` are currently defined after `Function`, but when used in a class variable they must already exist when `Function` is parsed. Personally, I don't think it's worth the effort and it's cleaner to have `Function` on top. What do you think?

src/frequenz/sdk/timeseries/formulas/_parser.py
simonvoelcker: Removing mentions of "factor", "unary" and "primary (expression)" is something Sahas and I agreed on last week - The intention is to hide technical terms from possibly-not-so-technical users.

src/frequenz/sdk/timeseries/formulas/_exceptions.py
```diff
@@ -0,0 +1,57 @@
...
+
+        return (
+            "Formula syntax error:\n"
+            f"Formula: {formula_line}\n"
+            f"Error:   {error_line}"
+        )
```
shsms, line 53:
```suggestion
        return (
            "Formula syntax error:\n"
            f"  Formula: {formula_line}\n"
            f"    Error: {error_line}"
        )
```

Although I wonder if we need to say "Error" at all.  Maybe it can just be

```suggestion
        return (
            "Formula syntax error:\n"
            f"  Formula: {formula_line}\n"
            f"           {error_line}"
        )
```

src/frequenz/sdk/timeseries/formulas/_exceptions.py
shsms: Gotta see how that would look in journald

src/frequenz/sdk/timeseries/formulas/_parser.py
shsms: Yes, factor,  unary etc might be very confusing for people who are only interested in writing the formulas.

src/frequenz/sdk/timeseries/formulas/_exceptions.py
```diff
@@ -0,0 +1,57 @@
...
+            error_line = error_line[num_chars_to_truncate:]
+            # Shift formula line by the same amount and insert leading ellipsis
+            formula_line = f"... {formula_line[num_chars_to_truncate+4:]}"
+
+        if len(formula_line) > char_limit:
+            # Truncate the formula line, insert ellipsis
+            formula_line = f"{formula_line[:char_limit-4]} ..."
+
+        return (
+            "\033[1mFormula syntax error:\033[0m\n"
```
shsms: I don't think this is the part that needs to be bold.  May be useful to make the relevant token bold, but we already add a highlight in the next line, so probably not necessary.

Also we're not targeting the terminal, but journald instead, and possible the UI, so it is very unlikely that these terminal control characters will get rendered properly.

src/frequenz/sdk/timeseries/formulas/_exceptions.py
simonvoelcker: I did some testing. journald stores the raw bytes, so it is down to how those are rendered later. E.g. when I run `journalctl --output=cat`, the highlighting works fine.

src/frequenz/sdk/timeseries/formulas/_exceptions.py
simonvoelcker: So better revert to plain text - @ela-kotulska-frequenz FYI :)

tests/timeseries/_formulas/test_formulas.py
simonvoelcker: Good challenge, 4/5 currently pass validation. I'll extend the parser as needed.

For `"1 2"` we probably want to see "Expected operator" in the output.

simonvoelcker: As it turns out, all four failing cases are addressed by one new bit of logic that I added now. We were basically parsing the first expression in the given string, eagerly, and then ignored anything that came after. I added a check which rejects any subsequent tokens as "Unexpected token".

src/frequenz/sdk/timeseries/formulas/_exceptions.py
simonvoelcker: As discussed further down - I removed bold output again.

src/frequenz/sdk/timeseries/formulas/_exceptions.py
shsms: I guess the couple of prefix blank spaces got missed, but everything else looks good to me.

src/frequenz/sdk/timeseries/formulas/_exceptions.py
ela-kotulska-frequenz: > May be useful to make the relevant token bold, but we already add a highlight in the next line

This is what I meant : To remove the extra line with `^^`  and just bold the invalid part of expression.
But if this does not render properly, then ok. Thanks for checking it :)

src/frequenz/sdk/timeseries/formulas/_exceptions.py
simonvoelcker: Oh, I didn't even know these were intentional. My take is that every character of prefix will shorten the formula line (due to the total char limit), so I'd keep it as is.

shsms: It doesn't really, we don't need to stick to an 80 char limit, it can be 82 or 90 or 100.  Nobody uses a terminal window that can only display 80 chars for reading logs.

Whereas the prefix really helps distinguish them from other logs, and because of just how much logs an app with several actors can produce, it is nice to clarify that these lines are part of a single message.

shsms: And it is also the style we use for other multi-line messages.

ela-kotulska-frequenz (dismissed):
LGTM

src/frequenz/sdk/timeseries/formulas/_exceptions.py
simonvoelcker: I put the spaces back in.

shsms (approved):
Awesome, congrats on the first PR! 🎉

llucax:
@simonvoelcker sorry for arriving late to the part and congrats on your first PR merged in the SDK! :tada: 

One super tiny general comment, in general we try to avoid adding references to issues in commits, we used to do it but when iterating over a commit doing amends, the issue will be polluted with every commit amend it is done, so it was a bit annoying. To avoid this, we prefer putting references to issues in the PR description, this way the cross-reference happens only once. As usual, this is just a general recommendation, and it is totally fine to put references to issues in commits when it makes sense (very low chances of being amended, or maybe a single commit affecting one particular issue in a PR with many other commits that is supposed to fix something else are clear examples).

Just as an example, this is how it looks the activity in #1356 now :laughing: 

image

llucax:
BTW, should that issue be closed or was this only a partial fix?

src/frequenz/sdk/timeseries/formulas/_exceptions.py
```diff
@@ -0,0 +1,53 @@
...
+
+        if len(formula_line) > char_limit:
+            # Truncate the formula line, insert ellipsis
+            formula_line = f"{formula_line[:char_limit-4]} ..."
+
+        return (
+            "Formula syntax error:\n"
+            f"  Formula: {formula_line}\n"
+            f"           {error_line}"
+        )
```
llucax, line 53: It feels a bit strange to me to do so much formatting in `__str__`. I think a better approach could be to provide a formatting function, where we can pass the max line length for example instead of using a hardcoded 80. And then, call this function where error reporting for humans is needed.

I see there is no explicit `except FormulaSyntaxError` here, so maybe this was hard to plug into the place where the error is actually being captured and printed for human consumption.

src/frequenz/sdk/timeseries/formulas/_functions.py
llucax, line 94: Nitpick: This should have been probably named `get_function_class_by_name()`, usually functions/methods should be named with verbs (if it was a class attribute, then the current name makes sense). I think it can also be shortened to `get_class()`, as being a function class can be implied because it is a `Function` method and the `name` is part of the function signature (to enforce extra clarity on the caller site, you could add a bit more verbosity and make it kw-only, like `get_class_for(function_name="MAX")`)

llucax (commented):
This is a pretty awesome improvement! I would have made it more prominent in the release notes, listing it (also?) in the new features section, with an example of how an error is being displayed now.

PR #1371, opened 2026-03-05
Remove channel registry

simonvoelcker:
This removes the `ChannelRegistry` class.

# Background

This is part of an initiative to close unused channels to save resources. The `ChannelRegistry` was a de-facto singleton class for storing channels by name. Request objects such as `ComponentMetricRequest` or `ReportRequest` have a `get_channel_name()` method, and the returned channel name acted as a contract between sender and receiver - With the name, they could obtain the same channel from the `ChannelRegistry`. By getting rid of the registry, we are moving towards a target state where only references to _senders_ and _receivers_, but no _channels_ are being held.

# Implementation

## ComponentMetricRequest

`ComponentMetricRequest` objects represent a request to receive metrics from a component - Sometimes also referred to as telemetry stream. To serve such a request without relying on the `ChannelRegistry`, the request got a new attribute: `telem_stream_sender: Sender[Receiver[Sample[Quantity]]]`. This is the _sender_ of a one-shot channel that is used to send the _receiver_ of the telemetry stream back to the original requester - Similar to a callback function. The receiver of a component metric request should therefore create (and own) the channel that sends the metric data, and use the one-shot channel to send a receiver of that channel "back".

In a few places, component metric requests were created and sent from synchronous functions which returned the receiver. To make this work with the round-trip outlined above, an additional channel was added that merely forwards the metric data via a `Pipe`. This allows us to immediately return a sender while the channel that serves the metric data is being constructed asynchronously.

## ReportRequest

`ReportRequest` objects are a different kind of request that was handled similarly.

## Channel ownership

The receivers of request objects create the channels and own them. For instance, in case of `ComponentMetricRequests`, MicrogridApiSource now maintains a dictionary (`channel_lookup: dict[str, Broadcast]`) as a replacement for the channel registry. It is necessary to have this lookup dictionary to avoid creating duplicate channels for similar metric requests. In this case, a new receiver is created from an existing channel found in the lookup dictionary.

Forwarding channels are owned by their creator (e.g. `GridFrequency`). These (and the Pipes that connect them to the other channels) are currently never closed. Closing these when all receivers or senders are closed is a TODO for a later iteration.


pyproject.toml
```diff
@@ -32,7 +32,7 @@ dependencies = [
   "frequenz-client-microgrid >= 0.18.1, < 0.19.0",
   "frequenz-microgrid-component-graph >= 0.3.4, < 0.4",
   "frequenz-client-common >= 0.3.6, < 0.4.0",
-  "frequenz-channels >= 1.6.1, < 2.0.0",
+  "frequenz-channels @ git+https://github.com/shsms/frequenz-channels-python.git@oneshot",
```
simonvoelcker: I won't merge before this is merged 👀

src/frequenz/sdk/_internal/_channels.py
llucax: I think this one should be gone too, right? But I guess if it should, it should be detected by the linter, so maybe it is used somewhere outside this file?

src/frequenz/sdk/microgrid/_data_sourcing/_component_metric_request.py
```diff
@@ -51,6 +55,9 @@ class ComponentMetricRequest:
     If None, only live data is streamed.
     """
 
+    telem_stream_sender: Sender[Receiver[Sample[Quantity]]]
+    """Sender of a oneshot channel used to send the data stream back to the requester."""
```
llucax: Nitpick: Maybe just `receiver_sender`? I guess we are domed and this will be confusing no matter what 😆 

Also I wouldn't mention oneshot channel here as it is really not important there is a channel in the middle or which type of channel. The only important thing is that sender will be used to send a receiver from which the requester can receive the data, right?

I also think it is worth updating the class docstring to explain this.

src/frequenz/sdk/microgrid/_data_sourcing/microgrid_api_source.py
```diff
@@ -177,11 +168,13 @@ def __init__(
...
         """The dictionary of component IDs to asyncio tasks."""
 
-        self._registry = registry
         self._req_streaming_metrics: dict[
             ComponentId, dict[Metric | TransitionalMetric, list[ComponentMetricRequest]]
         ] = {}
 
+        self._channel_lookup: dict[str, Broadcast[Sample[Quantity]]] = {}
+        """ Channel cache for reuse (map channel name to channel)."""
+
```
llucax: I would name this just `_channels` or `_active_channels`, this is not really only used for lookup, this class is owning those channels, right? If so I would also update the attribute docstring to reflect this.

llucax:
Sorry to be a party pooper.  I just started with the review and looking at the data sourcing actor and I think we might be overcomplicating things.

Why can't we just reverse the ownership, and make the requester own the "data receiving channel"? So instead of creating a oneshot channel[^1], the requester creates a normal channel and gives the other end the sender. That's it! No back and forth sending and receiving receivers.

The only disadvantage of this approach is we can't de-duplicate channels so to speak, so no broadcasting like we do now (the data sourcing actor creates one channel per metric-cid combination I think, and create new receivers for any new party asking for data). We can still de-duplicate by subscribing to the microgrid API once and send sending to multiple senders, but we'll still have a channel per sender. But for broadcast channels we actually use one queue per receiver, so I guess in terms of memory usage it shouldn't change much.

So, this is a very early thought, I didn't think of all the other potential issues, but by looking at the changes it really feels to me that we are introducing non-trivial complexities with this change. For example the new requirement to being async to subscribe to data is gone with this approach. I think that's my main concern, we are converting something that is inherently sync (create a new receiver to receive data) into something async, and that's very disruptive.

I will continue with my review, but I wanted to mention this early, as we are planning to build on top of this, I want to raise it as soon as possible.

[^1]: See @sahas-subramanian-frequenz? I didn't thought about it and wrote basically `create_oneshot_channel()`, that must be a good name :trollface:

simonvoelcker:
@llucax I completely agree that this change adds regrettable complexity, and we should explore ways of avoiding it.

It is my understanding that channels must be de-duplicated somehow because we might be dealing with hundreds of components in practice. The channels must then be owned centrally, whether that's near the source or in a singleton registry. Maybe we can revisit the registry idea, but store _senders_ there instead of channels? That would avoid the oneshot channels.

Regarding sync-vs-async: The creation of channels was async before and stays async, no? We're sending ComponentMetricRequests either way, it's just that in one case we discussed today (GridFrequency.new_receiver) a receiver must be returned synchronously, so we have to use the forwarding channel.

llucax:
The problem with the singleton channel registry for me is when things go wrong it is very hard to find out when a channel was first created, who requested it, etc. This problem came up and was very hard to debug. This is why I created the option to log when a channel is created in the registry (including the stack trace), but that turns out to pollute the logs excessively when enabled, as you can't really filter it by channel name, so it is printed for all channels.

This is why I think getting rid of the channel registry is a good goal, but I think we agree that not at all cost.

About the de-duplication, as I mentioned, it might be worth taking a deeper look at how much overhead really means having one channel per metric-cid vs. the current approach of having one receiver per metric-cid. As for broadcast channels we have a queue per receiver, and the sender just iterates over all receiver and put the new message there, I'm not sure it will be a big difference between having 1 broadcast channel with N receivers vs having N "unicast channels" (for now I guess we'll use a broadcast channel too.

> [!NOTE]
> We explored in the past implementing a unicast channel but I think it was as slow as the broadcast channel so we kept the broadcast channel only. I guess for unicast we can just use a [asyncio queue](https://docs.python.org/3/library/asyncio-queue.html), not sure if that was what we used to implement the unicast channel.

benchmarks/timeseries/benchmark_datasourcing.py
```diff
@@ -105,18 +105,21 @@ async def consume(channel: Receiver[Any]) -> None:
 
     for evc_id in mock_grid.evc_ids:
         for component_metric_id in COMPONENT_METRIC_IDS:
+            telem_stream_sender, telem_stream_receiver = make_oneshot(
+                Receiver[Sample[Quantity]]  # type: ignore[type-abstract]
```
llucax: This is definitely not nice from the new interface. I guess sending/receiving receivers is not a widely used case, but still a disadvantage of having to pass the type as a variable instead of a type argument (like with classes). I wonder if we can somehow hack it to make a channel an object but that only holds a sender and a receiver, basically a tuple, maybe an actual NamedTuple so we can still do `sender, receiver` but more like `sender, receiver = create_oneshot_channel[T]()`, or even going back to `sender, receiver = OneshotChannel[T]()` and avoid the open/make/create discussion :P

But I guess this is a topic for https://github.com/frequenz-floss/frequenz-channels-python/pull/502.

llucax (commented):
I will send this partial review, as all commits were split while I was mid-review. Will wait until things settle down to continue the review.

src/frequenz/sdk/microgrid/_data_sourcing/_component_metric_request.py
shsms: I think this would be so much better typed as `OneshotSender[BroadcastReceiver[Sample[Quantity]]`.

Right now, Merge, Timer etc are exposed as themselves, but not the channel senders and receivers.  I think that needs to change.  This helps with readability and with exposing other quirks of these types.

I've already updated the Oneshot PR.  Will do with the upcoming Broadcast PR as well.

benchmarks/timeseries/benchmark_datasourcing.py
shsms: `sender, receiver = create_oneshot_channel[T]()`  <- This doesn't work, Python doesn't like subscript operator for functions.

`sender, receiver = OneshotChannel[T]()` <- I didn't try as a NamedTuple,  will see if that helps.

But the channel registry also had this problem, you had to specify the type.

src/frequenz/sdk/microgrid/_data_sourcing/_component_metric_request.py
simonvoelcker: Updated the types now.

llucax (commented):
I see this now needs rebasing and it is using the old oneshot names, so I will wait until it is updated for the final review. Can you requrest a review from me again when it is ready @simonvoelcker?

simonvoelcker:
@llucax rebased. Regarding the "old" names, IIRC the plan was to make the changes in two steps. And I'll have to wait until something is merged in the channels repo anyways, so I can update the dependency.

benchmarks/timeseries/benchmark_datasourcing.py
llucax, line 1: Just curious, did you run the benchmark before and after the changes to see if there are any significant performance changes?

benchmarks/timeseries/benchmark_datasourcing.py
```diff
@@ -16,15 +16,17 @@
 from time import perf_counter
 from typing import Any
 
-from frequenz.channels import Broadcast, Receiver, ReceiverStoppedError
+from frequenz.channels import Broadcast, OneshotChannel, Receiver, ReceiverStoppedError
+from frequenz.channels._broadcast import BroadcastReceiver
```
llucax:
```suggestion
from frequenz.channels import BroadcastReceiver
```

Always use public imports. If the symbols are not exported publicly, we need to fix it in the channels library.

src/frequenz/sdk/microgrid/_data_sourcing/_component_metric_request.py
```diff
@@ -6,13 +6,18 @@
 from dataclasses import dataclass
 from datetime import datetime
 
+from frequenz.channels._broadcast import BroadcastReceiver
+from frequenz.channels._oneshot import OneshotSender
```
llucax: Once more, I will stop it here as I guess you get the point, but this should be fixed in all files.

```suggestion
from frequenz.channels import BroadcastReceiver
from frequenz.channels import OneshotSender
```

src/frequenz/sdk/microgrid/_data_sourcing/microgrid_api_source.py
```diff
@@ -413,18 +407,23 @@ def _get_metric_senders(
...
+            senders = []
+            for request in req_list:
+                channel_name = request.get_channel_name()
+                # Create missing channels and inform the requesting side via oneshot
+                if channel_name not in self._channels:
+                    telem_stream: Broadcast[Sample[Quantity]] = Broadcast(
+                        name=channel_name
+                    )
+                    self._channels[channel_name] = telem_stream
+                    await request.telem_stream_sender.send(telem_stream.new_receiver())
```
llucax, line 420: Shouldn't this be outside of the `if` block? If the channel already exists, we should probably send the receiver too via the oneshot channel right? Or am I missing something?

If what I say is correct, we probably need to add a test for this case.

src/frequenz/sdk/microgrid/_data_sourcing/microgrid_api_source.py
```diff
@@ -397,11 +391,11 @@ def _get_data_extraction_method(
...
         raise ValueError(err)
 
-    def _get_metric_senders(
+    async def _get_metric_senders(
         self,
         category: ComponentCategory | int,
         requests: dict[Metric | TransitionalMetric, list[ComponentMetricRequest]],
-    ) -> list[tuple[Callable[[Any], float], list[Sender[Sample[Quantity]]]]]:
+    ) -> list[tuple[Callable[[Any], float], list[BroadcastSender[Sample[Quantity]]]]]:
         """Get channel senders from the channel registry for each requested metric.
```
llucax: "Channel registry" here. It might be worth doing a global search for "channel.*registry" or something like that to make sure to remove all mentions of the channel registry.

src/frequenz/sdk/microgrid/_power_managing/_power_managing_actor.py
```diff
@@ -70,8 +67,8 @@ def __init__(  # pylint: disable=too-many-arguments
         self._bounds_subscription_receiver = bounds_subscription_receiver
         self._power_distributing_requests_sender = power_distributing_requests_sender
         self._power_distributing_results_receiver = power_distributing_results_receiver
-        self._channel_registry = channel_registry
         self._proposals_receiver = proposals_receiver
+        self._channel_lookup: dict[str, Broadcast[_Report]] = {}
```
llucax: Nit: Could also be renamed `_channels`.

src/frequenz/sdk/microgrid/_power_managing/_power_managing_actor.py
```diff
@@ -233,18 +230,27 @@ async def _run(self) -> None:
...
+                    subscription: ReportRequest,
+                ) -> Broadcast[_Report]:
+                    channel_name = subscription.get_channel_name()
+                    if channel_name not in self._channel_lookup:
+                        report_channel = Broadcast[_Report](name=channel_name)
+                        report_channel.resend_latest = True
+                        self._channel_lookup[channel_name] = report_channel
+                        await subscription.report_stream_sender.send(
+                            report_channel.new_receiver()
+                        )
```
llucax: Same here, shouldn't we reply always, not only when the channel gets created?

Also I would move the replying out of this method, for me it is surprising that it replies, specially in the in case I'm right and we need to reply when the channel already exists.

src/frequenz/sdk/microgrid/_resampling.py
```diff
@@ -68,28 +85,24 @@ async def _subscribe(self, request: ComponentMetricRequest) -> None:
         """
         request_channel_name = request.get_channel_name()
 
-        # If we are already handling this request, there is nothing to do.
-        if request_channel_name in self._active_req_channels:
+        # If we are already handling this request, answer the request by sending a
+        # new receiver from the existing channel.
+        if data_sink_channel := self._data_sink_channels.get(request_channel_name):
+            await request.telem_stream_sender.send(data_sink_channel.new_receiver())
```
llucax, line 96: It would be nice to send the reply only in one place, to make it more obvious that we always do, but I guess in this case it would mean removing this short-circuit and put all the following code inside the `if` block instead, just to send the reply after the `if` block, so I'm not sure it is worth it/it makes it more readable, so I'm good with leaving it as it is, just bringing it up in case you have other ideas.

src/frequenz/sdk/timeseries/battery_pool/_battery_pool.py
```diff
@@ -74,6 +78,9 @@ def __init__(
         unique_id = str(uuid.uuid4())
         self._source_id = unique_id if name is None else f"{name}-{unique_id}"
         self._priority = priority
+        self._report_stream_receiver: Receiver[Receiver[_Report]] | None = None
+        # Keep a reference to prevent garbage collector from destroying pipe
+        self._pipe: Pipe[_Report] | None = None
```
llucax, line 88: Oh, so the battery pool also needs a pipe, is not just the grid frequency? I see the `create_task` to hack our way outside the sync world already existed so I guess it is the same case as for grid frequency, but here we are wrapping a @property, so we'll need to change it to a method too to make it async in the future.

So, there is also no stopping of the `_pipe` for now, ever, right? I guess there is no stopping of the pools either, so we can pretend the problem doesn't exist 😆 

Also I see this happens for all the pools.

src/frequenz/sdk/timeseries/_grid_frequency.py
```diff
@@ -107,7 +97,47 @@ def new_receiver(self) -> Receiver[Sample[Frequency]]:
...
+        component_metric_request = ComponentMetricRequest(
+            "grid-frequency",
+            self._source_component.id,
+            Metric.AC_FREQUENCY,
+            None,
+            telem_stream_sender,
+        )
+        await self._request_sender.send(component_metric_request)
+        receiver = await telem_stream_receiver.receive()
+        return self._map_frequency_samples(receiver)
```
llucax: I found this a bit confusing, I thought you were somehow mapping values here (so the mapping happened right now, not for each received value), not calling `receiver.map()`.

I would change the function so you can write this instead

```suggestion
        return receiver.map(receiver, self._map_frequency_samples)
```
That makes the intention more clear IMHO.

tests/actor/test_resampling.py
```diff
@@ -176,11 +176,88 @@ async def test_duplicate_request(
...
+        await _assert_resampling_works(
+            timeseries_sender=telem_stream.new_sender(),
+            timeseries_receiver=await telem_stream_receiver.receive(),
+            fake_time=fake_time,
+        )
+
+        await resampling_actor._resampler.stop()  # pylint: disable=protected-access
+
+
+async def test_resubscribe(fake_time: time_machine.Coordinates) -> None:
```
llucax, line 197: OK, it seems we need this test also for the MicrogridApiSource (although it is going away, so maybe just manual validation is OK for now), and the power manager.

RELEASE_NOTES.md
```diff
@@ -19,3 +19,5 @@
 - This fixes a rare power distributor bug where some battery inverters becoming unreachable because of network outages would lead to excess power values getting set.  This is fixed by measuring the power of the unreachable inverters through their fallback meters and excluding that power from what is distributed to the other inverters.
 
 - Fixed stopping formulas: It will now also stop the evaluator and sub-formulas correctly.
+
+- Removed ChannelRegistry. Streams are not being set up using one-shot channels and owned by the data source.
```
llucax: Users should not be affected by this change, so we probably should skip adding release notes.

llucax (commented):
OK, done with the review. Mostly minor comments but a few important too:

1. Make sure we always reply requests to the oneshot channel (unless I'm missing something)
2. Add tests for 1
3. Avoid importing *private* symbols (underscore prefix)

src/frequenz/sdk/microgrid/_data_sourcing/microgrid_api_source.py
simonvoelcker: I think you're right, the receiver should be sent regardless of whether the channel is new. The tests pass even after changing it, proving that the nuance is not covered by a test case yet.

The next planned change that builds on top of this is proper teardown when all receivers disconnect, so I think it makes sense to add the test case only then, because this is basically the "reconnect" use case. What do you think?

src/frequenz/sdk/timeseries/battery_pool/_battery_pool.py
simonvoelcker: The pipe is needed for the same reason it was needed in grid frequency: Because the receiver method (in this case _power_status_) is sync. So yes, when we move this to async we don't need the pipe, as far as I understand.

Closing the pipe is something I'd handle in the next step when we handle closed senders.

src/frequenz/sdk/timeseries/_grid_frequency.py
simonvoelcker: Much better indeed :+1:

src/frequenz/sdk/microgrid/_data_sourcing/microgrid_api_source.py
llucax: Doesn't this also affects connecting multiple times to the same channel? But I guess it doesn't otherwise that **must** be caught by some test...

benchmarks/timeseries/benchmark_datasourcing.py
simonvoelcker: I tried it now, but apparently the data sourcing benchmark is broken, and has been so for a while already. Will try to repair it.

simonvoelcker: So I've been toying around with this for quite a while now, looking for a version that worked so I can find out what exactly broke, and when. However, I see different errors depending on far I go back, and it never works for me unless I make a change (whether this is actually a fix remains unclear).

Here's the _only_ working setup I found:
- Go back 500 commits (!).
- Create `MockMicrogrid` with `api_client_streaming=True`. Otherwise it won't stream any data.

@llucax I'm at a loss. Do you think it makes sense to get to the bottom of this, or should we perhaps better create a spin-off task?

llucax: I wouldn't block this on the non-working benchmark, let's just create an issue to fix it. Just out of curiosity I'm asking an agent to investigate.

llucax:
I guess we the channels release this is ready for a final review, let me know (request a review) when it is ready for another round.

src/frequenz/sdk/microgrid/_data_sourcing/microgrid_api_source.py
simonvoelcker: One problem is see here is that this class is holding on to a reference to each component metric request indefinitely (because the teardown is not implemented yet).

The defensive thing to do then, in my opinion, is to send oneshot message every time (=outside the if) and catch a possible SenderClosedError in case the message had already been sent before.

src/frequenz/sdk/microgrid/_resampling.py
simonvoelcker: I see your point, but can't think of a clean solution either. Maybe it will be more obvious when we handle closing channels.

src/frequenz/sdk/microgrid/_data_sourcing/microgrid_api_source.py
llucax: I didn't get why the channel could be closed? I think this PR went out of my context window and I'm having trouble to remember the full context :P.

In any case catching sender closed errors is always a good thing.

benchmarks/timeseries/benchmark_datasourcing.py
llucax: * https://github.com/frequenz-floss/frequenz-sdk-python/pull/1389

simonvoelcker:
@llucax I think it's ready again. Tbh I also lost some context since the last time I worked on this and also kinda ran out of steam in the end. Would appreciate if we can move this over the line so I can continue with phase 2.

llucax:
Oh, needs to resolve a conflict in the benchmark now, but I will review anyway, it is a very isolated part.

src/frequenz/sdk/microgrid/_data_sourcing/microgrid_api_source.py
llucax, line 541: It seems we have a problem here. Found by AI but seems to make sense:

If the requested metric was already requested or the component ID is unknown (`category is None`), we just `return`, which means we don't call [`self._update_streams()`](https://github.com/frequenz-floss/frequenz-sdk-python/pull/1371/changes#diff-c2bf9d047ad0208040ac19de203626f6b3a5d8ca8517f120a8e1793d83fb0734R493), which means we don't call [`self._handle_data_stream()`](https://github.com/frequenz-floss/frequenz-sdk-python/pull/1371/changes#diff-c2bf9d047ad0208040ac19de203626f6b3a5d8ca8517f120a8e1793d83fb0734R428), which means we don't call [`self._get_metric_senders()`](https://github.com/frequenz-floss/frequenz-sdk-python/pull/1371/changes#diff-c2bf9d047ad0208040ac19de203626f6b3a5d8ca8517f120a8e1793d83fb0734R393), which means we don't call `await request.telem_stream_sender.send(telem_stream.new_receiver())`, which means we never reply to the requester.

I guess we can reply in here before the return, but this is getting very hard to follow, so not sure if we can find a better approach. Maybe do the replying completely outside all that call stack, and in this function, right after we either created the new channel or got the existing channel, but I guess this will imply awaiting on the task that is creating the new channel.

<details>
<summary>Full AI analysis including a proposed fix (long)</summary>

## What goes wrong

A direct caller now creates a fresh oneshot channel for **every** subscription request and waits for a `BroadcastReceiver` to be handed back.

You can see that in:

- `GridFrequency.subscribe()` at `src/frequenz/sdk/timeseries/_grid_frequency.py:116-130`
- `GridFrequency.new_receiver()`’s background path at `src/frequenz/sdk/timeseries/_grid_frequency.py:107-114, 132-140`
- `ResampledStreamFetcher.fetch_stream()` at `src/frequenz/sdk/timeseries/formulas/_resampled_stream_fetcher.py:52-65`

Each of those callers expects this sequence:

1. create `OneshotChannel`
2. send `ComponentMetricRequest(..., telem_stream_sender=...)`
3. wait on `telem_stream_receiver.receive()`

The request object itself makes the dedupe key explicit:

- `ComponentMetricRequest.get_channel_name()` is based on `namespace`, `component_id`, `metric`, and optional `start_time`
- see `src/frequenz/sdk/microgrid/_data_sourcing/_component_metric_request.py:59-73`

So two requests for the same logical stream have the same `get_channel_name()`, but **different oneshot senders**.

## The exact failure path

In `MicrogridApiSource.add_metric()`:

```python
for existing_request in self._req_streaming_metrics[comp_id][request.metric]:
    if existing_request.get_channel_name() == request.get_channel_name():
        # the requested metric is already being handled, so nothing to do.
        return
```

That is at:
- `src/frequenz/sdk/microgrid/_data_sourcing/microgrid_api_source.py:531-534`

The bug is that “nothing to do” is false now.

Why? Because the **only** place that replies to a caller’s oneshot sender is later, in `_get_metric_senders()`:

- `src/frequenz/sdk/microgrid/_data_sourcing/microgrid_api_source.py:412-423`

Specifically:

```python
await request.telem_stream_sender.send(telem_stream.new_receiver())
```

But `_get_metric_senders()` only iterates over the requests that are still present in `req_list`.

So if request B is deduped against request A:

- A stays in `req_list`
- B is dropped early
- only A gets its oneshot reply
- B waits forever on `receive()`

## Why this hangs forever

The second caller is waiting on a oneshot receiver that will never get a value.

There is no timeout in:
- `GridFrequency.subscribe()` (`_grid_frequency.py:128-129`)
- `GridFrequency._send_request()` (`_grid_frequency.py:138-140`)
- `ResampledStreamFetcher.fetch_stream()` (`_resampled_stream_fetcher.py:64-65`)

So if no one ever sends on that oneshot sender, the waiter just sits there.

## Why `subscribe()` + deprecated `new_receiver()` is broken too

`GridFrequency` has **two** request paths now:

### New path
`subscribe()` creates a fresh oneshot sender every time:
- `_grid_frequency.py:118-126`

### Deprecated path
`new_receiver()` uses a pre-created oneshot receiver stored on the instance and starts `_send_request()` once:
- `_grid_frequency.py:107-114`
- `_grid_frequency.py:132-140`

Both paths generate requests with the same logical stream key:

- namespace = `"grid-frequency"`
- same component id
- same metric = `AC_FREQUENCY`

So they collide in `add_metric()`.

That means:

- if `new_receiver()` gets there first, then `subscribe()` can be deduped and hang forever
- if `subscribe()` gets there first, then the background `_send_request()` started by `new_receiver()` can hang forever waiting on `self._telem_stream_receiver.receive()`

So the returned receiver from `new_receiver()` may exist, but it never gets wired to the underlying telemetry stream.

## Practical impact

This is high severity because it is:

- **deterministic** once the duplicate order happens
- **silent** — no exception, just a permanent wait
- **user-facing** — it affects public subscription APIs
- **sticky** — the waiting task doesn’t recover on its own

## The fix

The right non-breaking fix is:

> **dedupe the underlying stream, but still reply to every caller**

In other words:
- keep only one upstream telemetry channel per `get_channel_name()`
- but hand out a fresh `BroadcastReceiver` to **each** caller’s oneshot sender

### Important subtlety

A naïve fix would be “just append the duplicate request too”.

That does **not** work, because `_get_metric_senders()` would then create multiple senders for the same broadcast channel and send duplicate samples into the same stream.

So the fix has to separate:

1. **canonical stream registration**
2. **pending caller replies**

## Suggested fix shape

Introduce a pending-replies map keyed by `channel_name`, for example:

```python
self._pending_replies: dict[
    str, list[OneshotSender[BroadcastReceiver[Sample[Quantity]]]]
] = {}
```

### In `add_metric()`

- always register the caller’s oneshot sender in `_pending_replies[channel_name]`
- if this is a duplicate and the channel already exists, reply immediately
- if this is a duplicate and the channel does not exist yet, return and let the eventual stream setup flush all pending replies
- only the first request should be appended to `_req_streaming_metrics`

Sketch:

```python
async def add_metric(self, request: ComponentMetricRequest) -> None:
    comp_id = request.component_id
    category = await self._get_component_category(comp_id)
    if category is None:
        _logger.error("Unknown component ID: %d in request %s", comp_id, request)
        return

    channel_name = request.get_channel_name()
    self._pending_replies.setdefault(channel_name, []).append(
        request.telem_stream_sender
    )

    self._req_streaming_metrics.setdefault(comp_id, {}).setdefault(request.metric, [])

    for existing_request in self._req_streaming_metrics[comp_id][request.metric]:
        if existing_request.get_channel_name() == channel_name:
            if channel_name in self._channels:
                await self._flush_pending_replies(channel_name)
            return

    self._req_streaming_metrics[comp_id][request.metric].append(request)
    await self._update_streams(comp_id, category)
```

### Helper to answer everyone waiting for that logical stream

```python
async def _flush_pending_replies(self, channel_name: str) -> BroadcastSender[Sample[Quantity]]:
    if channel_name not in self._channels:
        self._channels[channel_name] = Broadcast(name=channel_name)

    telem_stream = self._channels[channel_name]

    for sender in self._pending_replies.pop(channel_name, []):
        try:
            await sender.send(telem_stream.new_receiver())
        except SenderClosedError:
            pass

    return telem_stream.new_sender()
```

### In `_get_metric_senders()`

Deduplicate by `channel_name`, reply to all pending callers for that channel, and create only **one** broadcast sender per logical stream:

```python
async def _get_metric_senders(...):
    all_senders = []
    for metric, req_list in requests.items():
        extraction_method = self._get_data_extraction_method(category, metric)
        senders = []
        seen_channel_names: set[str] = set()

        for request in req_list:
            channel_name = request.get_channel_name()
            if channel_name in seen_channel_names:
                continue
            seen_channel_names.add(channel_name)

            senders.append(await self._flush_pending_replies(channel_name))

        all_senders.append((extraction_method, senders))

    return all_senders
```

## Why this fix is good

It preserves all the good parts of the current design:

- still one underlying stream per logical metric request
- still one broadcast per logical stream
- still one fresh receiver per caller
- still non-breaking API

And it fixes both cases:

- duplicate direct subscriptions
- `subscribe()` mixed with deprecated `new_receiver()`

## One important note

This fix addresses the **duplicate request** bug.

It does **not** fix the separate invalid-request bug where data sourcing can return/raise without replying to the oneshot at all. That still needs its own explicit error/close path.

</details>

simonvoelcker:
@llucax yeah, I saw those and resolved them already locally. I'll push this after your review then, so as not to interfere. It's isolated, as you say.

src/frequenz/sdk/microgrid/_data_sourcing/microgrid_api_source.py
```diff
@@ -397,34 +390,40 @@ def _get_data_extraction_method(
...
+                channel_name = request.get_channel_name()
+                # Create missing channels
+                if channel_name not in self._channels:
+                    self._channels[channel_name] = Broadcast(name=channel_name)
+                telem_stream = self._channels[channel_name]
+                # Ensure request sender gets the telemetry stream (once)
+                try:
+                    await request.telem_stream_sender.send(telem_stream.new_receiver())
+                except SenderClosedError:
+                    pass
```
llucax, line 422: I wonder how this could happen and not be an issue, someone would have to create the oneshot channel, send the request and close the receiver/sender before receiving. I guess that could be a valid way to *cancel* a request. In any case, I think I would add at least a DEBUG log here saying why we are ignoring this exception (something like "assuming the request was cancelled by the requester").

src/frequenz/sdk/timeseries/battery_pool/_battery_pool.py
```diff
@@ -329,22 +341,48 @@ def power_status(self) -> ReceiverFetcher[BatteryPoolReport]:
...
-                self._pool_ref_store._power_manager_bounds_subscription_sender.send(sub)
+                self._send_request(forwarding_channel.new_sender(), request)
             )
         )
-        channel = self._pool_ref_store._channel_registry.get_or_create(
-            _power_managing._Report, sub.get_channel_name()
+
+        return forwarding_channel
+
+    async def _send_request(
```
llucax, line 367: It seems like we have a rece condition here. Also AI assisted finding but looks legit to me.

If `power_status` is called twice very fast, `self._report_stream_receiver` is overwritten, losing the reference to the first oneshot channel receiver, and the same could happen for `_pipe()`, but for this to happen the first call should have had time to actually give execution time to the `_send_request()` task so `self._pipe` is actually assigned, and while `await self._pipe.start()` is waiting, a new task could override it too. I guess this is much more unlikely, but yet still possible.

And there is also a potential race for `self._pool_ref_store._power_bounds_subs[sub.get_channel_name()]` too, which can lead to losing the reference to the task object. But I think this issue predates this PR, as we already had the task and `self._pool_ref_store._power_bounds_subs[sub.get_channel_name()]`. And I think this demonstrate why I didn't like very much the approach to create tasks to make async stuff look like async, but we have many of these before, so I guess it will a big undertaking of its own to get rid of this and make our async code more sound.

As for the fix, we should probably make some stuff task-local by passing it as an argument instead of using the instance shared state.

Since it took me a bit to understand the whole dimension of this, in case it is useful, this is the whole AI analysis (which includes details on a possible fix):

<details>
<summary>AI analysis and fix proposal (long)</summary>

## The Race Condition in `power_status` — Detailed Explanation

### The Setup

Each pool class (`BatteryPool`, `EVChargerPool`, `PVPool`) stores two pieces of per-request handshake state as **instance attributes**:

```python
self._report_stream_receiver: Receiver[Receiver[_Report]] | None = None
self._pipe: Pipe[_Report] | None = None
```

When `power_status` is accessed, it:
1. Creates a `OneshotChannel` pair (sender + receiver)
2. Stores the receiver on `self._report_stream_receiver` (overwrites any previous value)
3. Creates a `ReportRequest` carrying the sender
4. Spawns an async task (`_send_request`) that sends the request to the power manager, then **reads from `self._report_stream_receiver`** to get the reply

### The Race

`get_channel_name()` returns `f"power_manager.report.{self.component_ids=}.{self.priority=}"` — it's **deterministic** for a given pool instance. Two calls on the same instance produce the same key. Here's the interleaving:

```
Call A:                                     Call B:
──────                                      ──────
sender_A, recv_A = OneshotChannel()
self._report_stream_receiver = recv_A  ←── overwritten below
                                            sender_B, recv_B = OneshotChannel()
                                            self._report_stream_receiver = recv_B
task_A = create_task(_send_request(...))
                                            power_bounds_subs[key] = task_B  ←── overwrites task_A ref
                                            task_B = create_task(_send_request(...))

─── async tasks resume ───

task_A: sends request_A via power manager
        await self._report_stream_receiver.receive()
        ⬆ reads recv_B (not recv_A!) — self was overwritten

task_B: sends request_B via power manager
        await self._report_stream_receiver.receive()
        ⬆ also reads recv_B — contends with task_A
```

**Three bugs compound:**

| What gets overwritten | Consequence |
|---|---|
| `self._report_stream_receiver` | Task A reads from the wrong oneshot (recv_B). The actor correctly sends reply_A through sender_A → recv_A, but nobody reads recv_A. Task A either gets reply_B (cross-wired) or hangs forever waiting on recv_B if task B consumes it first. |
| `self._pipe` | When task B creates its Pipe and stores it in `self._pipe`, task A's Pipe loses its only reference and becomes eligible for GC. The comment on line 87 even says *"Keep a reference to prevent garbage collector from destroying pipe"* — the race defeats this exact protection. |
| `power_bounds_subs[key]` | The dict uses `get_channel_name()` as key, which is identical for both calls. Task A's handle is silently replaced. The task still runs (the event loop holds it), but it can no longer be cancelled via `stop()`. |

### Why This Hasn't Blown Up (Yet)

In practice, `power_status` is typically accessed once per pool instance during setup. The race requires two calls before the first task's `await` yields — a narrow window. But it's a latent correctness bug that becomes a hard-to-diagnose hang under any workload that re-subscribes.

---

## The Fix

The fix follows the pattern already used elsewhere in this codebase (`_resampling.py:63–110`, `_grid_frequency.py:116–130`): **make the oneshot receiver task-local** by passing it as a parameter, and **collect pipes in a list** instead of a single slot.

Here are the exact changes for all three files:

### 1. `_battery_pool.py`

**`__init__`** — replace the two instance attrs with a pipe list:

```python
# BEFORE (lines 86–88):
        self._report_stream_receiver: Receiver[Receiver[_Report]] | None = None
        # Keep a reference to prevent garbage collector from destroying pipe
        self._pipe: Pipe[_Report] | None = None

# AFTER:
        # Keep references to prevent garbage collector from destroying pipes
        self._pipes: list[Pipe[_Report]] = []
```

**`power_status`** — make the receiver local, pass it to the task:

```python
# BEFORE (lines 344–365):
        report_stream_sender, self._report_stream_receiver = OneshotChannel[
            BroadcastReceiver[_Report]
        ]()
        request = _power_managing.ReportRequest(
            source_id=self._source_id,
            priority=self._priority,
            component_ids=self._pool_ref_store._batteries,
            report_stream_sender=report_stream_sender,
        )

        forwarding_channel = Broadcast[_Report](
            name=request.get_channel_name() + ":Forwarded",
            resend_latest=True,
        )

        self._pool_ref_store._power_bounds_subs[request.get_channel_name()] = (
            asyncio.create_task(
                self._send_request(forwarding_channel.new_sender(), request)
            )
        )

        return forwarding_channel

# AFTER:
        report_stream_sender, report_stream_receiver = OneshotChannel[
            BroadcastReceiver[_Report]
        ]()
        request = _power_managing.ReportRequest(
            source_id=self._source_id,
            priority=self._priority,
            component_ids=self._pool_ref_store._batteries,
            report_stream_sender=report_stream_sender,
        )

        forwarding_channel = Broadcast[_Report](
            name=request.get_channel_name() + ":Forwarded",
            resend_latest=True,
        )

        self._pool_ref_store._power_bounds_subs[request.get_channel_name()] = (
            asyncio.create_task(
                self._send_request(
                    forwarding_channel.new_sender(),
                    request,
                    report_stream_receiver,
                )
            )
        )

        return forwarding_channel
```

**`_send_request`** — accept receiver as parameter, append pipe to list:

```python
# BEFORE (lines 367–385):
    async def _send_request(
        self,
        forwarding_sender: Sender[_Report],
        request: ReportRequest,
    ) -> None:
        """Send the report request and receive the report channel.

        Connect it via pipe to the channel that was returned in power_status.
        """
        await self._pool_ref_store._power_manager_bounds_subscription_sender.send(
            request
        )

        assert self._report_stream_receiver is not None
        report_receiver: Receiver[_Report] = (
            await self._report_stream_receiver.receive()
        )
        self._pipe = Pipe(report_receiver, forwarding_sender)
        await self._pipe.start()

# AFTER:
    async def _send_request(
        self,
        forwarding_sender: Sender[_Report],
        request: ReportRequest,
        report_stream_receiver: Receiver[Receiver[_Report]],
    ) -> None:
        """Send the report request and receive the report channel.

        Connect it via pipe to the channel that was returned in power_status.
        """
        await self._pool_ref_store._power_manager_bounds_subscription_sender.send(
            request
        )

        report_receiver: Receiver[_Report] = await report_stream_receiver.receive()
        pipe = Pipe(report_receiver, forwarding_sender)
        self._pipes.append(pipe)
        await pipe.start()
```

### 2. `_ev_charger_pool.py`

The same three-part change, adjusted for this file's naming conventions:

**`__init__`** (lines 73–75):
```python
# BEFORE:
        self._report_stream_receiver: Receiver[Receiver[_Report]] | None = None
        # Keep a reference to prevent garbage collector from destroying pipe
        self._pipe: Pipe[_Report] | None = None

# AFTER:
        # Keep references to prevent garbage collector from destroying pipes
        self._pipes: list[Pipe[_Report]] = []
```

**`power_status`** (lines 186–207):
```python
# BEFORE:
        report_stream_sender, self._report_stream_receiver = OneshotChannel[
            BroadcastReceiver[_Report]
        ]()
        request = _power_managing.ReportRequest(
            source_id=self._source_id,
            priority=self._priority,
            component_ids=self._pool_ref_store.component_ids,
            report_stream_sender=report_stream_sender,
        )

        forwarding_channel = Broadcast[_Report](
            name=request.get_channel_name() + ":Forwarded",
            resend_latest=True,
        )

        self._pool_ref_store.power_bounds_subs[request.get_channel_name()] = (
            asyncio.create_task(
                self._send_request(forwarding_channel.new_sender(), request)
            )
        )

        return forwarding_channel

# AFTER:
        report_stream_sender, report_stream_receiver = OneshotChannel[
            BroadcastReceiver[_Report]
        ]()
        request = _power_managing.ReportRequest(
            source_id=self._source_id,
            priority=self._priority,
            component_ids=self._pool_ref_store.component_ids,
            report_stream_sender=report_stream_sender,
        )

        forwarding_channel = Broadcast[_Report](
            name=request.get_channel_name() + ":Forwarded",
            resend_latest=True,
        )

        self._pool_ref_store.power_bounds_subs[request.get_channel_name()] = (
            asyncio.create_task(
                self._send_request(
                    forwarding_channel.new_sender(),
                    request,
                    report_stream_receiver,
                )
            )
        )

        return forwarding_channel
```

**`_send_request`** (lines 209–225):
```python
# BEFORE:
    async def _send_request(
        self,
        forwarding_sender: Sender[_Report],
        request: ReportRequest,
    ) -> None:
        """Send the report request and receive the report channel.

        Connect it via pipe to the channel that was returned in power_status.
        """
        await self._pool_ref_store.power_manager_bounds_subs_sender.send(request)

        assert self._report_stream_receiver is not None
        report_receiver: Receiver[_Report] = (
            await self._report_stream_receiver.receive()
        )
        self._pipe = Pipe(report_receiver, forwarding_sender)
        await self._pipe.start()

# AFTER:
    async def _send_request(
        self,
        forwarding_sender: Sender[_Report],
        request: ReportRequest,
        report_stream_receiver: Receiver[Receiver[_Report]],
    ) -> None:
        """Send the report request and receive the report channel.

        Connect it via pipe to the channel that was returned in power_status.
        """
        await self._pool_ref_store.power_manager_bounds_subs_sender.send(request)

        report_receiver: Receiver[_Report] = await report_stream_receiver.receive()
        pipe = Pipe(report_receiver, forwarding_sender)
        self._pipes.append(pipe)
        await pipe.start()
```

### 3. `_pv_pool.py`

Identical structure to EV charger:

**`__init__`** (lines 68–70):
```python
# BEFORE:
        self._report_stream_receiver: Receiver[Receiver[_Report]] | None = None
        # Keep a reference to prevent garbage collector from destroying pipe
        self._pipe: Pipe[_Report] | None = None

# AFTER:
        # Keep references to prevent garbage collector from destroying pipes
        self._pipes: list[Pipe[_Report]] = []
```

**`power_status`** (lines 158–179):
```python
# BEFORE:
        report_stream_sender, self._report_stream_receiver = OneshotChannel[
            BroadcastReceiver[_Report]
        ]()
        ...
        self._pool_ref_store.power_bounds_subs[request.get_channel_name()] = (
            asyncio.create_task(
                self._send_request(forwarding_channel.new_sender(), request)
            )
        )

# AFTER:
        report_stream_sender, report_stream_receiver = OneshotChannel[
            BroadcastReceiver[_Report]
        ]()
        ...
        self._pool_ref_store.power_bounds_subs[request.get_channel_name()] = (
            asyncio.create_task(
                self._send_request(
                    forwarding_channel.new_sender(),
                    request,
                    report_stream_receiver,
                )
            )
        )
```

**`_send_request`** (lines 181–197):
```python
# BEFORE:
    async def _send_request(
        self,
        forwarding_sender: Sender[_Report],
        request: ReportRequest,
    ) -> None:
        ...
        await self._pool_ref_store.power_manager_bounds_subs_sender.send(request)

        assert self._report_stream_receiver is not None
        report_receiver: Receiver[_Report] = (
            await self._report_stream_receiver.receive()
        )
        self._pipe = Pipe(report_receiver, forwarding_sender)
        await self._pipe.start()

# AFTER:
    async def _send_request(
        self,
        forwarding_sender: Sender[_Report],
        request: ReportRequest,
        report_stream_receiver: Receiver[Receiver[_Report]],
    ) -> None:
        ...
        await self._pool_ref_store.power_manager_bounds_subs_sender.send(request)

        report_receiver: Receiver[_Report] = await report_stream_receiver.receive()
        pipe = Pipe(report_receiver, forwarding_sender)
        self._pipes.append(pipe)
        await pipe.start()
```

---

### Optional Further Hardening: Unique Task Keys

The `power_bounds_subs` dict uses `get_channel_name()` as key, which is the same for repeated calls on the same instance. This silently drops the old task reference. While the task still runs (the event loop keeps it alive), `stop()` can no longer cancel it. A straightforward fix would be to append a uuid to the key:

```python
task_key = f"{request.get_channel_name()}:{uuid.uuid4()}"
self._pool_ref_store._power_bounds_subs[task_key] = asyncio.create_task(...)
```

This is lower severity since it only affects cleanup, but it prevents orphaned tasks from surviving `stop()`.

---

### Summary

| Root cause | Shared mutable state (`self._report_stream_receiver`, `self._pipe`) written by sync code, read by async tasks |
|---|---|
| Fix principle | Make per-request state **task-local**: pass the oneshot receiver as a function parameter, collect pipes in a list |
| Files changed | 3 files, ~15 lines each, all mechanical |
| Risk | Very low — the change is strictly narrower in scope (removes shared state), follows existing codebase patterns, and the external contract (`ReceiverFetcher` return type) is unchanged |
| Precedent in codebase | `_resampling.py:63–110`, `_grid_frequency.py:116–130`, `_voltage_streamer.py:139–166` all use this local-oneshot pattern correctly |

</details>

> [!NOTE]
> This applies to the other pools too.

src/frequenz/sdk/microgrid/_resampling.py
```diff
@@ -50,15 +51,36 @@ def __init__(  # pylint: disable=too-many-arguments
...
+        sender, receiver = OneshotChannel[BroadcastReceiver[Sample[Quantity]]]()
+        data_source_request = ComponentMetricRequest(
+            namespace=request.namespace + ":Source",
+            component_id=request.component_id,
+            metric=request.metric,
+            start_time=request.start_time,
+            telem_stream_sender=sender,
+        )
+        await self._data_sourcing_request_sender.send(data_source_request)
+        return await receiver.receive()
```
llucax, line 83: It seems this is introducing a new subtle concurrency bug (also found with AI).

`ComponentMetricsResamplingActor` processes subscription requests strictly serially in `_process_resampling_requests()` via `async for request ...: await self._subscribe(request)` ([`src/frequenz/sdk/microgrid/_resampling.py:112-115`](https://github.com/frequenz-floss/frequenz-sdk-python/pull/1371/changes#diff-2e16f5caaa07abb06b56644d58c11c8f3c4d023ca1093a75b5bf751047b21486R112-R115)). 

Inside that, `_subscribe()` calls `_subscribe_to_data_source()` (this method), which sends a data-sourcing request and then waits synchronously for a oneshot reply with `return await receiver.receive()`. From the channels source, that oneshot wait has **no built-in timeout**: it wakes only if someone calls `send()` or `aclose()`. If neither happens, it can **wait forever**.

Before the PR a slow upstream setup might delay **data arrival**, but not **subscription registration** in the resampling actor. The resampling actor could still keep draining its incoming request queue because it only needed the channel name and local registry wiring.

After the PR a slow upstream setup delays **subscription registration itself**. Now the resampling actor cannot process later requests until the current one gets its oneshot reply.

<details>
<summary>Full AI analysis and fix proposal (long)</summary>

- **What actually stalls**
  - Resampling request A arrives.
  - `_process_resampling_requests()` starts handling A and blocks in `_subscribe_to_data_source()` waiting for the data-sourcing reply.
  - Requests B, C, D are still accepted into the resampling actor’s request channel, but they are not processed because the loop never gets past A.
  - The queue is bounded at `500` (`src/frequenz/sdk/microgrid/_data_pipeline.py:54-59`, `460-468`), so after enough blocked requests, senders themselves start back-pressuring.

- **Why the reply can disappear**
  - The happy path is `MicrogridApiSource._get_metric_senders()` sending `telem_stream.new_receiver()` back through `request.telem_stream_sender` (`src/frequenz/sdk/microgrid/_data_sourcing/microgrid_api_source.py:415-420`).
  - But there are bad paths where no reply is sent at all. One concrete one is the unknown-component case in `add_metric()`, which logs and returns without replying (`microgrid_api_source.py:521-525`).
  - So the resampling actor can end up waiting forever on a reply that will never come.

- **Implications / risk**
  - This is broader than “one request fails”: one bad subscription can delay **unrelated** later subscriptions behind it.
  - It does **not** necessarily crash the actor, which makes it worse operationally: a stuck `await receiver.receive()` just hangs; it doesn’t trigger restart/recovery.
  - In practice this means callers waiting for new resampled streams can hang, and under load the request queue can fill and propagate backpressure outward.

## Recommended fix

I’d treat this as a two-layer fix:

1. **Fix the root cause in data sourcing**  
   Data sourcing should never silently drop a request without replying. On invalid requests, it should explicitly fail the oneshot handshake instead of just returning.

2. **Make resampling robust against missing replies anyway**  
   Even if data sourcing is fixed, resampling should not let one bad handshake stall all later requests.

The safest non-breaking design is:

### A. Stop processing subscriptions inline
Instead of this:

```python
async for request in self._resampling_request_receiver:
    await self._subscribe(request)
```

process each request in its own task and supervise those tasks.

### B. Coalesce duplicate in-flight subscriptions
Because concurrent tasks can race on the same `request_channel_name`, keep a `pending_subscriptions` map so repeated requests share the same initialization task/future instead of creating duplicate source subscriptions.

### C. Bound the oneshot wait
Wrap the data-sourcing reply wait in a timeout, e.g. `asyncio.timeout(...)`, so a lost reply becomes a logged failure instead of a permanent hang.

## Suggested shape

Something like this internally:

```python
self._pending_subscriptions: dict[str, asyncio.Task[Broadcast[Sample[Quantity]]]] = {}
self._subscription_tasks: set[asyncio.Task[None]] = set()
```

Then:

```python
async def _subscribe_to_data_source(
    self, request: ComponentMetricRequest
) -> BroadcastReceiver[Sample[Quantity]]:
    sender, receiver = OneshotChannel[BroadcastReceiver[Sample[Quantity]]]()
    data_source_request = ComponentMetricRequest(
        namespace=request.namespace + ":Source",
        component_id=request.component_id,
        metric=request.metric,
        start_time=request.start_time,
        telem_stream_sender=sender,
    )
    await self._data_sourcing_request_sender.send(data_source_request)
    async with asyncio.timeout(5.0):
        return await receiver.receive()
```

And coalesced initialization:

```python
async def _ensure_subscription(
    self, request: ComponentMetricRequest
) -> Broadcast[Sample[Quantity]]:
    request_channel_name = request.get_channel_name()

    if channel := self._data_sink_channels.get(request_channel_name):
        return channel

    if pending := self._pending_subscriptions.get(request_channel_name):
        return await pending

    async def init() -> Broadcast[Sample[Quantity]]:
        data_source = await self._subscribe_to_data_source(request)
        data_sink_channel = Broadcast(name=request_channel_name, resend_latest=True)
        self._data_sink_channels[request_channel_name] = data_sink_channel
        self._resampler.add_timeseries(
            name=request_channel_name,
            source=data_source,
            sink=data_sink_channel.new_sender().send,
        )
        return data_sink_channel

    task = asyncio.create_task(init(), name=f"init:{request_channel_name}")
    self._pending_subscriptions[request_channel_name] = task
    try:
        return await task
    finally:
        self._pending_subscriptions.pop(request_channel_name, None)
```

Then the request loop can stay responsive:

```python
async def _handle_subscription_request(self, request: ComponentMetricRequest) -> None:
    channel = await self._ensure_subscription(request)
    await request.telem_stream_sender.send(channel.new_receiver())

async def _process_resampling_requests(self) -> None:
    async for request in self._resampling_request_receiver:
        task = asyncio.create_task(
            self._handle_subscription_request(request),
            name=f"subscribe:{request.get_channel_name()}",
        )
        self._subscription_tasks.add(task)
        task.add_done_callback(self._subscription_tasks.discard)
```

## Why this is the best fix

- It removes the **head-of-line blocking**.
- It preserves the current external API.
- It prevents duplicate initialization races for the same request.
- It turns “wait forever” into a bounded failure.
- It still works even if data sourcing regresses again later.

If you want, I can also turn this into a minimal patch against `_resampling.py` and sketch the matching data-sourcing change.

</details>

src/frequenz/sdk/microgrid/_power_managing/_power_managing_actor.py
```diff
@@ -233,18 +237,26 @@ async def _run(self) -> None:
...
+                    report_channel = Broadcast[_Report](name=channel_name)
+                    report_channel.resend_latest = True
+                    self._channels[channel_name] = report_channel
+                channel = self._channels[channel_name]
+
+                # Ensure receiver is sent via oneshot channel (once)
+                try:
+                    await sub.report_stream_sender.send(channel.new_receiver())
+                except SenderClosedError:
+                    pass
```
llucax, line 252: Also in here, I would always *document* why we ignore an error with a debug log line. If feel if something goes wrong here we'll be too blind otherwise when trying to figure out what happened.

llucax (commented):
I hate to revive old discussions, but the more we insist with this approach of using oneshot to replace the channel registry, the more I'm convinced this is not the right approach.

In this review round *I* found 3 serious and **very obscure** concurrency bugs. To be honest, I'm not sure if I would have caught these issues if I didn't used AI, as some issues can only be understood when looking at parts of the code that are not even in the diff (I asked for the AI review first, so we'll never know). I still spent about 3 hours trying to understand and validate the issues, so it is also not like I just asked AI and it was zero effort. That said it is also a possibility that I didn't get it right and some issue is not that serious, or extremely unlikely, but still...

This makes me very little confident that we even uncovered all potential issues with this approach, so I suggest stepping back again and thinking if we really want to move forward with this.

My feeling is we are planting a minefield here, and for what? Why are we removing the channel registry? If it is because of auto-closing, can't we just introduce auto-closing while keeping the channel registry? Maybe we can keep the channel class as non-deprecated and allow both usages of channels, one owning the whole channel and one driven by senders/receivers, or we can have some other mechanism to be able to spawn more senders/receivers for an existing channel as long as it is not closed.

I'm sure we can find a solution, but to me, adding this level of complexity is not acceptable, and will hunt us in the future even if we can get this PR right and bug free, we'll need to add more code that will have the same likelihood of getting obscure concurrency bugs.

In any case, I also asked the AI to add tests cases to expose these issues, I think it makes sense to add them, and here is a PR to add them to v1.x.x:

* https://github.com/frequenz-floss/frequenz-sdk-python/pull/1392 

<details>
<summary>New test cases diff for this branch</summary>

> [!NOTE]
> This is an old iteration, I wanted to update it but I'm hitting some rate limiting :D 

```diff
diff --git a/tests/actor/test_resampling.py b/tests/actor/test_resampling.py
index 6e8ee9389..f9e6dce73 100644
--- a/tests/actor/test_resampling.py
+++ b/tests/actor/test_resampling.py
@@ -266,3 +266,47 @@ def _now
         )
 
         await resampling_actor._resampler.stop()  # pylint: disable=protected-access
+
+
+async def test_stalled_request_does_not_block_later_subscriptions() -> None:
+    """Ensure one stalled data-source handshake doesn't block later requests."""
+    data_source_req_chan = Broadcast[ComponentMetricRequest](name="data-source-req")
+    data_source_req_recv = data_source_req_chan.new_receiver()
+    resampling_req_chan = Broadcast[ComponentMetricRequest](name="resample-req")
+    resampling_req_sender = resampling_req_chan.new_sender()
+
+    async with ComponentMetricsResamplingActor(
+        data_sourcing_request_sender=data_source_req_chan.new_sender(),
+        resampling_request_receiver=resampling_req_chan.new_receiver(),
+        config=ResamplerConfig2(
+            resampling_period=timedelta(seconds=0.2),
+            max_data_age_in_periods=2,
+        ),
+    ):
+        first_sender, _ = OneshotChannel[BroadcastReceiver[Sample[Quantity]]]()
+        second_sender, _ = OneshotChannel[BroadcastReceiver[Sample[Quantity]]]()
+
+        first_request = ComponentMetricRequest(
+            namespace="Resampling-A",
+            component_id=ComponentId(9),
+            metric=Metric.BATTERY_SOC_PCT,
+            start_time=None,
+            telem_stream_sender=first_sender,
+        )
+        second_request = ComponentMetricRequest(
+            namespace="Resampling-B",
+            component_id=ComponentId(10),
+            metric=Metric.AC_ACTIVE_POWER,
+            start_time=None,
+            telem_stream_sender=second_sender,
+        )
+
+        await resampling_req_sender.send(first_request)
+        stalled_data_source_request = await data_source_req_recv.receive()
+        assert stalled_data_source_request.component_id == ComponentId(9)
+
+        await resampling_req_sender.send(second_request)
+        later_data_source_request = await asyncio.wait_for(
+            data_source_req_recv.receive(), timeout=0.1
+        )
+        assert later_data_source_request.component_id == ComponentId(10)
diff --git a/tests/microgrid/test_data_sourcing.py b/tests/microgrid/test_data_sourcing.py
index c18031bc9..577eee20c 100644
--- a/tests/microgrid/test_data_sourcing.py
+++ b/tests/microgrid/test_data_sourcing.py
@@ -11,7 +11,13 @@
 
 import pytest
 import pytest_mock
-from frequenz.channels import Broadcast, BroadcastReceiver, OneshotChannel
+from frequenz.channels import (
+    Broadcast,
+    BroadcastReceiver,
+    OneshotChannel,
+    Receiver,
+    ReceiverStoppedError,
+)
 from frequenz.client.common.microgrid import MicrogridId
 from frequenz.client.common.microgrid.components import ComponentId
 from frequenz.client.microgrid.component import (
@@ -37,11 +43,30 @@
 )
 from frequenz.sdk.timeseries import Sample
 
+from ..utils.receive_timeout import Timeout
+
 T = TypeVar("T", bound=ComponentData)
+U = TypeVar("U")
 
 _MICROGRID_ID = MicrogridId(1)
 
 
+class Stopped:
+    """Sentinel for a closed receiver."""
+
+
+async def _receive_or_timeout(
+    receiver: Receiver[U], timeout: float = 0.1
+) -> U | type[Timeout] | type[Stopped]:
+    """Receive a message, or return a sentinel if the receiver stops or times out."""
+    try:
+        return await asyncio.wait_for(receiver.receive(), timeout=timeout)
+    except ReceiverStoppedError:
+        return Stopped
+    except asyncio.TimeoutError:
+        return Timeout
+
+
 @pytest.fixture
 def mock_connection_manager(mocker: pytest_mock.MockFixture) -> mock.Mock:
     """Fixture for getting a mock connection manager."""
@@ -124,6 +149,102 @@ def mock_connection_manager
             assert expected_sample_value + i == sample.value.base_value
 
 
+async def test_duplicate_requests_reply_to_all_callers(
+    mock_connection_manager: mock.Mock,  # pylint: disable=redefined-outer-name,unused-argument
+) -> None:
+    """Ensure duplicate direct subscriptions reply to every requester."""
+    req_chan = Broadcast[ComponentMetricRequest](name="data_sourcing_requests")
+    req_sender = req_chan.new_sender()
+
+    async with DataSourcingActor(req_chan.new_receiver()):
+        first_sender, first_receiver = OneshotChannel[
+            BroadcastReceiver[Sample[Quantity]]
+        ]()
+        second_sender, second_receiver = OneshotChannel[
+            BroadcastReceiver[Sample[Quantity]]
+        ]()
+
+        request = ComponentMetricRequest(
+            "test-namespace",
+            ComponentId(4),
+            Metric.AC_ACTIVE_POWER,
+            None,
+            first_sender,
+        )
+        duplicate_request = ComponentMetricRequest(
+            "test-namespace",
+            ComponentId(4),
+            Metric.AC_ACTIVE_POWER,
+            None,
+            second_sender,
+        )
+
+        await req_sender.send(request)
+        _ = await asyncio.wait_for(first_receiver.receive(), timeout=1.0)
+
+        await req_sender.send(duplicate_request)
+        _ = await asyncio.wait_for(second_receiver.receive(), timeout=1.0)
+
+
+async def test_unknown_component_request_does_not_wait_forever(
+    mock_connection_manager: mock.Mock,  # pylint: disable=redefined-outer-name,unused-argument
+) -> None:
+    """Ensure invalid component requests fail fast instead of hanging callers."""
+    req_chan = Broadcast[ComponentMetricRequest](name="data_sourcing_requests")
+    req_sender = req_chan.new_sender()
+
+    async with DataSourcingActor(req_chan.new_receiver()):
+        telem_stream_sender, telem_stream_receiver = OneshotChannel[
+            BroadcastReceiver[Sample[Quantity]]
+        ]()
+        request = ComponentMetricRequest(
+            "test-namespace",
+            ComponentId(999),
+            Metric.AC_ACTIVE_POWER,
+            None,
+            telem_stream_sender,
+        )
+
+        await req_sender.send(request)
+        result = await _receive_or_timeout(telem_stream_receiver)
+        assert result is not Timeout
+
+
+async def test_invalid_metric_request_does_not_block_later_requests(
+    mock_connection_manager: mock.Mock,  # pylint: disable=redefined-outer-name,unused-argument
+) -> None:
+    """Ensure a bad request doesn't stall later valid subscriptions."""
+    req_chan = Broadcast[ComponentMetricRequest](name="data_sourcing_requests")
+    req_sender = req_chan.new_sender()
+
+    async with DataSourcingActor(req_chan.new_receiver()):
+        invalid_sender, _ = OneshotChannel[BroadcastReceiver[Sample[Quantity]]]()
+        valid_sender, valid_receiver = OneshotChannel[
+            BroadcastReceiver[Sample[Quantity]]
+        ]()
+
+        invalid_request = ComponentMetricRequest(
+            "test-namespace-invalid",
+            ComponentId(4),
+            Metric.BATTERY_SOC_PCT,
+            None,
+            invalid_sender,
+        )
+        valid_request = ComponentMetricRequest(
+            "test-namespace-valid",
+            ComponentId(4),
+            Metric.AC_ACTIVE_POWER,
+            None,
+            valid_sender,
+        )
+
+        await req_sender.send(invalid_request)
+        await req_sender.send(valid_request)
+
+        result = await _receive_or_timeout(valid_receiver)
+        assert result is not Timeout
+
+
 def _new_meter_data(
     component_id: ComponentId, timestamp: datetime, value: float
 ) -> MeterData:
diff --git a/tests/timeseries/test_power_status_subscription_races.py b/tests/timeseries/test_power_status_subscription_races.py
new file mode 100644
index 000000000..a99ce5c02
--- /dev/null
+++ b/tests/timeseries/test_power_status_subscription_races.py
@@ -0,0 +1,140 @@
+# License: MIT
+# Copyright © 2026 Frequenz Energy-as-a-Service GmbH
+
+"""Regression tests for pool power_status subscription races."""
+
+import asyncio
+from types import SimpleNamespace
+
+from frequenz.channels import Broadcast, Receiver, Sender
+from frequenz.client.common.microgrid.components import ComponentId
+from frequenz.quantities import Power
+
+from frequenz.sdk import timeseries
+from frequenz.sdk.microgrid._power_managing import ReportRequest, _Report
+from frequenz.sdk.timeseries.battery_pool._battery_pool import BatteryPool
+from frequenz.sdk.timeseries.ev_charger_pool._ev_charger_pool import EVChargerPool
+from frequenz.sdk.timeseries.pv_pool._pv_pool import PVPool
+
+
+def _new_report(target_power_watts: float) -> _Report:
+    """Create a distinct report for subscription assertions."""
+    target_power = Power.from_watts(target_power_watts)
+    return _Report(
+        target_power=target_power,
+        _inclusion_bounds=timeseries.Bounds(target_power, target_power),
+        _exclusion_bounds=None,
+    )
+
+
+async def _assert_same_instance_subscriptions_are_independent(
+    pool: BatteryPool | EVChargerPool | PVPool,
+    requests_rx: Receiver[ReportRequest],
+) -> None:
+    """Ensure two subscriptions on the same pool instance remain independent."""
+    existing_tasks = {
+        task for task in asyncio.all_tasks() if task is not asyncio.current_task()
+    }
+
+    try:
+        first_status_rx = pool.power_status.new_receiver()
+        second_status_rx = pool.power_status.new_receiver()
+
+        await asyncio.sleep(0)
+
+        first_request = await asyncio.wait_for(requests_rx.receive(), timeout=1.0)
+        second_request = await asyncio.wait_for(requests_rx.receive(), timeout=1.0)
+
+        first_report_channel = Broadcast[_Report](name="first-report", resend_latest=True)
+        second_report_channel = Broadcast[_Report](
+            name="second-report", resend_latest=True
+        )
+
+        await first_request.report_stream_sender.send(first_report_channel.new_receiver())
+        await second_request.report_stream_sender.send(
+            second_report_channel.new_receiver()
+        )
+
+        await asyncio.sleep(0)
+
+        first_report = _new_report(100.0)
+        second_report = _new_report(200.0)
+
+        await first_report_channel.new_sender().send(first_report)
+        await second_report_channel.new_sender().send(second_report)
+
+        received_first_report = await asyncio.wait_for(
+            first_status_rx.receive(), timeout=0.1
+        )
+        received_second_report = await asyncio.wait_for(
+            second_status_rx.receive(), timeout=0.1
+        )
+
+        assert received_first_report.target_power == first_report.target_power
+        assert received_second_report.target_power == second_report.target_power
+    finally:
+        created_tasks = [
+            task
+            for task in asyncio.all_tasks()
+            if task is not asyncio.current_task() and task not in existing_tasks
+        ]
+        for task in created_tasks:
+            task.cancel()
+        await asyncio.gather(*created_tasks, return_exceptions=True)
+
+
+async def test_battery_pool_power_status_same_instance_subscriptions_are_independent() -> None:
+    """Ensure battery pool power_status subscriptions don't race each other."""
+    requests_channel = Broadcast[ReportRequest](name="battery-pool-requests")
+    pool = BatteryPool(
+        pool_ref_store=SimpleNamespace(
+            _batteries=frozenset({ComponentId(8), ComponentId(18)}),
+            _power_manager_bounds_subscription_sender=requests_channel.new_sender(),
+            _power_bounds_subs={},
+        ),
+        name="battery-pool",
+        priority=5,
+    )
+
+    await _assert_same_instance_subscriptions_are_independent(
+        pool,
+        requests_channel.new_receiver(),
+    )
+
+
+async def test_ev_charger_pool_power_status_same_instance_subscriptions_are_independent() -> None:
+    """Ensure EV charger pool power_status subscriptions don't race each other."""
+    requests_channel = Broadcast[ReportRequest](name="ev-pool-requests")
+    pool = EVChargerPool(
+        pool_ref_store=SimpleNamespace(
+            component_ids=frozenset({ComponentId(12), ComponentId(22)}),
+            power_manager_bounds_subs_sender=requests_channel.new_sender(),
+            power_bounds_subs={},
+        ),
+        name="ev-pool",
+        priority=5,
+    )
+
+    await _assert_same_instance_subscriptions_are_independent(
+        pool,
+        requests_channel.new_receiver(),
+    )
+
+
+async def test_pv_pool_power_status_same_instance_subscriptions_are_independent() -> None:
+    """Ensure PV pool power_status subscriptions don't race each other."""
+    requests_channel = Broadcast[ReportRequest](name="pv-pool-requests")
+    pool = PVPool(
+        pool_ref_store=SimpleNamespace(
+            component_ids=frozenset({ComponentId(28), ComponentId(38)}),
+            power_manager_bounds_subs_sender=requests_channel.new_sender(),
+            power_bounds_subs={},
+        ),
+        name="pv-pool",
+        priority=5,
+    )
+
+    await _assert_same_instance_subscriptions_are_independent(
+        pool,
+        requests_channel.new_receiver(),
+    )
```

</details>

simonvoelcker:
@llucax thank you for raising these points. I agree that we're adding a lot of potentially problematic code here, and it will be even more in phase 2 (teardown of closed channels / propagating closed senders) which I have lying around on a branch locally.

I am okay with closing this PR. I am not married to it just because I wrote it, this was simply my second task overall and I did not oversee the consequences yet or have enough of an overview myself to propose a better alternative. @shsms what do you think?

llucax:
Sorry for the late reply, I somehow missed the notification.

> @llucax thank you for raising these points. I agree that we're adding a lot of potentially problematic code here, and it will be even more in phase 2 (teardown of closed channels / propagating closed senders) which I have lying around on a branch locally.

OK, interesting, I was expecting auto-closing to actually simplify the code, so you are saying it makes it even more complex? Maybe it is worth creating a draft PR with it so we can have a better idea on how it makes things worse/better, at least if you have something that could be useful to share.

> I am okay with closing this PR. I am not married to it just because I wrote it, this was simply my second task overall and I did not oversee the consequences yet or have enough of an overview myself to propose a better alternative. 

Yeah, to be honest I also didn't saw this coming. I think removing the channel registry is still a good thing, it has it own issues, it is just I wouldn't remove it at all cost.

PR #1372, opened 2026-03-05
Implement EventResampler for Cascaded Resampling

malteschaaf:
## Problem

When cascading Timer-based resamplers (e.g., 1s → 10s) with `align_to=UNIX_EPOCH`, samples can be lost at window boundaries due to timing synchronization issues.

**Example:**
- Resampler 1 emits samples at: 1s, 2s, ..., 10s, 11s, ...
- Resampler 2 has windows: [0-10s), [10-20s)
- A sample from Resampler 1 arriving at exactly t=10.0s may be excluded from the [0-10s) window
- Result: **DATA LOSS** 

This occurs because both resamplers use independent timers aligned to the same epoch, but their tick times are not synchronized.

## Solution

`EventResampler` uses event-driven window management instead of fixed timer intervals. Windows are emitted when a sample arrives with a timestamp >= current window_end, not on a timer schedule.

**Benefits:**
- No data loss at boundaries (samples trigger window closure)
- Seamless cascading (no synchronization issues)
- Window alignment maintained through simple addition
- Event-driven (no polling, resource-efficient)

### Core Components

1. **EventResampler class**: Event-driven resampler inheriting from `Resampler`
   - Overrides `resample()` with event-driven loop
   - Implements `_process_sample()` for window state management
   - Uses `_calculate_window_end()` (simplified for event-driven scenario)

2. **Sample callback mechanism**: Added to `StreamingHelper`
   - `register_sample_callback()` method
   - Callback invoked when sample arrives
   - Enables event-driven processing without polling

3. **Refactored `_emit_window()`**: Extracted from `Resampler.resample()`
   - Shared by both Timer-based and Event-driven resamplers
   - Reduces code duplication
   - Clear separation of concerns

### Design Decision: asyncio.Queue + Callbacks

After evaluating multiple approaches (polling, direct channels, event signals, task-based multiplexing), I chose `asyncio.Queue` with callbacks because:

- **Encapsulation**: No protected-access violations. Communication through public APIs only.
- **Event-Driven**: Truly async/await pattern (`await queue.get()`), not polling.
- **Simplicity**: Minimal code, immediately understandable, no complex state tracking.
- **Type Safety**: Generic `Queue[Sample[Quantity]]` ensures type correctness.
- **Decoupling**: Components don't need to know internal implementation details.
- **Built for Purpose**: `asyncio.Queue` is designed specifically for this use case.

#### StreamingHelper
- Added `_sample_callback` attribute to store callback function
- Added `register_sample_callback()` method to register async callback
- Modified `_receive_samples()` to invoke callback when sample arrives

#### Resampler
- Extracted `_emit_window()` method from `resample()`
- Refactored `resample()` to delegate window emission to `_emit_window()`

#### EventResampler (New)
- Event-driven resampler for cascaded scenarios
- Uses sample queue fed by callbacks instead of timer ticks
- Maintains window alignment through simple addition
- Simplified `_calculate_window_end()` (no warm-up period needed)

#### Tests
- Comprehensive test suite for EventResampler
- Parametrized tests for both aligned and non-aligned scenarios
- Tests verify no data loss at boundaries, proper alignment, and window management

## Open Question: Task Cancellation in `remove_timeseries()`

**Observation:** When `remove_timeseries()` is called, the `_receiving_task` continues running in the background. It will:
- Continue reading from the source
- Continue adding samples to the buffer
- Continue invoking registered callbacks (EventResampler)

**Current behavior:**
```python
def remove_timeseries(self, source: Source) -> bool:
    try:
        del self._resamplers[source]
    except KeyError:
        return False
    return True
```

**Impact:**
- **Memory Leak**: Samples accumulate in the buffer but are never consumed
- **Resource Waste**: Task runs unnecessarily
- **Critical for EventResampler**: Queue grows unbounded with samples from removed sources
- **Relevant for TimerResampler**: Buffer eventually stabilizes due to maxlen, but still wastes resources

**Question for reviewers:**
Should `remove_timeseries()` cancel the receiving task for both `TimerResampler` and `EventResampler`? If yes, we need to either:
1. Make `remove_timeseries()` async and call `await helper.stop()`
2. Add a synchronous `cancel_receiving()` method to `_StreamingHelper`

This should be addressed as a separate issue for both resamplers, not just EventResampler.

## Backward Compatibility

- No breaking changes
- `Resampler` API unchanged
- `StreamingHelper` additions are backward compatible (callback defaults to None)
- `EventResampler` is opt-in (users must explicitly use it)

## Usage
```python
from datetime import timedelta
from frequenz.sdk.timeseries import EventResampler, ResamplerConfig

# Configure for cascaded resampling
config = ResamplerConfig(
    resampling_period=timedelta(seconds=10),
    resampling_function=my_resampling_func,
)

# Use EventResampler in second+ stage
resampler = EventResampler(config)
resampler.add_timeseries("my_source", source, sink)
await resampler.resample()
```

## Testing

- Unit tests cover all core functionality
- Parametrized tests verify behavior with and without `align_to`
- Key test demonstrates no data loss at boundaries
- All existing tests continue to pass

### DIFF
--- a/src/frequenz/sdk/timeseries/_resampling/_event_resampler.py
+++ b/src/frequenz/sdk/timeseries/_resampling/_event_resampler.py
@@ -1,0 +1,213 @@
+# License: MIT
+# Copyright © 2026 Frequenz Energy-as-a-Service GmbH
+
+"""Event-driven resampler for cascaded resampling stages."""
+
+import asyncio
+import logging
+from datetime import datetime, timedelta, timezone
+
+from frequenz.quantities import Quantity
+
+from .._base_types import Sample
+from ._base_types import Sink, Source
+from ._config import ResamplerConfig
+from ._resampler import Resampler, _ResamplingHelper, _StreamingHelper
+
+_logger = logging.getLogger(__name__)
+
+
+class EventResampler(Resampler):
+    """Event-driven resampler for cascaded resampling stages.
+
+    Unlike the standard Timer-based Resampler which uses fixed wall-clock
+    intervals, EventResampler is triggered by incoming data. Windows are
+    emitted when a sample arrives that falls outside the current window,
+    not on a fixed timer schedule.
+
+    Problem Solved:
+    When cascading Timer-based resamplers (e.g., 1s → 10s) with
+    align_to=UNIX_EPOCH, samples can be lost at window boundaries due to
+    timing synchronization issues. EventResampler eliminates this by
+    opening/closing windows based on actual data arrival.
+
+    Important: This resampler is optimized for continuous data streams
+    where samples arrive at regular or semi-regular intervals. It is not
+    suitable for handling raw, irregular data directly from sources.
+
+    Best Used:
+    Stage 1: Timer-based Resampler (handles raw, irregular data)
+    Stage 2+: Event-based Resampler (handles continuous data from Stage 1)
+
+    Example:
+        config = ResamplerConfig(
+            resampling_period=timedelta(seconds=10),
+            resampling_function=...,
+        )
+        resampler = EventResampler(config)
+        resampler.add_timeseries("my_source", source, sink)
+        await resampler.resample()
+
+    Note: If a long gap occurs without incoming samples (no data for multiple periods),
+    the corresponding windows will be emitted all at once when data resumes. This is
+    acceptable for cascaded resampling since the input typically comes from another
+    Resampler with guaranteed continuous output.
+    """
+
+    # pylint: disable=super-init-not-called
+    def __init__(self, config: ResamplerConfig) -> None:
+        """Initialize EventResampler.
+
+        This does not call super().__init__() to avoid starting any timers
+
+        Args:
+            config: Resampler configuration
+        """
+        self._config = config
+        """The configuration for this resampler."""
+
+        self._resamplers: dict[Source, _StreamingHelper] = {}
+        """A mapping between sources and the streaming helper handling that source."""
+
+        window_end, _ = self._calculate_window_end()
+        self._window_end: datetime = window_end
+        """The time in which the current window ends.
+
+        This is used to make sure every resampling window is generated at
+        precise times. We can't rely on the timer timestamp because timers will
+        never fire at the exact requested time, so if we don't use a precise
+        time for the end of the window, the resampling windows we produce will
+        have different sizes.
+
+        The window end will also be aligned to the `config.align_to` time, so
+        the window end is deterministic.
+        """
+
+        self._window_lock = asyncio.Lock()
+        """Lock protecting access to `_window_end` during window state transitions."""
+
+        self._sample_queue: asyncio.Queue[Sample[Quantity]] = asyncio.Queue()
+        """Queue for samples awaiting processing. Filled by `_StreamingHelper` callbacks,
+        consumed by the event loop in `resample()`.
+        """
+
+    # OVERRIDDEN: Register callback to receive samples asynchronously for
+    # event-driven window management.
+    def add_timeseries(self, name: str, source: Source, sink: Sink) -> bool:
+        """Start resampling a new timeseries.
+
+        Registers the timeseries and sets up a sample callback to enqueue
+        incoming samples for event-driven processing.
+
+        Args:
+            name: The name of the timeseries (for logging purposes).
+            source: The source of the timeseries to resample.
+            sink: The sink to use to send the resampled data.
+
+        Returns:
+            `True` if the timeseries was added, `False` if the timeseries was
+            not added because there already a timeseries using the provided
+            receiver.
+        """
+        if source in self._resamplers:
+            return False
+
+        resampler = _StreamingHelper(
+            _ResamplingHelper(name, self._config), source, sink
+        )
+
+        # Register the callback to receive samples from the streaming helper.
+        resampler.register_sample_callback(self._enqueue_sample)
+
+        self._resamplers[source] = resampler
+        return True
+
+    async def _enqueue_sample(self, sample: Sample[Quantity]) -> None:
+        """Add a sample to the processing queue.
+
+        Args:
+            sample: The sample to enqueue.
+        """
+        await self._sample_queue.put(sample)
+
+    # OVERRIDDEN: no warm-up period needed for event-driven sample accumulation.
+    def _calculate_window_end(self) -> tuple[datetime, timedelta]:
+        """Calculate the end of the first resampling window.
+
+        Calculates the next multiple of resampling_period after the current time,
+        respecting align_to configuration.
+
+        Returns:
+            A tuple (window_end, delay_time) where:
+            - window_end: datetime when the first window should end
+            - delay_time: always timedelta(0) for EventResampler
+        """
+        now = datetime.now(timezone.utc)
+        period = self._config.resampling_period
+        align_to = self._config.align_to
+
+        if align_to is None:
+            return (now + period, timedelta(0))
+
+        elapsed = (now - align_to) % period
+
+        return (
+            (now + (period - elapsed), timedelta(0))
+            if elapsed > timedelta(0)
+            else (now, timedelta(0))
+        )
+
+    async def resample(self, *, one_shot: bool = False) -> None:
+        """Start event-driven resampling.
+
+        Processes incoming samples from the queue continuously. Windows are
+        emitted when a sample arrives with a timestamp >= current window_end.
+        This is in contrast to Timer-based resampling which emits windows at
+        fixed intervals regardless of data arrival.
+
+        Args:
+            one_shot: If True, waits for the first window to be emitted, then exits.
+
+        Raises:
+            asyncio.CancelledError: If the task is cancelled.
+        """
+        try:
+            while True:
+                sample = await self._sample_queue.get()
+                emmitted = await self._process_sample(sample)
+
+                if one_shot and emmitted:
+                    return
+
+        except asyncio.CancelledError:
+            _logger.info("EventResampler task cancelled")
+            raise
+
+    async def _process_sample(
+        self,
+        sample: Sample[Quantity],
+    ) -> bool:
+        """Process an incoming sample and manage window state.
+
+        This method checks if the incoming sample falls outside the current
+        window and emits completed windows as needed. Returns True if any
+        windows were emitted.
+
+        Args:
+            sample: Incoming sample to process
+
+        Returns:
+            True if at least one window was emitted, False otherwise.
+        """
+        async with self._window_lock:
+            emmitted = False
+            while sample.timestamp >= self._window_end:
+                _logger.debug(
+                    "EventResampler: Sample at %s >= window end %s, closing window",
+                    sample.timestamp,
+                    self._window_end,
+                )
+                await self._emit_window(self._window_end)
+                self._window_end += self._config.resampling_period
+                emmitted = True
+            return emmitted
--- a/src/frequenz/sdk/timeseries/_resampling/_resampler.py
+++ b/src/frequenz/sdk/timeseries/_resampling/_resampler.py
@@ -12,7 +12,7 @@ import math
 from bisect import bisect, bisect_left
 from collections import deque
 from datetime import datetime, timedelta, timezone
-from typing import assert_never
+from typing import Awaitable, Callable, assert_never
 
 from frequenz.channels.timer import Timer, TriggerAllMissed, _to_microseconds
 from frequenz.quantities import Quantity
@@ -537,6 +537,9 @@ class _StreamingHelper:
         self._helper: _ResamplingHelper = helper
         self._source: Source = source
         self._sink: Sink = sink
+        self._sample_callback: Callable[[Sample[Quantity]], Awaitable[None]] | None = (
+            None
+        )
         self._receiving_task: asyncio.Task[None] = asyncio.create_task(
             self._receive_samples()
         )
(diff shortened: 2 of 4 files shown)


src/frequenz/sdk/timeseries/_resampling/_resampler.py
```diff
@@ -554,6 +557,22 @@ async def stop(self) -> None:
         """Cancel the receiving task."""
         await cancel_and_await(self._receiving_task)
 
+    def register_sample_callback(
```
simonvoelcker, line 560: To me, _register_ would suggest that multiple callbacks can exist. How about _set_ instead?

src/frequenz/sdk/timeseries/_resampling/_event_resampler.py
```diff
@@ -0,0 +1,213 @@
...
+    the corresponding windows will be emitted all at once when data resumes. This is
+    acceptable for cascaded resampling since the input typically comes from another
+    Resampler with guaranteed continuous output.
+    """
+
+    # pylint: disable=super-init-not-called
+    def __init__(self, config: ResamplerConfig) -> None:
+        """Initialize EventResampler.
+
+        This does not call super().__init__() to avoid starting any timers
```
simonvoelcker, line 61: `self._timer` is not used at all, right? I think it would be cleaner then to extract a common base class from the existing `Resampler` and `EventResampler`.

src/frequenz/sdk/timeseries/_resampling/_event_resampler.py
```diff
@@ -0,0 +1,213 @@
...
+        Args:
+            one_shot: If True, waits for the first window to be emitted, then exits.
+
+        Raises:
+            asyncio.CancelledError: If the task is cancelled.
+        """
+        try:
+            while True:
+                sample = await self._sample_queue.get()
+                emmitted = await self._process_sample(sample)
```
simonvoelcker, line 177: emitted

A few more occurrences below.

src/frequenz/sdk/timeseries/_resampling/_event_resampler.py
```diff
@@ -0,0 +1,213 @@
...
+        window and emits completed windows as needed. Returns True if any
+        windows were emitted.
+
+        Args:
+            sample: Incoming sample to process
+
+        Returns:
+            True if at least one window was emitted, False otherwise.
+        """
+        async with self._window_lock:
```
simonvoelcker, line 202: `self._window_end` is only written to in `__init__` and here, correct? So the lock is needed to protect concurrent calls to `_process_sample`? Just trying to understand it :)

tests/timeseries/_resampling/test_event_resampler.py
```diff
@@ -0,0 +1,298 @@
...
+    """Alignment point for windows. If None, windows are aligned to the first sample time."""
+
+    first_window_end: datetime
+    """Expected end time of the first window based on the configuration and start time."""
+
+
+@pytest.fixture
+def now() -> datetime:
+    """Fixture providing a fixed current time for testing."""
+    return datetime(2024, 1, 1, 12, 0, 5, tzinfo=timezone.utc)
```
simonvoelcker, line 35: Could it make sense to use something like [freezegun](https://pypi.org/project/freezegun/) in these tests? Or that `fake_time` fixture used elsewhere in this repo?

llucax (commented):
Unfortunately I don't think this is the way to go. To me it looks too hacky to use/inherit from the resampler to implement this by adding the callback function. If we are going for this, we'll probably need to rethink the whole resampler and probably write the new one from scratch. For example, the whole time-jump logic is not needed if we just receive samples and trigger a resampling when the fist sample with timestamp >= window_end arrives.

There are other complexities introduced that feel unnecessary, like adding locks and queues, these are not good signs.

I also asked AI to analyze the PR, and it found some important bugs. I didn't validate them yet (because I think the reasons above are enough to try to find another approach), but I will leave them here in case you want to look at them (I wouldn't, because I think we need to find another approach even if those bugs aren't real).

<details>
<summary>AI review</summary>

The biggest blocker is a **multi-timeseries correctness bug**. `EventResampler` keeps one global `_window_end` and one global queue for all sources. Every `_StreamingHelper` registers the same callback, the queue stores only `Sample` objects without source identity, and when any queued sample crosses the boundary, `_process_sample()` calls `_emit_window()` for **all** registered helpers. In practice, the first source whose `t=10` sample arrives will close the `t=10` window for every other source too, so a slightly slower source can still miss its boundary sample. That recreates the same class of boundary-loss bug this PR is trying to solve, just at the “multiple sources in one resampler” level. ([[GitHub](https://github.com/frequenz-floss/frequenz-sdk-python/pull/1372/files)][1])

A second blocker is the **lifecycle/remove bug**. `remove_timeseries()` still only deletes the dict entry and does not stop the receiving task; the PR description explicitly calls out that removed sources keep reading, buffering, and invoking callbacks, and that the queue can grow unbounded for `EventResampler`. With the current shared-queue design, that is worse than a leak: removed sources can continue to enqueue samples that advance the global window for the remaining active sources. I would not merge this without cancelling/unregistering the helper on removal. ([[GitHub](https://github.com/frequenz-floss/frequenz-sdk-python/pull/1372)][2])

There is also a **public API/docs mismatch**. The PR description shows `from frequenz.sdk.timeseries import EventResampler`, but the package `frequenz.sdk.timeseries.__init__` on the PR branch still does not export `EventResampler`; the new tests also import the private module path directly. So either the advertised import path is broken, or this class is intentionally still private and the PR text/release notes are overstating availability. Add the public export plus a public-import test, or document it as private/internal for now. ([[GitHub](https://github.com/frequenz-floss/frequenz-sdk-python/pull/1372)][2])

I also see an **edge-case semantic gap around idle/end-of-stream behavior**. The event-driven loop only emits windows from `_process_sample()` after a later sample arrives, and the class doc explicitly says missed windows are emitted “when data resumes.” I do not see a path that flushes the final pending window on source stop, shutdown, or long idle periods without a future sample. That is a behavior difference from the timer-based resampler and should be either handled explicitly or clearly documented and tested. This is an inference from the control flow, not something the PR text states directly. ([[GitHub](https://raw.githubusercontent.com/malteschaaf/frequenz-sdk-python/event_resampler/src/frequenz/sdk/timeseries/_resampling/_event_resampler.py)][3])

On **security**, I do not see injection/auth/crypto problems here. The main issue is **resource exhaustion**: `_sample_queue` is an unbounded `asyncio.Queue`, and the PR itself already notes unbounded growth after `remove_timeseries()`. A noisy, stalled, or orphaned source can therefore consume memory without limit. ([[GitHub](https://github.com/frequenz-floss/frequenz-sdk-python/pull/1372/files)][1])

For **breaking changes**, I did not find an existing public API signature change, removed/renamed public method, return-type break, DB schema change, or config-surface break in the files changed. The main API problem is with the *new* feature’s availability and import path, not with existing consumers. ([[GitHub](https://github.com/frequenz-floss/frequenz-sdk-python/pull/1372/files)][1])

**Release notes are present, but not quite right.** It is good that `RELEASE_NOTES.md` was updated, but two of the entries are private implementation details: `StreamingHelper.register_sample_callback()` and `Resampler._emit_window()`. Those do not look like user-facing release-note items. I would keep the `EventResampler` note, add caveats about intended usage/limitations, and drop or re-scope the private internals. ([[GitHub](https://github.com/frequenz-floss/frequenz-sdk-python/pull/1372/files)][1])

**Documentation looks incomplete.** The PR changes only four files: release notes, two implementation files, and tests. I do not see README/public API doc updates, and the only user-facing usage example I can find is the PR description itself. For a feature presented as a new resampler users can adopt, that is thin. ([[GitHub](https://github.com/frequenz-floss/frequenz-sdk-python/pull/1372/files)][1])

**Testing is the other major gap.** The new tests are good for boundary arithmetic, but they mostly unit-test `_process_sample()` with `_emit_window()` mocked out. They do not cover the highest-risk integration paths: multiple active sources, `add_timeseries()`/`_StreamingHelper` end-to-end behavior, `remove_timeseries()`, source stop, long idle gaps, one-shot semantics, backpressure/queue growth, error propagation, or the public import path. So the current tests miss the areas most likely to fail in production. ([[GitHub](https://raw.githubusercontent.com/malteschaaf/frequenz-sdk-python/event_resampler/tests/timeseries/_resampling/test_event_resampler.py)][4])

My must-fix list before merge would be: fix the multi-source window-closing logic, fix helper cancellation/unregistration on removal, decide whether `EventResampler` is public and wire exports/docs accordingly, and add integration tests for multi-source and lifecycle behavior.

[1]: https://github.com/frequenz-floss/frequenz-sdk-python/pull/1372/files "Implement EventResampler for Cascaded Resampling by malteschaaf · Pull Request #1372 · frequenz-floss/frequenz-sdk-python · GitHub"
[2]: https://github.com/frequenz-floss/frequenz-sdk-python/pull/1372 "Implement EventResampler for Cascaded Resampling by malteschaaf · Pull Request #1372 · frequenz-floss/frequenz-sdk-python · GitHub"
[3]: https://raw.githubusercontent.com/malteschaaf/frequenz-sdk-python/event_resampler/src/frequenz/sdk/timeseries/_resampling/_event_resampler.py "raw.githubusercontent.com"
[4]: https://raw.githubusercontent.com/malteschaaf/frequenz-sdk-python/event_resampler/tests/timeseries/_resampling/test_event_resampler.py "raw.githubusercontent.com"

</details>

tests/timeseries/_resampling/test_event_resampler.py
llucax: We use [async-solipsism](https://github.com/bmerry/async-solipsism) and [time-machine](https://github.com/adamchainz/time-machine/). There should be test in there using them.

malteschaaf:
Hey @llucax, thanks a lot for the detailed feedback! It's really helpful and you've identified some fundamental issues I should have caught earlier.

I want to clarify my initial thinking, which I admittedly didn't communicate well in the PR: My goal was to approach this in two steps to minimize disruption.

**Step 1 (current PR attempt):** Introduce `EventResampler` as a subclass without breaking changes. This allows downstream components (like `MovingWindow`) to work seamlessly with both timer-based and event-driven resamplers without modification. The idea was to keep the existing `Resampler` mostly intact.

**Step 2 (planned, not yet attempted):** Once `EventResampler` is working, refactor the architecture:
- Extract the shared logic (config, timeseries management, etc.) into a base `Resampler` class that acts as a factory
- Move timer-based logic into a subclass e.g `TimerResampler` 
- `EventResampler` stays as its own implementation
- Users can configure which one they want, but the default stays timer-based for backward compatibility

However, you're absolutely right that the current PR is too hacky and tries to squeeze everything into the existing code. The global state, shared queues, and locks are all symptoms of trying to this. I think you're absolutely right that this needs a bit more "rethinking the whole resampler".

I still think it's important that `EventResampler` is a `Resampler` type (or at least has the same interface) to enable seamless integration downstream. For example, `MovingWindow` should be able to use either implementation without code changes. Do we agree on this or am I missing some reason to have a totally independent `EventResampler`?

Thinking about it more: maybe we should start with the architecture refactoring right away instead of trying the incremental approach? The current PR's problems suggest that trying to retrofit it won't work well. 

I've been sketching a different approach: what if each source runs its own independent event loop with per-source state (window_end, buffer, etc.) instead of global shared state?

The idea is to have an abstract `_StreamingHelper` base class that both resamplers implement:
```
_StreamingHelper (abstract)
├─ _TimerStreamingHelper (passive, for timer-based)
│  ├─ Buffers samples in background task
│  ├─ Waits for Timer to call resample()
│  └─ Timer decides when to emit
│
└─ _EventStreamingHelper (active, for event-driven)
├─ Per-source state (_window_end, buffer)
├─ Runs independent event loop
├─ Decides when to emit based on sample timestamps
└─ Completely independent from other sources
```

This way:
- **TimerResampler** uses `_TimerStreamingHelper` (passive pattern, external timer drives it)
- **EventResampler** uses `_EventStreamingHelper` (active pattern, each source drives itself)
- Both inherit from `_StreamingHelper` and share `_ResamplingHelper` logic
- Each source is completely independent (no multi-source correctness bugs)
- `remove_timeseries()` just cancels that one handler (clean lifecycle)
- No need for locks or shared queues (much simpler)
- If one source has an issue, it doesn't affect others

The key insight is that both resamplers follow the same abstraction pattern, just with different execution models. This keeps the architecture clean and extensible without trying to force everything into one implementation.

Does this abstraction-first direction feel right to you? And would you still want `EventResampler` to implement the `Resampler` interface for downstream compatibility, or would you prefer a completely independent class?

Looking forward to your thoughts! 🙌

llucax:
> Step 1 (current PR attempt): Introduce EventResampler as a subclass without breaking changes. This allows downstream components (like MovingWindow) to work seamlessly with both timer-based and event-driven resamplers without modification. The idea was to keep the existing Resampler mostly intact.

I think this experimentation phase is a good idea but I'm not convinced we need to do it in the SDK. If you only want to experiment with this using the moving window, I think a better approach, that will also let you have a much faster iteration over the solution, would be to just copy & paste the Resampler to a project where you want to use this, then apply the changes you made here (although I think they still need some changes, and after it is working as you'd like, we think about how to move it to the SDK.

Do you think this would be a viable option?

llucax:
> For example, MovingWindow should be able to use either implementation without code changes. Do we agree on this or am I missing some reason to have a totally independent EventResampler?

Without having thought about it, I'm not convinced we **should** *be able to use either implementation without code changes*. All the downstream classes were designed having only *real-time resampling* in mind, not batched resampling. So I wouldn't be surprised if we need to make code changes. Keeping the `MovingWindow` public interface the same might be a nice goal, but I would also not discard improving its interface to better fit both models.

What we probably need is the `MovingWindow` being able to start with batched/historical data and switch to real-time in the same object/interface. I guess this is really the baseline and only *hard* requirement, right?

llucax:
> maybe we should start with the architecture refactoring right away instead of trying the incremental approach?

No, I like the approach of doing small incremental changes, I just think the SDK is not the right place for that, as we need to commit to public interfaces.

PR #1374, opened 2026-03-13
Coalesce function unsubscribes only after receiving 3 samples

### LINKED ISSUE #1332: Coalesce nodes should wait for 3 subsequent values before unsubscribing from previous node
### What happened?

On slow component streams, the resampler takes time to stabilize on a value.  During this time, there will be intermittent values in original streams.  This causes the coalesce nodes to cycle between subscribing and unsubscribing from next components.

Coalesce nodes should unsubscribe from secondary sources only when the primary has stabilized with at least 3 subsequent good values.

### Affected version(s)

_No response_

### Affected part(s)

### Extra information

_No response_

simonvoelcker:
Addresses https://github.com/frequenz-floss/frequenz-sdk-python/issues/1332.

# Background

The Coalesce function subscribes and unsubscribes from components based on data availability. If it receives None-values, it will subscribe to the next component, if any, from its parameter list. Once the other components start sending data again, we may unsubscribe from "later" components. We currently unsubscribe from all components "after" a given one once that given one sends us one non-None sample. This is undesired because we may receive intermittent None values for various reasons and would then unsubscribe and resubscribe excessively.

# Change

The solution is to count non-None samples from earlier streams and unsubscribe only once a threshold (3) is reached.

# Discussion

If there are only two components (main, backup), it is quite straightforward. With more components, although probably rarely seen in practice, one could come up with a more sophisticated solution that tracks each components samples and marks them as stable when 3 samples are reached. Happy to discuss, but I think it's overkill.


tests/timeseries/_formulas/test_formula_validation.py
```diff
@@ -0,0 +1,186 @@
+# License: MIT
+# Copyright © 2025 Frequenz Energy-as-a-Service GmbH
```
simonvoelcker: 2026

src/frequenz/sdk/timeseries/formulas/_functions.py
```diff
@@ -156,6 +164,17 @@ async def subscribe(self) -> None:
...
 
+    async def _count_sample(self) -> None:
+        """Count samples from parameters other than the last.
+
+        If a reasonable number of samples is reached, unsubscribe
+        from the last parameter, whose samples were not counted.
+        """
+        self.num_samples += 1
+
+        if self.num_samples >= 3:
```
llucax: I would make this number a constant, like `MINIMUM_STABLE_SAMPLES` (there is probably a better name).

src/frequenz/sdk/timeseries/formulas/_functions.py
```diff
@@ -156,6 +164,17 @@ async def subscribe(self) -> None:
         if self.num_subscribed == 0:
             await self._subscribe_next_param()
 
+    async def _count_sample(self) -> None:
```
llucax: I find it surprising that a "count" function can actually trigger unsubscribing from somewhere. Since this is used only once, I think I would just move it to where it is called, or change the name of the function, but I guess it is hard to find a name that is not `_count_sample_and_unsubscribe_if_we_have_more_than_three()` 😆

This looks reasonable to me:
```py
                case Sample(timestamp, value):
                    if value is not None:
                        # Found a non-None value, unsubscribe from subsequent params
                        # Found a non-None value
                        if ctr < self.num_subscribed:
                            await self._unsubscribe_all_params_after(ctr)
                            self.num_samples += 1
                            if self.num_samples >= MINIMUM_STABLE_SAMPLES:
                                await self._unsubscribe_last_param()
                        return arg
                    ts = timestamp
```

tests/timeseries/_formulas/test_formula_validation.py
```diff
@@ -0,0 +1,186 @@
...
+
+import pytest
+from frequenz.quantities import Quantity
+
+from frequenz.sdk.timeseries.formulas._exceptions import FormulaSyntaxError
+from frequenz.sdk.timeseries.formulas._parser import parse
+
+
+class TestFormulaValidation:
+    """Tests for Formula validation."""
```
llucax: This is a case where I would drop the class, as it is the only class in the file, it will only make the test name much longer and redundant: `tests/timeseries/_formulas/test_formula_validation.py::TestFormulaValidation::test_parser_validation`.

tests/timeseries/_formulas/test_general.py
```diff
@@ -907,3 +734,129 @@ async def test_compound(self) -> None:
                 ([15.0, 17.0, None, 5.0], None),
             ],
         )
+
+
+class TestCoalesceFunction:
```
llucax, line 739: Since you split out some tests into their own file, I wonder what's the relationship between formulas and functions. Wouldn't it make also sense to separate test for coalesce function to its own file, or make these tests part of the existing `TestFormulas`? Sorry if the question is silly, I'm not familiar with this code (so I didn't follow all the test logic) to assess it.

tests/timeseries/_formulas/test_general.py
```diff
@@ -907,3 +734,129 @@ async def test_compound(self) -> None:
...
+    """Test coalesce function subscribe/unsubscribe behavior."""
+
+    class CoalesceSample(NamedTuple):
+        """Helper class to represent expected behavior of coalesce function."""
+
+        values: list[float | None]
+        expected_subscriptions: list[bool]
+
+    async def run_test(  # pylint: disable=too-many-locals
```
llucax, line 748: I guess you tried to keep the current file style, so I'm OK if you prefer to keep it as it is, but I'd recommend using `@pytest.mark.parametrize()` for new tests.

<details>
<summary>Example conversion done by AI (with some extra test case wrapping)</summary>

```py
class CoalesceSample(NamedTuple):
    """Helper class to represent expected behavior of coalesce function."""

    values: list[float | None]
    expected_subscriptions: list[bool]


class CoalesceFunctionTestCase(NamedTuple):
    """A full coalesce function test case."""

    name: str
    formula_str: str
    samples: list[CoalesceSample]


class TestCoalesceFunction:
    """Test coalesce function subscribe/unsubscribe behavior."""

    @pytest.mark.parametrize(
        "test_case",
        [
            CoalesceFunctionTestCase(
                name="subscribe_when_none_encountered",
                formula_str="COALESCE(#0, #1, #2, 0.0)",
                samples=[
                    CoalesceSample(
                        values=[10.0, None, None],
                        expected_subscriptions=[True, False, False],
                    ),
                    # No need to subscribe unless stream #1 gives None
                    CoalesceSample(
                        values=[10.0, 12.0, 15.0],
                        expected_subscriptions=[True, False, False],
                    ),
                    # If None is encountered, one subscription is added per sample
                    CoalesceSample(
                        values=[None, None, 15.0],
                        expected_subscriptions=[True, True, False],
                    ),
                    CoalesceSample(
                        values=[None, None, 15.0],
                        expected_subscriptions=[True, True, True],
                    ),
                ],
            ),
            CoalesceFunctionTestCase(
                name="unsubscribe_only_after_3_samples",
                formula_str="COALESCE(#0, #1, #2, 0.0)",
                samples=[
                    # First subscription is added before the first sample.
                    # Every sample can add one subscription.
                    CoalesceSample(
                        values=[None, None, 15.0],
                        expected_subscriptions=[True, True, False],
                    ),
                    CoalesceSample(
                        values=[None, None, 15.0],
                        expected_subscriptions=[True, True, True],
                    ),
                    # After 3 samples, the last subscription is dropped.
                    CoalesceSample(
                        values=[None, 12.0, 15.0],
                        expected_subscriptions=[True, True, True],
                    ),
                    CoalesceSample(
                        values=[10.0, None, 15.0],
                        expected_subscriptions=[True, True, True],
                    ),
                    CoalesceSample(
                        values=[None, 12.0, 15.0],
                        expected_subscriptions=[True, True, False],
                    ),
                ],
            ),
        ],
        ids=lambda tc: tc.name,
    )
    async def test_coalesce_subscriptions(
        self,
        test_case: CoalesceFunctionTestCase,
    ) -> None:
        """Test coalesce subscribe/unsubscribe behavior."""
        # Component IDs are 0, 1, 2 for convenience.
        channels: list[Broadcast[Sample[Quantity]]] = [
            Broadcast(name=str(num)) for num in range(3)
        ]
        senders = [channel.new_sender() for channel in channels]
        receivers: list[Receiver[Sample[Quantity]] | None] = [None, None, None]

        def new_receiver(component_id: ComponentId) -> Receiver[Sample[Quantity]]:
            """Create a new receiver, overwriting any existing one.

            When Coalesce unsubscribes, it closes its receiver.
            """
            comp_id = int(component_id)
            receiver = channels[comp_id].new_receiver()
            receivers[comp_id] = receiver
            return receiver

        telem_fetcher = MagicMock(spec=ResampledStreamFetcher)
        telem_fetcher.fetch_stream = AsyncMock(side_effect=new_receiver)
        formula = parse(
            name="f2",
            formula=test_case.formula_str,
            create_method=Quantity,
            telemetry_fetcher=telem_fetcher,
        )

        result_chan = formula.new_receiver()
        await asyncio.sleep(0.1)
        now = datetime.now()

        async def send_sample(values: list[float | None]) -> None:
            nonlocal now
            now += timedelta(seconds=1)
            _ = await asyncio.gather(
                *[
                    senders[comp_id].send(
                        Sample(
                            now,
                            None if value is None else Quantity(value),
                        )
                    )
                    for comp_id, value in enumerate(values)
                ]
            )
            _ = await result_chan.receive()

        for index, sample in enumerate(test_case.samples):
            await send_sample(sample.values)
            active_subscriptions = [
                receiver is not None and not getattr(receiver, "_closed", True)
                for receiver in receivers
            ]
            assert active_subscriptions == sample.expected_subscriptions, (
                f"{test_case.name}, step {index}: values={sample.values}"
            )

        await formula.stop()
```

</details>

RELEASE_NOTES.md
```diff
@@ -15,5 +15,5 @@
 ## Bug Fixes
 
 - Improved formula validation: Consistent error messages for invalid formulas and conventional span semantics.
-
 - This fixes a rare power distributor bug where some battery inverters becoming unreachable because of network outages would lead to excess power values getting set.  This is fixed by measuring the power of the unreachable inverters through their fallback meters and excluding that power from what is distributed to the other inverters.
+- Better data stream subscription handling in Coalesce function: Only unsubscribe from stream after other streams have received 3 samples.
```
llucax: I would rephrase it to emphasize a bit more around it making fallback behave less erratically during unstable periods or something like that. Remember to keep it focused on the person that's using the SDK and want to know what effect will have in their program.

llucax (commented):
LGTM in general, all comments are minor enough to be optional for me, but I would still leave the final approval to @shsms as I'm not very familiar with formulas code.

src/frequenz/sdk/timeseries/formulas/_functions.py
simonvoelcker: I moved this bit out because a linter error informed me that there were too many branches (13, one more than allowed). Do you think it is better so silence it?

llucax: Yeah, or look for a deeper refactor. Maybe moving all code for each `case` to its own function. But I would only do that if you feel the method got too complex. We can disagree with the linter from time to time...

src/frequenz/sdk/timeseries/formulas/_functions.py
simonvoelcker: I did not extract because this is the kind of very specific thing for which there is no obvious name, and adding one for the sake of it often makes it more confusing. If you have a good name and think that it is helpful to use that, please share.

tests/timeseries/_formulas/test_general.py
simonvoelcker: I split out the validation tests because (a) the file went over the 1000 line limit and (b) the validation tests were the most obvious candidate to move to their own file. What is left is indeed a bit of a mix of tests for formula _evaluation_ (includes functions used in formulas) and tests for the _subscribe/unsubscribe behavior of the coalesce function specifically_. If would probably make sense to have `test_formula_evaluation.py` for the former, but then, what is the other one? `test_formulas.py` is the lowest common abstraction, IMO.

tests/timeseries/_formulas/test_general.py
simonvoelcker: I'm a big fan of parametrize, but I would not use it in this case, personally. It is possible only because the setup happens to be identical for subscribe and unsubscribe, but if that ever changes we're better off with two test cases. I also find it difficult to debug non-trivial parametrized tests (which this one would be).

RELEASE_NOTES.md
simonvoelcker: I've only learned today that it is possible to omit a release note entry for changes that are neither features nor bug fixes - Now I feel like this change qualifies as well. I'll remove it.

src/frequenz/sdk/timeseries/formulas/_functions.py
simonvoelcker: I resolved this by removing a piece of dead code in this function that I had spotted earlier but wanted to handle in a separate PR. It brings the number of statements down, so I could inline the sample counting as you suggested.

src/frequenz/sdk/timeseries/formulas/_functions.py
llucax: I think it is worth extracting even if the name is not great to give it more visibility, as magic numbers buried deep in the code might be easy to miss otherwise. I would ask AI for a name to see if they can do better than us...

tests/timeseries/_formulas/test_general.py
llucax: I think we can keep the name for general tests without a better grouping. For example

I also just noticed these tests are under a `_formulas` directory, so we can also either keep `test_formula.py` or use something like `test_general.py`, and I would remove the prefix from other files, like `_formulas/test_evaluation.py`, `_formulas/test_validation.py`, again, to keep the resulting test names as short as possible (and not redundant).

tests/timeseries/_formulas/test_general.py
llucax: But you are sharing the code now, si if it changes you'll need to rewrite the `run_test` too. It is just a change in form.

But as I said, I'm good with keeping it as it is for the other reasons too.

What makes me curious now is what did you find hard to debug when using parametrization, as I never felt parametrization interfered with debugging.

simonvoelcker: I briefly forgot that it is possible to run the individual test cases in a parametrized test - without that ability, breakpoints trigger for each case before the one of interest is even reached. So yes, it is not difficult per se, but it adds a layer of indirection.

shsms:
Sorry, I'll take a look in the morning.

tests/timeseries/_formulas/test_general.py
```diff
@@ -907,3 +909,129 @@ async def test_compound(self) -> None:
...
+                    expected_subscriptions=[True, True, True],
+                ),
+                self.CoalesceSample(
+                    values=[10.0, None, 15.0],
+                    expected_subscriptions=[True, True, True],
+                ),
+                self.CoalesceSample(
+                    values=[None, 12.0, 15.0],
+                    expected_subscriptions=[True, True, False],
+                ),
```
shsms: This might not be enough.  Could you please also track the earliest_non_nil_param, for which the num_samples is valid?  If earliest non-nil param changes, then you reset the num_samples.

Then we'll know that we have at least one stable stream, before we unsubscribe.   Then you don't have to remove `_unsubscribe_all_params_after` either.

shsms: Could you also add a test that cycles between sub and unsub a couple of times, with some stable period, at least some with more than 3 stable samples, before they change again?  I think this is complicated logic, and now that we depend on 3 samples for each state change, the tests need a lot more samples.

simonvoelcker: My thinking here was that we can unsubscribe from the last parameter when the other parameters are "stable together", i.e. we got 3 samples from them and therefore didn't need the last one. And I wanted to keep complexity down.

simonvoelcker: @shsms the new logic leads to potentially surprising behavior, described in [this test case](https://github.com/frequenz-floss/frequenz-sdk-python/pull/1374/changes#diff-51f90d1a37caaba8008a6ca3f10140a623067cdaefbcb50d52c7bf570f59f741R903-R904).

But it is implemented and looks nice otherwise. Also not too complex.

simonvoelcker: Added more test cases.

shsms: Damn, that doesn't look nice, because it could prevent us from switching to the second component for ever.  But I feel like there's an easy fix.  Let me do some quick tests.

shsms (approved):
🎉

PR #1376, opened 2026-03-16
Switching from our own Peekable implementation to more-itertools.peekable.

### LINKED ISSUE #1325: Clarify that the `Peekable` takes ownership of the iterator
Maybe it is worth clarifying that the `Peekable` takes ownership of the iterator, otherwise if someone iterates it externally too it might *lose* data (removed from the iterator and stored in the internal buffer).

_Originally posted by @llucax in https://github.com/frequenz-floss/frequenz-sdk-python/pull/1295#discussion_r2606359570_

### LINKED ISSUE #1326: Make internal `Peekable` more robust or switch to `more-itertools`
The internal peekable implementation doesn't support having `None` as a valid value (for example `Peekable[int | None]` will not work). We could fix it, but there is also an implementation in `more-itertools` which seems popular and well maintained, so we could just switch to using that one.

_Originally posted by @llucax in https://github.com/frequenz-floss/frequenz-sdk-python/pull/1295#discussion_r2606364249_

simonvoelcker:
Addresses https://github.com/frequenz-floss/frequenz-sdk-python/issues/1325 and https://github.com/frequenz-floss/frequenz-sdk-python/issues/1326.

Switching to more-itertools.peekable as suggested in the [original thread](https://github.com/frequenz-floss/frequenz-sdk-python/pull/1295#discussion_r2606364249).


RELEASE_NOTES.md
```diff
@@ -17,3 +17,5 @@
 - Improved formula validation: Consistent error messages for invalid formulas and conventional span semantics.
 
 - This fixes a rare power distributor bug where some battery inverters becoming unreachable because of network outages would lead to excess power values getting set.  This is fixed by measuring the power of the unreachable inverters through their fallback meters and excluding that power from what is distributed to the other inverters.
+
+- Using more-itertools.peekable instead of our own implementation.
```
llucax: The release notes are intended to document changers that affect downstream users. I don't think this change should appear in the release notes, as it is only internal. Although actually the new dependency might affect downstream users, as if they already use `more-itertools`, but a version < 10.8.0, they will need to upgrade (it is technically a *breaking change*, but I guess it is low risk enough to pretend it is not for the next release).

llucax (commented):
Just one comment about the release notes, the rest LGTM.

llucax (approved):

Issue #949, opened 2024-05-15
BatteryPool should be more explicit about using all components or a given set of component ids


Marenz:
### What's needed?

Working a bit with the scenario of possibly changing component ids in the FCR actor, I found it cumbersome that 
the empty set had the special meaning of "use all components".

### Proposed solution

I propose we:
* Add an explicit parameter `use_all_components: bool`
* Also add a new getter `using_all_components: bool` to explicitly know if it is using all components or a a specific set

llucax:
So calling without either some `component_ids` or `use_all_components` will be an error? I don't like that it makes the interface more complicated and that you can end up with some exception if you call it with the wrong combination of arguments.

Also, why it is important to know if it was created with all available components or not? It would be good to explain the user case a little bit.

Marenz:
It's fundamentally different if you use the pool with "use everything you can, even new things" and "use this fixed set of ids" and right now it's non-trivial to find out which of those two cases it is, once the pool is created. 

and because it is so fundamentally different, I feel it should be setup differently, too.

> So calling without either some component_ids or use_all_components will be an error?

No I think it' still fine to have `use_all` as default, it just shouldn't be `use_all` if you pass an empty set or `None`, imo, that mixes the two different cases too much.

> Also, why it is important to know if it was created with all available components or not? It would be good to explain the user case a little bit.

Probably becoming less relevant once the SDK does pool caching or so, but it would help to then decide if you want to recreate the pool because you need a different set of ids (or all)

llucax:
OK, I see your point now, but I if the issue is the interface being confusing and error prone, then I think your suggestion introduces other set of issues, as we have an overlapping set of options.

What about creating a sentinel value `USE_ALL` and make the signature: `new_pool(batteries: USE_ALL | Set[int] = USE_ALL)`? Then we can make passing an empty set an error and I think the interface should be pretty clear and explicit.

Then we can expose the requested `batteries` as a property (which might be different to the "currently used batteries", we should expose this as a separate property), which will have the info if you requested all or a fixed set.

llucax:
@jh2007github this might be a good issue for you to pick up.

shsms:
> It's fundamentally different if you use the pool with "use everything you can, even new things" and "use this fixed set of ids" and right now it's non-trivial to find out which of those two cases it is, once the pool is created.

[...]

PR #1391, opened 2026-04-22
Extract abstract base classes from pools and pool ref stores

simonvoelcker:
This is preparation for the introduction of a new subclass for steam boiler pools.

There was a lot of duplicated code between the three existing types of pools and their respective references stores, with just a bit of drift.


src/frequenz/sdk/timeseries/component_pool/_component_pool.py
```diff
@@ -0,0 +1,157 @@
...
+        """
+        self._pool_ref_store = pool_ref_store
+        unique_id = str(uuid.uuid4())
+        self._source_id = unique_id if name is None else f"{name}-{unique_id}"
+        self._priority = priority
+
+    @property
+    def component_ids(self) -> abc.Set[ComponentId]:
+        """Return component IDs of all PV inverters managed by this PVPool.
+
```
Marenz, line 54: Docstring copy-paste from `PVPool` — this is the abstract base, so it should not mention PV inverters. Something like "Return the set of component IDs managed by this pool." would fit.

src/frequenz/sdk/timeseries/abstract_pool/_abstract_pool.py
```diff
@@ -0,0 +1,157 @@
...
+        return channel
+
+    @property
+    def power_distribution_results(self) -> ReceiverFetcher[_power_distributing.Result]:
+        """Get a receiver to receive power distribution results.
+
+        Returns:
+            A receiver that will stream power distribution results for the pool's set of
+            PV inverters.
+        """
```
Marenz: Copy-paste from `PVPool` again — "PV inverters" should be "components in the pool" (or similar) in the abstract base.

src/frequenz/sdk/timeseries/component_pool/_component_pool_reference_store.py
```diff
@@ -0,0 +1,126 @@
...
+
+        self.namespace = self.get_namespace()
+        self.formula_pool = FormulaPool(
+            self.namespace,
+            self.channel_registry,
+            self.resampler_subscription_sender,
+        )
+        self.bounds_channel: Broadcast[SystemBounds] = Broadcast(
+            name=f"System Bounds for {self.get_component_type_name_plural()}: {component_ids}",
+            resend_latest=True,
```
Marenz, line 99: Two issues with this broadcast name:

1. `component_ids` here refers to the **constructor argument**, which is `None` when the caller lets the store auto-discover IDs. The name then becomes e.g. `"System Bounds for EV chargers: None"`. Should be `self.component_ids` (the resolved frozenset).
2. `get_component_type_name_plural()` returns `"EV chargers"` / `"PV inverters"` (lowercase), whereas the previous per-class strings used Title Case (`"EV Chargers"`, `"PV Inverters"`). Minor user-visible regression in log/metric names. Worth picking one convention and sticking to it.

src/frequenz/sdk/timeseries/ev_charger_pool/_ev_charger_pool_reference_store.py
```diff
@@ -114,5 +59,6 @@ def __init__(  # pylint: disable=too-many-arguments
     async def stop(self) -> None:
         """Stop all tasks and channels owned by the EVChargerPool."""
         await self.formula_pool.stop()
-        await self.bounds_tracker.stop()
+        if self.bounds_tracker is not None:
```
Marenz: The EV and PV `stop()` methods are byte-for-byte identical (modulo a stale docstring in PV that still says "PVInverterPool"). Natural candidate to hoist into `AbstractPoolReferenceStore.stop()`; Battery would then `await super().stop()` and add its extras (`_active_methods`, `_update_battery_status_task`). Would also fix the PV docstring typo for free.

src/frequenz/sdk/timeseries/component_pool/_component_pool.py
```diff
@@ -0,0 +1,157 @@
...
+                creation_time=asyncio.get_running_loop().time(),
+            )
+        )
+
+    @property
+    @abstractmethod
+    def power(self) -> Formula[Power]:
+        """Fetch the total power for the components in the pool.
+
+        Returns:
```
Marenz, line 98: This widens the public return type. Previously `BatteryPool.power_status` returned `ReceiverFetcher[BatteryPoolReport]`, and similarly for PV/EV — those are `typing.Protocol`s (not aliases for `_Report`); `BatteryPoolReport` in particular documents extra methods like `adjust_to_bounds(...)`. With the inherited `ReceiverFetcher[_power_managing._Report]`, downstream code typed as `BatteryPoolReport` loses those methods at the type level (mypy will reject `.adjust_to_bounds(...)`) and `_Report` — a private name — is now exposed in the public API.

Suggest making `AbstractPool` generic over the report protocol:

```python
TReport = TypeVar("TReport", bound=_power_managing._Report)

class AbstractPool(ABC, Generic[TReport]):
    @property
    def power_status(self) -> ReceiverFetcher[TReport]:
        ...

class BatteryPool(AbstractPool[BatteryPoolReport]): ...
class PVPool(AbstractPool[PVPoolReport]): ...
class EVChargerPool(AbstractPool[EVChargerPoolReport]): ...
```

This preserves each pool's public API exactly, keeps `_Report` out of the signatures, and sets up a clean pattern for `SteamBoilerPool`. (Same generics trick would also let us drop the `isinstance(self._pool_ref_store, ...)` asserts in `BatteryPool` by parametrising over the ref-store type too — worth considering in the same pass.)

Marenz: Wrong line, this should be  at power status line 101

src/frequenz/sdk/timeseries/component_pool/_component_pool.py
simonvoelcker: Good catch. I though I had done a full text search, but apparently I did it only in the other new abstract class.

src/frequenz/sdk/timeseries/component_pool/_component_pool_reference_store.py
simonvoelcker: 1. Noted, great catch.
2. I'll go with the lowercase versions, then. In a way, we are gaining consistency by using the same capitalization here as in the ValueError above.

src/frequenz/sdk/timeseries/component_pool/_component_pool.py
simonvoelcker: A very good idea and I've applied the proposed changes almost verbatim. It got a bit more complicated because the report subclasses actually inherit from Protocol, so I extracted another base class there. I hope I didn't go too far astray there.

simonvoelcker:
@Marenz could you have another look please? :upside_down_face:

llucax (commented):
Draft review with inline findings.

llucax (commented):
Draft review for inline findings.

llucax (commented):
Draft review for inline findings.

src/frequenz/sdk/timeseries/pv_pool/_pv_pool.py
```diff
@@ -25,41 +19,19 @@ class PVPoolError(Exception):
...
-        self._pool_ref_store = pool_ref_store
-        unique_id = uuid.uuid4()
-        self._source_id = str(unique_id) if name is None else f"{name}-{unique_id}"
-        self._priority = priority
-
+    @override
     async def propose_power(
         self,
         power: Power | None,
+        *,
```
llucax: Sadly this is a breaking change.

I see 2 paths forward:

1. Create a new v1.0.0rc2300 branch to accumulate breaking changes and merge there
2. Not sure if we can somehow allow passing bounds as positional arguments only for this class (but it is required when called via the abstract base class), then we can add runtime deprecation if the args are being passed as positional (but for this I guess we need to change the signature to use `*args, **kwargs`)
3. Check if any of our projects is using bounds as positional arguments, and if they are not just ship the breaking change as part of a "point release" (but we might have external users using this)
4. Move fast and break things: just make v1.0.0rc2300  the next release and deal with the complicated upgrade of all dependencies. Maybe is not that bad, as we can probably declare the dependency for downstream projects just as `frequenz-sdk >= v1.0.0rc2200, < v1.0.0rc2400` if when updating the dependency we make sure we call `propose_power()` with `bounds as kw arguments, that will work with both rc22xx and rc23xx releases.

1 feels like postponing the unavoidable. 2 feels extremely hacky and high effort to be worth it, 3 seems risky, and after writing all of that, I guess 4 is not that bad (last famous words), so I'm inclined to 4.

src/frequenz/sdk/timeseries/battery_pool/_battery_pool_reference_store.py
```diff
@@ -81,81 +81,63 @@ def __init__(  # pylint: disable=too-many-arguments
...
-            self._batteries = all_batteries
+        super().__init__(
+            channel_registry=channel_registry,
+            resampler_subscription_sender=resampler_subscription_sender,
+            status_receiver=batteries_status_receiver,
+            power_manager_requests_sender=power_manager_requests_sender,
+            power_manager_bounds_subs_sender=power_manager_bounds_subscription_sender,
+            power_distribution_results_fetcher=power_distribution_results_fetcher,
+            component_ids=batteries_id,
+        )
```
llucax, line 92: We have another subtle breaking change here. But this one should be easier to fix.

Before, passing `[]` was interpreted as *please I want all batteries*, but now, because `AbstractPoolReferenceStore` uses `if component_ids is not None` instead of `if component_ids`, passing `[]` will create an empty pool instead.

We can easily fix this by checking `if batteries_id is not None and batteries` (this sounds super weird BTW, it should be `battery_ids` 😆 ) and emit a deprecation warning if that's the case, passing `None` to `super.__init__()` instead of the empty list.

We could also deprecate `batteries_id` and add `component_ids`, as I see it is already used by other pools. We can easily do it by accepting both and emitting a deprecation warning if `batteries_id` was passed. We can use a sentinel like:

```py

class _NotSet: pass

class BatteryPoolReferenceStore(...):
        self,
        *,
        resampler_subscription_sender: Sender[ComponentMetricRequest],
        batteries_status_receiver: Receiver[ComponentPoolStatus],
        power_manager_requests_sender: Sender[Proposal],
        power_manager_bounds_subscription_sender: Sender[ReportRequest],
        power_distribution_results_fetcher: ReceiverFetcher[Result],
        min_update_interval: timedelta,
        batteries_id: Set[ComponentId] | None | _NotSet = _NotSet(),
        component_ids: Set[ComponentId] | None = None,
    ) -> None:
        """ DOCS SHOULD SAY IT IS DEPRECATED TOO """
        if isinstance(batteries_id, _NotSet) and component_ids is not None:
            raise ValueError("use only 'component_ids', 'batteries_id' is deprecated")
        if not isinstance(batteries_id, _NotSet):
            warnings.warn(
                "'batteries_id' is deprecated, use 'component_ids' instead.",
                DeprecationWarning,
                stacklevel=2,
            )
        ...
```

Maybe we can even use `None` instead of a sentinel, the sentinel is just more reliable for cases where maybe `None` is used explicitly (or indirectly, like setting to `None` to only assign it in a `if`).

src/frequenz/sdk/timeseries/component_pool/_component_pool_reference_store.py
```diff
@@ -0,0 +1,133 @@
...
+            if not self.component_ids.issubset(all_components):
+                unknown_ids = self.component_ids - all_components
+                raise ValueError(
+                    f"Unable to create {self.get_pool_type_name()}. These component IDs "
+                    + f"are either not {self.get_component_type_name_plural()} or are unknown: "
+                    + f"{unknown_ids}"
+                )
+        else:
+            self.component_ids = all_components
+
```
llucax, line 88: What about an empty set (in both `if` cases), shouldn't that be validated too? It seems it wasn't validated before, so maybe we have a case for it, or maybe it is validated at a deeper level?

src/frequenz/sdk/timeseries/component_pool/_component_pool_report.py
llucax, line 1: Still not *abstract enough*, many mentions to `PV` in the docstrings 😄

src/frequenz/sdk/timeseries/abstract_pool/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+"""Manage a pool of components."""
+
+from ._abstract_pool import AbstractPool
+from ._abstract_pool_reference_store import AbstractPoolReferenceStore
+from ._abstract_pool_report import AbstractPoolReport
+
+__all__ = [
+    "AbstractPool",
+    "AbstractPoolReferenceStore",
+    "AbstractPoolReport",
```
llucax: I can see the use of having the `AbstractPool` public, but do we need to expose these publicly? It feels like it is too internal to expose, I would try to keep the public surface as small as possible.

src/frequenz/sdk/timeseries/component_pool/_component_pool.py
```diff
@@ -0,0 +1,161 @@
...
+            self._pool_ref_store.power_distribution_results_fetcher,
+            lambda recv: recv.filter(
+                lambda x: x.request.component_ids == self._pool_ref_store.component_ids
+            ),
+        )
+
+    @property
+    def _system_power_bounds(self) -> ReceiverFetcher[SystemBounds]:
+        """Return a receiver fetcher for the system power bounds."""
+        return self._pool_ref_store.bounds_channel
```
llucax, line 151: Is this private property really necessary? Why can't sub-classes just access `self._pool_ref_store.bounds_channel`, is it just for convenience or was to avoid updating code that was using a different name?

src/frequenz/sdk/timeseries/battery_pool/_battery_pool.py
```diff
@@ -1,5 +1,5 @@
 # License: MIT
-# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+# Copyright © 2026 Frequenz Energy-as-a-Service GmbH
```
llucax: The copyright year should be the oldest one, or a range (2023-2026) but I would keep the oldest for simplicity. In practice it is probably irrelevant as the years the file was updated is in git history.

Same for the other copyright year updates.

src/frequenz/sdk/timeseries/battery_pool/messages.py
```diff
@@ -16,11 +16,12 @@
     Success,
 )
 from .._base_types import Bounds
+from ..abstract_pool import AbstractPoolReport
 
 
 # This class is used to expose the generic reports from the PowerManager with specific
 # documentation for the battery pool.
-class BatteryPoolReport(typing.Protocol):
+class BatteryPoolReport(AbstractPoolReport, typing.Protocol):
```
llucax: I'm not sure we need the `Protocol` now if this is a an abstract class, right? Not sure though, I remember there was some controversy around this usage of `Protocol`. /cc @shsms just in case.

llucax (commented):
Pretty nice! 🎉 

The main concern is the breaking changes. If we decide to add any, we would need to update the release notes to mention it too. Maybe adding tests for the parts that didn't fail even if we introduced breaking changes would be a good idea too.

I'm not completely against a "Address PR comments" commit but since amending the previous commit should be trivial, I would go with that. Also "Make AbstractPool generic over ref store and report types. Extract abstract base class for reports." feels like maybe the second part should be the commit body instead of part of a (long) title?

Finally I wonder if something like `ComponentPool` could make more sense than `AbstractPool`, thinking about maybe at some point we want to use them more generically, maybe some lib can accept any pool, not really caring about which kind of components are in it, and just wants to propose power, maybe `ComponentPool` reads more easily than `AbstractPool`. Or if we really want to highlight the abtractness, maybe use `AbstractComponentPool`?

src/frequenz/sdk/timeseries/pv_pool/_pv_pool.py
simonvoelcker: 5. Allow bounds to be passed as positional argument in all pool classes (remove the *) and create a ticket?

src/frequenz/sdk/timeseries/component_pool/_component_pool_report.py
simonvoelcker: Whoops, I added this while addressing marenz' comments, forgot to run another search :sweat_smile: 

Good catch

src/frequenz/sdk/timeseries/abstract_pool/__init__.py
simonvoelcker: This may be a dumb question, but how would one achieve that without importing private modules where we use these?

src/frequenz/sdk/timeseries/battery_pool/_battery_pool.py
simonvoelcker: Interesting, in my previous jobs this was commonly updated to the current year. I'll just stop editing these henceforth.

src/frequenz/sdk/timeseries/component_pool/_component_pool_reference_store.py
simonvoelcker: It wasn't validated and is also not covered by any test case as far as I can see.

src/frequenz/sdk/timeseries/component_pool/_component_pool.py
simonvoelcker: I think the idea is to provide a consistent interface across all pool classes (battery pool implements it differently). I do wonder whether it could be made public though, because it is used in a few places other than subclasses already.

src/frequenz/sdk/timeseries/battery_pool/_battery_pool_reference_store.py
simonvoelcker: Good catch, the `batteries_id` parameter was bugging me as well. Therefore I went with the idea of deprecating it, mostly the way you suggested.

src/frequenz/sdk/timeseries/pv_pool/_pv_pool.py
llucax: ~Technically yes, but we don't want that, we made them kw-only on purpose. We try to use kw-only arguments when they are built-in types and it would not be very clear what's the meaning of an argument on the call side, as we target a not very experience audience, we want to encourage readable code.~

OK, I thought this was about the ctor. I'm not sure why we used kw-only arguments here, as bounds is a class and should be pretty clear on the call site if we pass a Bounds() object that is a bounds argument, so I guess that could be an option too. :+1:

src/frequenz/sdk/timeseries/component_pool/_component_pool_reference_store.py
llucax: I guess it is a good idea to validate and test, but maybe out of scope for this PR.

There was some issue talking about how to improve this, but also probably out of scope for this PR:

* https://github.com/frequenz-floss/frequenz-sdk-python/issues/949

src/frequenz/sdk/timeseries/abstract_pool/__init__.py
llucax: Not sure I understand the question, but for extra clarification, I'm talking about these 2: `AbstractPoolReferenceStore`, `AbstractPoolReport`, assuming they are only used internally (as we allow importing from private modules in the same codebase, we are doing this everywhere).

src/frequenz/sdk/timeseries/component_pool/_component_pool.py
llucax: Then I agree it should be public. @shsms?

src/frequenz/sdk/timeseries/battery_pool/_battery_pool.py
llucax: Yeah, I spent some time researching this and it seems like copyright years are really not that meaningful as long as you can prove when you created and edited the code, which in the era of SCMs is pretty easy to do.

A quick new search (emphasis by me):

> Should I include a copyright notice or register the copyright in my work?
> 
> Although **no longer required for copyright protection**, ’using a copyright notice is a good idea. A notice informs others that the work is copyrighted and by whom, thus potentially deterring infringement and facilitating requests for permission. In addition, in the event another party uses the work without permission, the presence of the copyright notice will preclude that person from using a defense of innocent infringement.
> 
> A proper notice generally requires the symbol “©” or the word “Copyright,” together with the copyright holder’s name and the year of **first publication**—for example, “© 2023 President and Fellows of Harvard College.” This designation should appear on or near the title page in printed works, and on an early screen in electronic works.

[Harvard University - Copyright and Fair Use](https://ogc.harvard.edu/pages/copyright-and-fair-use)

src/frequenz/sdk/timeseries/abstract_pool/__init__.py
simonvoelcker: Made them private :+1:

src/frequenz/sdk/timeseries/component_pool/_component_pool.py
simonvoelcker: (made it public)

src/frequenz/sdk/timeseries/pv_pool/_pv_pool.py
simonvoelcker: Made the change (5).

examples/battery_pool.py
```diff
@@ -31,7 +31,7 @@ async def main() -> None:
         battery_pool.soc.new_receiver(limit=1),
         battery_pool.capacity.new_receiver(limit=1),
         # pylint: disable-next=protected-access
```
llucax: Nitpick: This can be removed now.
```suggestion
```

src/frequenz/sdk/microgrid/_power_managing/_power_managing_actor.py
```diff
@@ -150,19 +150,19 @@ def _add_system_bounds_tracker(self, component_ids: frozenset[ComponentId]) -> N
                 priority=-sys.maxsize - 1, component_ids=component_ids
             )
             # pylint: disable-next=protected-access
```
llucax:
```suggestion
```

src/frequenz/sdk/microgrid/_power_managing/_power_managing_actor.py
```diff
@@ -150,19 +150,19 @@ def _add_system_bounds_tracker(self, component_ids: frozenset[ComponentId]) -> N
...
-            bounds_receiver = battery_pool._system_power_bounds.new_receiver()
+            bounds_receiver = battery_pool.system_power_bounds.new_receiver()
         elif issubclass(self._component_class, EvCharger):
             ev_charger_pool = _data_pipeline.new_ev_charger_pool(
                 priority=-sys.maxsize - 1, component_ids=component_ids
             )
             # pylint: disable-next=protected-access
```
llucax:
```suggestion
```

src/frequenz/sdk/microgrid/_power_managing/_power_managing_actor.py
```diff
@@ -150,19 +150,19 @@ def _add_system_bounds_tracker(self, component_ids: frozenset[ComponentId]) -> N
...
-            bounds_receiver = ev_charger_pool._system_power_bounds.new_receiver()
+            bounds_receiver = ev_charger_pool.system_power_bounds.new_receiver()
         elif issubclass(self._component_class, SolarInverter):
             pv_pool = _data_pipeline.new_pv_pool(
                 priority=-sys.maxsize - 1, component_ids=component_ids
             )
             # pylint: disable-next=protected-access
```
llucax:
```suggestion
```

src/frequenz/sdk/timeseries/battery_pool/_battery_pool_reference_store.py
```diff
@@ -77,84 +79,80 @@ def __init__(  # pylint: disable=too-many-arguments
...
-            batteries_id: Subset of the batteries that should be included in the
-                battery pool. If None or empty, then all batteries from the microgrid
-                will be used.
-
-        Raises:
-            ValueError: If any of the specified batteries is not present in the
-                microgrid.
+            batteries_id: Deprecated. Subset of the batteries that should be included
+                in the battery pool. If None or empty, then all batteries from the
+                microgrid will be used.
```
llucax: I would just forward to the new name, without explaining how to use the old name.

```suggestion
            batteries_id: Deprecated, use `component_ids` instead.
```

llucax (commented):
Damn, I just realized `batteries_id` is an argument for the reference pool, not the battery pool (I always think the pool should be constructed passing the IDs 😱).

Since reference stores are private (not part of the public modules), we don't actually need the deprecation path, we can just rename it (so we can remove the override for the ctor).

Sorry I didn't realized this before 😬 

Besides these very minor comments, this is good to go for me.

src/frequenz/sdk/timeseries/component_pool/_component_pool_reference_store.py
llucax: I think this one we can leave it for when we implement #949, then we can make sure empty sets are an error.

src/frequenz/sdk/timeseries/battery_pool/_battery_pool_reference_store.py
llucax: This is not really relevant if we remove `batteries_id` without going through the deprecation, of course.

llucax (approved):
LGTM, removing the protocol, if possible, is optional and low-prio IMHO, so approving.

src/frequenz/sdk/timeseries/battery_pool/messages.py
simonvoelcker: To be honest, I can't tell.

PR #1392, opened 2026-04-22
Add tests for some edge cases to prevent regressions

llucax:
Add targeted tests adapted to the v1.x.x branch that verify behavior for issues identified in the pull request #1371:

- Data sourcing: Verify duplicate subscription and unknown components with same channel name does not hang. There is also a test for invalid metrics, but this one is currently failing (it hangs), but since we plan to remove the data sourcing actor, we delay the fix for now.

- Resampling: Verify invalid metric request does not block later valid request.

- Power status subscriptions (battery, EV charger, PV pools): Verify same-instance subscriptions share the registry channel correctly.

### DIFF
--- a/tests/actor/test_resampling.py
+++ b/tests/actor/test_resampling.py
@@ -184,3 +184,48 @@ async def test_duplicate_request(
         )
 
         await resampling_actor._resampler.stop()  # pylint: disable=protected-access
+
+
+async def test_stalled_request_does_not_block_later_subscriptions() -> None:
+    """Ensure resampling does not head-of-line block."""
+    channel_registry = ChannelRegistry(name="test")
+    data_source_req_chan = Broadcast[ComponentMetricRequest](name="data-source-req")
+    data_source_req_recv = data_source_req_chan.new_receiver()
+    resampling_req_chan = Broadcast[ComponentMetricRequest](name="resample-req")
+    resampling_req_sender = resampling_req_chan.new_sender()
+
+    async with ComponentMetricsResamplingActor(
+        channel_registry=channel_registry,
+        data_sourcing_request_sender=data_source_req_chan.new_sender(),
+        resampling_request_receiver=resampling_req_chan.new_receiver(),
+        config=ResamplerConfig2(
+            resampling_period=timedelta(seconds=0.2),
+            max_data_age_in_periods=2,
+        ),
+    ):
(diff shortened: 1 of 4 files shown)


tests/actor/test_resampling.py
```diff
@@ -184,3 +184,48 @@ async def test_duplicate_request(
         )
 
         await resampling_actor._resampler.stop()  # pylint: disable=protected-access
+
+
+async def test_stalled_request_does_not_block_later_subscriptions() -> None:
+    """Ensure resampling does not head-of-line block on v1.x.x."""
```
simonvoelcker: Can you explain the significance of the 1.x.x branch here? Was the behavior different on 0.x.x and/or do we keep backwards compatibility?

tests/timeseries/_pv_pool/test_pv_pool.py
```diff
@@ -0,0 +1,84 @@
...
+
+from frequenz.channels import Broadcast
+from frequenz.client.common.microgrid.components import ComponentId
+from frequenz.quantities import Power
+from pytest_mock import MockerFixture
+
+from frequenz.sdk import timeseries
+from frequenz.sdk._internal._channels import ChannelRegistry
+from frequenz.sdk.microgrid._power_managing import ReportRequest, _Report
+from frequenz.sdk.timeseries.pv_pool._pv_pool import PVPool
```
simonvoelcker:
```python
from frequenz.sdk.timeseries.pv_pool import PVPool
```

This should work, same in the line below and in the similar EV charger tests.

simonvoelcker (dismissed):
Big fan of more tests 👍

tests/actor/test_resampling.py
llucax: AI glitch, it was on comparison with #1371. Will remove.

tests/timeseries/_pv_pool/test_pv_pool.py
llucax: Damn, will remove other potential internal imports. Good catch!

llucax:
[Fixed](https://github.com/frequenz-floss/frequenz-sdk-python/compare/83caed478e44fbb8bdef22ab878e8a05da45a44b..b32620a0a996729c3b5f53c8eeaad918a6785bbd).

simonvoelcker (approved):

PR #1396, opened 2026-04-29
Hold last known power bounds during transient data outages

Marenz:
When the microgrid API loses contact with a component (e.g. a Modbus timeout), the `PowerBoundsCalculator` emits `SystemBounds` with `inclusion_bounds=None`. The `PowerManagingActor`'s bounds tracker propagated this `None` value to the Matryoshka allocation algorithm, which clamped it to `[0, 0]` (via the `else: Power.zero()` branch in `_calc_targets`). All downstream actors then saw zero available power and dropped their setpoints to 0 W, even though the hardware was capable of full power and the telemetry interruption lasted only a few seconds.

The root cause is a semantic mismatch: "no data" is not the same as "zero power available". Treating them as equivalent means a short telemetry glitch collapses the energy throughput of an entire site.

**Fix**

`_should_update_bounds()` returns `False` when `inclusion_bounds` becomes `None` after having been valid. The bounds tracker holds the last known good bounds and only updates when real data resumes. Transient telemetry outages no longer collapse power allocations.

**Testing**

The logic is extracted into the static helper `PowerManagingActor._should_update_bounds()` with 8 unit tests that cover:

- first-ever bounds accepted
- identical bounds skipped (no-op dedup)
- genuine bound changes propagated
- `inclusion_bounds=None` after valid bounds — ignored
- `inclusion_bounds=None` -> `None` — still ignored
- valid bounds after `None` — accepted
- real `[0, 0]` bounds from actual sensor data — still propagated (the guard only blocks the `None` -> `[0, 0]` path)

### DIFF
--- a/src/frequenz/sdk/microgrid/_power_managing/_power_managing_actor.py
+++ b/src/frequenz/sdk/microgrid/_power_managing/_power_managing_actor.py
@@ -128,15 +128,43 @@ class PowerManagingActor(Actor):
         """
         last_bounds: SystemBounds | None = None
         async for bounds in bounds_receiver:
-            if (
-                last_bounds is not None
-                and bounds.inclusion_bounds == last_bounds.inclusion_bounds
-            ):
+            if not self._should_update_bounds(bounds, last_bounds):
+                last_bounds = bounds
                 continue
             last_bounds = bounds
             self._system_bounds[component_ids] = bounds
             await self._send_updated_target_power(component_ids, None)
             await self._send_reports(component_ids)
+
+    @staticmethod
+    def _should_update_bounds(
+        new_bounds: SystemBounds,
+        last_bounds: SystemBounds | None,
+    ) -> bool:
+        """Decide whether the bounds tracker should propagate *new_bounds*.
+
+        This is a no-op if *new_bounds* is identical to *last_bounds*.  When
+        the system loses contact with a component (e.g.  a Modbus timeout) the
+        BoundsCalculator emits ``inclusion_bounds=None``.  If we already know
+        good bounds we hold them — "no data" is not the same as "zero power
+        available".
+        """
+        if last_bounds is None:
+            return True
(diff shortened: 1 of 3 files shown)


src/frequenz/sdk/microgrid/_power_managing/_power_managing_actor.py
```diff
@@ -128,16 +128,44 @@ async def _bounds_tracker(
...
             last_bounds = bounds
             self._system_bounds[component_ids] = bounds
             await self._send_updated_target_power(component_ids, None)
             await self._send_reports(component_ids)
 
+    @staticmethod
+    def _should_update_bounds(
+        new_bounds: SystemBounds,
+        last_bounds: SystemBounds | None,
+    ) -> bool:
```
llucax, line 143: I wonder if old bounds should be only kept for some fixed period of time instead of forever. There was also the idea of somehow attaching a timestamp or a confidence level to values, but I guess that will be more complicated to implement and not backwards compatible.

src/frequenz/sdk/microgrid/_power_managing/_power_managing_actor.py
```diff
@@ -128,16 +128,44 @@ async def _bounds_tracker(
...
+        """Decide whether the bounds tracker should propagate *new_bounds*.
+
+        This is a no-op if *new_bounds* is identical to *last_bounds*.  When
```
llucax, line 146: Might be misleading, it only compares `inclusion_bounds`, not the whole bounds.

Marenz:
Honestly, after the thorough investigation by Sahas & Tiyash, i am no longer sure this PR is needed or even helpful

shsms:
I agree, and if we need this sort of thing in the future, it should be implemented in the battery pool's bound stream.

llucax:
So close?

PR #1400, opened 2026-05-06
Bump component-graph version to 0.4

shsms:


Marenz (approved):

llucax:
Are there an SDK user-visible changes with this bump? If so, we should document them properly in the RELEASE NOTES, pointing to the component graph release too for more details. If there are breaking changes we should add migration steps in the release notes.

PR #1403, opened 2026-05-15
Extract mixin from PV and EV component managers

simonvoelcker:
PV inverter manager, EV charger manager and also battery manager have similar functionality for setting power levels in the microgrid API. These have drifted apart to different degrees, especially the battery manager implementation differs a lot. The other two were close enough to extract them into a util function. This will be reusable when we introduce a similar manager for steam boilers.

### DIFF
--- a/src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_ev_charger_manager/_ev_charger_manager.py
+++ b/src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_ev_charger_manager/_ev_charger_manager.py
@@ -17,7 +17,6 @@ from frequenz.channels import (
     selected_from,
 )
 from frequenz.client.common.microgrid.components import ComponentId
-from frequenz.client.microgrid import ApiClientError, MicrogridApiClient
 from frequenz.client.microgrid.component import EvCharger
 from frequenz.quantities import Power, Voltage
 from typing_extensions import override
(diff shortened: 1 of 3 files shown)


llucax (commented):
LGTM. I wonder if it makes more sense to just include it in `ComponentManager` and make `BatteryManager` override it if it needs something special, instead of introducing a separate mixin.

Will leave @shsms the final approval as he's more familiar with this code.

simonvoelcker:
@llucax I was considering that, but the battery manager is structured very differently. The closest match to the extracted method, `_set_api_power`, is [_set_distributed_power](https://github.com/frequenz-floss/frequenz-sdk-python/blob/v1.x.x/src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_battery_manager.py#L740-L745), and you will find that its signature already differs quite a lot.

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_set_power_mixin.py
```diff
@@ -0,0 +1,113 @@
...
+
+import asyncio
+import logging
+from datetime import datetime, timedelta
+
+from frequenz.client.base.exception import ApiClientError
+from frequenz.client.common.microgrid.components import ComponentId
+from frequenz.quantities import Power
+
+from frequenz.sdk.microgrid import connection_manager
```
shsms:
```suggestion
from ... import connection_manager
```

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_set_power_mixin.py
```diff
@@ -0,0 +1,113 @@
...
+from ..result import PartialFailure, Result, Success
+
+_logger = logging.getLogger(__name__)
+
+
+class SetPowerMixin:
+    """Mixin for setting component powers via microgrid API."""
+
+    @staticmethod
+    async def _set_api_power(  # pylint: disable=too-many-locals,too-many-arguments
```
shsms: Optional, but I wonder if this should just be a function instead, the mixin is a lot of machinary for a static method that doesn't need self.

shsms (commented):
We use relative import paths, and one optional comment.

shsms (dismissed):

shsms:
looks like CI is failing

shsms (approved):

PR #1407, opened 2026-05-28
Allow `ComponentDataSamples` carrying only states

### LINKED ISSUE #1406: Error state gets lost when converting API data to old_component_data types
### What happened?

A protobuf component data message with no metrics but with a component state was not converted and sent downstream, so downstream didn't learn about the component's error status as soon as it happened.

```
ERROR    frequenz.sdk.microgrid._old_component_data:206: Error while creating 'InverterData' from samples: ComponentDataSamples(component_id=ComponentId(1503), metric_samples=[], states=[ComponentStateSample(sampled_at=datetime.datetime(2026, 5, 21, 12, 37, 40, 107899, tzinfo=datetime.timezone.utc), states=frozenset({<ComponentStateCode.ERROR: 10>}), warnings=frozenset(), errors=frozenset())])
  File "/venv/lib/python3.13/site-packages/frequenz/sdk/microgrid/_old_component_data.py", line 204, in _receive_logging_errors
    return cls.from_samples(samples)
           ~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/venv/lib/python3.13/site-packages/frequenz/sdk/microgrid/_old_component_data.py", line 708, in from_samples
    raise ValueError("No metrics in the samples.")
ValueError: No metrics in the samples.
```

This should not have failed to construct an `InverterData` instance, it should have created an instance with default values and Error state.

llucax:
If a `ComponentDataSample` doesn't have any `metric_samples` we raise an exception, but that's not correct, as it is perfectly valid for them to carry only states.

This commits allows this case, using the timestamp of the last state (as we do with metrics). The only caveat is the data itself will be created in `*Data` object using defaults (zeros).

We also add some regression tests, although the affected code should be gone soon.

Fixes #1406.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -14,4 +14,4 @@
 
 ## Bug Fixes
 
-<!-- Here goes notable bug fixes that are worth a special mention or explanation -->
+* Accept `ComponentDataSamples` that only carry state changes (the state changes were dropped before).
--- a/src/frequenz/sdk/microgrid/_old_component_data.py
+++ b/src/frequenz/sdk/microgrid/_old_component_data.py
@@ -104,19 +104,26 @@ class ComponentData(ABC):
     @staticmethod
     def _from_samples(class_: type[T], /, samples: ComponentDataSamples) -> T:
         """Create a new instance from a component data object."""
-        if not samples.metric_samples:
-            raise ValueError("No metrics in the samples.")
+        if not samples.metric_samples and not samples.states:
+            raise ValueError("No metrics and no states in the samples.")
 
         # FIXME: This might not be true forever, but the service sends all metrics with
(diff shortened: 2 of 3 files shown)


Marenz (approved):
lgtm. nice catch removing the redundant per-subclass `metric_samples` checks too, the parent guard + no-op `for sample in metric_samples` loops make that safe.

one thing, not blocking: when both `metric_samples` and `states` are present, timestamp comes from metrics and state timestamp mismatches are silently ignored. fits the "one timestamp for now" `FIXME` but nothing tests it.

llucax:
> one thing, not blocking: when both metric_samples and states are present, timestamp comes from metrics and state timestamp mismatches are silently ignored. fits the "one timestamp for now" FIXME but nothing tests it.

True, this should be rare and hopefully this code will be gone soon, so I hope we can skip it for now.

PR #1408, opened 2026-05-29
Speed up `from_samples` metric dispatch

shsms:
Each `from_samples` matches a sample's metric against ~40 `case Metric.X` arms, and every arm evaluates `Metric.X` through the slow `EnumType.__getattribute__`/`_name_map` machinery — once per case, per sample, per component, on every resampling tick. Profiling showed this to be a dominant CPU consumer (~45% of Python CPU at a realistic component count and sample rate).

Resolve the matched members once into a plain holder class `_M` and match on `case _M.X`, so the patterns use ordinary (fast) attribute access. This roughly halves steady-state Python CPU in profiling (43 components @ 5 Hz); behaviour is unchanged.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,11 +2,7 @@
 
 ## Summary
 
-<!-- Here goes a general summary of what this release is about -->
-
-## Upgrading
-
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+Along with the new `tick_delay` resampler config option, this release also includes some performance improvements in the data pipeline.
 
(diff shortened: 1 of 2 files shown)


shsms:
Internal change, no release notes necessary.

llucax (dismissed):
Wow, this is bizarre if it really has such a performance impact 😱

llucax:
Optional, adding a short release note saying we improved performance.

shsms:
Done, taking it in.  Will release after this.

llucax:
FYI, tick delay was already released, I forgot to reset the notes I guess 🙈

PR #123, opened 2022-12-19
resampling: Don't subscribe twice to the same source

leandro-lucarella-frequenz:
When receiving a subcription, we need to check if we are already handling a subscription with the same request parameters to avoid pushing the data twice to the same channel.

Fixes #115.

PR #1411, opened 2026-06-01
Fix unreachable-power formula leak in the power distributor

shsms:
The power distributor runs a formula for each set of unreachable components.  They were created on the shared logical-meter pool, which never evicts unused formulas.

On teardown only the cache was stopped, not the formula — so a formula-evaluating actor leaked per distinct set, and CPU climbed over time until a restart.

## Changes

* `FormulaPool`: `stop()` now also stops string formulas; new `stop_string_formula()` stops + evicts a single one.
* Each `ComponentManager` lazily owns a dedicated `FormulaPool` (new `_DataPipeline.new_formula_pool`) and stops its formulas on teardown, via a per-type `_unreachable_power_formula` hook. Battery and PV use it.
* Fix `from_reactive_power_formula` checking the wrong cache dict, which re-created and leaked the grid reactive-power formula on every access.

## Breaking

* PV distribution now subtracts the power of unreachable PV inverters from the target, so PV results differ when some inverters are unreachable.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -10,8 +10,9 @@
 
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+* The PV inverter manager now accounts for the power measured on unreachable PV inverters when distributing power, so the reachable inverters compensate for it (matching the battery manager's behavior).
 
 ## Bug Fixes
 
-<!-- Here goes notable bug fixes that are worth a special mention or explanation -->
+* Fixed a resource leak in the power distributor: the formulas created for unreachable batteries were never stopped, so CPU usage slowly climbed over time until the application was restarted.
+* Fixed the grid reactive-power formula being recreated and leaked on every access instead of being reused from the cache.
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -7,6 +7,7 @@
 ## Upgrading
 
 * Update microgrid client to v0.18.3+, which fixes a problem with missing steam boilers on formula generation.
+* The PV inverter manager now subtracts the power measured on unreachable PV inverters from the distribution target, so the power sent to the reachable inverters can change when some requested PV inverters are unreachable.
 
 ## New Features
 
--- a/src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
+++ b/src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
@@ -46,6 +46,7 @@ class PVManager(ComponentManager):
             api_power_request_timeout: Timeout to use when making power requests to
                 the microgrid API.
         """
+        super().__init__()
         self._results_sender = results_sender
         self._api_power_request_timeout = api_power_request_timeout
         self._pv_inverter_ids = self._get_pv_inverter_ids()
@@ -90,6 +91,7 @@ class PVManager(ComponentManager):
         )
         if self._component_pool_status_tracker:
             await self._component_pool_status_tracker.stop()
+        await self._stop_all_unreachable_power_subscriptions()
 
     @override
     async def distribute_power(self, request: Request) -> None:
@@ -119,10 +121,21 @@ class PVManager(ComponentManager):
                 "Cannot distribute power to PV inverters without any inverters"
             )
 
-        working_components: list[ComponentId] = []
-        for inv_id in self._component_pool_status_tracker.get_working_components(
+        working = self._component_pool_status_tracker.get_working_components(
             request.component_ids
-        ):
+        )
+        await self._subscribe_to_unreachable_power(request.component_ids, working)
+        unreachable_power = self._unreachable_power(request.component_ids)
+        if unreachable_power is not None:
+            remaining_power -= unreachable_power
+            _logger.debug(
+                "Excluding %s measured on unreachable PV inverters from the power to "
+                "distribute on working PV inverters.",
+                unreachable_power,
+            )
+
+        working_components: list[ComponentId] = []
+        for inv_id in working:
             if self._component_data_caches[inv_id].has_value():
                 working_components.append(inv_id)
             else:
@@ -185,6 +198,13 @@ class PVManager(ComponentManager):
             component_category="PV inverter",
         )
         await self._results_sender.send(result)
+
+    @override
+    def _unreachable_power_formula(
+        self, component_ids: collections.abc.Set[ComponentId]
+    ) -> str:
+        """Return the formula for the active power of the given PV inverters."""
+        return connection_manager.get().component_graph.pv_formula(component_ids)
 
     def _get_pv_inverter_ids(self) -> collections.abc.Set[ComponentId]:
         """Return the IDs of all PV inverters present in the component graph."""
--- a/src/frequenz/sdk/timeseries/formulas/_formula_pool.py
+++ b/src/frequenz/sdk/timeseries/formulas/_formula_pool.py
@@ -136,7 +136,7 @@ class FormulaPool:
         Returns:
             A formula that evaluates the given formula.
         """
-        if channel_key in self.power_formulas:
+        if channel_key in self._reactive_power_formulas:
             return self._reactive_power_formulas[channel_key]
 
         if formula_str == "0.0":
--- a/src/frequenz/sdk/timeseries/formulas/_formula_pool.py
+++ b/src/frequenz/sdk/timeseries/formulas/_formula_pool.py
@@ -61,7 +61,7 @@ class FormulaPool:
             resampler_subscription_sender
         )
 
-        self._string_formulas: dict[str, Formula[Quantity]] = {}
+        self._string_formulas: dict[tuple[str, Metric], Formula[Quantity]] = {}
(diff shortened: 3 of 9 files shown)


src/frequenz/sdk/timeseries/formulas/_formula_pool.py
```diff
@@ -246,8 +246,31 @@ def from_current_3_phase_formula(
...
+        Callers that create short-lived formulas (e.g. one per changing set of
+        components) must call this when the formula is no longer needed —
+        otherwise the underlying formula-evaluating actor keeps running and the
+        pool's cache grows without bound.
+
+        Args:
+            formula_str: The formula string passed to `from_string`.
+            metric: The metric passed to `from_string`.
+        """
+        channel_key = formula_str + str(metric.value)
```
llucax: `formula_str + str(metric.value)` is ambiguous. The same key is also used by `from_string()` earlier in this class, so distinct `(formula, metric)` pairs can collide. For example, with `frequenz-client-microgrid` v0.18.3, `Metric.DC_POWER.value == 3` and `Metric.AC_POWER_APPARENT_PHASE_1.value == 23`, so `"#12" + "3"` and `"#1" + "23"` both become `#123`. That can return the wrong cached formula, or make this method stop/evict the wrong actor while leaving the intended one running.

* Suggested fix: use a structured key, for example `dict[tuple[str, Metric], Formula[Quantity]]`, and use the same tuple in `from_string()` and `stop_string_formula()`.

* Suggested test: create two formulas with colliding concatenated keys, assert they are distinct, then stop one and assert the other still has its evaluator/receiver usable.

src/frequenz/sdk/timeseries/formulas/_formula_pool.py
```diff
@@ -246,8 +246,31 @@ def from_current_3_phase_formula(
...
+        formula = self._string_formulas.pop(channel_key, None)
```
llucax: This removes the formula from the pool before `formula.stop()` completes. If stopping is cancelled or raises, the pool has lost the handle needed to retry cleanup, which can leave the evaluator running. There is a related race in `stop()` below: it snapshots `_string_formulas`, awaits each stop, and then clears the dict, so a concurrent `from_string()` after the snapshot can be cleared without being stopped. 

* Suggested fix: guard formula creation and shutdown with a lock or a “stopping” state, and only evict entries after successful/complete cleanup. At minimum, avoid clearing entries that were added after the snapshot.

* Suggested tests: use a fake or patched formula whose `stop()` blocks/raises, call `stop_string_formula()` or `stop()` concurrently with `from_string()`, and assert no running formula is lost from the pool.

RELEASE_NOTES.md
```diff
@@ -10,8 +10,9 @@
 
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+* The PV inverter manager now accounts for the power measured on unreachable PV inverters when distributing power, so the reachable inverters compensate for it (matching the battery manager's behavior).
```
llucax, line 14: The PR body calls this a breaking behavior change because PV distribution results now differ when some requested inverters are unreachable, but the release notes only list it as a new feature.

* Suggested fix: add an Upgrading or Breaking Changes note explicitly saying PV distribution results can change when requested PV inverters are unreachable.

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
```diff
@@ -119,10 +121,21 @@ async def distribute_power(self, request: Request) -> None:
                 "Cannot distribute power to PV inverters without any inverters"
             )
 
-        working_components: list[ComponentId] = []
-        for inv_id in self._component_pool_status_tracker.get_working_components(
+        working = self._component_pool_status_tracker.get_working_components(
             request.component_ids
-        ):
+        )
+        await self._subscribe_to_unreachable_power(request.component_ids, working)
```
llucax, line 127: This new unreachable-PV compensation path is not covered by tests in this PR. Please add regression coverage showing that measured unreachable PV power is subtracted from the target for the reachable inverters, including a boundary case where the unreachable PV power already exceeds or satisfies the requested target. That would also document the behavior change called out in the PR body.

Suggested tests:

- `FormulaPool.stop()` stops formulas created by from_string().
- `stop_string_formula()` stops only the requested string formula and evicts it.
- `from_reactive_power_formula()` returns the cached formula on repeated calls.
- Battery unreachable-power subscriptions are stopped when the unreachable set changes and on manager shutdown.
- PV distribution subtracts measured unreachable PV power from the reachable inverter target.
- PV distribution handles the case where unreachable PV production already exceeds the requested production.
- Invalid/error paths after creating an unreachable-power subscription do not leave formulas running.

llucax (commented):
This is my AI-powered review, tomorrow I will give it a look myself.

llucax:
I think this is a direct consequence on how we didn't get asyncio concurrency in Python very well at the beginnings of the SDK. I hope we can get some time to clear things up and make concurrency (including cleanup) safer.

src/frequenz/sdk/timeseries/formulas/_formula_pool.py
shsms: fixed.

src/frequenz/sdk/timeseries/formulas/_formula_pool.py
shsms: fixed.

RELEASE_NOTES.md
shsms: added an upgrading entry also.

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_pv_inverter_manager/_pv_inverter_manager.py
shsms: added tests.

llucax (commented):
LGTM. AI says there is still some scenario where we can leak formulas. I'll let you decide if you want to move to a separate issue of fix in this PR.

---

## Finding: manager shutdown can still leak formula actors if a cache stop fails

**Severity:** Medium
**Location:** `src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_component_manager.py`, `_stop_all_unreachable_power_subscriptions()`

The single-subscription teardown path was improved correctly: `_stop_unreachable_power_subscription()` uses `try/finally`, so the pooled formula is stopped even if `subscription.cache.stop()` raises. 

But the bulk shutdown path still does this:

```python
async def _stop_all_unreachable_power_subscriptions(self) -> None:
    """Stop every unreachable-power subscription and the owned formula pool."""
    for subscription in self._unreachable_power_subscriptions.values():
        await subscription.cache.stop()
    self._unreachable_power_subscriptions.clear()
    if self._unreachable_power_pool is not None:
        await self._unreachable_power_pool.stop()
        self._unreachable_power_pool = None
```

If any `subscription.cache.stop()` raises, the function exits before clearing subscriptions and, more importantly, before `self._unreachable_power_pool.stop()` runs. That means the formula-evaluating actors owned by the pool can remain running during manager shutdown. This is the same class of leak the PR is trying to fix, just in the manager-level teardown path. 

### Impact

During normal stop/teardown of `BatteryManager` or `PVManager`, a cache teardown failure could prevent the owned `FormulaPool` from stopping. Battery and PV managers both call `_stop_all_unreachable_power_subscriptions()` during `stop()`: battery after stopping component caches and the status tracker, PV after stopping inverter caches and the status tracker.  

In practice, this could leave formula actors and resampler subscriptions alive after a component manager is stopped. If managers are recreated, CPU/memory usage can still grow over time.

### Suggested fix

Make the pool cleanup happen in a `finally`, and avoid losing later cleanup because one cache failed. For example:

```python
async def _stop_all_unreachable_power_subscriptions(self) -> None:
    subscriptions = list(self._unreachable_power_subscriptions.values())
    self._unreachable_power_subscriptions.clear()

    cache_stop_errors: list[BaseException] = []
    try:
        for subscription in subscriptions:
            try:
                await subscription.cache.stop()
            except BaseException as err:
                cache_stop_errors.append(err)
    finally:
        if self._unreachable_power_pool is not None:
            try:
                await self._unreachable_power_pool.stop()
            finally:
                self._unreachable_power_pool = None

    if cache_stop_errors:
        raise cache_stop_errors[0]
```

A variant using `asyncio.gather(..., return_exceptions=True)` would also be fine, as long as the formula pool is always stopped and an error is still surfaced.

### Suggested test

Add a test parallel to the existing single-subscription failure test, but exercise `manager.stop()` / `_stop_all_unreachable_power_subscriptions()`:

1. Create two unreachable subscriptions.
2. Make the first `LatestValueCache.stop()` raise.
3. Assert the owned pool’s `stop()` is still called.
4. Assert the pool reference is cleared or otherwise left in a retryable, well-defined state.
5. Assert the original cache-stop error is not silently swallowed.

llucax (approved):

llucax:
Approving in case you want to merge as is, but disable auto-merge to really give you the choice :D

shsms:
They're talking about leaks during power distributor shutdown, which basically means during app shutdown, so I've turned it into an issue: https://github.com/frequenz-floss/frequenz-sdk-python/issues/1412

Pushed again to fix signatures.

PR #1429, opened 2026-06-08
Add a steam boiler pool

shsms:
Steam boilers are controllable electrical loads the SDK could not monitor or
manage. This adds a steam boiler pool with power readings and power control,
following the structure of the EV charger and PV pools.

## Changes

- New public API: `microgrid.new_steam_boiler_pool()` returning a
  `SteamBoilerPool` with `power`, `power_status`, and a consumption-only
  `propose_power` (negative power is rejected).
- `SteamBoilerData` component data type and data sourcing support.
- Power manager wiring using the plain matryoshka algorithm, with
  `DefaultPower.ZERO` when no actor is proposing.
- `SteamBoilerManager` splits the target power equally across the working
  boilers, keeping every allocation inside the boiler's inclusion bounds: a
  share below a boiler's minimum operating power is raised to the minimum, or
  the boiler is kept off. Like the PV manager, it subtracts the measured draw
  of unreachable boilers from the target.
- `SteamBoilerStatusTracker` for per-boiler health, mirroring the PV tracker.

## Worth a look

- The single-pass allocation reports power as excess when a high-minimum
  boiler can't be afforded instead of topping up earlier boilers
  (`test_unaffordable_minimum_strands_excess` pins this).
- With no usable boilers a request is dropped with only an error log — the
  same behavior as the PV and EV managers; a follow-up could make all three
  send an `Error` result the way the battery manager does.

### DIFF
--- a/src/frequenz/sdk/microgrid/_data_sourcing/microgrid_api_source.py
+++ b/src/frequenz/sdk/microgrid/_data_sourcing/microgrid_api_source.py
@@ -24,6 +24,7 @@ from .._old_component_data import (
     EVChargerData,
     InverterData,
     MeterData,
+    SteamBoilerData,
     TransitionalMetric,
 )
 from ._component_metric_request import ComponentMetricRequest
@@ -123,6 +124,38 @@ _EV_CHARGER_DATA_METHODS: dict[
 }
 
 _CHP_DATA_METHODS: dict[Metric | TransitionalMetric, Callable[[ChpData], float]] = {
+    Metric.AC_ACTIVE_POWER: lambda msg: msg.active_power,
+    Metric.AC_ACTIVE_POWER_PHASE_1: lambda msg: msg.active_power_per_phase[0],
+    Metric.AC_ACTIVE_POWER_PHASE_2: lambda msg: msg.active_power_per_phase[1],
+    Metric.AC_ACTIVE_POWER_PHASE_3: lambda msg: msg.active_power_per_phase[2],
+    TransitionalMetric.ACTIVE_POWER_INCLUSION_LOWER_BOUND: lambda msg: (
+        msg.active_power_inclusion_lower_bound
+    ),
+    TransitionalMetric.ACTIVE_POWER_EXCLUSION_LOWER_BOUND: lambda msg: (
+        msg.active_power_exclusion_lower_bound
+    ),
+    TransitionalMetric.ACTIVE_POWER_EXCLUSION_UPPER_BOUND: lambda msg: (
+        msg.active_power_exclusion_upper_bound
+    ),
+    TransitionalMetric.ACTIVE_POWER_INCLUSION_UPPER_BOUND: lambda msg: (
+        msg.active_power_inclusion_upper_bound
+    ),
+    Metric.AC_CURRENT_PHASE_1: lambda msg: msg.current_per_phase[0],
+    Metric.AC_CURRENT_PHASE_2: lambda msg: msg.current_per_phase[1],
+    Metric.AC_CURRENT_PHASE_3: lambda msg: msg.current_per_phase[2],
+    Metric.AC_VOLTAGE_PHASE_1_N: lambda msg: msg.voltage_per_phase[0],
+    Metric.AC_VOLTAGE_PHASE_2_N: lambda msg: msg.voltage_per_phase[1],
+    Metric.AC_VOLTAGE_PHASE_3_N: lambda msg: msg.voltage_per_phase[2],
+    Metric.AC_FREQUENCY: lambda msg: msg.frequency,
+    Metric.AC_REACTIVE_POWER: lambda msg: msg.reactive_power,
+    Metric.AC_REACTIVE_POWER_PHASE_1: lambda msg: msg.reactive_power_per_phase[0],
+    Metric.AC_REACTIVE_POWER_PHASE_2: lambda msg: msg.reactive_power_per_phase[1],
+    Metric.AC_REACTIVE_POWER_PHASE_3: lambda msg: msg.reactive_power_per_phase[2],
+}
+
+_STEAM_BOILER_DATA_METHODS: dict[
+    Metric | TransitionalMetric, Callable[[SteamBoilerData], float]
+] = {
     Metric.AC_ACTIVE_POWER: lambda msg: msg.active_power,
     Metric.AC_ACTIVE_POWER_PHASE_1: lambda msg: msg.active_power_per_phase[0],
     Metric.AC_ACTIVE_POWER_PHASE_2: lambda msg: msg.active_power_per_phase[1],
@@ -306,6 +339,31 @@ class MicrogridApiSource:
                 connection_manager.get().api_client, comp_id
             )
 
+    async def _check_steam_boiler_request(
+        self,
+        comp_id: ComponentId,
+        requests: dict[Metric | TransitionalMetric, list[ComponentMetricRequest]],
+    ) -> None:
+        """Check if the requests are valid steam boiler metrics.
+
+        Raises:
+            ValueError: if the requested metric is not available for steam boilers.
+
+        Args:
+            comp_id: The id of the requested component.
+            requests: A list of metric requests received from external actors
+                for the given steam boiler.
+        """
+        for metric in requests:
+            if metric not in _STEAM_BOILER_DATA_METHODS:
+                err = f"Unknown metric {metric} for steam boiler id {comp_id}"
+                _logger.error(err)
+                raise ValueError(err)
+        if comp_id not in self.comp_data_receivers:
+            self.comp_data_receivers[comp_id] = SteamBoilerData.subscribe(
+                connection_manager.get().api_client, comp_id
+            )
+
     async def _check_meter_request(
         self,
         comp_id: ComponentId,
@@ -362,6 +420,8 @@ class MicrogridApiSource:
             await self._check_meter_request(comp_id, requests)
         elif category == ComponentCategory.CHP:
             await self._check_chp_request(comp_id, requests)
+        elif category == ComponentCategory.STEAM_BOILER:
+            await self._check_steam_boiler_request(comp_id, requests)
         else:
             err = f"Unknown component category {category}"
             _logger.error(err)
@@ -393,6 +453,8 @@ class MicrogridApiSource:
             return _EV_CHARGER_DATA_METHODS[metric]
         if category == ComponentCategory.CHP:
             return _CHP_DATA_METHODS[metric]
+        if category == ComponentCategory.STEAM_BOILER:
+            return _STEAM_BOILER_DATA_METHODS[metric]
         err = f"Unknown component category {category}"
         _logger.error(err)
         raise ValueError(err)
--- a/src/frequenz/sdk/microgrid/_old_component_data.py
+++ b/src/frequenz/sdk/microgrid/_old_component_data.py
@@ -1131,6 +1131,270 @@ class ChpData(ComponentData):  # pylint: disable=too-many-instance-attributes
 
 
 @dataclass(kw_only=True)
+class SteamBoilerData(ComponentData):  # pylint: disable=too-many-instance-attributes
+    """A wrapper class for holding steam boiler data.
+
+    Steam boilers are controllable electrical loads.  Their active power and its
+    inclusion/exclusion bounds, along with the per-phase power, current, voltage and
+    frequency, are reported through the corresponding `AC_*` metrics.
+    """
+
+    active_power: float = 0.0
+    """The total active 3-phase AC power, in Watts (W).
+
+    Represented in the passive sign convention.
+
+    * Positive means consumption from the grid.
+    * Negative means supply into the grid.
+    """
+
+    active_power_per_phase: PhaseTuple = (0.0, 0.0, 0.0)
+    """The per-phase AC active power for phase 1, 2, and 3 respectively, in Watt (W).
+
+    Represented in the passive sign convention.
+
+    * Positive means consumption from the grid.
+    * Negative means supply into the grid.
+    """
+
+    reactive_power: float = 0.0
+    """The total reactive 3-phase AC power, in Volt-Ampere Reactive (VAr).
+
+    * Positive power means capacitive (current leading w.r.t. voltage).
+    * Negative power means inductive (current lagging w.r.t. voltage).
+    """
+
+    reactive_power_per_phase: PhaseTuple = (0.0, 0.0, 0.0)
+    """The per-phase AC reactive power, in Volt-Ampere Reactive (VAr).
+
+    The provided values are for phase 1, 2, and 3 respectively.
+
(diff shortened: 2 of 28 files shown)


src/frequenz/sdk/microgrid/_data_sourcing/microgrid_api_source.py
```diff
@@ -152,6 +153,38 @@
     Metric.AC_REACTIVE_POWER_PHASE_3: lambda msg: msg.reactive_power_per_phase[2],
 }
 
+_STEAM_BOILER_DATA_METHODS: dict[
```
cwasicki, line 156: Why do we need to repeat these for every device?

cwasicki (commented):
Any advice what to review here? On a first glimpse this looks like a lot of boiler-plate code.

src/frequenz/sdk/microgrid/_old_component_data.py
llucax, line 1: Oh, damn. I sill need to remove all this cr*p... 🤦 ☹️

src/frequenz/sdk/microgrid/_data_sourcing/microgrid_api_source.py
llucax: "historical reasons". This comes from the old days where the API have one message per component for the data. Now it just needs modernization. Hopefully soon™️.

shsms: It would be a lot of work to do it differently, which is not worth it, because this stuff will all be rewritten in rust in the next several months.

shsms:
> Any advice what to review here? On a first glimpse this looks like a lot of boiler-plate code.

The big things to check are:

- the power distribution algorithm, which allocates equally, starting from the smallest component, here: https://github.com/frequenz-floss/frequenz-sdk-python/pull/1429/changes#diff-acd36e353c52229236c9db5604976d972cb6e7cef600da43ef0d385af0f47b41R103
- the default power and power manager algorithm here: https://github.com/frequenz-floss/frequenz-sdk-python/pull/1429/changes#diff-9f32f637c4ac702aa43914f3ab2949c4b6adef41b482c1d300a4262c06b1f995R146-R147

I'll need to do some testing as well.

cwasicki (approved):
LGTM, the PR description mentions a wrong default though.

llucax (commented):
LGTM. AI review with a few findings coming up.

BTW, sorry for the delay, I started the review but never submitted it and forgot about it.

src/frequenz/sdk/timeseries/steam_boiler_pool/_system_bounds_tracker.py
```diff
@@ -0,0 +1,138 @@
...
+                    for data in self._latest_component_data.values()
+                )
+            ),
+            upper=Power.from_watts(
+                sum(
+                    data.active_power_exclusion_upper_bound
+                    for data in self._latest_component_data.values()
+                )
+            ),
+        )
```
llucax, line 89: 🟡➡️ Should aggregate bounds and allocation handle each boiler's exclusion zone?

The tracker sums exclusion bounds across boilers, while the distributor considers only each boiler's inclusion bounds. Is this simplification intentional while we gather operational data and choose the final distribution algorithm?

For example, if two boilers allow `[0, 0] ∪ [2, 10] kW` and `[0, 0] ∪ [4, 10] kW`, the aggregate exclusion zone becomes `0..6 kW`. A `7 kW` target is accepted, but equal allocation can send `3.5 kW` to both boilers, which is inside the second boiler's exclusion zone. Conversely, a feasible `5 kW` target can be rejected even though the first boiler could handle it alone.

#### Impact
The initial algorithm can reject feasible aggregate targets or send a per-boiler target that the API rejects. If this is a deliberate first approximation, documenting that constraint would make the intended scope explicit.

#### Related locations

- [`_steam_boiler_manager.py:181-205`](https://github.com/shsms/frequenz-sdk-python/blob/ba9e6186577af48ee95f3959e4bf99bbb997cb47/src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_steam_boiler_manager/_steam_boiler_manager.py#L181-L205): Allocation uses inclusion bounds but not exclusion bounds.

#### Suggested tests
Add a heterogeneous exclusion-bounds case to record the intended behavior, whether the current approximation is accepted or the allocator is changed.

---

🟡 Medium Severity: Probably best to address before merging, or explicitly accept and document the initial limitation.
➡️ Medium Effort: A complete solution requires choosing how non-convex per-device ranges should be aggregated and allocated.

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_steam_boiler_manager/_steam_boiler_manager.py
```diff
@@ -0,0 +1,239 @@
...
+                component_data.get().active_power_inclusion_lower_bound
+            )
+            if Power.zero() < allocated_power < lower_bound:
+                allocated_power = (
+                    lower_bound
+                    if lower_bound <= upper_bound and remaining_power >= lower_bound
+                    else Power.zero()
+                )
+            allocations[comp_id] = allocated_power
+            remaining_power -= allocated_power
```
llucax, line 205: 🟡➡️ Is intentionally stranding feasible power part of the initial simple algorithm?

This single pass can leave a feasible target mostly unallocated. With boilers supporting `0..1 kW` and `4..10 kW`, a `4.5 kW` target is reported as `1 kW` allocated and `3.5 kW` excess, although `0.5 + 4 kW` satisfies the target. The new `test_unaffordable_minimum_strands_excess` explicitly preserves this behavior.

Is this intentional as part of starting with a simple distribution algorithm while gathering operational data, or is it an oversight? If intentional, it would be useful to document the tradeoff because callers may otherwise interpret excess power as unavailable capacity.

The result can also depend on iteration order when boilers have equal upper bounds, because the stable sort starts from a set. Visiting a high-minimum boiler first can fulfill the request, while visiting it later can strand power.

#### Impact
The pool can under-consume significantly despite having enough available capacity, and tied bounds can make fulfillment dependent on component ordering.

#### Related locations

- [`test_steam_boiler_manager.py:229-265`](https://github.com/shsms/frequenz-sdk-python/blob/ba9e6186577af48ee95f3959e4bf99bbb997cb47/tests/microgrid/power_distributing/test_steam_boiler_manager.py#L229-L265): The test preserves the known limitation.

#### Suggested tests
Add permutations with tied upper bounds to make the chosen order and expected approximation deterministic.

---

🟡 Medium Severity: Probably best to clarify before merging because this affects observable control behavior.
➡️ Medium Effort: Clarifying and documenting the behavior is small; improving allocation requires a broader algorithm decision.

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_steam_boiler_manager/_steam_boiler_manager.py
```diff
@@ -0,0 +1,239 @@
...
+
+        result = await _set_component_power(
+            request=request,
+            target_power=request.power,
+            allocations=allocations,
+            api_request_timeout=self._api_power_request_timeout,
+            remaining_power=remaining_power,
+            component_category="steam boiler",
+        )
+        await self._results_sender.send(result)
```
llucax, line 221: 🟡⬇️ Forward set-power failures to the status tracker

`_set_component_power()` can return a `PartialFailure`, but the manager only publishes the result. It never calls `ComponentPoolStatusTracker.update_status()`, so the new `SteamBoilerStatusTracker._handle_set_power_result()` path is unreachable.

#### Impact
A boiler whose API request failed remains `WORKING` and can be selected again immediately instead of being marked `UNCERTAIN` and temporarily excluded.

#### Suggested fix (⬇️ effort)
Pass the succeeded and failed component sets from the result to `self._component_pool_status_tracker.update_status()` before publishing the result, as the battery manager does.

#### Suggested tests
Fail one boiler's API request while another succeeds, then verify that the failed boiler becomes `UNCERTAIN` and is excluded from the next allocation.

---

🟡 Medium Severity: Probably best to fix before merging because the status tracker's failure recovery currently cannot run.
⬇️ Low Effort: The status tracker and result data already exist; only the forwarding call and regression test are missing.

src/frequenz/sdk/timeseries/steam_boiler_pool/_system_bounds_tracker.py
```diff
@@ -0,0 +1,138 @@
...
+        self._component_pool_status = ComponentPoolStatus(set(), set())
+
+    def start(self) -> None:
+        """Start the steam boiler system bounds tracker."""
+        self._tasks.add(asyncio.create_task(run_forever(self._run)))
+
+    async def _send_bounds(self) -> None:
+        """Calculate and send the aggregate system bounds if they have changed."""
+        if not self._latest_component_data:
+            return
```
llucax, line 61: 🟡➡️ Clear bounds when the last boiler becomes unavailable

After the final cached component is removed, `_send_bounds()` returns without publishing a replacement. Consumers and the power manager therefore retain the previous nonzero capacity.

#### Impact
`power_status` continues to advertise stale capacity. A subsequent request reaches the distributor with no usable boilers, where it only logs and returns, so `power_distribution_results` emits nothing for that request.

#### Related locations

- [`_steam_boiler_manager.py:154-159`](https://github.com/shsms/frequenz-sdk-python/blob/ba9e6186577af48ee95f3959e4bf99bbb997cb47/src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_steam_boiler_manager/_steam_boiler_manager.py#L154-L159): Requests are dropped without a result when no boiler is usable.

#### Suggested fix (➡️ effort)
Publish unavailable system bounds when the cache becomes empty. Also emit an `Error`, or an explicit result with all requested power as excess, when no boiler can accept the request.

#### Suggested tests
Start with one working boiler, transition it to `NOT_WORKING`, verify that its old capacity disappears, then propose power and assert that a distribution result arrives.

---

🟡 Medium Severity: Probably best to fix before merging because stale capacity can persist indefinitely.
➡️ Medium Effort: The state transitions are straightforward, but the desired empty-bounds and no-result semantics need to be selected consistently.

src/frequenz/sdk/microgrid/_old_component_data.py
```diff
@@ -1130,6 +1130,270 @@ def to_samples(self) -> ComponentDataSamples:
...
+    * Negative power means inductive (current lagging w.r.t. voltage).
+    """
+
+    reactive_power_per_phase: PhaseTuple = (0.0, 0.0, 0.0)
+    """The per-phase AC reactive power, in Volt-Ampere Reactive (VAr).
+
+    The provided values are for phase 1, 2, and 3 respectively.
+
+    * Positive power means capacitive (current leading w.r.t. voltage).
+    * Negative power means inductive (current lagging w.r.t. voltage).
```
llucax, line 1173: 🟡⬇️ Correct the reactive-power sign convention

These descriptions reverse the Microgrid API convention. The API defines negative reactive power as capacitive, with current leading voltage, and positive reactive power as inductive, with current lagging voltage. Both the total and per-phase docstrings currently say the opposite.

API reference: [`microgrid.proto:183-187`](https://github.com/frequenz-floss/frequenz-api-microgrid/blob/479beed1b4d45ceddf8666eb844d552f8aa9160e/proto/frequenz/api/microgrid/v1alpha18/microgrid.proto#L183-L187).

#### Impact
Users can interpret measured reactive power with the wrong physical direction and build control or reporting logic around the reversed convention.

#### Suggested fix (⬇️ effort)
Change both docstrings to say that negative is capacitive and positive is inductive.

---

🟡 Medium Severity: Probably best to fix before merging because this documents a public data type with the opposite convention from the API.
⬇️ Low Effort: This requires changing four documentation lines.

llucax (commented):
🤖 **rAIview** (AI review here!)

The overall structure is consistent with the existing component pools. I found three concrete control-path issues around failure feedback, stale bounds when all boilers become unavailable, and reversed reactive-power documentation. I also left two questions about whether the deliberately simple allocation and exclusion-bound behavior should be accepted and documented for this initial version.

src/frequenz/sdk/timeseries/steam_boiler_pool/_system_bounds_tracker.py
shsms: When we discussed, we decided to keep it simple and distribute equally and consider later if we want more complicated algorithms.

We don't know that there will be multiple steam boilers, if they will have exclusion bounds, and what the correct algorithm is, based on what metrics etc.

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_steam_boiler_manager/_steam_boiler_manager.py
shsms: This also, I would not worry about now.

src/frequenz/sdk/microgrid/_power_distributing/_component_managers/_steam_boiler_manager/_steam_boiler_manager.py
shsms: I guess at some point we stopped caring about the distributor results.  As long as there are meaningful logs, operations and debugging are no problem.

It is like this in PV and EV as well.  Currently the steam boiler is based on the PV pool for many things.  If we want to fix it, I think we can do it separately covering the other pools as well.

src/frequenz/sdk/timeseries/steam_boiler_pool/_system_bounds_tracker.py
shsms: This corresponds with PV and it doesn't matter, because the component becomes NOT_WORKING, and then any other param is irrelevant.

src/frequenz/sdk/microgrid/_old_component_data.py
shsms: I guess we're contradicting the proto everywhere. in the old_component_data.py file, not just here.

Besides, both the proto and the SDK docs are completely wrong about the meaning of reactive power signs.  So both need to be fixed, but separate project I guess.

src/frequenz/sdk/microgrid/_old_component_data.py
shsms: Maybe you can forget about this for now.  There's a big cleanup coming soon anyway, with the moving of the algorithmic parts to be behind the microgrid API.

src/frequenz/sdk/microgrid/_old_component_data.py
llucax: Are there issues for this? I seems nasty to document stuff that is not true.

llucax (approved):
OK, approving, assuming there were no code changes except for resolving any conflicts or adapting the necessary bits to the fixes for exhaustive match.

shsms:
Yes, it was only a rebase.

PR #1455, opened 2026-09-08
Make `match` statements pass mypy's `exhaustive-match` check

llucax:
#1448 updates `frequenz-repo-config` to 0.19.0, which enables mypy's `exhaustive-match` error code, and CI fails on 11 `match` statements that mypy can't prove to be exhaustive. These are the code changes needed to make it pass.

Most of the reported statements match over a tuple, and mypy can't prove *any* of those exhaustive: it narrows tuple items inside a matching case, but can't subtract combinations across cases (python/mypy#12364), so writing the missing case does not help. Those either stop matching over a tuple, or get an explicit last case. The rest were genuinely missing cases, which are now spelled out.

One bug came out of it, in `COALESCE()`, and it has its own commit and release notes entry. Everything else is behaviour preserving, except that the formula operators now raise instead of silently evaluating to nothing when an operand has an unexpected type, which can only happen if something violates the type hints at runtime.

### DIFF
--- a/src/frequenz/sdk/timeseries/formulas/_functions.py
+++ b/src/frequenz/sdk/timeseries/formulas/_functions.py
@@ -10,7 +10,7 @@ import asyncio
 import logging
 from dataclasses import dataclass, field
 from datetime import datetime
-from typing import Generic
+from typing import Generic, assert_never
 
 from frequenz.quantities import Quantity
 from typing_extensions import override
@@ -117,7 +117,7 @@ class Coalesce(Function[QuantityT]):
         """Return the name of the function."""
         return "COALESCE"
 
-    @override
+    @override  # pylint: disable-next=too-many-branches
     async def __call__(self) -> Sample[QuantityT] | QuantityT | None:
         """Return the first non-None argument."""
         ts: datetime | None = None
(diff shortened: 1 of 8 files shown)


llucax:
Adding a new commit to enable relative cross refs in mkdocs as a quick fix. It seems for some reason properdocs is being used, not sure if that is related to parsing docs from a dependency or not.

llucax:
Dammit! This didn't help because the quantities docs relative references are really incorrect. I have no idea why the CI didn't fail in the quantities repo. See:

* https://github.com/frequenz-floss/frequenz-quantities-python/pull/170#pullrequestreview-5139931001

llucax:
Removing the last commit, it is not helping, it seems to be a bug in mkdocstrings relative-cross referencing :grimacing:

llucax:
Green now.

llucax:
@shsms some changes affect formulas including a bug fix, so it would be good to have your review here.

shsms:
All commits make sense and the first one is particularly nice because it simplifies a public interface.

Still looking into the second commit.

src/frequenz/sdk/timeseries/formulas/_functions.py
```diff
@@ -132,16 +132,29 @@ async def __call__(self) -> Sample[QuantityT] | QuantityT | None:
...
+                        # unsubscribe from the last parameter if any earlier
+                        # one produces at least
+                        # `REQUIRED_CONSECUTIVE_STABLE_SAMPLES` samples,
+                        # regardless of intermittent non-None values received
+                        # from other params.
+                        used_arg = (
+                            args[self.used_param - 1] if self.used_param > 0 else None
+                        )
+                        match used_arg:
+                            case None | Sample(value=None):
```
shsms, line 146: I don't think there are any cases that would produce a plain None.  So this is improving type coverage, not fixing a bug.  I think an alternate solution is to drop the None from the types completely, and that would make some of the other nodes simpler as well.

But happy to accept it as it is, because it is only a minor cost.

shsms (approved):

src/frequenz/sdk/timeseries/formulas/_functions.py
shsms: And the unit test is also just making up some random stuff, because it cannot produce a case to test this by going through the parser.

llucax: I might not have looked at the tests :grimacing:

